PR #1, opened 2016-02-11
Fix PHP warnings during Chado install

njbooher:


spficklin:
Thanks for providing some code fixes!

PR #2, opened 2016-02-24
Added license

mestato:


spficklin:
In order for the file to be properly overwritten by Drupal (so we don't have two license files), it must be named LICENSE.txt

spficklin:
Thanks for noticing the missing file

PR #3, opened 2016-05-04
Fix PHP warnings

njbooher:


spficklin:
Thanks for catching these!   I'll merge them in

PR #4, opened 2016-08-01
Fix "theme is missing" warning

abretaud:
On a fresh install of tripal 2.0 or 2.1b3 I get the following warning on most of the pages, even if I only enable the tripal_core module:

```
User warning: The following theme is missing from the file system: tripal. In order to fix this, put the theme back in its original location. For more information, see the documentation page. in _drupal_trigger_error_with_delayed_logging() (line 1128 of /var/www/html/includes/bootstrap.inc).
```

As it seems like there is no tripal theme anymore, I simply removed the line that triggered the warning, I don't know if it still should be declared somehow (for compatibility maybe?)

### DIFF
--- a/tripal_core/tripal_core.module
+++ b/tripal_core/tripal_core.module
@@ -81,11 +81,9 @@ function tripal_core_init() {
   }
 
   // add some variables for all javasript to use for building URLs
-  $theme_dir = drupal_get_path('theme', 'tripal');
   $clean_urls = variable_get('clean_url', 0);
   drupal_add_js(
     " var baseurl  = '$base_url';
-      var themedir = '$theme_dir';
       var isClean  =  $clean_urls;",
     'inline', 'header');
 


MingChen0919:
I got exactly the same warning when installing Tripal 2.0.

Ming

On Mon, Aug 1, 2016 at 9:44 AM, Anthony Bretaudeau <notifications@github.com

MingChen0919:
this error has been fixed in the github tripal version but not in the
drupal repo. if you use drush to install tripal, you will have this error.
To fix this, you can git clone tripal-7.x-2.x and then drush install it.

Ming

On Mon, Aug 1, 2016 at 9:44 AM, Anthony Bretaudeau <notifications@github.com

spficklin:
Hi Ming,
Thanks for reporting. Is this on Tripal v2 or Tripal v3?  What drush command are you using to install from Drupal?

MingChen0919:
Hi Stephen,

This is on Tripal v2. This tripal theme missing error was fixed by Anthony,
and the pull request was merged on github. But it seems the tripal module
on drupal has not been updated yet. So if I use 'drush dl' to download the
module directly from drupal, it will generate the tripal theme error.

Thanks,
Ming

spficklin:
Oh yes.  We have yet to make another release for the Tripal 2 version.  The updated code should be on Drupal.org but in the development version 7.x-2.x.  You're right, once we make a new release it will be more easily accessible.

PR #5, opened 2016-08-02
Customize auto-created protein names with regex

abretaud:
Two new fields in the GFF loader to allow customizing the name of the automatically created protein features using a regex.
I talked about that in this issue: https://www.drupal.org/node/2386685

### DIFF
--- a/tripal_feature/includes/tripal_feature.gff_loader.inc
+++ b/tripal_feature/includes/tripal_feature.gff_loader.inc
@@ -330,7 +330,7 @@ function tripal_feature_gff3_load_form_submit($form, &$form_state) {
   $args = array($gff_file, $organism_id, $analysis_id, $add_only,
     $update, $refresh, $remove, $use_transaction, $target_organism_id,
     $target_type, $create_target, $line_number, $landmark_type, $alt_id_attr,
-    $re_mrna, $re_protein, $create_organism);
+    $create_organism, $re_mrna, $re_protein);
 
   $type = '';
   if ($add_only) {
@@ -411,10 +411,6 @@ function tripal_feature_gff3_load_form_submit($form, &$form_state) {
  *   Sometimes lines in the GFF file are missing the required ID attribute that
  *   specifies the unique name of the feature. If so, you may specify the
  *   name of an existing attribute to use for the ID.
- * @param $re_mrna A
- *          regular expression to extract portions from mRNA id
- * @param $re_protein A
- *          replacement string to generate the protein id
  * @param $create_organism
  *   The Tripal GFF loader supports the "organism" attribute. This allows
  *   features of a different organism to be aligned to the landmark sequence of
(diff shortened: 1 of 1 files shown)


laceysanderson:
This looks like an awesome feature addition -Thanks for writing this up!

Just a couple of comments:
1. I noticed that you changed the order of the parameters to tripal_feature_load_gff3() For backwards compatibility you should add your optional parameters to the end (right before job_id) rather than before $create_organism.
2. Also, within tripal_feature_load_gff3() you set the name to be equal to the transformed uniquename where the original functionality was for the mRNA & protein names to be the same. I think by default the name should be the same regardless of whether a regex was supplied. If you really need that feature perhaps you might want to add a checkbox to the loading form that allows users to say they want the name=uniquename (this could even be a separate feature entirely).

Otherwise, it looks good :-) Though Drupal coding standards enforce that each element of the Form API array be on it's own line (picky detail that wouldn't stop merging).

Note: I'll wait to test this patch until those changes are incorporated.

abretaud:
Thanks for reviewing!

> This looks like an awesome feature addition -Thanks for writing this up!
> 
> Just a couple of comments:
> 1. I noticed that you changed the order of the parameters to tripal_feature_load_gff3() For backwards compatibility you should add your optional parameters to the end (right before job_id) rather than before $create_organism.

Ok, I changed it

> 1. Also, within tripal_feature_load_gff3() you set the name to be equal to the transformed uniquename where the original functionality was for the mRNA & protein names to be the same. I think by default the name should be the same regardless of whether a regex was supplied. If you really need that feature perhaps you might want to add a checkbox to the loading form that allows users to say they want the name=uniquename (this could even be a separate feature entirely).

Ok, I didn't know this default behaviour, I switched back to the name value. If I need this feature later, I'll make another PR

> Otherwise, it looks good :-) Though Drupal coding standards enforce that each element of the Form API array be on it's own line (picky detail that wouldn't stop merging).

Did you mean this:      d70270a ?

> Note: I'll wait to test this patch until those changes are incorporated.

laceysanderson:
As far as coding standards go, you're still not quite there after d70270a. You have:

``` php
  $form['advanced'] = array(
    '#type' => 'fieldset','#title' => t('Advanced Options'),
    '#collapsible' => TRUE,'#collapsed' => TRUE
  );
```

Whereas it should be:

``` php
  $form['advanced'] = array(
    '#type' => 'fieldset',
    '#title' => t('Advanced Options'),
    '#collapsible' => TRUE,
    '#collapsed' => TRUE
  );
```

Notice how each element of the array is on it's own line. As I said, picky detail ;-)

Thanks for making the other changes, I'll try to test this either today or tomorrow :-)

abretaud:
Oh yes, I had missed the multiple elements on the same line, it should be ok now!

abretaud:
Hi,
Did you find some time to test this?
I have also a few PR waiting for tripal_analysis_blast and tripal_analysis_interpro, should I ping someone to have a look at it?
Thanks
Anthony

laceysanderson:
Hi Anthony, sorry for the lag time! I've actually been gone on holidays and while I meant to look at this before leaving obviously that didn't happen. Clearing up these issues is top on my list

laceysanderson:
I have incorporated these changes into the main dev branch with a small amount of re-organization on the form to make things flow a little better. Thanks for the contribution!

abretaud:
Ok no problem! Thank you for merging!

PR #7, opened 2016-08-17
Fix permissions on tripal_db_admin_dbs view

njbooher:
Fixes [Drupal.org issue #2765775](https://www.drupal.org/node/2765775).

I decided instead of adding a new permission it'd probably be better just to change the view to match what the rest of the module uses.


laceysanderson:
Looks good, thanks!

PR #8, opened 2016-09-01
Properly detect chado 1.31

abretaud:
Hi, 
Just a little fix for this issue: https://www.drupal.org/node/2793501
It should be ok if (as I guess) 1.3 and 1.31 versions of chado are very similar


spficklin:
I'm surprised we missed that.  Thanks for the fix!

abretaud:
Thanks for merging! It seems like it didn't land on the drupal git repo: https://www.drupal.org/node/1337878/commits
Would it possible to push it there too? (or should I just wait a bit?)

PR #10, opened 2016-09-28
Fix theme_table() call when displaying list of features

abretaud:
Hi,
This is a fix for an error I got while trying to access a list of features with an url like this:
http://example.com/feature/XXX000254
I guess the theme_table function signature got changed in drupal 7?


spficklin (approved):

spficklin:
Yeah, it did get changed in D7.  Despite our efforts to catch all of these we missed one.  Thanks for finding and correcting.

PR #11, opened 2016-10-26
Fix duplicate files directories (7.x-2.x)

njbooher:
Currently when using tripal_get_files_dir an empty identically-named subdirectory is automatically created within the actual files directory (e.g. real files directory is files/tripal_organism and this is what is returned, but another folder files/tripal_organism/tripal_organism is also created). This fixes that.


spficklin:
Thanks for providing the fix for both Tripal 2 and 3!

PR #15, opened 2016-10-27
Replace views patch with nicer one that doesn't break other modules

njbooher:
The current views patch Tripal uses always turns the right table in a join into a string. This breaks relationships like views_handler_relationship_node_term_data which use a subquery as a right table.

The attached patch avoids interfering with other modules that aren't using search_path by only prepending the search_path to the right table if search_path is defined.

https://www.drupal.org/node/1971160#comment-11752459


spficklin:
I've sent a request to have Lacey review this one as she's are Views expert.

laceysanderson:
Finally found the time to test the patch and it works great :-) Thanks for the contribution!

PR #18, opened 2016-11-30
Allow Admin Users to Execute Jobs Using the GUI

almasaeed2010:
Hello,

We added a new feature to the Tripal Jobs module to allow users to execute jobs from the control panel. The execute link shows up on both the jobs list page and the job page. The feature becoms available only for jobs that have not started executing yet.

**Job page:**
screen shot 2016-11-30 at 10 40 18 am

**List page:**
screen shot 2016-11-30 at 10 40 50 am

I hope you find this useful.

Thanks,
Abdullah


spficklin:
Thanks for helping improve the core code! It's much appreciated.  I tested this branch and I have a few requests if you can make them.  Can you remove the 'Execute' link from jobs that are not in a 'Waiting' state?  User's get the impression they can execute completed jobs because the link is there but in reality they can't.  Also, I tested executing a new job. I click the 'Execute' link for the job and it redirects me to the job page but there is no message indicating if the job run was successful.  Could you add a drupal_set_mesage() function call to let the user know the status of the run?  That way it will be obvious if there was a problem or if it completed successfully.

almasaeed2010:
I added the message to inform the user of the completed action. 

Since the jobs list page is using views, I am not sure how to limit the ```execute``` link to appear on specific conditions. I think this issue also concerns the ```cancel``` link because it shouldn't appear for jobs that have been completed or terminated.

Any suggestions on how to do it?

Thanks!

spficklin:
Okay. I merged!  I forgot that that page was provided by views.  We'll have to add that as an issue and fix it for the future.

spficklin:
Thank you!

almasaeed2010:
No problem

PR #19, opened 2016-12-01
Small cleanups

abretaud:
Here are a few things:
-some typos
-fixes for "undefined variable" notices
-a small patch to avoid doing some unnecessary sql queries (not sure if the query results were cached though, and I haven't benchmarked this)


spficklin:
Thanks for the fixes!

PR #20, opened 2016-12-05
Fix Function Call Typo

almasaeed2010:
Hello,

I found an error in the ```chado_select_record()``` function where a call to ```tripal_report_error()``` was intended but ```chado_select_record()``` was used causing a fatal error.

Thanks!


spficklin:
Hi almasaeed2010.  Thanks for submitting the pull request!  It's very much appreciated.  I'm sorry for the slow reply.  I'm not able to see where the fix occurred  because there are too many formatting changes to the code.  Some of the formatting changes are good and some still need adjusting.  To help me out, can you redo this pull request with only the fix to the code?

Thanks

almasaeed2010:
Hello @spficklin  
No problem. I created a new pull request that contains only the fix #23 

Thanks!

PR #22, opened 2017-02-01
Fixes for gff_loader: wrong landmark and strand assignation for protein + "Illegal string offset" error

abretaud:
I just stumbled upon 2 bugs in the gff loader:
 - the attr_others was initalized as a string, but used as an array => it triggered an "Illegal string offset" error on my setup
 - a bit more scary, the proteins created when loading a gff were positioned on the wrong reference sequence and with wrong strand too. This because it reused the $landmark and $strand variables used while parsing the gff => it reused the values from the last gff line for all proteins.

For this last bug, I guess it could be cleaner to modify the cds/protein temp tables to add a landmark column, but it would break compatibility with older databases...


spficklin:
Thank you for these fixes!  Great catches.

PR #23, opened 2017-02-06
Fix function call

almasaeed2010:
This is a replica of #20 but with only the error. No reformatting this time.

Thanks!


spficklin:
What a silly mistake we made.  Thanks for noticing and for the fix!

almasaeed2010:
No problem!

PR #25, opened 2017-02-13
Tripal Job Launcher changes: --max_jobs and --single

ekcannon:
Added --max_jobs to limit # of concurrent jobs, and --single to indicate job launcher should run only one job on each invocation; both intended to be interpreted in conjunction with --parallel=1

--max_jobs=# is to protect the system from being overloaded with concurrent jobs (a possible hazard with lots of users and with the BLAST UI functionality now exposed via Web services)

--single=1 is to prevent a backlog of queued jobs from running concurrently. If not set, the job launcher will loop through each queued job sequentially. If set, multiple queued jobs will be picked up by subsequent invocations of the job launcher.

### DIFF
--- a/tripal_core/api/tripal_core.jobs.api.inc
+++ b/tripal_core/api/tripal_core.jobs.api.inc
@@ -212,6 +212,51 @@ function tripal_get_job_end($job) {
 
   return $end;
 }
+
+
+/**
+ * Check for too many concurrent jobs
+ *
+ * @param $max_jobs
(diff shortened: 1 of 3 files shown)


laceysanderson:
See branch "LegumeFederation-multijobs" in the tripal repo for fixes to the merge conflict. I'm currently testing functionality now.

laceysanderson:
See updated pull request for a few small changes... Thanks for the great contribution! I've made sure your group is given proper attribution in the new pull request :-)

PR #51, opened 2017-03-31
Tripal Job Launcher changes: --max_jobs and --single

laceysanderson:
This pull request includes #25 with a small fix to the max jobs functionality. The original contribution allowed 3 jobs if max-jobs was set to 2 due to max_jobs_exceeded only returning true when [number of running jobs] > [max jobs].

Proper attribution is given in this pull request giving most of the credit to LegumeFederation for the original contribution --Thanks!!

### DIFF
--- a/tripal_core/api/tripal_core.jobs.api.inc
+++ b/tripal_core/api/tripal_core.jobs.api.inc
@@ -212,6 +212,51 @@ function tripal_get_job_end($job) {
 
   return $end;
 }
+
+
+/**
+ * Check for too many concurrent jobs
+ *
+ * @param $max_jobs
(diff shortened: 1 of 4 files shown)


laceysanderson:
I've tested the following cases and everything works for me
1) --single when not running parallel jobs only runs one job per drush call and only if nothing is already running.
2) --max_jobs=2 --parallel=1 without --single only allows two jobs to be run concurrently.
3) --max_jobs=2 --parallel=1 --single only allows two jobs to be run concurrently and only runs one job per drush call.

I give this LegumeFederation contribution my stamp of approval 👍

spficklin:
Thanks @laceysanderson and @ekcannon !!

PR #146, opened 2017-08-11
Fix small warning

abretaud:
Hi,
A small fix for a warning I get on this line: "A non-numeric value encountered tripal_core.chado_nodes.api.inc:886    [warning]"
I'm not sure what exactly is supposed to do ".-", I guess it was just a typo


spficklin:
Thanks for the fix!

abretaud:
It would be nice to have the drupal git repo updated for this too

spficklin:
Right... It's now been pushed to the drupal git repo as well.

PR #149, opened 2017-08-23
Fix Warning: in tripal_bulk_loader_load

almasaeed2010:
screen shot 2017-08-23 at 8 23 49 am

This simple fix gets rid of the above error in php 7.1 and 7.0.

Thanks!


almasaeed2010:
Also added a fix for a chado_query error where the column chado_organism is called as chado.chado_organism rather than the public chado_organism.

spficklin:
Thanks much for the fixes! The first fix was also needed in Tripal v2 so I've committed it there as well.

Issue #154, opened 2017-08-28
Tripal 3 loader error


bradfordcondon:
Currently getting the following error with all new Loaders (Chado FASTA loader, GFF3 loader)...  after pressing "Upload File" on both local dev and our Hardwoods dev site:
`SyntaxError: Unepected token { in JSON at position 48`

![screen shot 2017-08-28 at 10 02 06 am](https://user-images.githubusercontent.com/7063154/29776869-15b5673a-8bd8-11e7-924e-010e70b767e9.png)

Currently investigating `tripal/theme/js/TripalUploadFile.js` to see if I can figure out the problem but if anyone knows the issue please let me know!

bradfordcondon:
output: I have 1 chunk successfully written. Below is the contents of `tripal_upload.log`
 
`a:1:{s:14:"chunks_written";a:1:{i:0;s:2:"27";}}`

spficklin:
Is this the case where the file is only the size of a single chunk?  Otherwise, I've not encountered this problem before.

bradfordcondon:
Yes it was.  However the chunk size is pretty large (2GB I believe?) so i'd imagine it will occur.

Edit:  oops its 2MB.  `this.chunk_size = (1024 * 2000); // 2024MB` the human name is 2GB though :P

spficklin:
Hi @bradfordcondon I submitted a fix for this problem that is different from your pull request #155.  I'll comment on that pull request but the fix proposed didn't quite fix the real issue which was an accidental incorrect naming of a function that made it look like a hook when it wasn't.

bradfordcondon:
Hi Stephen, the current branch works for us, thank you so much!

PR #155, opened 2017-08-28
fix tripal 3 loader  #154

bradfordcondon:
The JSON error was from returning malformed JSON: check was running when it wasnt supposed to.  To fix this, we make sure the check only runs when the action is check. 

see #154

### DIFF
--- a/tripal/includes/tripal.upload.inc
+++ b/tripal/includes/tripal.upload.inc
@@ -22,33 +22,37 @@ function tripal_file_upload($type, $filename, $action = NULL, $chunk = 0) {
 
   // Allow the module that will own the file to make some checks. The module
   // is allowed to stop the upload as needed.
-  $hook_name = $module . '_file_upload_check';
-  if (function_exists($hook_name)) {
-    $details = array(
-      'filename' => $filename,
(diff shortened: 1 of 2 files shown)


spficklin:
Hey @bradfordcondon Can you give an example of when this problem occurred?

spficklin:
Oh, I see you referenced issue #154 .. let me look at that.

bradfordcondon:
We didn't test files over 2MB so we very well may break the loader in those (much more important) cases, let's test first

bradfordcondon:
I just tested with 200MB and had no problems (with my version).  When i reverted back to master, I had the same error (see screenshot).

screen shot 2017-08-28 at 3 56 27 pm

spficklin:
I'm going to close this pull request as  I just submitted a fix for the problem of issue #154 that is a bit different from this one.  This request adds an if statement around the call to the hook_file_upload_check to make sure it was a 'check' action.  But that hook is meant to be called for all 3 actions not just for a 'check'.  The problem was that the tripal_file_upload_check function was incorrectly named to be the same as the hook and it was being called incorrectly.

PR #157, opened 2017-09-07
Add events to tripal ds

almasaeed2010:
Hello,

I added an event to the Tripal DS module that allows other modules to detect size changes of their container. We encountered an issue where D3JS could not compute the width of its container because of the size changes during the animation. Since d3js is commonly used with Tripal, this event will be helpful to many other modules.

Problem: Chart prematurely computed dimensions of container
screen shot 2017-09-07 at 10 45 26 am

Fixed with listening to the event and recomputing the container width post animation completion:
screen shot 2017-09-07 at 10 50 45 am

Usage example:
```js
// Container expanded and animations completed
$(document).on('tripal_ds_pane_expanded', function(event) {
  // React to completion of animation
  // For example, redraw the chart with the correct dimensions
  // The id attribute of the pane is stored in event.id
});

// Container collapsed and animations completed
$(document).on('tripal_ds_pane_collapsed', function(event) {
  // React to completion of animation
  // The id attribute of the pane is stored in event.id
});
```

Thanks!


shawnawsu:
@almasaeed2010 We've merged this into the 7.x-3.x branch, can you please test that it's still successfully addressing the issue with D3js content render? Thanks!

almasaeed2010:
Thanks for merging!

It works great with our d3js libraries when hooked into. Here is an example of how we are using it:
```
$(document).on("tripal_ds_pane_expanded", function(event) {
    redrawChart();
});
```

Issue #162, opened 2017-09-14
TripalImporter upload and AJAX calls


bradfordcondon:
I am using the Tripal Importer to upload XML files for biosamples/biomaterials, and utilize a "check" button that performs an AJAX call to sample the file and look in the db for CVTerms that match XML attributes.

After rebuilding part of the form, the File Uploader goes haywire (screenshot below).  It loses sight of the uploaded file, and adds a second "upload file" column.

To make matters weirder, I can still submit the form after this... but the file is no longer the correct file (it runs with a different file in the files folder)!  

<img width="1229" alt="screen shot 2017-09-14 at 2 40 47 pm" src="https://user-images.githubusercontent.com/7063154/30447826-1b2ece78-995b-11e7-8eca-eb492f1b0529.png">

It should be noted I'm a bit behind in my tripal branch because of my unmerged PR #155

bradfordcondon:
An even simpler (but possibly related) problem:  If you upload a file, then submit a form that won't pass validation, you lose the file in the form (but you dont get that duplication of headers).

spficklin:
I merged your pull request #163 so that should fix this problem for you.  If not just re-open this issue.

PR #163, opened 2017-09-14
Prevent ajax replacement

almasaeed2010:
This is a solution to issue #162.

The solution simply prevents TripalUploader from attaching to html form inputs more than once.

Thanks!


spficklin:
Thanks for catching this.

Issue #168, opened 2017-09-27
label callback function doesnt exist?


bradfordcondon:
The relation module is calling the [`entity_label` method.](https://api.drupal.org/api/drupal/includes%21common.inc/function/entity_label/7.x)  I see that Tripal entities have the label callback function correctly defined as `tripal_entity_label`.  However, `function_exists(tripal_entity_label)` returns false and I cant find the callback in the code.  Am I missing something or is this still on the To Do list?

edit:  Maybe we are supposed to define the callback function for each bundle?  That would make some sense but it there's no implementation of it for the core Tripal entity types.

edit 2: I just made a simple PR that returns the node title.  I imagine you all have a solution with more finesse in mind, interested in hearing it.

Thanks!

spficklin:
Pull request has been applied.  Thanks for providing the fix.

PR #169, opened 2017-09-27
Label callback

bradfordcondon:
Very simple fix for #168.


spficklin:
Thanks for the fix.  I think this is the proper way to handle this as we use the title for the label in web services as well.

PR #183, opened 2017-11-06
Various fixes

abretaud:
Just a few fixes for crashes or warnings I found while working on using tripal3 in https://github.com/erasche/docker-tripal


spficklin:
Thanks for catching these and making fixes!

PR #187, opened 2017-11-10
allow many extensions for Fasta Importer

bradfordcondon:
Fasta files are coming from a variety of sources, and they are famously inconsistent in extension naming.  Having to rename all extensions to `.fasta` is a hassle.  I propose updating the importer to allow all of the extensions I've seen in my journeys.


bradfordcondon:
Note to self dont delete branches you make PRs from

spficklin:
Sounds like a good fix to me!

PR #189, opened 2017-11-15
Remove empty custom panes

almasaeed2010:
Hello all,

We noticed that custom tripal panes built through the UI do not get removed if their content is empty. Therefore, we are left with a sidebar link that when clicked nothing happens.

This pull request fixes the problem by checking if a pane has been created and removes the link if none exists. It also address an issue where `prevObj` is undefined.

Thanks!

### DIFF
--- a/tripal_ds/theme/js/tripal_ds.js
+++ b/tripal_ds/theme/js/tripal_ds.js
@@ -36,28 +36,34 @@
       });
       // Move the tripal pane to the first position when its TOC item is clicked.
       $('.tripal_pane-toc-list-item-link').each(function (i) {
+        var id = '.tripal_pane-fieldset-' + $(this).attr('id');
+        if ($(id).length === 0) {
+            $(this).parents('.views-row').first().remove();
+            return;
+        }
         $(this).click(function() {
-          var id = '.tripal_pane-fieldset-' + $(this).attr('id');
           var prevObj = $(id).prev().attr('class');
-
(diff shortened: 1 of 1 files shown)


shawnawsu:
Hi @almasaeed2010 ,

"We noticed that custom tripal panes built through the UI do not get removed if their content is empty. " When you say this do you mean the pane or the menu item in the table of contents block? Can you supply a screenshot of what you mean?

Thanks!
Shawna

almasaeed2010:
Hello @shawnawsu,

I mean the link that appears on the left side (TOC). Please see the following link for an example:
https://media.giphy.com/media/l2QDMZa1nJ8ItYBmE/giphy-hd.mp4

In this case, the "Reference Genome" and "Physical and Genetic Maps" have no content attached to them and therefore clicking them causes a JS error:
```
Uncaught TypeError: Cannot read property 'indexOf' of undefined
    at HTMLAnchorElement.<anonymous> (tripal_ds.js?ozks6n:44)
    at HTMLAnchorElement.dispatch (jquery.min.js?v=1.10.2:5)
    at HTMLAnchorElement.v.handle (jquery.min.js?v=1.10.2:5)
```

The proposed solution removes any TOC links that don't have an associated pane.

Please let me know if this isn't clear.

Thanks!

shawnawsu:
That is clear, your original pull request communicated that the actual pane was showing and causing a problem. Since the code submitted didn't match with that I wanted to check the intention.

Thanks!
Shawna

PR #192, opened 2017-11-15
Fix analysis migration 

bradfordcondon:
Hi simple fix for an issue we were having with migration not working properly for analysis types.

## Problem
* Migrating Analysis explicitly creates a new bundle, local:analysis.
* Migrating "All" content type works fine.

## Solution
replace ` $key = urlencode('tv3_content_type--local--analysis--analysis');` with `  $key  = urlencode('tv3_content_type--operation--2945--analysis');` in the form that gets passed down


bradfordcondon:
After changing analysis I noticed two locations where the local biomaterial term was used and changed them to the new biological sample term.

PR #195, opened 2017-11-21
Fix some warnings

abretaud:
Hi,
A few small fixes I had to do while working on porting python-tripal to tripal3.
I also have this warning:
`Undefined index: file TripalImporter.inc:418                            [notice]`

Looking at https://github.com/tripal/tripal/blob/7.x-3.x/tripal/includes/TripalImporter.inc#L418 I have the feeling that `arguments['file']` should be `arguments['files']` in multiple lines, but I'm not 100% sure, can you have a look? Also there are some `$this->arguments['file']['file_remote']` that look odd to me (compared to `$this->arguments['file'][$i]['file_remote']` a few lines above)

If you leave this PR open for a few days, maybe I'll add other small fixes I find while testing.


spficklin:
Thanks for the fixes!  Much appreciated

Issue #200, opened 2017-11-29
Tripal.chado.cv.api inserts into undefined table


bradfordcondon:
### Error
I'm trying to get tripal analysis expression to install on tripal 3 with a clean bill of health, and i have the following error:

```
exception 'PDOException' with message 'SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "tripal_cv_defaults" does not exist                   [error]
LINE 1: DELETE FROM tripal_cv_defaults
```

Note this is using Ming's docker as a base, which is i believe the quick, pre-install of tripal 3 in the .info site.

### Proposed cause and solution
the tripal 3 chado.cv function `tripal_set_default_cv` deletes and inserts into `tripal_cv_defaults`.  However this schema only appears to be defined in legacy tripal_chado, not in the tripal 3 tripal_chado module.

Assuming we still want this functionality, we should define the schema, i guess in `tripal_chado.install`

bradfordcondon:
consensus is that this method is deprecated in Tripal 3.  So rather than defining the schema, `tripal-set_default_cv` should be deprecated.

bradfordcondon:
duplicate issue with #206

PR #201, opened 2017-11-29
add the cv defaults table that was in the legacy tripal cv

bradfordcondon:
This PR resolves issue #200


shawnawsu:
Hi Bradford, 

Since this is for a specific module I think the table creation should go in the module's install file. But I'm relatively knew to tripal so @spficklin  please correct me if I am wrong. 

Thanks,
Shawna

bradfordcondon:
if that is the case then you should deprecate `tripal_set_default_cv`  which is in tripal 3 and requires this table.

shawnawsu:
I agree, especially because it's an unused function, but let me confirm with Stephen what his preference is.

spficklin:
We don't want to bring back the CV defaults table for Tripal v3 as we no longer need it.  Can you give an example for why you want it?

bradfordcondon:
I don't even know why you would want to set a default CV, i was just fixing install errors and it looked like this method was still invoked.  If its no longer necessary I'll remove the calls in my modules instead.

bradfordcondon:
I'm closing this PR.  Sounds like the correct fix is to remove `tripal_set_default_cv` instead.

Issue #202, opened 2017-12-01
API to create organism, analysis, ...


abretaud:
Hi,
I'm working on porting python-tripal to tripal3. I need to create new organisms (or analysis, ...) using the API.
With tripal2, I was using the Services module which exposed CRUD methods for nodes.
Now that organisms are no longer nodes in tripal3, I wonder how to create them programmatically.

Is there in the Tripal code a magic function like tripal_add_organism which would create the entity and populate the chado table? If not, what's the easiest way?

(I've started implementing in tripal_ws some code to catch POST HTTP request and use the posted json array to create entities. We could create a new organism like this:
```
curl -X POST -H "Content-Type: application/json" -d '{"genus":"Mygenus", "species":"somespecies", "common_name":"Cool org", "description": "blablabla <br> test"}' http://localhost:8500/tripal/web-services/content/v0.1/Organism/
```
What do you think about it? I can also implement it in tripal_rest_api module if you prefer)

shawnawsu:
HI @abretaud ,

Just so I understand, you're asking about creating new organisms (entity) when the organism content type (bundle) already exists?

If you are looking to do entities have you looked at using the tripal.importer.api? That will load your info then you could use the tripal.entities.api to load them. 

Thanks,
Shawna

abretaud:
Hi

> Just so I understand, you're asking about creating new organisms (entity) when the organism content type (bundle) already exists?

Yes that's it!

> If you are looking to do entities have you looked at using the tripal.importer.api? That will load your info then you could use the tripal.entities.api to load them.

Yes I know about the new importer mechanism, I use it now for loading fasta, gff, blast, interpro and go.
I didn't want (but maybe I'm wrong!) to use it for organism-like entities because I was looking for a more lightweight way to create them, without creating a job.
I've looked at tripal.entities.api but it's not clear to me how I can use it :/

Anthony

spficklin:
Hi Anthony,

This is a very good question.  There is currently no API function to create organisms programatically.  We need to do this.  Technically, you can do this now by creating an entity object programmatically.   For this you create a new entity instance and as long as you add the fields with appropriate values you can save the entity and the Tripal code will try to do the insert. It will work as long as you setup the entity object correctly.  If you want to try to explore this route I can share with you some code examples.  

[...]

PR #211, opened 2017-12-07
Implement POST method in API to create entities

abretaud:
Hi,
Following the discussion in #202, I've written this code as a first attempt to implement a POST method to create entities.
Here's a quick curl command to create an organism (supposing anonymous users have the permission in the tripal instance):

```
curl -X POST -i -H "Content-Type: application/json" -d '{"genus":"Testus", "species":"species", "description": "OMG\nit works", "common_name":"Joe the fly", "abbreviation": "G. speciesa"}' http://localhost:8500/tripal/web-services/content/v0.1/Organism/
```

You get something like this as a response:

```
HTTP/1.1 201 Created
Server: nginx/1.11.7
Date: Thu, 07 Dec 2017 16:43:03 GMT
Content-Type: application/ld+json
Content-Length: 0
Connection: keep-alive
X-Drupal-Cache: MISS
Expires: Sun, 19 Nov 1978 05:00:00 GMT
Cache-Control: no-cache
X-Content-Type-Options: nosniff
Link: <http://localhost:8500/tripal//web-services/vocab/v0.1>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"
Location: http://localhost:8500/tripal//web-services/content/v0.1/Organism/10
```

You can then get the details of the created organism with the Location returned by the POST query (ie 201 redirection):

```
curl -X GET http://localhost:8500/tripal//web-services/content/v0.1/Organism/10 | jq
```

I still haven't tested much more complicated fields like infra specific taxons, but any feedback would be really appreciated, I'm sure there are many things to improve or do differently.

While looking at the code, I've also fixed several typos, I added them to this PR to.

Anthony

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalEntityService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalEntityService_v0_1.inc
@@ -46,6 +46,11 @@ class TripalEntityService_v0_1 extends TripalWebService {
    */
   public function handleRequest() {
 
+    // Redirect POST requests to dedicated function
+    if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+        return $this->handlePostRequest();
+    }
+
     // Get the content type.
     $ctype     = (count($this->path) > 0) ? $this->path[0] : '';
     $entity_id = (count($this->path) > 1) ? $this->path[1] : '';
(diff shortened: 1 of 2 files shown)


abretaud:
Hi,
I have problems with populating "subfields": for example source_data for analysis or infraspecific_taxon for organism.

For the moment my code works for simple fields (=a single value mapped to a chado column in a chado table) using https://github.com/tripal/tripal/pull/211/files#diff-be613f9a069dda7b6e4634ec2c197c5eR153

For more complicated fields, I just don't know how to properly fill this `'und'` structure (supposing it's the way to create new entities...).

For example, I'd like to set the `sourceuri` for a new analysis. In the field definition code, I see 3 ways to name it: `sourceuri`, `chado-analysis__sourceuri`, `data:1047`. 

- What should the `'und'` structure look like? I've seen things like `['und'][$delta]['source_data']['sourceuri']`, `['und'][$delta]['chado-analysis__sourcename']`, ...
- What is the correct way to explore the fields available for a given entity type? I'm playing with `field_info_instances`, and I tried also `field_info_field` or `tripal_load_include_field_class` + `elementInfo()`, but it feels really complicated, and I'm not sure to find all the needed data. Ideally I'd like to have a GET which would return the list of available fields with proper ids, that I could use in a POST query to create a new entity.

I would really love some help on this, it's blocking me for deploying new data

abretaud:
In https://github.com/tripal/tripal/pull/211/commits/d3b0b528cbce21b6b6f84dc38b3424775f2e8e12 I've added some draft code.

Supposing I'm in the right direction, I don't see how I can discover that I should use the 'chado-analysis__sourceXXX' keys.

I'm also wondering if there should be some changes to the array returned by elementInfo(): as the element keys are ontology term ids, it means that you can't have 2 elements of the same type in a field. For example for the source data field, if I wanted to add another version-like elements, I wouldn't be able to add 2 items with key 'IAO:0000129' in the same hash.

spficklin:
Hi @abretaud .  Thanks for the pull request.  I'm excited to see your solution to the POST for web services. We hadn't got to that yet.  So, thanks for contributing.   

Could you help me out a bit with this.  There are multiple fixes and topics going on with this pull request.  To help me focus on just the bit about the POST could you submit that code as a separate pull request?  You've also made other minor corrects.  Could you submit those as a separate pull request?   And your question about programmatically inserting, can you add that as an new issue in our issue queue?  I will prioritize working on this if you could do that.   

Thanks!

abretaud:
Thanks @spficklin, this PR is cleaned up now, I've moved the noise to #225 and #226
There's still some draft non-working code around https://github.com/tripal/tripal/pull/211/files#diff-be613f9a069dda7b6e4634ec2c197c5eR164, but I'll look at what you propose in #202

spficklin:
Hi @abretaud. I'm going to close out this pull request.  I just committed some code on the 7.x-3.x-entity-crud-api branch that merges in this pull request, but I made some major changes to it. I'll comment on issue #202

Issue #210, opened 2017-12-06
tripal 3 entity chado  FK fields  are numerical input instead of dropdown


bradfordcondon:
As an example, the phylogenetic trees bundle has a numerical input box for analysis_id.  But i think this happens for all auto-generated bundles.  The protocols module arraydesign table, for example, has a column for protocol_id where the widget is a raw input box.

Clearly it would be better to have a dropdown of analyses, protocols, etc, so do we need to create a field for each individually?

![screen shot 2017-12-06 at 5 29 38 pm](https://user-images.githubusercontent.com/7063154/33688891-37e52198-daab-11e7-9afe-9388ed8c6028.png)

spficklin:
Hi @bradfordcondon. Yes this occurs when we have no field other than the default one that Tripal makes. For example, when a new content type is mapped to  a Chado table, Tripal will create a Drupal field for every single column of that table, but by default these fields are just simple Drupal field types.  If a a field needs something more than the Drupal default then we need a special TripalField for it.  This is how the feature.organism_id and stock.organism_id columns are supported.  The **obi__organism** field takes care of those columns.  You don't see a textfield expecting an integer for those fields.  They have organism drop-downs instead.   

We need to create the fields that will support analyses and protocols.   These should be fairly easy to create and we can and should certainly add that to Tripal v3 for the next release.

bradfordcondon:
OK.   So we do require field for each.

I can contribute these fields to the main tripal module and use the `obi__organism` field as as reference.  Unless you all are already working on this?

Cheers

bradfordcondon:
Took a stab at Protocol #213

bradfordcondon:
- [ ] Analysis FK field (figure out #217) first
- [x] Protocol FK  field #254 

I'll track additional FK fields as necessary...

spficklin:
Just checking on this. @bradfordcondon do you want to contribute your fields to the tripal_chado module? Are they sufficently generic?

bradfordcondon:
what happened was i needed help understanding the field methods related to searching/sorting and you said wait for the documentation. 

I'm sure if I revisit this I'd find that the fields all need work...

bradfordcondon:
Specifically:

>But the field is missing an elementInfo(), query() and queryOrder() functions. Do you think you could add those and add another pull request?

laceysanderson:
So then this is currently blocked by #266.

bradfordcondon:
Added a PR with the requested methods.

[...]

PR #213, opened 2017-12-07
Protocol field

bradfordcondon:
Creates a protocol field with dropbox widget.

Looking at the organism field, there's a lot of code there that i dont really understand what its doing.  If I've left out something important it would be good to know what and why.

please see #210 for context


spficklin:
As per @bradfordcondon suggestion... sending back to rework.

PR #219, opened 2017-12-12
if contact field  empty set field to null

bradfordcondon:
The contact field throws errors for biosamples without entries, and displays an empty table.  I set the value to null if there is no contact.

Also the query methods looks for arguments that dont exist, so I've removed those.


spficklin:
Committed a fix for this.

PR #225, opened 2017-12-14
Fix minor typos

abretaud:
@spficklin you're right, #211 was a mess :)
Here's a first PR extracted from it with only the small typos I found while looking at the code. Most of them are in the comments + a few ones in the code, but it should have 0 impact


spficklin:
Thanks again!

PR #229, opened 2017-12-15
Fixes for entity crud api

abretaud:
3 little fixes for problems I found while testing the entity-crud-api branch, discussed in #202


spficklin:
Nice catch.  Thanks for testing and fixing!

Issue #242, opened 2018-01-04
PDO exception on editing a cvterm


bradfordcondon:
### Error

```

PDOException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: 
"edit" LINE 1: SELECT name FROM chado.cvterm WHERE cvterm_id = 'edit' ^: SELECT name FROM chado.cvterm WHERE cvterm_id = :id; Array ( [:id] => edit ) in chado_query()
 (line 1677 of /Users/chet/UTK/tripal/sites/all/modules/tripal/tripal_chado/api/tripal_chado.query.api.inc)

```

### Solution
the edit url: 
http://localhost:8081/organism/226#overlay=admin/tripal/loaders/chado_vocabs/chado_cv/12/cvterm/edit/2789
Pretty sure the arguments are off by and is easily corrected by shifting the page arguments down for `  $items['admin/tripal/loaders/chado_vocabs/chado_cv/%/cvterm/edit/%']`

bradfordcondon:
@shawnawsu has verified this bug exists

bradfordcondon:
fixed

PR #243, opened 2018-01-04
Fix PDO exception on editing a cvterm

bradfordcondon:
Simple fix for #242 . 

Simply, editing a cvterm returns a PDO error.  This is because the argument given is one off from the wildcard.  Shifting the args down resolves the error.


spficklin:
Has been fixed already.

Issue #250, opened 2018-01-12
newick loader reroutes to node


bradfordcondon:
The newick loader menu link brings you to a node, so the loader should be upgraded to a tripal 3 loader.

After converting the loader I realized that there is a command-line drush command.  I assume we want to preserve this?  Can drush instantiate the loader class, or do i need to not put the methods to load the newick tree in the loader class so both the tripal 3loader and the drush loader can access them?

edit : duh yeah the drush command can just instantiate the loader...

final edit:  I was being foolish: the thing to do is to just keep the phylotree loading api where it was.  PR submitted.

bradfordcondon:
resolved in https://github.com/tripal/tripal/pull/252

PR #252, opened 2018-01-12
Migrate newick loader

bradfordcondon:
define newick tripalimporter and remove admin menu for old one  #250


spficklin:
All looks good, except it looks like you have some debugging code in the tripal_chado_field_storage_write_table() function.  Can you remove it?

bradfordcondon:
Was in a rush to push before the demo: sorry about that and thanks for the catch.

spficklin:
Much appreciated!  Thanks for converting this over to the new file importer class.

PR #254, opened 2018-01-16
Protocol field

bradfordcondon:
After review, this field was better than i remembered it.  (see #213 which i closed because i thought i had hardcoded the tables.  Turns out i didnt!)

It's a flexible field for chado tables with a `protocol_id` column.

this pr is part of #210 but not the only dropdown field we need...


spficklin:
Looks good. Thanks for submitting! I'm going to merge.  But the field is missing an elementInfo(), query() and queryOrder() functions.  Do you think you could add those and add another pull request?

Issue #257, opened 2018-01-18
[bug] fatal error:   Class 'WebServicesField' not found i


bradfordcondon:
I think the latest patch @spficklin fixing the jobs queue broke the jobs admin area at  `admin/tripal/tripal_jobs`. 

 Visiting there returns 

`Fatal error: Class 'WebServicesField' not found in /data/www/html/hardwoods_dev/drupal-7.52/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data.inc on line 9`

Same error also prevents cache from being cleared.

spficklin:
Fixed with pull request by @bradfordcondon

PR #258, opened 2018-01-18
[bugfix] unbreak cache: uncomment required files in web services

bradfordcondon:
Addresses bug #257 

Could not find the webservices field class because all of the web services.module includes were commented out.


spficklin:
Ooops... Thanks for catching that.

Issue #206, opened 2017-12-04
Remove function tripal_set_default_cv from tripal_chado.cv.api.inc


shawnawsu:
Per the conversation in the closed pull request, https://github.com/tripal/tripal/pull/201, we need to remove function tripal_set_default_cv (line 1617 of tripal/trupal_chado/api/modules/tripal_chado.cv.api.inc) as it's no longer used and we do not want to continue with the use of the tripal_cv_defaults table.

bradfordcondon:
note #200 duplicate issue , I'll just close that one

spficklin:
Yes, this should be removed the Tripal v3 but it should get moved back to the Legacy API for backwards compatibility.

spficklin:
Closing as @bradfordcondon fixed this with a pull request.

PR #259, opened 2018-01-18
move tripal_set_default_cv to legacy and remove from tripal_chado

bradfordcondon:
Cut and paste job to resolve issue #206 .  I also added the in-code ingroup parameter to match the legacy cv group.


spficklin:
Thanks for tackling that one!

bradfordcondon:
maybe more of a "light shove" than a tackle.  But no problem :)

Issue #251, opened 2018-01-12
suggestion: issue template


bradfordcondon:
At the first day of the user's meeting, many people requested the ability to add more information to their issues: they wanted to tag them.   Abdullah suggested writing a github issue template instead.  
Check out the issue template [on the react native repo](https://github.com/facebook/react-native/blob/master/.github/ISSUE_TEMPLATE.md) for example.  Basically we just stick a markdown file in `.github/ISSUE_TEMPLATE.md` and we're in business.

Things the template could ask users to do:

* name issue with [Question] or [bug]
* include a v3 or v2 tag
* redirect issues to a submodule ....

bradfordcondon:
I created an example ticket template on my tripal_alchemist page.

[heres when you create a new issue]( https://github.com/statonlab/tripal_alchemist/issues/new)
[heres a blank ticket](https://github.com/statonlab/tripal_alchemist/issues/11) i just filled out

spficklin:
I'm a bit worried if we complicate adding of issues it may deter folks who may be overwhelmed by a template.  But I don't want to discount the idea.  @laceysanderson any thoughts?

laceysanderson:
I think it could be awesome but we would have to be careful not to overwhelm them! I would be tempted to do something like this:
```
-- Please enter the following information (if known) --
Tripal Version:
Drupal Version:
PostgreSQL Version:
PHP Version:
Operating System:

-- Please describe your issue here :-) 
    Some information you might want to include: the page you're seeing the issue on, 
    what behaviour you're experiencing versus what you expect, steps to reproduce... 
    really anything you think might best help us help you! --

-- PS. Also feel free to ask questions, start discussions, suggest improvements
           We welcome any and all feedback! --
```

This way we can give them suggestions which might speed up the debugging process but we're very friendly and clear that all information is optional :-) I would actually be more likely to file an issue if I came across something inviting rather then an empty box. 

Thoughts @spficklin? I agree we don't want to make it more complicated but at least with these templates they can completely ignore it if they want.

spficklin:
Okay, I can be persuaded.   Can we adapt the tripal_alchemist template with Lacey's suggestions and add this text at the top


INSTRUCTIONS: The following template is meant to provide the information that will help other Tripal developers diagnose and reproduce your issue.  Follow the directions below to complete the template.  If the template is not appropriate for your issue you may remove it and describe your issue.

bradfordcondon:
[...]

PR #263, opened 2018-01-20
issue template first draft

bradfordcondon:
per discussion #251


spficklin:
I like it.  Let's give it a go!

PR #273, opened 2018-01-29
Fix typo that results in 'Use of undefined constant tripal_feature' error

almasaeed2010:
This request fixes an error that's caused by a type.

Thanks!


spficklin:
Good catch. Thanks for the fix

Issue #268, opened 2018-01-24
Coding sequence (CDS) from alignment not correct


srobb1:
### System information
* Tripal Version: 2
* Drupal Version: 7
* PostgreSQL Version: 9.2.14
* PHP Version:  5.4.16 (I know I need to upgrade :( )
* Operating System: CentOS7

### Issue description
Has anyone noticed that the CDS sequence derived from the alignment of the mRNA to the genome is not correct? A user just pointed this out to me.

The CDS derived from the alignment is 4 of the correct CDSs concatenated. I am not sure this happens all the time, but I had a user point out a few records where this occurs.

### general info about this feature.
This mRNA has 4 CDS regions (maybe this has something to do with the 4x the whole CDS is repeated?) -- Edited: Okay, I just checked this theory. I found a mRNA with 2 CDS regions, this feature's displayed "CDS derived from alignment" is the total CDS concatenated 2x. 

I wanted to make sure I did not have extraneous record in chado so I did some queries.
I have only one uniqueID for my CDS, but it is in the featureloc table 4x, just as I expected.
```
drupal=# select cvt.name, uniquename, feature_id from feature f, cvterm cvt where uniquename like 'Nv3_00000721%' and type_id = cvt.cvterm_id;
      name       |           uniquename            | feature_id
-----------------+---------------------------------+------------
 mRNA            | Nv3_00000721-RA                 |    6969777
 gene            | Nv3_00000721                    |    6969776
 polypeptide     | Nv3_00000721-RA-protein         |    7128590
 CDS             | Nv3_00000721-RA:cds             |    6969784
 five_prime_UTR  | Nv3_00000721-RA:five_prime_utr  |    6969783
 exon            | Nv3_00000721-RA:3               |    6969780
 exon            | Nv3_00000721-RA:4               |    6969781
 exon            | Nv3_00000721-RA:2               |    6969779
 exon            | Nv3_00000721-RA:1               |    6969778
 exon            | Nv3_00000721-RA:5               |    6969782
 three_prime_UTR | Nv3_00000721-RA:three_prime_utr |    6969785
(11 rows)

drupal=# select * from featureloc where feature_id = 6969784;
 featureloc_id | feature_id | srcfeature_id |  fmin   | is_fmin_partial |  fmax   | is_fmax_partial | strand | phase | residue_info | locgroup | rank
---------------+------------+---------------+---------+-----------------+---------+-----------------+--------+-------+--------------+----------+------
[...]

PR #274, opened 2018-01-30
added check for already processed feature_ids for subseq concat

srobb1:
Hi,
Here is my fix for my overly concatenated CDS issue #268  that is caused by having one CDS feature with multiple featureloc locations. This seems to work for the CDS and not alter the mRNA sequence. I am not sure how this will affect other feature types. Not sure what other feature types to look at or if I have any that might be good to check in my db.

Sofia


spficklin:
Thanks @srobb1 for submitting this fix!

Issue #277, opened 2018-02-01
Tripal 3 sio__Annotation field


bradfordcondon:
The sio__annotation field builds an accession linkout using the db info but then doesnt include it in the table.  The result is there are no linkouts in the table.

```
  // Build the accession.
        $accession = $dbxref->db_id->name . ':' . $dbxref->accession;
        if ($dbxref->db_id->urlprefix) {
          $accession = l($accession, tripal_get_dbxref_url($dbxref), array('attributes' => array('target' => '_blank')));
        }

```

This is an obvious fix that I'll PR.

I have another question though.  In my database, all of my GO and IPR terms are loaded in without defenitions.  Instead, there is a descriptive name.

<img width="828" alt="screen shot 2018-02-01 at 2 51 24 pm" src="https://user-images.githubusercontent.com/7063154/35699940-7fd0dda4-075f-11e8-9bcf-cfdff44fe46a.png">

I'm not sure why this is the case: is it true for everyone else?  If so, can we alter this field to display the name rather than the description?

Check out a [feature page](http://160.36.205.61:8095/feature/QlobV10scaffold115670000020-RA) to see what i mean.  The go and IPR terms are there, but without a human readable name (which is present).  None of my terms have definitions, just names.

spficklin:
That's odd that you have no definitions. If you search for the terms via Admin > Tripal > Data Loaders > Chado Vocabularieis > Controlled Vocabulary Terms do you see the definitions there?

bradfordcondon:
it was null in Chado, so no.  we've since overwritten the definition with the name.

bradfordcondon:
I think we can close this, given that its pretty fixable and hackable.  Just wasn't sure if others also had lots of terms without definitions for some vocabularies.

PR #279, opened 2018-02-01
use accession that was built with linkout

bradfordcondon:
please see #277 . But I'd also like to address the name vs definition issue. or at least determine if that part is an HWG issue...


bradfordcondon:
Hi all/ @spficklin

Just checking in on this, I keep having to patch it on HWG.

Questions?

Thanks

spficklin:
Thanks for the fix!

PR #283, opened 2018-02-07
Prevent tripal_cron_notification job from getting duplicated

almasaeed2010:
We noticed the cron jobs are getting duplicated again due to API call changes. This PR fixes that.

Thanks!


almasaeed2010:
Already fixed in master ... please ignore

spficklin:
We made a mistake and didn't make the tripal_add_job() call backwards compatible. This fix that you noticed was correcting that.   We made an adjustment to the Tripal Galaxy module as well to support this change.  Sorry about the fix causing an issue for you.

PR #294, opened 2018-02-20
Allow cvalue id in chado property

bradfordcondon:
The [next version of Chado](https://github.com/GMOD/Chado/pull/54) will include a `cvalue_id` column for each prop table.  Supporting this is a pressing goal for us at HWG so I've already added support to our [Biomaterial Importer](https://github.com/tripal/tripal_analysis_expression/pull/198).  For those unaware, this means that instead of only support free-text for the property value in the `value` column, you can instead supply a cvterm for the value as well as the property type.

However we use `chado_insert_property` in our module.  So, this PR allows `cvalue_id` support.  It has no effect on calls to this method that don't supply a `cvalue_id.


bradfordcondon:
Sorry, not sure why all our other commits show up on this PR?  We obviously dont want that commit history in this PR just the code change (which is minor).

laceysanderson:
I think this is because your statonlab/7.x-3.x has diverged from core tripal/7.x-3.x. Thus when you forked this PR branch you ended up with your divergent commits as well as the small fix you wanted to submit...

bradfordcondon:
😿  I can see how to fix it on the tripal repo side:  Merge the pull request using "Squash and Merge" instead of the default "Merge Pull Request.


Not sure how to fix it on the statonlab fork side though- I think rebasing is the answer but wasn't successful.

laceysanderson:
Maybe it would be easiest to create a new branch that you never commit to and matches Tripal Core. That way you can always branch from it making clean pull requests? Something like (untested):
```bash
# Create clean 7.x-3.x branch
git remote add tripal_core https://github.com/tripal/tripal.git
git fetch tripal_core
git checkout -b core-7.x-3.x tripal_core/7.x-3.x

# Create your pull request branch from it
git checkout core-7.x-3.x
git checkout -b [pull request branch]
```

bradfordcondon:
a cleaner PR is made on #295 

Need to be mindful of which master branch (core not staton) i base my branches off of for PRs to core i guess.

thank you lacey!

PR #295, opened 2018-02-20
add cvalue support for properties

bradfordcondon:
The [next version of Chado](https://github.com/GMOD/Chado/pull/54) will include a `cvalue_id` column for each prop table.  Supporting this is a pressing goal for us at HWG so I've already added support to our [Biomaterial Importer](https://github.com/tripal/tripal_analysis_expression/pull/198).  For those unaware, this means that instead of only support free-text for the property value in the `value` column, you can instead supply a cvterm for the value as well as the property type.

However we use `chado_insert_property` in our module.  So, this PR allows `cvalue_id` support.  It has no effect on calls to this method that don't supply a `cvalue_id`.

Thanks @laceysanderson for the easy workaround cleaning up my PR #294

### DIFF
--- a/tripal_chado/api/tripal_chado.property.api.inc
+++ b/tripal_chado/api/tripal_chado.property.api.inc
@@ -169,6 +169,7 @@ function chado_insert_property($record, $property, $options = array()) {
   $cv_id       = array_key_exists('cv_id', $property) ? $property['cv_id'] : '';
   $value       = array_key_exists('value', $property) ? $property['value'] : '';
   $rank        = array_key_exists('rank', $property) ? $property['rank'] : 0;
+  $cvalue_id   = array_key_exists('cvalue_id', $property) ? $property['cvalue_id'] : '';
 
   $update_if_present = array_key_exists('update_if_present', $options) ? $options['update_if_present'] : FALSE;
   $force_rank = array_key_exists('force_rank', $options) ? $options['force_rank'] : FALSE;
@@ -249,6 +250,20 @@ function chado_insert_property($record, $property, $options = array()) {
       array('%property' => print_r($property, TRUE)));    return FALSE;
   }
 
+
+    //Check that the cvalue property exists
+    if ($cvalue_id){
+        $term = chado_select_record('cvterm', array('cvterm_id'), array('cvterm_id' => $cvalue_id), $options);
+       if (!$term or count($term) == 0) {
+            tripal_report_error('tripal_chado', TRIPAL_ERROR, "chado_insert_property: " .
+                "Cannot find the term for the property value described by: %property.",
+                array('%property' => print_r($property, TRUE)));
+           return FALSE;
+    }
+    $values['cvalue'] = $cvalue_id;
+  }
+
+
   // Get the foreign key for this property table
   $table_desc = chado_get_schema($base_table . 'prop');
   $fkcol = key($table_desc['foreign keys'][$base_table]['columns']);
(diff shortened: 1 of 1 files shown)


spficklin:
Hey @bradfordcondon What about the chado_linker__prop field.  That handles inserts of field data attached to bundles.  Were you planning to deal with that as well?

Also, did @scottcain approve of that addition (i.e.) has it already been merged in to the master branch?

And finally, would you mind adding a comment to your code that indicates the status of that cvalue_id field (that it's in Chado for a future release) but it's being supported now for early adopters.

bradfordcondon:
Hi Stephen: Thanks for the feedback.

>What about the chado_linker__prop field

Certainly both the widget and the formatter (and probably therefore the core load) need to be updated.  I wasn't planning on tackling that in this PR but I could if you'd like.  I think it's not as easy as I'd like to recycle the existing CVterm AJAX lookup box for use with Cvalue but I'd have to take another look at it. 


>did scottcain approve of that addition (i.e.) has it already been merged in to the master branch?

No and No.  however I think of all the changes this is the one thats most likely to go into the next release as you did the work creating the diff files (it will be the first PR to be merged).

>And finally, would you mind adding a comment to your code that indicates the status of that cvalue_id field (that it's in Chado for a future release) but it's being supported now for early adopters.

Absolutely. Will do tomorrow am

bradfordcondon:
hi @spficklin - trying to implement the cvterm lookup for the widget is getting complicated. 

We need to create dynamic callbacks and wrappers so that the cvterm lookup will work for each property.  I can see how to do that for the wrapper:

```

    $widget['cvalue_term']['select_button'] = [
      '#type' => 'button',
      '#value' => t('Lookup Term'),
      '#name' => 'select_cvterm',
      '#ajax' => [
        'callback' => "cvalue_ajax_callback",
        'wrapper' => "chado-' . $field_table .'_' . $type_id . '__cvalue_term",
        'effect' => 'fade',
        'method' => 'replace'
      ],
    ];
```

But what about the callback?  I cant use an anonymous function.  Abdullah has suggested i create a new drupal ajax event and write the js to handle the event.  

What do you think?  Is there an easier way?

Maybe we can make a seperate issue/PR to discuss?

spficklin:
I'm not sure I understand the issue with the AJAX in the widget.  Perhaps that could be opened as a new issue and we can deal with that separate.  I agree with you that this will be added to Chado in the future, but my preference would be to wait to merge this until it's at least in the master branch of Chado.  Updates to Chado can move very slowly and if we add anything to the core of Tripal I'd like a bit more insurance that it's not just another side feature that never comes to fruition.  It looks like we had a pull request issued:  https://github.com/GMOD/Chado/pull/53, but it was closed with no reason why. Perhaps we should reissue that pull request. Get the changes merged into master of Chado and then we can merge this into Tripal . @laceysanderson do you agree?

bradfordcondon:
That PR was duplicate with https://github.com/GMOD/Chado/pull/54.  

>Updates to Chado can move very slowly and if we add anything to the core of Tripal I'd like a bit more insurance that it's not just another side feature that never comes to fruition

This is reasonable.  I guess I can just merge into the staton branch for now.

That said... if this feature doesn't get implemented in Chado its going to be a big problem for me because supporting cvterms in this way is one of the PGRP/DIBBs goals, and its not something that is a neat custom table or anything like that.

Scott said he prefers a slower pace for a CHADO release (as a feature not a bug)... you'd know better than I if we can expect this sooner or later.


I'll create a seperate issue for the widget then so I can stick my code somewhere and not have it gunk up this fairly straightforward change.

bradfordcondon:
happy to report cvalue support has been merged into master for Chado, freeing this PR to be merged I believe?

spficklin:
merged!

Issue #208, opened 2017-12-06
Cron issues


spficklin:
In the past for Tv3, the cron would perform a few checks which could end up slowing the site down as Drupal will sometimes initiate the cron as user's visit the page.  So, we moved some of those long running tasks into Tripal jobs and now the cron just submitts them.  But now they pile up in the job queue!   I had 97 jobs to run this morning on my dev site.  

We should adjust that code to not submit the cron jobs again if they are already in the Tripal job queue.

bradfordcondon:
Any update on this?
i had 25,000 jobs this evening on the dev site!  These are the three jobs:

```
tripal_expire_collections()
tripal_galaxy_check_status()
tripal_tripal_cron_notification()
```

spficklin:
Yikes, 25K jobs!  I just committed a fix that should stop this from happening.  If the job already exists in the queue it won't add it again.  When adding a new job you can override this behavior, if you want, by setting the ignore_duplicate parameter to FALSE when using the tripal_add_job() function.    I've tested it and it works, so I'll close the issue, but if you still have a problem please reopen.

bradfordcondon:
Hi Stephen,

Looking at this with @almasaeed2010 ... it looks like manually calling `tripal_add_job()` wont pass the `ignore_duplicate` parameter.

This is because when the job is created:
```
  try {
    $job->create(array(
      'job_name' => $job->getJobName(),
      'modulename' => $job->getModuleName(),
      'callback' => $job->getCallback(),
      'arguments' => $arguments,
      'uid' => $user_id,
      'priority' => $job->getPriority(),
      'includes' => $includes
    ));
```

the ignore duplicate would be in $arguments.

But looking at the create method, it checks for ` $details[‘ignore_duplicate’]` and not   `$details[‘arguments’][‘ignore_duplicate’]`

spficklin:
Good catch.  I forgot to extend that argument to the tripal_add_job() function.  I've fixed it.   Although the code in the TripalJob.create() method should be $details['ignore_duplicate'] as the 'arguments' key is meant for arguments that get passed to the callback function and ignore_duplicate is meant for the create function.  This time I'll leave things open until you're happy with it.

bradfordcondon:
looks good to me, thanks Stephen.  I can't close the issue myself anymore lol so please go ahead and close it when you can ready.

e- oops, it was never mine, go figure.  anyway, we're happy so close away

laceysanderson:
[...]

PR #306, opened 2018-03-12
Issue #208: Fix ignoring of duplicates for cron jobs.

laceysanderson:
There was a parameter shift when using tripal_add_jobs() to add the cron jobs. Specifically, the `priority` was missing which caused `ignore_duplicates` to use it's default value. When this was changed for backwards compatibility this error resurfaced.

This pull request sets the `priority` to `1` (the lowest priority) and thus shifts the parameters so `ignore_duplicates` is `TRUE`. This is assuming that the `uid` should be set to `1` to ensure these jobs are submitted as the administrator user.


spficklin:
Thanks Lacey!

PR #307, opened 2018-03-13
Collections

spficklin:
This pull request is to merge in code to support multi-site data collections, including creation of downloadable files (CSV, TAB, FASTA) that are mashups of data from multiple sites.

The file creations are really slow still and that needs looking into.  It would take days to build a data collection with 100,000 remote genes.  That's not good enough, but it is functional as far as I have tested.


spficklin:
Some further info... it took approximately 3 minutes to create a data collection of about 5000 "cached" genes local to my machine.  If the records are not cached it takes longer....

bradfordcondon:
* The "contains" parameter is case sensitive by default.  Not a big fan.
* I currently end up with empty collections with no supported file types creating collections. Let's hold off on merging until we can atl east determine if this is an error on my end.

spficklin:
What 'contains' parameter do you mean @bradfordcondon

spficklin:
Okay, I fixed the issue with collections not being created with the Data Search Views.  Also, I added some code to the unit test case to get it to at least run, but we really should plan out how to organize these tests.  I was going to write a few tests for data collections but I have to have data in the site and so we'd need to pre-populate a dataset.  So, I think we need to write unit tests for the GFF3 and FASTA loaders first.  Once that's done then we can use that data loaded to do the unit tests on data collections.

bradfordcondon:
>contains parameter

shoot im sorry i realize tahts for the site search, not related to collections.


>testing organization and datasets

First of all I totally agree we need to discuss our plan for organizing tests.

Regarding data for tests.  I had originally envisioned using dockers pre-populated with data as an "official test docker".  However, I've come to believe that best practice is for each test suite is to insert and remove test data in the buildUp/tearDown (which gets run at the start/end of each test in the class).  Because you dont know what order tests will run in, relying on data inserted by a different class isn't great.  Additionally, if a FASTA loader didnt remove features the first run through, then the test would probably fail the second run through because the features were already in the database.  This wouldnt happen on Travis, but it would happen for local testing.

Anyway obviously we need to discuss more and we'll do so elsewhere.

Thanks for pushing changes I'll test again.

tripal/includes/TripalFieldDownloaders/TripalFieldDownloader.inc
```diff
@@ -4,67 +4,159 @@
...
+      ->fetchObject();
 
     // Make sure the user directory exists
+    $user = user_load($this->collection->uid);
     $user_dir = 'public://tripal/users/' . $user->uid;
+
+    // Set the collection ID of the collection that this downloader will use.
+    $this->collection_id = $collection_id;
+    $this->outfile = $user_dir . '/' . $outfile_name;
+    $this->selected_fields = $selected_fields;
```
bradfordcondon: `$selected_fields` not defined

tripal/includes/TripalFieldDownloaders/TripalFieldDownloader.inc
```diff
@@ -76,7 +168,36 @@ abstract class TripalFieldDownloader {
...
+   *   A field info array.
+   */
+  public function isFieldSupported($field, $instance) {
+    $field_name = $field['field_name'];
+    $field_type = $field['type'];
+
+    // If a field is a TripalField then check its supported downloaders.
+    if (tripal_load_include_field_class($field_type)) {
+      $formatters = $field_type::$download_formatters;
+      if (in_array($formatter, $settings['download_formatters'])) {
```
bradfordcondon: `$formatter`, `$settings` not defined?

tripal/includes/TripalFieldDownloaders/TripalFieldDownloader.inc
```diff
@@ -91,58 +212,156 @@ abstract class TripalFieldDownloader {
...
+    // But, only add the usage if it doens't already exists.
+    $usage = file_usage_list($file);
+    if (array_key_exists('tripal', $usage)) {
+      if (!array_key_exists('data-collection', $usage['tripal'])) {
+        file_usage_add($file, 'tripal', 'data-collection', $fid);
+      }
+    }
+
+    if ($job) {
+      $job->setItemsHandled($num_handled);
```
bradfordcondon: `$num_handled` not defined?

tripal/includes/TripalFieldDownloaders/TripalFieldDownloader.inc
```diff
@@ -152,6 +371,113 @@ abstract class TripalFieldDownloader {
...
+
+    foreach ($this->fields as $site => $bundles) {
+      foreach ($bundles as $bundle_name => $bundle_fields) {
+        foreach ($bundle_fields as $field_id => $info) {
+          $instance = $info['instance'];
+          $accession = $instance['settings']['term_vocabulary'] . ':' . $instance['settings']['term_accession'];
+          if (!array_key_exists($accession, $this->printable_fields)) {
+            // Only include fields that support this downloader type in
+            // or list of printable fields.
+            if ($this->isFieldSupported($field, $instance)) {
```
bradfordcondon, line 474: $field not defined.  `$site` maybe?

spficklin:
Okay, 3rd time's the charm.... Fixed issues with messages on deleting a collection, PHP 7.2 warnings and the items pointed out by @bradfordcondon

tripal/includes/TripalEntityCollection.inc
```diff
@@ -400,73 +455,124 @@ class TripalEntityCollection {
...
+
+          // Write the same entity to all the formatters that are supported.
+          foreach ($formatters as $class => $formatter) {
+            //if ($class == 'TripalTabDownloader') {
+              $formatter->writeEntity($entity, $job);
+            //}
           }
         }
+        if ($job) {
+          $job->setItemsHandled($num_handled + count($slice));
```
bradfordcondon, line 566: I don't think this $num_handled is defined

bradfordcondon (approved):
tested local collections a bunch and it all looks good!

Issue #228, opened 2017-12-15
PDO Exception errors in Tripal 3.0 dev


risharde:
Running on Drupal 7.56 and Tripal 7.x-3.0-rc2+63-dev

The following error message is displayed when trying to edit ANY Chado Views Integration table choice

PDOException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "views-integration" LINE 1: SELECT * FROM tripal_views WHERE setup_id = 'views-integrati... ^: SELECT * FROM {tripal_views} WHERE setup_id = :setup; Array ( [:setup] => views-integration ) in tripal_chado_views_integration_form() (line 244 of /var/www/Drupal/sites/all/modules/tripal/tripal_chado_views/includes/tripal_chado_views_integration_UI.inc).

spficklin:
Okay. Thanks for pointing this out. We'll get it fixed.

spficklin:
You beat us to it! Your pull request has been applied. Thanks for fixing it.   I'm closing this out but reopen if you still have issues.

PR #312, opened 2018-03-15
fix the form params for chado/views-integration

isumoc:
This is a fix for https://github.com/tripal/tripal/issues/228

Right now the views-integration form is unusable - it passes the wrong param throwing a PDO error.


spficklin:
Thanks for the fix!

Issue #305, opened 2018-03-08
Sudden issue with contact record not being created + error


risharde:
HI guys and @spficklin , suddenly I have come across a problem regarding creating a contact. Cannot track this back to any change made. When creating the contact by entering data and clicking submit, I get the following error:
```
Could not save the entity: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "" LINE 1: ...chado_bio_data_8 (entity_id, record_id) VALUES ('65297', '') ^
Cannot save entity
```
After some debugging, I have traced it back to the file:
sites/all/modules/tripal/tripal_chado/includes/tripal_chado.field_storage.inc
```
Line:
$base_record_id = tripal_chado_field_storage_write_table($base_table, $values, $base_table);
```

**$base_record_id is blank**. After some further debugging, it seems
$values is empty for $base_table which shows up as 'contact'

Further backtracking shows that the following array does not contain any data which I think might be causing the problem. Cannot understand why this is happening.
$field_vals[$base_table] returns empty

So in essence $field_vals['contact'] returns empty which I think is why it's not able to create a $base_record_id

Any assistance would be appreciated? This was working before and all of a sudden, it stopped, couldn't trace it back to a custom module or anything like that. Currently contact submission is crippled for me :(

Please help?
Thanks

bradfordcondon:
You're creating contacts directly (ie from add tripal content -> contact) right? Or is this a field widget of some sort?  If thats the case are you able to create contacts directly?

I'm able to create contacts in this manner on current master.  But your error looks to me like a misconfigured field - possibly the column id for contact (contact_id) wasn't specified?

risharde:
Hey @bradfordcondon thanks for your contribution, yes I am trying to create directly from add tripal content -> contact
I thought so as well but no field seems to be misconfigured since no changes were made to the fields since prior to working
A bit stumped on what could have done this

spficklin:
Do you have a custom field that works with records in the contact table?

risharde:
Hey @spficklin , no custom field that works with records in the contact table. It's like I just woke up today to test it and it just spurts the error. It seems as though because it cannot get the $values[] array populated, nothing gets created in the contact table and so no record id is returned for additional inserts

spficklin:
What is bio_data_8 on your site... is it the contact content type?

risharde:
@spficklin Correct!

spficklin:
[...]

PR #313, opened 2018-03-15
Fixed issue #305

spficklin:
This is a fix reported for issue #305.  Chado properties added to field settings were not being preserved when the field settings were updated using the UI.  This caused problems inserting records afterwards.


bradfordcondon:
Tested the fix, works as expected.  Code looks reasonable.  👍 

edit: woops guess i should say this in the "review" box

bradfordcondon (approved):
Tested the fix, works as expected. Code looks reasonable. 👍

spficklin:
Thanks @bradfordcondon I've merged it.

PR #316, opened 2018-03-22
Upload file size 2 g limit

MingChen0919:
Disable uploading files larger than 2G, this fixes an issue from the tripal galaxy module: https://github.com/tripal/tripal_galaxy/issues/34.

### DIFF
--- a/tripal/theme/js/TripalUploader.js
+++ b/tripal/theme/js/TripalUploader.js
@@ -44,6 +44,7 @@
 
       if (file.size > 2147483648) {
         alert('Please upload a file less than 2048M (2G).');
+        return null;
       }
 
       // Make sure the file type is allowed.  If there are no file types
--- a/tripal/theme/js/TripalUploader.js
+++ b/tripal/theme/js/TripalUploader.js
@@ -41,7 +41,11 @@
       var i = options['i'];
       var url = options['url'];
       var self = this;
-      
+
+      if (file.size > 2147483648) {
+        alert('Please upload a file less than 2048M (2G).');
+      }
+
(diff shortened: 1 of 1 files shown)


spficklin:
I don't think we should have a hard-coded limit at 2GB upload.  This Tripal file uploader is meant to handle really large files and uses HTML5 chunking to do that.  The issue with chunks being lost on issue tripal/tripal_galaxy#34 needs to be fixed but we can probably do that with fixes to the JS code for the uploader.  I don't think we need to set a limit just yet.

mestato:
I don't think this was a well described issue, at least as I initially requested it (sorry Ming). The goal as I requested in https://github.com/tripal/tripal_galaxy/issues/34 was to have an admin option where they could first, select if they want to have a limit (or not), and then if they do want a limit, set what that limit should be. This could be important for reasons other than the chunking, it is just a nice admin option if you are trying to manage limited file storage on a tripal site and a galaxy site.

spficklin:
Yes, I agree that an admin option would be good.  We had discussed moving some of the file limiting code from the Tripal Galaxy module down to Tripal because it would be useful in general.  Perhaps the solution is to do that? If so,  I think we could close this out again and create a new issue for that work?

MingChen0919:
I agree this PR is not a good solution. I'll work on a better solution for the file upload limit. I am closing this one.

spficklin:
@MingChen0919 & @mestato I'm currently in the process of moving the quota and files admin tabs out of the galaxy module and into Tripal.  Once that's done we can add another setting to set a default upload limit.

PR #329, opened 2018-04-09
Fix path in module load function in phylotree importer

almasaeed2010:
The phylotree importer gives an error when running the job caused by an incorrect path to the parser file. This PR fixes the issue by correcting the path.

This is the error that happened before the fix:
```
NOTE: Loading of this tree file is performed using a database transaction. 
If the load fails or is terminated prematurely then the entire set of 
insertions/updates is rolled back and will not be found in the database

WD php: Error: Call to undefined function tripal_phylogeny_parse_newick_file() in tripal_phylogeny_import_tree_file() (line 863 of          [error]
/var/www/html/sites/all/modules/tripal/tripal_chado/api/modules/tripal_chado.phylotree.api.inc).
Cannot modify header information - headers already sent by (output started at phar:///usr/local/bin/drush/includes/output.inc:40)           [warning]
bootstrap.inc:1486
Error: Call to undefined function tripal_phylogeny_parse_newick_file() in tripal_phylogeny_import_tree_file() (line 863 of /var/www/html/sites/all/modules/tripal/tripal_chado/api/modules/tripal_chado.phylotree.api.inc).
Drush command terminated abnormally due to an unrecoverable error.   
```

Thanks!

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
@@ -859,7 +859,7 @@ function tripal_phylogeny_import_tree_file($file_name, $format, $options = array
     // Parse the file according to the format indicated.
     if ($format == 'newick') {
       // Parse the tree into the expected nested node format.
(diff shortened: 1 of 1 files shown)


spficklin:
Thanks @almasaeed2010 !

PR #335, opened 2018-04-12
Update patch for drupal views 7.x-3.19

abretaud:
An update for the patch to drupal views as there was a new release a few days ago


laceysanderson:
Awesome! Tested it: applies cleanly and works great 👍 Thanks!

PR #346, opened 2018-04-24
Add db transaction to tripal_update_cvtermpath

almasaeed2010:
We've added db transaction support in the cvtermpath builder function to avoid having to clear the table when an error occurs during the building process.

Thanks!

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.cv.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.cv.api.inc
@@ -403,10 +403,10 @@ function tripal_update_cvtermpath($cv_id, $job_id = NULL){
     // Rollback any database changes
     $transaction->rollback();
     throw $e;
+  } finally {
+    // Set the database back
+    chado_set_active($prev_db);
   }
-
-  // Set the database back
-  chado_set_active($prev_db);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


laceysanderson:
Looks like a good PR although I'm not the person to test it :-) I'm not sure where the addition of the "in progress" tag came from -maybe my ZenHub adds it automatically?

almasaeed2010:
I should add that I have ran the script on cellular_component and it didn't cause any problems (takes about 6 hours to complete the test unfortunately)

Thanks!

spficklin:
This looks good to me.  Can you make one adjustment non-related to your fix but since you're doing a pull request on the function?  The line "chado_set_active($prev_db);" appears on line 405, in the catch block and on line 411.  I think we only need it on line 411.  Would you mind adjusting your pull request to remove that entry on line 405?

almasaeed2010:
Sorry for the delay @spficklin! I removed the extra call in the latest commit.

Thanks!

almasaeed2010:
Come to think about it, it is re-throwing the error so it's in fact needed in both areas or in a `finally {}` statement otherwise it won't execute since throwing an error stops the script. I'll add a `finally` which executes even in case of an error.

bradfordcondon (approved):

Issue #350, opened 2018-04-26
Update CV Term Path Job Fails


almasaeed2010:
### System information

* Tripal Version: 3 (7.x-3.x)
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description
We've encountered the following error while attempting to populate cvtermpath for cellular_component (part of GO).

```
Calling: tripal_update_cvtermpath(57)
Job execution failed: SQLSTATE[23505]: Unique violation: 7 ERROR:    [error]
duplicate key value violates unique constraint "cvtermpath_c1"
DETAIL:  Key (subject_id, object_id, type_id, pathdistance)=(56014,
56014, 52, 1) already exists.
```

Worth noting that biological_process didn't have this particular issue but that could be because we updated it before anything else.

We've attempted to clear cvtermpath table `TRUNCATE chado.cvtermpath` and rerunning the job but the issue persisted.

Please let me know if you have any questions.

Thanks!

almasaeed2010:
Fixed in PR #352

PR #352, opened 2018-04-26
Fix issue #350

almasaeed2010:
Hello,

This PR fixes issue #350 by preventing duplicate records from getting inserted.

Please let me know if you have any questions.

Thanks!

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.cv.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.cv.api.inc
@@ -739,21 +739,30 @@ function tripal_update_cvtermpath_loop_increment(
         }
       }
 
-      $query = db_insert('cvtermpath')
-        ->fields([
-          'object_id' => $origin,
-          'subject_id' => $child_id,
-          'cv_id' => $cv_id,
-          'type_id' => $type_id,
-          'pathdistance' => $depth,
-        ]);
+      $find_query = db_select('chado.cvtermpath', 'CVTP');
+      $find_query->fields('CVTP', ['subject_id']);
+      $find_query->condition('object_id', $origin);
+      $find_query->condition('subject_id', $child_id);
+      $find_query->condition('type_id', $type_id);
+      $find_query->condition('pathdistance', $depth);
+      $exists = $find_query->execute()->fetchObject();
 
-      try {
-        $rows = $query->execute();
-      } catch (Exception $e) {
-        $error = $e->getMessage();
-        tripal_report_error('tripal_chado', TRIPAL_ERROR, "Could not fill cvtermpath term: @error", array('@error' => $error));
-        return false;
+      if(!$exists){
+        $query = db_insert('cvtermpath')
+          ->fields([
+            'object_id' => $origin,
+            'subject_id' => $child_id,
+            'cv_id' => $cv_id,
+            'type_id' => $type_id,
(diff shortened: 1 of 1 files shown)


spectacledweb:
I'm worried this will keep actual records that should be inserted from being inserted. The test you've setup is: 
```
$find_query = db_select('chado.cvtermpath', 'CVTP');
$find_query->fields('CVTP', ['subject_id']);
$find_query->condition('object_id', $origin);
$find_query->condition('subject_id', $child_id);
$find_query->condition('type_id', $type_id);
$find_query->condition('pathdistance', $depth);
$exists = $find_query->execute()->fetchObject();
```

This isn't considering the cv_id and I believe there will be records that have the same object_id, subject_id, type_id, pathdistance but different cv_id's.

Did you test this against a large data set like gene ontology and ensure you're getting the same number of records without and without this check?

almasaeed2010:
I've based this query on the fact that cvtermpath has only 1 unique constraint defined as such:
```
UNIQUE (subject_id, object_id, type_id, pathdistance)
```

Having this constraint in place means that a record that has all of those but a different cv_id cannot get inserted.

Thanks!

almasaeed2010:
Sorry forgot to answer the tests question.

This is currently running on our dev server to repopulate Gene Ontology (cellular_component, biological_process, mollecular_function). It has finished both biological_process and cellular_component successfully (without any errors but I have no other proof that the content produced is accurate).

Thanks!

spectacledweb:
I am aware of that constraint, it was the reason why the GoSlim wouldn't load because that unique constraint was failing because the cv's were different but everything else was the same. I believe @spficklin put in a ticket about it (am I remembering that correctly?).

The errors don't worry me, it's the number of records that get written. If the GO fully loads, ie exactly the same number as a previously loaded GOs, is what I am curious about.

almasaeed2010:
I see your point. You think that chado might change that constraint so that ontologies like GoSlim can get inserted?

Unfortunately, we couldn't test the number of records before the fix since it kept failing to insert all available terms. However, if you have a fully loaded GO, I can provide the number of terms in cvtermpath for a given cv that you can compare. I'll add the numbers here once the run is completed.

almasaeed2010:
**cellular_component**
```
select count(*) from chado.cvtermpath
   inner join chado.cv ON cv.cv_id = cvtermpath.cv_id
   where cv.name = 'cellular_component';


 count  
--------
 260030
(1 row)
```

**biological_process**
```
select count(*) from chado.cvtermpath
inner join chado.cv ON cv.cv_id = cvtermpath.cv_id
where cv.name = 'biological_process';

  count  
---------
 1872499
(1 row)
```

**molecular_function**
```
select count(*) from chado.cvtermpath
inner join chado.cv ON cv.cv_id = cvtermpath.cv_id
where cv.name = 'molecular_function';
 count  
--------
 107222
(1 row)
```

spficklin:
I'm just checking on this PR.  Do we need more validation?

shawnawsu:
Sorry, I lost visibility on this ticket when I went away on vacation. 

I checked the numbers and they line up. So it should be good to merge.

spficklin:
PR merged!  Thanks @almasaeed2010

almasaeed2010:
Great! Thanks @shawnamspoor and @spficklin

Issue #310, opened 2018-03-13
Tripal style guide


bradfordcondon:
I would appreciate some sort of standard Tripal codebase style guide.  I use PHPstorm which lets me set a style and auto-format the code to it.

The XML hosted [here](https://raw.githubusercontent.com/Erikdekamps/phpstorm-drupal/master/Phpstorm.xml)  might be on the money.  Applying the formatter to random lines of code mostly finds minor formatting errors in the code itself (extra spaces etc).

This isnt just anal retentiveness (honest): my PRs that touch core files are full of minor formatting changes from running my linter.... with the linked style, I'm changing 5-10 lines of code per file so I'd like to establish a "correct" format.

shawnawsu:
Drupal has extensive coding standards: https://www.drupal.org/docs/develop/standards/coding-standards

Given Tripal is a contributed module it's best practice to stay within the guidelines set out in their documentation.

I use Visual Studio Code setup using this as a guideline: https://www.drupal.org/docs/develop/development-tools/configuring-visual-studio-code

This page lists most of the popular IDEs with links to suggested configurations: https://www.drupal.org/docs/develop/development-tools/configuring-visual-studio-code

bradfordcondon:
Thanks Shawna.  Looking over the XML I linked, that reflects the style guidelines (which if you are a hack like me you don't want to read and conform to, you want to import and not worry about it).   The IDE guide you link has some configuration files for some editors but not others .

I know the coding standards link is in the developer guide.  I'll add a link to specific style import profiles as people want them.

bradfordcondon:
Summarizing slack discussion:

The new Drupal coding standard conventions use `[]` instead of `array()`.

[Drupal coding standards also discourage pushing changes that are for style only.] <--- edit, contradictory with below... 

edit: after reading the [coding standards](https://www.drupal.org/docs/develop/standards/coding-standards) I have pulled out the two relevant sections:

>Note: Do not squeeze coding standards updates/clean-ups into otherwise unrelated patches. Only touch code lines that are actually relevant. To update existing code for the current standards, always create separate and dedicated issues and patches.



>Please note, short array syntax is unsupported in versions of PHP prior to 5.4. This means that Drupal 7 core and Drupal 7 contributed projects without an explicit PHP 5.4+ requirement must use long array syntax.


Therefore:

* Only the tripal 7.x-3.x branch and future branches should change from `array()` to `[]`
[...]

PR #361, opened 2018-05-01
format all inc php module install files using Drupal code standards

bradfordcondon:
Oh man, get excited people. I took the liberty of creating a patch for updating the tripal code base with the Drupal style standard.

This was done after the discussion at #310 where after reading the Drupal guidelines we discovered the recommendation is to *not* push style changes with code changes, but to have a seperate patch.

I only propose changes for php files IE `.install .module .inc .php` files.

The most obvious changes to me are the switch from `array()` to `[]`, comment line length, and consistent spacing between characters.

Can't tell if i'm doing the community a service or disservice with this ,but here we go!


laceysanderson:
Wow! This ones a doozy and will take a while to look at! Thanks for the PR though -it will be nice to have it cleaned up so these stop sneaking into other PRs ;-)

bradfordcondon:
The code cleanup is automated using phpstorm's Drupal code standards setting.  I could generate separate pull requests for individual modules etc if it makes it more manageable.

This may appear trivial but as I laid out in the issue, for me, this is actually fairly important in terms of making contributing code easier.  When my PRs are full of unrelated style changes, its harder for everyone.  I try to remember to *not* run my code formatter to prevent this, which means I end up contributing code with formatting errors.

laceysanderson:
To be more clear of my 👍, I will bump this PR up in my priority queue :-)

spficklin:
So do we have a plan for validating this?  There are so many changes.  @bradfordcondon are you all running Hardwoods on this update?

bradfordcondon:
not yet.  we've got conflicts so I'll probably regenerate the PR off the current 7.x-3.x.

I'd be happy to put hardwoods on this update as a proof of concept.  Also we know that the build/install works thanks to travis.

I am super open to discussion on how to go about making these changes.  Multiple smaller PRs one module at a time.  Possibly one style change at a time (this might be harder to do...)

spficklin:
What if we make these changes on code that has unit tests?  We need to come up with a plan to divide up work creating unit tests.  I'm sure we can all squeeze in a little bit of time here and there to write some unit tests.

bradfordcondon:
> I'm sure we can all squeeze in a little bit of time here and there to write some unit tests.

I submitted it this way because the recommendation is to patch for style only in a dedicated patch.

We can not follow that recommendation and write unit tests, converting the code we tested to the new standard as we create the tests.

bradfordcondon:
I don't think its reasonable to manually check the quality of this PR.  That was never my intention.   We have two choices:

* Trust the code formatter.

* Format the code to new standards piece by piece alongside unit testing.


Personally I think trusting the code formatter is pretty reasonable.  It only makes style changes.  We 
know it doesnt break basic functionality because the site builds etc in Travis and on a developer site.  

I can understand not wanting to make a huge PR like this while our code test coverage is nonexistent.  So if we want to close this and say let's add tests and update style at same time, so be it.  **However** that is going to make that process all the harder, since PRs coming in are going to include style changes and literal code changes.

laceysanderson:
I think that we can trust the code formatter for the most part. I would like to see this PR tested through
1.  A quick glance at some of the changes to make to see the scope
2. Some basic use on a production/cloned site. 

Most (if not all) errors produced by such changes should be syntax errors which would WSOD the site. So as long as a couple of us confirm that our site doesn't WSOD through basic use, I would be content enough to merge it.

A model for "basic use":
 - site installs (covered by testing! YAY! 🎉 )
 - create new Tripal Content Type
 - create/view/edit/delete Tripal content
 - run a tripal job
 - load a file
 - use a search view

Do you feel this is sufficient @spficklin?

PS. I'm volunteering myself to be one of the testers (noted here for accountability ;-) )

spficklin:
This PR is a bit old and we've already merged in a few others and now has conflicts.  Perhaps we should cancel this out and redo it later once the code has settled and we're close to release?

bradfordcondon:
It is very easy for me to re-generate this pull request since its just a command in my IDE.

I can resubmit the PR whenever.  Let's decide to what extent this PR should be tested/reviewed and when we've agreed on that I'll resubmit (since I dont want to resubmit under the condition of "when we're not busy" since I think thats a surefire way to never do it).

@spficklin did you agree with lacey's suggested list of functional testing?  I think abdullah/myself are of the opinion that building in CI is confirmation enough given its only style changes but we should do whats agreed upon.

laceysanderson:
Also, we should wait to submit until the current list of PRs is cleared. This one will cause significant merge conflicts with every other PR ;-p

spficklin:
With the release date coming up and not enough time to test this type of a change, can we wait until after the release to redo and apply this PR?

bradfordcondon:
yes agreed, I'll reapply after the next release.

PR #365, opened 2018-05-05
7.x 3.x small fixes

spficklin:
This is just a variety of small formatting fixes or update to code to improve error/warning handling.  None of these are tied to an issue, just fixes I made as I encountered problems.


tripal_chado/api/tripal_chado.custom_tables.api.inc
```diff
@@ -128,6 +128,18 @@ function chado_edit_custom_table($table_id, $table_name, $schema, $skip_if_exist
...
     $mview_id = NULL, $redirect = TRUE) {
+  
+  if (!$table) {
+    throw new Exception('Please provide a value for the $table argument to the chado_create_custom_table() function'); 
+  }
+  if (!$schema) {
+    throw new Exception('Please provide a value for the $schema argument to the chado_create_custom_table() function');
+  }
+  if ($schema and !is_array($schema)) {
+    throw new Exception('Please provide an array for the $schema argument to the chado_create_custom_table() function');
```
laceysanderson: Should we be throwing exception outside of entity/field classes? Most of the code using these APIs has not been updated to the `try {} catch{}` method so I worry this is much less friendly to Tripal Admin then the original `tripal_report_error()`. Plus, we could always update `tripal_report_error()` to throw an exception in the future.

laceysanderson (commented):
Maybe we should make a best practices rule about this? My recommendation would be: throw exceptions within classes, use tripal_report_error() everywhere else. Thoughts?

spficklin:
Good point @laceysanderson.  I have replaced the lines where it was throwing exceptions with calls to tripoal_report_error(). It's ready for review again.

laceysanderson (approved):
Looks good code-wise :-) I tested there is no WSOD but with this many small fixes it was hard to test functionality...

bradfordcondon:
looking at the file changes seems a-ok to me.

This PR in particular makes me wonder if we want to say minor bugfixes only require 1x approval?...

bradfordcondon (approved):

PR #366, opened 2018-05-05
7.x 3.x files management

spficklin:
This pull request adds a user file quota system to Tripal.  The new file uploader allows developers to include HTML5 uploaders on any page they want.  But it is beneficial to have file quotas in some cases.   This pull request provides that. It adds two new interfaces.  First, there is a new 'Uploads' tab on the user profile page.  This allows the user to see all of the files they have uploaded through any uploader provided by the uploader code (i.e. all of the TripalImporter implementations).   Second, the site admin can set the user quota globally or on a per-user bases by going to Administration > Tripal > User File Management.

Users will not be able to upload files if they exceed their quote.   This code was originally part of the Tripal Galaxy Module, but it made more sense to move this to the Tripal level so that quotas for users could be enforced for all uploads, not just those for the Galaxy workflow uploads.

### DIFF
--- a/tripal/api/tripal.files.api.inc
+++ b/tripal/api/tripal.files.api.inc
@@ -125,3 +125,67 @@ function tripal_get_files_stream($module_name = FALSE) {
 
   return $stream;
 }
+
+/**
+ * Formats a size (in bytes) in human readable format.
+ *
+ * Function taken from php.net
+ *
+ * @param $bytes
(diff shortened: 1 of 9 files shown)


bradfordcondon (approved):
I tested all the declared functionality with no problems.

spficklin:
Thanks @bradfordcondon for the review.

PR #368, opened 2018-05-07
7.x 3.x api docs

shawnawsu:
This update is very large but I've tested to make sure all the basic functions are up and running. It still needs to be reviewed closely though.

PR #369, opened 2018-05-07
update test names

bradfordcondon:
I updated the test names to match your API changes.  This should cause your branch to pass (it passed on my statonlab copy).


spficklin:
Thanks. I merged as I'm doing the review of #368

PR #373, opened 2018-05-14
Bug Fix: Ensure that empty chado property fields are not rendered.

laceysanderson:
Expected Behavior: When all form fields for a specific chado property are left blank, they shouldn't be rendered on the page.

How to Test: Edit any entity with chado properties. Leave the field blank and save. Clear the cache! You shouldn't see the label for that field rendered. :-)

### DIFF
--- a/tripal_chado/includes/TripalFields/chado_linker__prop/chado_linker__prop.inc
+++ b/tripal_chado/includes/TripalFields/chado_linker__prop/chado_linker__prop.inc
@@ -127,6 +127,10 @@ class chado_linker__prop extends ChadoField {
 
     // Ensure there are no values if there are no properties.
     // This is necessary to make sure the field is not rendered when there are no properies.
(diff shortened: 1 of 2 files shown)


spficklin:
Hey Lacey,  Thanks for the fix. Can you add one item to it?  I don't think you should have to remove the field from the entity object as this fix does in the load() function.  None of the other fields need to do that.  We're busy so I'm keen to just leave your fix as is rather than take time to dig down into it.  So, could you add a TODO comment to revisit that code at a future date?

laceysanderson:
I added the todo as requested 👍. The unset() in load() was done to make it easier to check that the field was empty in the formatter without making assumptions. That said, consistency is king so if the other fields don't do this, it's definitely worth revisiting at a later date :-)

spficklin:
Thanks!

Issue #332, opened 2018-04-10
Tripal API Documentation updates required


shawnawsu:
The following api docs need a bit more information describing them. Feel free to just reply to this ticket or create a pull request on the 7.x-3.x-api-docs branch.

- [x] tripal_chado.analysis.api.inc
/**
 * @defgroup tripal_analysis_api Chado Analysis
 * @ingroup tripal_chado_api
 * @{

 * @}
 */

- [x] tripal_chado.contact.api.inc
/**
 * @defgroup tripal_contact_api Chado Contact
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.cv.api.inc
/**
 * @defgroup tripal_chado_cv_api Chado CV
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.db.api.inc
/**
 * @defgroup tripal_chado_database_api Chado DB
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.feature.api.inc
/**
 * @defgroup tripal_feature_api Chado Feature
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.organism.api.inc
/**
 * @defgroup tripal_organism_api Chado Organism
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.phylotree.api.inc
/**
 * @defgroup tripal_phylotree_api Chado Phylotree
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.pub.api.inc
/**
 * @defgroup tripal_pub_api Chado Publication
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal_chado.STOCK.api.inc
/**
 * @defgroup tripal_stock_api Chado Stock
 * @ingroup tripal_chado_api
 * @{
 * @}
 */

- [x] tripal.chado.api:
/**
 * @file
 *
 * This file provides the Chado API: a set of functions for
 * interacting with Chado.
 */

/**
 * @defgroup tripal_chado_api Chado
 *
 * @ingroup tripal_api
 * The Tripal Chado API: a set of functions for
 * interacting with Chado.
 *
 */

- [x] tripal.chado_views.api:
/**
 * @file
 * API functions for Chado Views Integration
 */
/**
 * @defgroup tripal_chado_views_api Chado Views Integration
 * @ingroup tripal_chado_api
[...]

PR #377, opened 2018-05-22
Legacy API functions not returning their expected values

laceysanderson:
The recent API update (#332) caused issues when running Tripal in legacy mode. It turns out many of the deprecated functions no longer returned the expected result. 

Deprecated functions should always return the result of the new function. Even if they don't return an object it is important any return code (true/false/null) be returned because existing legacy code may depend on this.


shawnawsu:
Thanks Lacey!

PR #379, opened 2018-05-24
Fixes the placeholder name collision between old and new data in chado_update_record() function.

guignonv:
When using chado_update_record and trying to change a field that is also used to select the record, it does nothing. The place holders used by the query argument for both the selection and the updated value use the field name as place holder name while they should be different. For instance, if I'd like to update stock_cvterm.cvterm_id, the argument array will use the key ':cvterm_id' for both the selection value and the new value and therefore, the selection value will be replaced by the new value in the argument array and the corresponding record will not be found and not updated. This fix just prefixes the selection value place holder name with 'old_' in order to avoid name collision. It has been tested and appears to work correctly.

Co-authored-by: SheilaBarron <sheilabarron@me.com>
Co-authored-by: guignonv <guignonv@gmail.com>

### DIFF
--- a/tripal_core/api/tripal_core.chado_query.api.inc
+++ b/tripal_core/api/tripal_core.chado_query.api.inc
@@ -708,8 +708,8 @@ function chado_update_record($table, $match, $values, $options = NULL) {
       $sql .= " $field = NULL AND ";
     }
     else {
-      $sql .= " $field = :$field AND ";
-      $args[":$field"] = $value;
+      $sql .= " $field = :old_$field AND ";
+      $args[":old_$field"] = $value;
     }
   }
   $sql = drupal_substr($sql, 0, -4);  // get rid of the trailing 'AND'


laceysanderson:
Good catch! Thanks for the fix :-) I'll try to get to testing it today

spficklin:
I confirm this PR works correctly!  I tested with the following code using 7.x-2.x branch and no change occurs as @guignonv indicates:

`drush php-eval "chado_update_record('cv', ['name' => 'analysis_property'],['name' => 'analysis_property2'])"
`
After testing this PR it works correctly.

spficklin:
I will make the same fix to the Tripal v3 equivalent code an add another Pull request for it.

laceysanderson (approved):
Sorry, it took me a bit to get my Tripal 2 site back up. I have now also confirmed this fix works 👍 
```php
$new = 'Analysis Date';
$old = 'analysis_date';
dpm(chado_select_record('cvterm', array('*'), array('name' => $old)), 'Original');
chado_update_record('cvterm', ['name' => $old],['name' => $new]);
dpm(chado_select_record('cvterm', array('*'), array('name' => $new)), 'New');
```
Before fix, "New" dpm showed no results. but afterwards it showed the correctly updated cvterm.

laceysanderson:
As we now have two approved reviews, according to the new guidelines, we can merge this in 🎉 

Thank you for the bug fix SheilaBarron  and guignonv!

PR #386, opened 2018-05-25
Made same fix as for PR #379

spficklin:
This is an equivalent code change for Tripal v3 as was submitted by @guignonv for pull request #379

### DIFF
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -875,8 +875,8 @@ function chado_update_record($table, $match, $values, $options = NULL) {
       $sql .= " $field = NULL AND ";
     }
     else {
-      $sql .= " $field = :$field AND ";
-      $args[":$field"] = $value;
+      $sql .= " $field = :old_$field AND ";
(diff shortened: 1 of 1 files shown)


laceysanderson:
Confirmed this works for me using
```php
$new = 'Analysis Date';
$old = 'analysis_date';
dpm(chado_select_record('cvterm', array('*'), array('name' => $old)), 'Original');
chado_update_record('cvterm', ['name' => $old],['name' => $new]);
dpm(chado_select_record('cvterm', array('*'), array('name' => $new)), 'New');
```
Where before fix, "New" dpm showed no results. but afterwards it showed the correctly updated cvterm.

Also checked using @spficklin original method:
```
drush php-eval "chado_update_record('cv', ['name' => 'analysis_property'],['name' => 'analysis_property2'])"
```
Confirmed this worked as well 👍

spficklin:
As this pull request is the same change as #379 and @laceysanderson 's test worked.  I think this is safe to merge without a second review?

bradfordcondon:
because this pR is identical to #379 im very in favor of merging as is.

Issue #141, opened 2017-08-02
TripalField namespace issue


spficklin:
If two TripalField classes have the same name there will be a scope problem. Morever, if two fields use the same term there will be potential problems as well.

laceysanderson:
The issue where two fields have the same term is actually quite a big one. For example, I had used rdfs:type in my property table to indicate the genetic marker type (e.g. Exome Capture). In this situation Tripal simply didn't create a property field displaying that data. There was no warning/error anywhere that I could see. At the very least we need to add some warnings :-)

almasaeed2010:
We've also encountered the same issue trying to create multiple `schema:description` fields for the `chado.protocol` bundle. The table has 3 description fields, `softwaredescription`, `hardwaredescription` and `protocoldescription`. The properties didn't show up on the bundle manage fields page and Tripal did not issue any warning. Further, it looks like Tripal creates the machine name of the field based on the cv term. Would it make sense to allow machine names to be specified by the developer in the tripal_create_bundle function?

Alternatively, tripal_create_bundle could perform additional checks to make sure all machine names it is creating are unique to the bundle.

Please let us know if we can help getting this functioning properly.

Thanks!

spficklin:
Hi @almasaeed2010 

In the case of the chado.protocol table I would recommend against using the 'schema:description' term for all three of those columns even if they are all descriptions.  The reason is that the 'schema:description' term is too generic for what data those columns contain.  I don't think it makes sense for an entity to have multiple fields with the same term.  If the same term is used then all values belonging to that term should then be provided in an array of values under a single field.  Make sense?  In that sense, you could combine all three of those description columns as an array of values, or concatenate them.  But, I would recommend you find or create a term for each column separately and that will solve the naming problem, and it informs the end-user better about the meaning of the columns.   

If you do find appropriate terms for each column of the chado.protocol table please let us know and we can see if they would make sense to incorporate into Tripal itself.

almasaeed2010:
Thanks @spficklin. This totally makes sense.
We did end up using more specific terms for each field but they are `local` terms. Searching through NCBI didn't yield anything that we thought was appropriate, unfortunately.

Thanks!

spficklin:
If you don't mind me asking, what were the terms you came up with?

almasaeed2010:
They were very straight forward:
```
local:protocol_description
local:software_description
local:hardware_description
```

spficklin:
To add to this, can you share the description you used for each of these as well?

[...]

PR #388, opened 2018-05-25
Fields with the same CV term -- issue #141

spficklin:
This is a fix for issue #141


almasaeed2010:
The warning works for me 👍

almasaeed2010 (approved):

laceysanderson:
I've confirmed that this PR shows a warning when two existing fields have the same term :-) However, it doesn't fix my original problem as described here: https://github.com/tripal/tripal/issues/141#issuecomment-392272290.

I'm merging in this PR as the warning it provides works great for @almasaeed2010's situation. We'll address my issue with another PR since I feel it is more related to the fact that a chado property is not created.

Issue #293, opened 2018-02-19
Tripal Collections Unable to Create Files Due to Permissions


almasaeed2010:
### System information

* Tripal Version: 3 (7.x-3.x)
* Drupal Version: 7.54
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

Creating a collection creates the correct folder for each user but fails to detect whether its writable or not. 

The error from the job log looks like this:
```
{"status":"failed","message":"Could not access the directory on the server for storing this file.","file_id":""}Job execution failed: Cannout open collection file:
```

Unfortunately a file name is not provided in the log.

The code block that fails:
```php
$user_dir = 'public://tripal/users/' . $user->uid;
    if (!file_prepare_directory($user_dir, FILE_CREATE_DIRECTORY)) {
      $message = 'Could not access the directory on the server for storing this file.';
     // WORKS WHEN WE COMMENT THIS OUT 
      watchdog('tripal', $message, array(), WATCHDOG_ERROR);
      drupal_json_output(array(
        'status'  => 'failed',
        'message' => $message,
        'file_id' => '',
      ));
      return;
     // UNTIL HERE
    }
```

However, the code functions properly if we ignore the returned value from `file_prepare_directory.` Also, adding `is_dir($user_dir )` and `is_writable($user_dir )` checks in the code both return `true`. So `file_prepare_directory` is failing to detect the correct permissions. Any idea why this is happening?

Thanks!

spficklin:
This is curious.... I'll take a look at it..

bradfordcondon:
hi all @spficklin any thoguhts/updates on this?

spficklin:
I just added a pull request to fix this.  If you can test that would be great.

bradfordcondon:
@almasaeed2010 where were we experiencing this, on HWG?

almasaeed2010:
All my tripal sites (mac os and centos). I test the PR

bradfordcondon:
thanks @spficklin .  merged, closing.

PR #393, opened 2018-05-25
Fix for File Collections File Permissions

spficklin:
This is a fix for issue #293.   Here is a summary of the changes:

1.  Two new API functions were added to the tripal.files.api.inc function for getting a User's file directory and for checking if it is writeable.  This unifies the code between the uploader and the collections with a single set of API calls.
2.   I added a few additional checks to see if the directory exists and is writeable to ensure that if the file_prepare_directory does not properly check we have the backup checks.
3.  I updated the uploader and data collections to use the new API functions.

You can check this by 

1.  Changing permissions in the default/sites/files/tripal/users so that it is not writeable by the user running the job queue.
2.  Creating a data collection
3.  Generating the files and running the job manually using drush.

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -644,7 +644,9 @@ function tripal_create_bundle_fields($bundle, $term) {
   foreach ($modules as $module) {
     $function = $module . '_bundle_fields_info';
     $temp = $function('TripalEntity', $bundle);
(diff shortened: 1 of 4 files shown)


almasaeed2010 (approved):

almasaeed2010:
I tested the branch and it worked perfectly. Thanks!

bradfordcondon (approved):
thanks stephen!

PR #399, opened 2018-05-31
Fixes for PHP 7.2 compatibility and typo on call to API function

spficklin:
These are very small fixes to fix PHP v7.2 errors and a typo when calling an API function call.

I realize that upgrading your site to PHP 7.2 just for testing this would be extremely time consuming and will have a lot of side effects.  I personally think just a quick look at the code changes would be a sufficient review.  Otherwise, to test:

To test you need PHP 7.2
1)  Add any job and you get an Drupal error
2)  Try deleting a record using chado_delete_record and be sure that the matching criteria are single values (no arrays)

To test the bad API Call  (PHP 7.2 not required):
3)  Just run the NCBI taxonomy importer. It will fail without this fix because of the bad function call for a function that doesn't exist.

### DIFF
--- a/tripal_chado/includes/TripalImporter/NewickImporter.inc
+++ b/tripal_chado/includes/TripalImporter/NewickImporter.inc
@@ -108,7 +108,6 @@ class NewickImporter extends TripalImporter {
       '#maxlength' => 255,
     ];
 
-    $type_cv = tripal_get_default_cv('phylotree', 'type_id');
     $so_cv = chado_get_cv(['name' => 'sequence']);
     $cv_id = $so_cv->cv_id;
     if (!$so_cv) {
--- a/tests/tripal_chado/TaxonomyImporterTest.php
+++ b/tests/tripal_chado/TaxonomyImporterTest.php
@@ -43,8 +43,31 @@ class TaxonomyImporterTest extends TripalTestCase {
     $this->assertNotEmpty($result);
 
   }
-//
-//  public function testImportOrganismFromTaxID() {
-//
-//  }
+
+  /**
+   * the importer can also load an array of pubmed ids.  We use the pillbug again.
+   *
+   * https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=96821
+   *
+   * @throws \Exception
+   */
+  public function testImportOrganismFromTaxID() {
+
+    $file = [];
+    $run_args = ['taxonomy_ids' => '96821']; //its the pillbug again!
+    $importer = new \TaxonomyImporter();
+
+    ob_start();
+    $importer->create($run_args, $file);
+    $importer->run();
+    ob_end_clean();
+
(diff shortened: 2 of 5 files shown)


bradfordcondon:
suggesting @almasaeed2010 as a reviewer since he loves upgrading versions :D

bradfordcondon:
👍 These changes look striaghtforward to me.

I did not test the PHP 7.2 component.

bradfordcondon (approved):
thanks!  NCBI loader runs and indeed fails on master.

It would be awesome if we could add a simple test for the loader.  Interested?  I can contribute if you'd like...

laceysanderson (approved):
The code changes look beautiful to me. I didn't test on PHP 7.2 nor have I got the chance to test on my current PHP version. That said, I think such minor changes are good with a single functional review 👍

laceysanderson:
I'm holding off merging though because I Love @bradfordcondon suggestion of a simple test for the loader! Unfortunately I don't have time to contribute one today :-(

spficklin:
A word of warning. Drupal is not yet PHP 7.2 compatible. So, if you upgrade don't do it on a production server, or even an active dev server.  You will get annoying errors constantly.

bradfordcondon:
I've got some time, ill add a test unless @spficklin is actively on it (please let me know).

If I can't contribute it by end of today for some reason let's merge anyway...

bradfordcondon:
i PR'd to stephens branch here rather than push directly to it  https://github.com/tripal/tripal/pull/400

We haven't discussed how review of this sort of thing should go?  I'd think that in this case stephen should soley review the pR to his branch?  but then does this branch need to be re-reviewed?  I dunno....

laceysanderson:
Good point. I think @spficklin should be the only reviewer for your PR #400 and then this PR is re-reviewed if the addition is substantial. In this case, I think we had discussed unit testing being one case where only one review is required and stephen's would count towards that? So maybe we call it good?

spficklin:
I say if the unit test passes then it's good!  Thanks @bradfordcondon for writing that. I will try to remind myself to write these tests for all my pull requests. It's not quite a habit yet for me.

tripal_chado/includes/TripalImporter/NewickImporter.inc
```diff
@@ -108,7 +108,6 @@ class NewickImporter extends TripalImporter {
       '#maxlength' => 255,
     ];
 
-    $type_cv = tripal_get_default_cv('phylotree', 'type_id');
```
bradfordcondon: is that why the newick importer broke?  I couldn't figure it out, i swore i had it running at the hackathon.

almasaeed2010 (approved):

almasaeed2010:
I tested on PHP 7.2 and it worked fine but you are right Drupal keeps crashing or throwing warnings on 7.2.

spficklin:
Thanks @almasaeed2010 !

PR #400, opened 2018-05-31
Na tv3 small fixes

bradfordcondon:
i discovered a couple of issues while writing the test.  I only testing the "update existing" method.

* `addDbxref` throws an error for pre-existing records, so I wrapped it in a check.

* is `chado_get_organism_scientific_name` working as intended?  the type gets appended to the end.  I had to assert that my test organism was null for type_id (instead of a random generated name).

### DIFF
--- a/tests/tripal_chado/TaxonomyImporterTest.php
+++ b/tests/tripal_chado/TaxonomyImporterTest.php
@@ -43,8 +43,31 @@ class TaxonomyImporterTest extends TripalTestCase {
     $this->assertNotEmpty($result);
 
   }
-//
-//  public function testImportOrganismFromTaxID() {
-//
-//  }
+
+  /**
+   * the importer can also load an array of pubmed ids.  We use the pillbug again.
+   *
+   * https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=96821
+   *
+   * @throws \Exception
+   */
+  public function testImportOrganismFromTaxID() {
+
+    $file = [];
+    $run_args = ['taxonomy_ids' => '96821']; //its the pillbug again!
+    $importer = new \TaxonomyImporter();
+
+    ob_start();
+    $importer->create($run_args, $file);
+    $importer->run();
+    ob_end_clean();
+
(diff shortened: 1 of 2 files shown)


bradfordcondon:
Third commit adds test to the other half of functionality (sorry i didnt just do right off, turns out it was very simple to test) providing a pubmed ID.

spficklin:
Hi @bradfordcondon the change to addDbxref function of the TaxonomyImporter looks good. I'm assuming that because you have a unit test for it that it's working and I don't need to manually test it myself.  Is that a safe assumption?

As per the chado_get_organism_scientific_name function.  It should do the following:
1) Genus goes first
2) species goes second.
3) If there is a type_id then it will add the taxonomic infraspecific rank that the type_id refers to after the species name and then that gets followed by the infraspecific name.  If the type_id is NULL then nothing should be added after the species.

spficklin:
I just realized this was a pull request to my branch and I don't have to be very picky about reviews :-)

bradfordcondon:
>Hi @bradfordcondon the change to addDbxref function of the TaxonomyImporter looks good. I'm assuming that because you have a unit test for it that it's working and I don't need to manually test it myself. Is that a safe assumption?


I really added 2 feature tests.  One tests that properties get added to existing organisms.  The other ensures that organisms get added given an ncbi taxon id.  I needed to make the addDbxref change to get either of these to run without warnings.  This was because I had a pre-existing organism in my db that had been looked up.  I didn't explicitly add a test for this, but if we wanted to, just run the importer 2x.

>As per the chado_get_organism_scientific_name function.

ok great.    you can see I make an organism explicitly declaring null for the type (otherwise the factory randomly sets a type with a new cvterm.  this is probably really bad behavior if thats the case...)
If the type isnt null, then ncbi won't find it (because it will be armadillo officinalis extra_nonsense.  

```
 $org = [
      'genus' => 'Armadillo',
      'species' => 'officinalis',
      'abbreviation' => 'A. officinalis',
      'common_name' => 'pillbug',
      'type_id' => null
    ];
```

If we wanted to be thorough we might supply an array of organisms to add and test (ie one that WOULD require a properly set type_id to look up in ncbi taxon.

Issue #322, opened 2018-03-28
Tripal 3 Chado preparation issue when reloading a Chado dump


guignonv:
* Tripal Version: 7.x-3.x
* Drupal Version: not related
* PostgreSQL Version: not related
* PHP Version: not related

Note: this is a minor issue as it should only happen to advanced users manipulating the database directly.

When you install and initialize Chado with Tripal (job) using "Prepare Chado" interface, materialized views are created correctly and everything is fine. However, if you clear the Chado schema in order to load a new one and want to re-"Prepare Chado" again, the materialized views may not be present in the dump while they are recorded as existing on the Drupal (public schema) side. Then, many errors will appear everywhere and it is not possible to "Prepare Chado" again.
Typical error messages: 'Please update the database using "drush updatedb" before continuing' and 'Materialized view, db2cv_mview, already exists. Skipping creation.'.

The workaround I used to solve that issue was to clear "tripal_mviews" table: "DELETE FROM tripal_mviews;" and re-run the "Prepare Chado" process.

Proposition: in tripal/tripal_chado/api/tripal_chado.mviews.api.inc around line 55, the presence/absence of a materialized view table is checked using `'SELECT mview_id FROM {tripal_mviews} WHERE name = :name'` to see if it is recorded in the tripal_mviews table. However, the table may not exist for real. A different approach should be used. What about:
```
SELECT EXISTS (
   SELECT 1
   FROM   information_schema.tables 
   WHERE  table_schema = :chado_schema
   AND    table_name = :name
   );
```
But the synchronization between existing materialized views and whats is recorded in the tripal_mviews table should be also checked in a way or another... Maybe a kind of "repair" function for materialized views should be added and interfaced?

laceysanderson:
I agree that a different means of checking this is a good idea! Perhaps using [chado_table_exists](http://api.tripal.info/api/tripal/tripal_chado%21api%21tripal_chado.schema.api.inc/function/chado_table_exists/3.x)?

We could alleviate the need for an interface if we added the following at line 55:
```php
// Check that the materialized view actually exists and if not, 
// remove the entry from tripal_mviews.
if ($mview_id AND !chado_table_exists($name)) {
   db_query(
    'DELETE FROM {tripal_mviews} WHERE mview_id = :id',
    array(':id' => $mview_id))->execute();
  $mview_id = FALSE;
}
```
[...]

PR #404, opened 2018-06-01
Fix for issue #322 -- Re-adding an mview that was manually deleted.

spficklin:
Issue #322 

## Type(s) of Change(s)

## Description

If a user creates a materialized view, then manually deletes the actual db table without using the Tripal API functions for removing a materialized view, and then attempts to re-add it, the mview table will never get created.  This is because the chado_add_mview() function checked the tripal_mviews table to see if the record exists, which it does and then decides all is good and so it doesn't create the table.   

This is a simple fix that checks to make sure that the table also exists. If it does not then the materialized view record is removed from the tripal_mviews table so the record can be readded and the table created.
## Testing?
I have added a new unit test class TripalChadoMViewsAPITest.  This class tests creation of a materialized view and also has a test specific for this issue.  It first creates the mview, then deletes it without using the proper API function and then tries to recreate it.  

The unit test can be run individually in this way:
 ```
export DRUPAL_ROOT=/var/www/html
./vendor/bin/phpunit Tests\tripal_chado\api\TripalChadoAPITest ./tests/tripal_chado/api/TripalChadoMViewsAPITest.php
```

## Additional Notes (if any):

### CODE REVIEW

File: tripal_chado/api/tripal_chado.mviews.api.inc [hunk 1]
```diff
@@ -40,23 +40,36 @@
...
   $mv_table = $mv_schema['table'];
-
+ 
   // See if the mv_table name already exsists.
-  $mview_id = db_query(
-    'SELECT mview_id FROM {tripal_mviews} WHERE name = :name',
-    array(':name' => $name))->fetchField();
-
+  $mview_id = db_select('tripal_mviews', 'tm')
+    ->fields('tm', array('mview_id'))
```
bradfordcondon, line 56: your'e sticking with the old style for this because you'd have to change the whole file otherwise, right?  I guess thats preferred, nevermind. 

I bring it up because we have the "style" box in the PR form...

tests/tripal_chado/api/TripalChadoMViewsAPITest.php
```diff
@@ -0,0 +1,215 @@
...
+    $this->assertFalse(chado_table_exists($mview_name), 
+      "Failed to manually remove the materialized view, cannot complete the test.");
+    
+    // Now try to read the mview. Previously, the behavior was the the mview
+    // table would not be created because Tripal thinks it's already there.
+    chado_add_mview($mview_name, $mview_module, $mview_schema, $mview_sql, $mview_comment, FALSE);
+    $this->assertTrue($success, "Failed to re-create materialized view: $mview_name");
+    
+    $this->assertTrue(chado_table_exists($mview_name),
+      "Manually removing a materialized views throws off the chado_add_mview function when the mview is re-added. See Issue #322");
```
bradfordcondon: Maybe  "Failed to re-add mview that was previously deleted manually. See Issue #322"

bradfordcondon (commented):
Looks great!  Nice tests and IMO makes it so much easier for me to see.

However my unit tests fail with this error:

```
phpunit
PHPUnit 6.5.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.14
Configuration: /Users/chet/UTK/tripal/sites/all/modules/tripal/phpunit.xml

....string(1) "1"

NOTE: publishing records is performed using a database transaction.
If the load fails or is terminated prematurely then the entire set of
is rolled back with no changes to the database

...........F...                                               19 / 19 (100%)

Time: 5.57 seconds, Memory: 46.00MB

There was 1 failure:

1) Tests\tripal_chado\api\TripalChadoMViewsAPITest::test_re_adding_deleted_mview_issue_322
Failed to create materialized view: analysis_organism_test
Failed asserting that false is true.

/Users/chet/UTK/tripal/sites/all/modules/tripal/tests/tripal_chado/api/TripalChadoMViewsAPITest.php:198
```

spficklin:
@bradfordcondon I don't understand why the test would fail for you but not for me.  I just committed a fix to the test but I don't think it will fix the issue for you.  Can you run this test too and see what you get:

`./vendor/bin/phpunit --filter test_chado_add_mview Tests\tripal_chado\api\TripalChadoAPITest ./tests/tripal_chado/api/TripalChadoMViewsAPITest.php 
`

spficklin:
You don't happen to already have a `analysis_organism_test` mview already do you?

bradfordcondon:
I do not.  I can see i CI that the test fails on php 7 and 7.1, but PASSES on 5.6
https://travis-ci.org/tripal/tripal/builds/386857523

Could this be why?


edit:  hold up, seems more likely that it would fail because the table is already there, and this is a "transactions arent behaving the way i expect them to".  Does adding an mview submit a job?  jobs use transactions, which breaks test suite's dbtransactions.

The travis would also fail because the jobs are executing in random order, and it passes when the mview hasnt been added yet.  I restarted the passing job on 5.6 to test....


Workaround for now would be to delete the test mview in each test.

bradfordcondon:
>./vendor/bin/phpunit --filter test_chado_add_mview Tests\tripal_chado\api\TripalChadoAPITest ./tests/tripal_chado/api/TripalChadoMViewsAPITest.php


this test passes for me

`phpunit --group 322` this also passes for me :|  the test works in isolation.

spficklin:
Okay, I think I've fixed it. The problem was actually caused by another API function:  `chado_table_exists`.  This table maintains a global variable for which chado tables it sees.  This way it doesn't have to do the same expensive lookups.   After each test, the database changes are rolled back, meaning the mview table is deleted, but this global array still has the entry in it.  This caused problems.  I'm not sure what to do about that.  I created a small fix in the test to get around it and I'll add a separate issue for the `chado_table_exists` problem

bradfordcondon (approved):

bradfordcondon:
suggesting reviewers based on participants in the issue.  one more review to merge this.

laceysanderson:
I wonder if it wouldn't be better to re-add the table to chado rather then remove the tripal_mviews record?

In the case where the table has gone missing due to a failed upgrade/restore, simply removing all the records from tripal_mviews would leave the site in a very broken state.

In the case originally reported, adding the table to chado would allow the tripal admin to remove it properly or edit it. Furthermore, when preparing chado re-creating the mview table is main goal.

spficklin:
@laceysanderson I believe the function does both. It will remove the entry and then it will re-create the table and then re-add the entry to the tripal_mviews table.

tripal_chado/api/tripal_chado.mviews.api.inc
laceysanderson: I think we should consider the currently coding style of the project the old style until your PR goes through changing the full code base. In order words, checking that checkbox means you've kept the coding style consistent with the current state of the code ;-p

laceysanderson (commented):
Code looks good 👍 A more careful review has confirmed that it will remove and then re-add the mview :facepalm: Functional testing to follow :-)

laceysanderson (approved):
Functional Testing:
1. I dropped the analysis_organism table at the database level.
2. Re-added the table definition through `admin/tripal/storage/chado/mviews/new`

- On submit I got a happy green message telling the mview was added :-)
- Checking the db showed that the table was re-created as expected :-)

laceysanderson:
This PR now officially has 2 reviews 🎉 thus I'm merging it in :-) Thanks @spficklin

Issue #411, opened 2018-06-04
OBO loader fails on plant trait ontology


shawnawsu:
The plant trait ontology, https://raw.githubusercontent.com/Planteome/plant-trait-ontology/master/to.obo, is not loading because of it's structure: 

```
[Term]
id: BFO:0000017 ! realizable entity
is_a: BFO:0000020 ! specifically dependent continuant
```

srobb1:
I wonder if it is not loading because it is missing a "name:" field. I think name is required.

PR #412, opened 2018-06-04
OBO Loader: support for ontologies with non-local terms (using EBI Lookups)

shawnawsu:
Issue #411 

## Type(s) of Change(s)

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description
The OBOImporter had previously been updated by me to search for ontology information when no default namespace was found. This update utilizes the existing ebi search function I previously wrote to address references in an imported ontology to another ontology. For example in the plant trait terms are defined as:
```
[Term]
id: BFO:0000017 ! realizable entity
is_a: BFO:0000020 ! specifically dependent continuant
```
The id needs to be looked up and the ontology that the term comes from needs to be added if it is not already in the local instance of chado. 

## Testing?
To test please load different types of OBO ontologies. I would suggest problematic ontologies like: 
https://raw.githubusercontent.com/Planteome/plant-trait-ontology/master/to.obo
https://raw.githubusercontent.com/obophenotype/planaria-ontology/master/src/ontology/plana.obo

I've tested both of these and loaded the full GO  & go-slim on this branch.

## Additional Notes (if any):
I'm not familiar with ontologies other than ones I've already tested, so if there is a particularly gnarly one that you know of please try it. 

### DIFF
--- a/tripal_chado/includes/TripalImporter/OBOImporter.inc
+++ b/tripal_chado/includes/TripalImporter/OBOImporter.inc
@@ -409,18 +409,18 @@ class OBOImporter extends TripalImporter {
 
     // Update the cv_root_mview materiailzed view.
     $this->logMessage("Updating the cv_root_mview materialized view...");
-    $mview_id = chado_get_mview_id('cv_root_mview');
-    chado_populate_mview($mview_id);
+    $mview_id = tripal_get_mview_id('cv_root_mview');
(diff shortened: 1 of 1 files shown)


spficklin:
Seems to work great!   But the EBI lookups for the Trait Ontology make it load extremely slow....

Issue #389, opened 2018-05-25
chado property fields attach load undefined property warning


bradfordcondon:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.58
* PostgreSQL Version: 9.6.1
* PHP Version: 7.1.14

### Issue description

Chado property fields give this error when null

```
Notice: Undefined property: TripalEntity::$tripal__ena_last_update in TripalEntityController->attachLoad() (line 658 of /var/www/html/sites/all/modules/custom/tripal/tripal/includes/TripalEntityController.inc).
```

You can see this on my "data seeded" tripal 3 docker https://hub.docker.com/r/statonlab/tripal3_seeded/ by visiting a biomaterial page.  These property fields were generated automagically by clicking "check for new fields".

Chado properties created manually (via the structure -> fields) as well as those "discovered" via check for new fields have this warning.

laceysanderson:
Can you describe this a little more? What do you mean by null... when the field is empty?

bradfordcondon:
yes.  So creating a new chado property field in the UI, attaching it, and visiting an entity of that content type will give the warning because its empty.

laceysanderson:
Hmmm... I'm unable to duplicate this. 

I added a chado property of type `tripal:analysis_blast_blastdb` to my organism content type. I confirmed that I have no properties in the db of that type. Then I navigated to an organism page (Lens culinaris actually :-)) and did not see an error. I also checked the edit page and the Drupal watchdog logs :-( I even tried clearning the cache a couple of times to see if that helped --no dice.

bradfordcondon:
When visiting an organism page:

```
Notice: Undefined property: TripalEntity::$local__lineage in TripalEntityController->attachLoad() (line 658 of /data/www/html/hardwoods_dev/drupal-7.52/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
Notice: Undefined property: TripalEntity::$local__division in TripalEntityController->attachLoad() (line 658 of /data/www/html/hardwoods_dev/drupal-7.52/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
Notice: Undefined property: TripalEntity::$local__mitochondrial_genetic_cod in TripalEntityController->attachLoad() (line 658 of /data/www/html/hardwoods_dev/drupal-7.52/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
[...]

PR #413, opened 2018-06-05
Cache set check

bradfordcondon:
Issue #389

## Type(s) of Change(s)

## Description

Fields that are not set for an entity caused this warning:

```
Notice: Undefined property: TripalEntity::$tripal__ena_last_update in TripalEntityController->attachLoad() (line 658 of /var/www/html/sites/all/modules/custom/tripal/tripal/includes/TripalEntityController.inc).
```

So for example, if you set a chado_property for a content type, and then open another entity with that property not set, you will get this error.  This error is present on my live site (with warnings enabled) and dev site.

However, @laceysanderson was not able to duplicate on her system.  
It's such a minor fix that I'm hoping that won't be an issue.

* add a chado property
* fill out for one entity but not another
* clear cache
* visit the entity lacking the entity's page.

before fix
screen shot 2018-06-05 at 1 33 47 pm

after fix
with_patch

## Additional Notes (if any):

- [x ] My code follows the code style of this project.

### DIFF
--- a/tripal/includes/TripalEntityController.inc
+++ b/tripal/includes/TripalEntityController.inc
@@ -655,7 +655,9 @@ class TripalEntityController extends EntityAPIController {
                   FIELD_LOAD_CURRENT, $options);
               // Cache the field.
               if ($field_cache) {
-                cache_set($cfid, $entity->$field_name, 'cache_field');
+                if (property_exists($entity, $field_name)) {
+                  cache_set($cfid, $entity->$field_name, 'cache_field');
+                }
               }
               $queried_entities[$id]->{$field_name}['#processed'] = TRUE;
             }


spficklin:
When you say "open another entity" do you mean create a new entity of the same content type that has the field attached?

bradfordcondon:
yes, or visit a pre-existing entity of that type (which should have null for the field).

spficklin:
I was able to duplicate this bug and this PR does fix it.  The source of the problem is in the chado_linker__prop.load() function.  At the end if the property does not exist then it manually removes the field from the $entity object.  This is the only field that does this and it's a hack to keep the property key from showing up in the TOC on the page when no properties are actually present.   I think the fix is fine and while we should probably dig into why we had to put that adjustment in the load() function of the chdao_linker__prop field, this fix corrects the problem and makes the Entity controller more robust.

spficklin:
This fix is so simple, I think it only needs one reviewer. I'm going to merge.

Issue #423, opened 2018-06-11
Remote Field Fails with Fatal Error


almasaeed2010:
### System information
* Tripal Version: 7.3 (master branch)
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

After adding a remote field to the organism bundle, I get the following error:

```
Notice: Undefined property: stdClass::$site_id in remote__data->makeRemoteRequest() (line 208 of /html/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data.inc).
Exception: Please provide a numeric site ID for the tripal_get_remote_content function. in tripal_get_remote_content() (line 217 of /html/Tripal/sites/all/modules/tripal/tripal_ws/api/tripal_ws.api.inc).
```

Seems like a simple error so should be easily fixable.

Thanks!

almasaeed2010:
Resolving the site id issue, produces the following non-fatal error:
```
Notice: Undefined index: label in remote__data_formatter->view() (line 106 of /html/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data_formatter.inc).
Notice: Undefined index: ItemPage in remote__data_formatter->view() (line 107 of /html/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data_formatter.inc).
```

almasaeed2010:
So the PR #424 addresses the first fatal error but I am not sure what's needed for the second formatter problem.

laceysanderson:
Removing PR submitted (it's merged btw) since now we need a fix for the formatter errors.
> Notice: Undefined index: label in remote__data_formatter->view() (line 106 of /html/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data_formatter.inc).
> Notice: Undefined index: ItemPage in remote__data_formatter->view() (line 107 of /html/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data_formatter.inc).

spficklin:
Thanks... I'm working on that...

laceysanderson:
Just making sure it didn't fall through the cracks :-)

spficklin:
I added a new pull request to fix the remaining problems.

laceysanderson:
PR Merged.

PR #424, opened 2018-06-11
Fix remote site fatal error by supplying the correct site id

almasaeed2010:
Issue #423 

## Type(s) of Change(s)

## Description

The error was caused by attempting to access an unavailable property `site_id` when it should've been `id`


spficklin:
Thanks for catching this one @almasaeed2010.

spficklin:
This PR only had one line of code and it is a correction to an obvious typo.  Only one review is needed.

Issue #204, opened 2017-12-02
Issue Creating Map Content Type 


spficklin:
When creating a content type using the featuremap table there is no type_id column.  This throws off the form. It doesn't automatically provide the ability to use the featuremapprop table.  The user is forced to selected the unittype_id field, then select '--None--' to get the property elements to appear.

laceysanderson:
This is fixed as we now have a map content type on install.

spficklin:
This unfortunately is still a problem.   We have the map type created on install, but if someone deletes it and wants to recreate it manually they cannot :-(

spficklin:
This should be fixed before stable release.

spficklin:
PR merged! closing.

PR #426, opened 2018-06-11
Fix for issue #204

spficklin:
Issue #204 

## Type(s) of Change(s)

## Description

Previously when creating content types if a table had a column with a FK to the cvterm table then it would assume that one of those fields was the type_id.  However for the featuremap table this is not the case. The column unitype_id is not a type for the map.  The create content type form should allow the user to skip the type and use the property table instead.
## Testing?

To test, create a new content type of type 'Genetic Map'.  Select the 'featuremap' table, and when it asks if all records in the table are of type 'Genetic Map' click 'No'.  Then click the checkbox to ignore the type column.  Finally use 'map type' for the property type and 'genetic' as the value.  This should allow you to create a content type of type 'Genetic Map' where it can be distinguished using the featuremapprop table.
## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/tripal_chado.field_storage.inc
+++ b/tripal_chado/includes/tripal_chado.field_storage.inc
@@ -43,6 +43,7 @@ function tripal_chado_field_storage_write($entity_type, $entity, $op, $fields) {
 
   // Convert the fields into a key/value list of fields and their values.
   $field_vals = tripal_chado_field_storage_write_merge_fields($fields, $entity_type, $entity);
+// dpm($field_vals);
 
   // First, write the record for the base table.  If we have a record id then
   // this is an update and we need to set the primary key.  If not, then this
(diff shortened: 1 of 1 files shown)


tripal_chado/includes/tripal_chado.field_storage.inc
```diff
@@ -43,6 +43,7 @@ function tripal_chado_field_storage_write($entity_type, $entity, $op, $fields) {
 
   // Convert the fields into a key/value list of fields and their values.
   $field_vals = tripal_chado_field_storage_write_merge_fields($fields, $entity_type, $entity);
+// dpm($field_vals);
```
bradfordcondon, line 46: can we remove this dpm?  even commented out it drives me crazy sometimes searching for my own dpms.

bradfordcondon (approved):

bradfordcondon:
looks straight forward enough for a 1 reviewer to me- if you agree @spficklin please merge.

spficklin:
Thanks @bradfordcondon merging on your recommendation that only one reviewer is needed.

Issue #375, opened 2018-05-17
Creating, Editing and Deleting Tripal Fields Cause 500 Error


almasaeed2010:
### System information
- Tripal Version: 7.x-3.x (latest)
- Drupal Version: 7.59
- PostgreSQL Version: 9.6
- PHP Version: 7.1

### Issue description
The issue does not happen with Chado and Drupal fields. When deleting a field that's "Supported By" Tripal, we get a 500 error. The same happens when we attempt to edit or delete. This issue only started happening after the latest update to tripal. I am not sure what's causing the issue and the logs don't show anything helpful. In fact I can't find anything in the logs regarding this error!

Any idea what might be causing this?

Thanks!

spficklin:
What Field did you try to delete? Is it one that comes with Tripal?

almasaeed2010:
I tried multiple fields but I believe `rdfs__type` comes with Tripal.

bradfordcondon:
@almasaeed2010  did we try the drush delete field command or only via GUI?

almasaeed2010:
No I haven't yet

bradfordcondon:
after pulling on master i was able to edit the rdfs:type field as well as create our field of interest (ontology_data).

Not sure if it was fixed or if there was something on our site that changed preventing me from duplicating issue, but closing as resolved for now.

bradfordcondon:
reopening: we dont have this issue on our dev site but its on our live site and updating hasn't resolved it :|

bradfordcondon:
the field **can** be deleted with drush.

bradfordcondon:
I edited the "label" field and brought down our live site :(

Adding error printing all pages have
```
PDOException: SQLSTATE[08006] [7] FATAL: remaining connection slots are reserved for non-replication superuser connections in lock_may_be_available() (line 167 of /var/www/html/includes/lock.inc).
```

I'm thinking hte field edit is doing something crazy taking up all db connections...

bradfordcondon:
@almasaeed2010 modified our database settings and now i am able to create the TripalField.  However now when I try to edit it to set the necessary settings I get this error:

```
Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 4096 bytes) in /var/www/html/sites/all/modules/tripal/tripal/includes/tripal.field_storage.inc on line 116
```

This error occurs on all our TripalFields.  It does not occur for Chado or Drupal fields.

bradfordcondon:
```

[...]

PR #428, opened 2018-06-12
Deleting Fields

spficklin:
Issue #375

## Type(s) of Change(s)

## Description

This is a fix for issue #375.   When deleting a TripalField (as opposed to a ChadoField) the tripal_field_storage_query() function is called but that function was not properly setup to handle all of the query filters and it caused slowness and problems.  This PR adds support for more of those filters and should fix the problem.
## Testing?

Delete a field associated with any Tripal content type.  The test works best with a large Tripal site with lots of content.   Try deleting the Resource Type field as this is a TripalField.  But don't do it on a production site as this field is necessary and it doesn't seem to come back if you "Check for new fields" afterwards.
 
## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/tripal_chado.field_storage.inc
+++ b/tripal_chado/includes/tripal_chado.field_storage.inc
@@ -727,6 +727,16 @@ function tripal_chado_field_storage_query($query) {
       } // end foreach ($field['bundles']['TripalEntity'] as $bundle_name) {
     } // end if ($sort['type'] == 'field') {
   } // end foreach ($query->order as $index => $sort) {
+  
+  
+  // Only include records that are deleted.  Tripal doesn't keep track of
+  // records that are deleted that need purging separately so we can do nothing
+  // with this.
+  if (isset($query->deleted)) {
+    // There won't ever be field data marked as deleted so just created a
+    // condition that always evaluates to false.
+    $cquery->where('1=0');
+  }
 
  //dpm($cquery->__toString());
  //dpm($cquery->getArguments());
--- a/tripal/includes/tripal.field_storage.inc
+++ b/tripal/includes/tripal.field_storage.inc
@@ -76,10 +76,18 @@ function tripal_field_storage_query($query) {
   $select->join('tripal_bundle', 'TB', 'TE.bundle = TB.name');
   $select->fields('TE', array('id'));
   $select->fields('TB', array('name'));
+  
+  // Apply any entity condition filters.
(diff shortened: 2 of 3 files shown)


bradfordcondon:
A safer way to test is to manually create a TripalField and then delete it.  FYI almsaeed is out of town and cannot test this week.

spficklin:
Oh, I also added a fix to the chado_generate_var() that fixes warning errors for PHP 7.2.  These warnings were all over when I would delete fields.

tripal_chado/includes/tripal_chado.field_storage.inc
```diff
@@ -727,6 +727,16 @@ function tripal_chado_field_storage_query($query) {
       } // end foreach ($field['bundles']['TripalEntity'] as $bundle_name) {
     } // end if ($sort['type'] == 'field') {
   } // end foreach ($query->order as $index => $sort) {
+  
+  
+  // Only include records that are deleted.  Tripal doesn't keep track of
+  // records that are deleted that need purging separately so we can do nothing
+  // with this.
+  if (isset($query->deleted)) {
```
bradfordcondon, line 735: can you elaborate on the point of this?  Will we some day be able to something with this?

bradfordcondon:
there might be more to the speed issue but this certainly fixes the linked issue of 500 errors/unable to redirect.

bradfordcondon (approved):

tripal_chado/includes/tripal_chado.field_storage.inc
spficklin: Drupal fields create their own tables for each field.  And each table has a 'deleted' column.  I believe this allows Drupal to mark field values attached to entities as deleted without taking the time to remove them.  Then during cron runs it will purge those deleted items.  That's what's actually happening here, Drupal tries to purge the first 10 deleted values when you delete the field from the content type.  However, our Tripal fields do not keep a deleted state.  We don't want the field query to return anything if the query is to find fields that are marked as deleted because our TripalFields don't do that.  So, it should always return nothing.

laceysanderson (approved):
I tested this on a moderate amount of content and was able to successfully delete a Tripal field 👍 
Do we want to specifically wait for @almasaeed2010 or is my review sufficient?

spficklin:
I think I'll merge since we have two reviews that were successful.  If @almasaeed2010 sees additional problems I can do another PR.

PR #443, opened 2018-06-15
Added analysis field 

spficklin:
Issue #210

## Type(s) of Change(s)

## Description

This adds an analysis field.  Currently the only content type that uses this is the Phylogenetic Tree.  It also fixes the accession field (corresponds to phylotree.dbxref_id column) as it is supposed to be required.

## Testing?

1)  Go to the Phylogenetic Tree content type page Structure > Tripal Content Types and click "manage fields" for the Phylogenetic Tree.  On that page, delete the Analysis (operation__analysis) and the Accession (data__accession) fields.  Once those are gone, click the "Check for new fields" and those field should be re-added.  Make sure that both fields appear on the page by reviewing the "Manage Display" tab.
4)  Create a new Phylogenetic Tree entity (page).  On the form, the Analysis element should be a select box that lists the available analyses.  You can select one, fill out the other fields and save the entity.  The page should appear as expected with an analysis value.
5)  Now that the entity exists, test editing it selecting a different analysis and saving.  You should see the updated analysis.
6) Test that the analysis shows up in web services.  Go to your site's web-services and navigate to the Phylogenetic Tree you added (using a JSON viewer in the browser helps navigate easily).  You should see the analysis name.
7) You should be able to search for your Phylogenetic Tree using the analysis name from a web services query like this:  http://[tripal_site]/web-services/content/v0.1/Phylogenetic_Tree?analysis,name=[analysis_name].  Experiment with changing the value of [analysis_name] to ensure that you see no results.
8) You can test searching for the Phylogenetic Tree using the analysis  in the Views based searches that Tripal creates for each content type.  Go to http://[tripal_site]/data_search/data_search/phylogenetic_tree, click the link when you mouse over the upper right of the form to edit the view.  Add the Analysis field as an exposed filter.  Save the view.  You should now be able to search for your Phylogenetic Tree using the Analysis name.  Experiment with correct and incorrect names to make sure the search is fully working.

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/TripalFields/data__accession/data__accession_widget.inc
+++ b/tripal_chado/includes/TripalFields/data__accession/data__accession_widget.inc
@@ -94,6 +94,14 @@ class data__accession_widget extends ChadoFieldWidget {
     $dbxref_id = $form_state['values'][$field_name]['und'][$delta]['chado-' . $field_table . '__dbxref_id'];
     $db_id = $form_state['values'][$field_name]['und'][$delta]['db_id'];
     $accession = $form_state['values'][$field_name]['und'][$delta]['accession'];
+    
+    // Is this field required?
+    if ($element['#required'] and !$db_id) {
+      form_set_error($field_name . '][und][0][db_id', "A database for the accession must be provided.");
+    }
+    if ($element['#required'] and !$accession) {
+      form_set_error($field_name . '][und][0][accession', "An accession number must be provided.");
+    }
 
     // If user did not select a database, we want to remove dbxref_id from the
     // field. We use '__NULL__' because this field is part of the base table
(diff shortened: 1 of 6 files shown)


laceysanderson (changes requested):
I confirmed that using the analysis field I can now add/edit linking an analysis to a phylotree.

However, when I leave the analysis blank I get the following error. The analysis should be optional for phylotree, right?
> Please specify an analysis.

Furthermore, I still do not see the dbxref of phylotree as required. There is no red asterisks and when I submit without entering it, I receive the following error:
> Could not save the entity: SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "dbxref_id" violates not-null constraint DETAIL: Failing row contains (4, null, testing 443, null, 483, null).
> Cannot save entity

spficklin:
Hi @laceysanderson Thanks for the review!  I have fixed the two issues.  Namely, you are no longer required to have an analysis. Also, the accession field now also indicates that it is required.

laceysanderson (approved):
Confirmed your changes fixed my concerns 👍

bradfordcondon (approved):
Functionality looks good.

Issue #436, opened 2018-06-14
Tripal Collections are very slow to create for Drupal Search API Views


laceysanderson:
## BUG/ERROR report

Follow-up Issue for #351.

### System information

* Tripal Version: 7.x-3.x
* Drupal Version: n/a
* PostgreSQL Version: n/a
* PHP Version: n/a

### Issue description

Currently creating a Tripal Collection from a Drupal Search API view is very slow (e.g. 5 min to create a collection with 400 members). 

This is due to needing to lookup the bundle for each item in the collection in order to create it.
https://github.com/tripal/tripal/blob/7.x-3.x/tripal/views_handlers/tripal_views_handler_area_collections.inc#L413-L421 
```php
  // First, determine the bundle for each entity in our resultset.
  $entities = array();
  $results = unserialize($form_state['values']['results']);
  foreach($results as $r) {
    // Retrieve the bundle for the current entity.
    // ASSUMPTION: all search results are TripalEntities.
    $wrapper = entity_metadata_wrapper('TripalEntity', $r->entity);
    $bundle = $wrapper->getBundle();
```

Additionally, we need to loop through all bundles to get the list of fields attached to them.
https://github.com/tripal/tripal/blob/7.x-3.x/tripal/views_handlers/tripal_views_handler_area_collections.inc#L450-L458
```php
  foreach($entities as $bundle_name => $bundle_col) {
    $field_ids = array();
    foreach (field_info_instances('TripalEntity', $bundle_name) as $field) {
      $field_ids[] = $field['field_id'];
    }
    $bundle_col['fields'] = $field_ids;
    $collection->addBundle($bundle_col);
  }
```

**NOTE: This is not an issue for non-Drupal Search API views, since they are created specific to a single bundle (no need to look it up) and fields are looked up during loading of the view (and cached) rather then during collection creation. We can't look up fields during loading of the view since we don't know which bundles our results include.**

#### Steps to reproduce

1. Set up the Drupal Search API as indicated in this tutorial: http://tripal.info/tutorials/v3.x/searching/search-api
2. Edit the search view created in the tutorial to add Global: Tripal Content Data Collections to the footer.
3. On the search page, create a tripal collection.

spficklin:
[...]

PR #444, opened 2018-06-16
Added warning for slow data collections when using Search API

spficklin:
Issue #436

## Type(s) of Change(s)

## Description

This isn't really a bug fix, as the code works as expected, it just adds a notice for the site developer warning that if they want to integrate data collections into the footer of the Search API results that they need to ensure that the highlighting filter is turned off.
## Testing?

There are two tests that need performing.  
1)  Go to your search view, and add the "Global: Tripal Content Data Collections" to the Footer section.  When adding you should see the blue highlighted warning message.
2)  We need to double check that in fact if you turn off highlighting that the data collections is built quickly for large result sets (> 400 results), and if it is on the data collections is built slowly.

This test is simple enough, and the code changes are small. I think only one reviewer is necessary.
## Additional Notes (if any):

### DIFF
--- a/tripal/views_handlers/tripal_views_handler_area_collections.inc
+++ b/tripal/views_handlers/tripal_views_handler_area_collections.inc
@@ -3,15 +3,17 @@
 class tripal_views_handler_area_collections extends views_handler_area_result {
 
   function options_form(&$form, &$form_state) {
-    // We have no options so we have to implement this function with
-    // nothing in it.
-    $form['content'] = array(
-      '#type' => 'markup',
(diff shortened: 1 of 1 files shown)


bradfordcondon:
Hi, while testing I found my highlighting was already turned off so i'm not sure what to say about this PR.  https://github.com/tripal/tripal/issues/436#issuecomment-398057923

PS I also found the message confusing so i re-worded, hope thats ok.

spficklin:
Turning off highlighting definitely helped on my end.  But not for @bradfordcondon.  So I'm going to close out this PR and we can revisit the issue.

PR #161, opened 2017-09-14
Add warning

almasaeed2010:
Regarding issue #141 

Added a warning to notify developers which fields didn't get automatically created.

Thanks!

PR #445, opened 2018-06-16
Fix for namespace conflicts (same term is used) for fields.

spficklin:
…e. the same term is used multiple times

          

Issue #141 

## Type(s) of Change(s)

## Description

It was possible for two fields to have the exact same term.  In a previous PR (#161) this was partially addressed by warning the user if two fields had the same term.  This PR fixes several problems.
1)  If a property in a prop table uses a term that is already used by another field then the Chado Propety field for that type doesn't show and no warning or error message is given.  The user is left wondering what happened to it.
2)  A user could manually add a new Chado property with a term that already existed.
3) Fields terms are used to build the property field names.  But if a term name is longer than 32 chars it was truncated.  This caused problems with fields such as mitochondrial_genetic_code and mitochondrial_genetic_code_name.  They both got truncated to the same length and one didn't get added because they looked like the same field.
4) Many of the fields were missing their term vocabs in the info arrays.  A check was added when creating new fields that made sure the term info was there and it caught this problem. This was corrected.

## Testing?

1) Go to the organism Content Type.   Click the 'check for new fields' link.  You should receive this message ```The term local:mitochondrial_genetic_code_name is in use by multiple fields: local__mitochondrial_2845, local__mitochondrial_genetic_cod. This is not allowed. Every field must have a different controlled vocabulary term. Please correct the term assignments.```. This occurs because of problem 3.  Delete the field with the machine name local__mitochondrial_genetic_cod and the message should go away.
2)  Add a new property  of type 'rdfs:type' to the organismprop table.  Use this SQL code provided by @laceysanderson in issue #141.  ``` INSERT INTO chado.organismprop (organism_id, type_id, value) 
SELECT 1 as organism_id, cvt.cvterm_id as type_id, 'Testing this out #141' as value 
FROM chado.cvterm cvt 
LEFT JOIN chado.cv ON cv.cv_id=cvt.cv_id 
WHERE cv.name='rdfs' AND cvt.name='type';```
Now back on the Organism content type page, click the 'Check for new fields' link.  You should see the following warning message: ```
Warning message
The field, rdfs__type, provided by the storage type, field_chado_storage, uses a term, rdfs:type, that is already in use on this content type and provided by another storage backend. The field instance cannot be added. Perhaps, consider a different term and adjust the data in the database.```
3)  Next, add a new Chado Property field to the content type. Name it whatever you want.  But set the term at the bottom to be 'TAXRANK:0000005' (i.e. the term genus).  It should not let you save the new field and should give you this error message: ```The term, TAXRANK:0000005, is already in use on this content type. A term can only be used once per content type. Please choose a different term.```
4) Without fixing the term for the new Chado Property, return to the Organism content type page.  The new field should be there, but without a term. You should now see a warning message: ```The field, [your field name], is missing a controlled vocabulary term. Please edit the field and set a term, otherwise this field may not work properly.```
## Additional Notes (if any):

I made one other change to the tripal_report_error API function. I added a 'drupal_set_message' option so that you can make the message appear on the screen within Drupal as well as log to watchdog.

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -628,6 +628,8 @@ function tripal_get_content_type($bundle_name) {
  *
  * @param $bundle_name
  *   The name of the bundle to refresh (e.g. bio_data_4).
+ * @param $term
+ *   The term object for the bundle.
  *
  * @return
  *   The array of field instance names that were added.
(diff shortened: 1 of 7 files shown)


laceysanderson (approved):
I tested all four scenarios and they work as advertised 👍 

However the message for my case (testing 2) would make it hard to debug this problem... It would be better if the field type (e.g. "Chado Property") were added to these messages.

spficklin:
Thanks Lacey for the review!

spficklin:
I fixed a merge conflict that occurred after merging other PRs.  Having this many PRs is causing a few problems with merge conflicts as they get merged in.

bradfordcondon (commented):
Tested all the warning scenarios, they all show up.  great job, thank you stephen.

bradfordcondon (approved):

Issue #415, opened 2018-06-06
Wrong CDS sequence order


abretaud:
### System information

* Tripal Version: 2.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.5
* PHP Version: 7.2.4

### Issue description

Hi,
I have a problem with one of our tripal 2 instance where the "Coding sequence (CDS) from alignment" displayed for some mRNA is wrong. After looking into it, it seems like it happens for mRNA on reverse strand. The sequence computed by tripal is composed of all the CDS blocks defined in the gff I loaded, but the CDS are in the wrong order, ie, the last CDS is displayed first, the first is displayed last, etc.
So the CDS sequences looks like KLMNFGHIJABCDE instead of ABCDEFGHIJKLMN.
Here's a sample of GFF that should allow to reproduce the bug.
I suspect the problem to be at https://github.com/tripal/tripal/blob/7.x-2.x/tripal_feature/api/tripal_feature.api.inc#L385 where the sequence is concatenated before doing the reverse complement. If it's the case (not sure), the same code is [still in tripal 3](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/api/modules/tripal_chado.feature.api.inc#L433), so I guess the bug will still be there too (haven't had the time to test though)

```
SomeContig	maker	gene	2723762	2730550	.	-	.	ID=FOOBAR9;Name=FOOBAR9
SomeContig	maker	mRNA	2723762	2730550	.	-	.	ID=FOOBAR9-RA;Parent=FOOBAR9;Name=FOOBAR9-RA
SomeContig	maker	three_prime_UTR	2723762	2724226	.	-	.	ID=FOOBAR9-RA:three_prime_utr;Parent=FOOBAR9-RA
SomeContig	maker	exon	2723762	2724371	.	-	.	ID=FOOBAR9-RA:exon:12912;Parent=FOOBAR9-RA
SomeContig	maker	CDS	2724227	2724371	.	-	1	ID=FOOBAR9-RA:cds;Parent=FOOBAR9-RA
SomeContig	maker	exon	2724568	2724639	.	-	.	ID=FOOBAR9-RA:exon:12911;Parent=FOOBAR9-RA
SomeContig	maker	CDS	2724568	2724639	.	-	1	ID=FOOBAR9-RA:cds;Parent=FOOBAR9-RA
[...]

PR #453, opened 2018-06-18
Fixed so__cds, so__transcript, added a local__gbrowse_image field.

spficklin:
Issue #415 

## Type(s) of Change(s)

## Description

Issue #415 is about a bad CDS sequence.  I'm not able to reproduce the problem nor do I have a fix for it, but in the process of trying to debug it in Tv3 I fixed the so__cds and so__transcript fields there were there just no working.  They now work.  I also added a local__gbrowse_image field to embed a GBrowse image of the feature alignment on the gene/mRNA page so I could more easily debug as well.   While I wait on submitter input for issue #415 I thought I'd submit these updates as a pull request.
## Testing?

Pull the branch for the PR, go to a gene or mRNA content type page (Administer > Structure > Tripal Content Types) and click the 'check for new fields' link.  You should get all three.  Click the 'Manage Display' tab move those new fields out of the Disabled section.  Finally, if you want to test the local__gbrowse_image field you need to edit the field on the 'Manage fields' tab of the content type and add a URL.  If you do not have access to a GBrowse instance for your data you can see that it works here for two different species:

Oryza sativa: http://gene-networks.org/gene/713648
Arabidopsis thaliana: http://gene-networks.org/gene/771463

## Additional Notes (if any):


laceysanderson:
As much as I love the GBrowse field, I worry it enters into the realm of feature creep... Thoughts @spficklin? 

You could move it into a "Tripal GBrowse" module or contribute it to another module -I'd be happy to have it in trpfancy_fields or tripal_jbrowse...

spficklin:
I thought you might say that @laceysanderson.  Do you have a similar field in your tripal_jbrowse module for jbrowse?  If you would like to take this field for one of your modules I'd be happy to let it move.

laceysanderson:
I will have a similar field for JBrowse in the near future. The newest version of JBrowse allows native embedding :-D so I just need some time to work that out and then I'll make a field as well.

Here's the repo's if you'd like to make a PR adding this field, I will happily accept :-)
https://github.com/tripal/tripal_jbrowse OR https://github.com/tripal/trpfancy_fields

spficklin:
Where's your preference for where it goes?  I'm thinking fancy fields  because it's specifically not JBrowse.  Would you agree?

laceysanderson:
I think so 👍

laceysanderson:
@spficklin this PR still needs the GBrowse field removed... Consider this a friendly ping :-)

spficklin:
Okay the local__gbrowse_image field has been removed and I resolved merge conflicts.

shawnawsu:
@spficklin I tested this and the so_cds field came through on mRNA but the so_transcript comes through on the gene content type (which according to the tripal_chado.fields.inc is correct as it's being attached to the gene bundle). I can't tell from the instructions if both these fields are meant to be on noth gene and mRNA? If this behaviour is correct then I approve this PR.

tripal_chado/includes/tripal_chado.fields.inc
```diff
@@ -1739,7 +1739,7 @@ function tripal_chado_bundle_instances_info_custom(&$info, $entity_type, $bundle
 
   // PROTEIN & CDS SEQUENCES.
   if ($table_name == 'feature' and
-      ($bundle->label == 'mRNA' or $bundle->label == 'transcript'))  {
+     ($bundle->label == 'mRNA' or $bundle->label == 'transcript' or $bundle->label == 'Transcript'))  {
```
bradfordcondon: shouldnt this instead us the SO term for mRNA and/or transcript? instead of checking the label, I mean.

bradfordcondon (approved):
I suggest using the type_id instead of `($bundle->label)` to support users who have renamed their content types.

We also do not support CDS that were not added via GFF (ie that dont have featureloc entries).    This is likely to be rare/unusual but we should at least comment in the code to alert users.

spficklin:
@bradfordcondon I have adjusted the code to look at the term name rather than the label.  Good suggestion.  @shawnawsu that sounds right!

spficklin:
Merging.  I got an approval from @shawnawsu and @bradfordcondon

Issue #452, opened 2018-06-17
Some field tokens don't work in URLs


spficklin:
### System information

* Tripal Version:  3
* Drupal Version:  7.x
* PostgreSQL Version:  N/A
* PHP Version: N/A

### Issue description

It seems some tokens don't work correctly when building URLs for content types.  They are missing from the URL.

#### Steps to reproduce
Setting the mRNA URL to be mRNA/[obi__organism]/[schema__name]

#### Error messages and screenshots

```
error message
```

laceysanderson:
PR Merged.

Issue #355, opened 2018-04-27
Tripal Upgrade from 2.x to 3.x - What happens to URLs, URL Aliases and Redirects ?


isumoc:
* Tripal Version:  latest via github
* Drupal Version:  7.56
* PostgreSQL Version: 9.5.10
* PHP Version: 7.1.13

When migrating content from Tripal 2.x to Tripal 3.x you have the option of migrating your old URL aliases from node/<node_id> to bio_data/<entity_id> - ie updating the url_alias table entries.  This option is on the Drupal menu at /admin/tripal/storage/chado/migrate

The work is done here:
https://github.com/tripal/tripal/blob/445cbb56c50fc80f098f82f9f99bee6e73307356/tripal_chado/includes/tripal_chado.migrate.inc#L828

This does not update a redirection from the bio_data/<entity_id> url.

It is possible to set up "bulk update" data paths/URL Alias and redirection from this menu:
/admin/structure/bio_data/manage/bio_data_<bundle_id>

The work is done here:
https://github.com/tripal/tripal/blob/0819c4eb3cae34e6503edeb0ab66f7c503fd3227/tripal/includes/TripalEntityController.inc#L238

This sets up aliases and invokes the redirect module to update a redirect for that URL.

However, you may only use a quite restrictive 5 tokens from the "Available Tokens" list to set up the alias.  

Thus,  it appears to me that you won't be able use your URL pattern(s) from Tripal 2.x.  when using Tripal 3.x unless you plan to manually update the aliases and redirections needed.

I am not clear if this is all by design or otherwise - but I wanted to put it here as an issue having gone through the most recent code etc.

Thanks in advance for any clarity on this.

laceysanderson:
HI @isumoc,

When you migrate URL alias' from nodes to Tripal Entities, it moves all the current alias' for the node to the entity [here](https://github.com/tripal/tripal/blob/445cbb56c50fc80f098f82f9f99bee6e73307356/tripal_chado/includes/tripal_chado.migrate.inc#L837-L839):

```php
            UPDATE url_alias
            SET source = 'bio_data/" . $entity->entity_id .
            "' WHERE source = 'node/" . $entity->nid . "'";
```

Example: if you had an alias `species/Lens/culinaris` set up for `node/14` and the corresponding TripalEntity is `bio_data_10/46`. Before upgrade public.url_alias would look like this:

| pid | source | alias  | language           |
|----|--------|-------|----------------|
[...]

PR #455, opened 2018-06-18
Tokens not working in URLs

spficklin:
Issue #452 and #355 

## Type(s) of Change(s)

## Description

There were two problems with using tokens in URLs.  

First, the tokens were not "deep" enough.  For example, when setting the URL for a gene or mRNA content type you had available the [obi__organism] token but that was it!  There was nothing for the genus and species.  If you used it then it an empty string would be substituted for the token.  This pull request fixes that problem.  Now, the tokens reference the sub elements of a field so we now have tokens for genus and species.  Tokens should exist for any field that has sub elements and those elements are defined in the elementInfo() function of the field.

Second, non alphanumeric characters were being rewritten as underscores.   That causes problems with migrating from Tv2 aliases. 

Finally, I fixed a typo in the TripalJob class that I discovered while testing this fix. It was preventing me from re-running jobs.  Just take a look at the code below and you'll see the problem.  The if statement tries to evaluate the $path but it is set in the line below:
https://github.com/tripal/tripal/blob/50270e6c9c9b850a217b75d6eefdea0fe747d3d4/tripal/includes/TripalJob.inc#L161-L165

## Testing

The easiest test is to edit the gene content type and set the URL alias to this:
```mRNA/[obi__organism,TAXRANK:0000005]/[obi__organism,TAXRANK:0000006]/[data__identifier]```
Before using that URL string, check the 'Available Tokens' to make sure you see the tokens used in the string above.  

It will take time to update your URLs if you have thousands of records.  So, testing on a small site or different content type would go fastest.

## Additional Notes (if any):

### CODE REVIEW

**Comment by almasaeed2010** [comment]
Note that this PR fixes issue #469

**Comment by almasaeed2010** [comment]
While testing this PR, I get the following fatal error when I visit `/admin/content/bio_data`:

```
Error messagePDOException: SQLSTATE[42703]: Undefined column: 7 ERROR: column tripal_entity.label does not exist LINE 1: ....id AS entity_id, tripal_entity.bundle AS bundle, tripal_ent... ^: SELECT tripal_entity.id AS entity_id, tripal_entity.bundle AS bundle, tripal_entity.label AS tripal_entitylabel, tripal_entity.changed AS tripal_entitychanged, :entity_type AS entity_type, NULL AS revision_id FROM {tripal_entity} tripal_entity WHERE (tripal_entity.bundle = :db_condition_placeholder_0) ORDER BY tripal_entity.label ASC, tripal_entity.changed DESC LIMIT 25 OFFSET 0; Array ( [:db_condition_placeholder_0] => bio_data_1 [:entity_type] => TripalEntity ) in tripal_content_overview_form() (line 330 of /Sites/Tripal/sites/all/modules/tripal/tripal/includes/TripalEntityUIController.inc).
```

bradfordcondon:
i made a PR that fixes this and it has been merged on master.  i think you can safely ignore assuming theres not a commit that *reverts* it in this pr.  https://github.com/tripal/tripal/pull/458

almasaeed2010:
(I meant to add the following comment here but went to another issue by mistake oops!)

Ok I reviewed your PR against this and I don't see the issue being reintroduced. Merging this to master should not revert the PR made by @bradfordcondon 👍

laceysanderson:
Ok, Added the Reviewer #1 Approval tag for @almasaeed2010's review --Thanks!

One more review required before merge...

bradfordcondon:
Code looks good to me.  I followed your example (replacing mrna with gene in the url) and the job ran successfully.
However I would think that for a gene with ID Contig10011, for Fraxinus excelsior, the below url should be the new location.  However I get a page not found error?

http://localhost:8083/gene/Fraxinus/excelsior/Contig10011

bradfordcondon:
NONE of the URL aliases work on my site.    We checked my url_alias table and it looks good there.

Since @almasaeed2010 has functional confirmation, and the code looks good to me, i'm going to aprove given our discussion re: number of required functional tests.

spficklin:
That's odd the URLs didn't work for you and warrants investigation if you have time.    Thanks for merging.

PR #459, opened 2018-06-19
Publication Importer fix Tv3

spficklin:
Issue #438 

## Type(s) of Change(s)

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] API-specific change (fix or addition to an API function)
- [ ] Updates documentation (inline or markdown files)
This publication does not import properly:  https://www.ncbi.nlm.nih.gov/pubmed/?term=29792315.  The problem is that it has embedded HTML in the title and abstract (e.g. ```<i>Juglans</i>```).  That threw off the XML importer.  This PR fixes the problem.

## Description

Use this drush command to import the publication:
```
drush trp-import-pubs --dbxref=PMID:29792315 --username=[username]
```
(substitute the [username] tag)

Then publish new entities of type "Publication" and find the new publication page.  You should see it has a title and an abstract.  Previously the import would fail completely and not import or it would import partially and the title would be truncated and the abstract missing.

This is such a simple PR and it has already been tested by @bradfordcondon with issue #438 that I think this only needs on reviewer. 

## Additional Notes (if any):

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #438, opened 2018-06-14
drush pub importer


bradfordcondon:
### System information

* Tripal Version:3

### Issue description
I was really excited about the command line drush importer for pubmed citations.  However the command fails with the below error.  However my sql query shows i only have one cv and one cvterm for the term "Journal Article".

```
                                                           ^
hardwood_dev=> select * from chado.cvterm cvterm INNER JOIN chado.cv cv ON cvterm.cv_id = cv.cv_id where cvterm. name='Journal Article';
2241	22	Journal Article	The predominant publication type for articles and other items indexed for NLM databases.	2721	0	0	22	tripal_pub	Tripal Publiation Ontology. A temporary ontology until a more formal appropriate ontology an be identified.
```
hardwood_dev=>

#### Steps to reproduce
`drush trp-import-pubs --dbxref=PMID:29792315 --username=youradmin`

#### Error messages and screenshots

```
$ drush trp-import-pubs --dbxref=PMID:29792315 --username=hwadmin

NOTE: Loading of publications is performed using a database transaction.
If the load fails or is terminated prematurely then the entire set of
insertions/updates is rolled back and will not be found in the database

WD tripal_cv_api: chado_get_cvterm: The identifiers you passed in were not unique. You passed in Array                                             [error]
(
    [name] => Journal Article
    [cv_id] => Array
        (
            [name] => tripal_pub
        )

)
.
[site http://160.36.205.61:8095] [TRIPAL ERROR] [TRIPAL_CV_API] chado_get_cvterm: The identifiers you passed in were not unique. You passed in Array(    [name] => Journal Article    [cv_id] => Array        (            [name] => tripal_pub        )).
WD tripal_pubmed: Cannot find a valid vocabulary term for the publication type: "Journal Article".                                                 [error]
[site http://160.36.205.61:8095] [TRIPAL ERROR] [TRIPAL_PUBMED] Cannot find a valid vocabulary term for the publication type: "Journal Article".
WD tripal_pub: chado_publication_exists(): The Publication Type is a required property but is missing                                              [error]
[site http://160.36.205.61:8095] [TRIPAL ERROR] [TRIPAL_PUB] chado_publication_exists(): The Publication Type is a required property but is missing
WD tripal_pub: The Publication Type is a required property but is missing                                                                          [error]
[site http://160.36.205.61:8095] [TRIPAL ERROR] [TRIPAL_PUB] The Publication Type is a required property but is missing

Syncing publications with Drupal...
```

[...]

PR #460, opened 2018-06-19
Publication Importer fix Tv2

spficklin:
Issue #438 

## Type(s) of Change(s)

This publication does not import properly:  https://www.ncbi.nlm.nih.gov/pubmed/?term=29792315.  The problem is that it has embedded HTML in the title and abstract (e.g. ```<i>Juglans</i>```).  That threw off the XML importer.  This PR fixes the problem.   This is the Tripal v2 fix that was applied in PR #459 for Tv3.

## Description

Use this drush command to import the publication:
```
drush trp-import-pubs --dbxref=PMID:29792315 --username=[username]
```
(substitute the [username] tag)

Tripal v2 should have automatically synced the publication So, go find the new publication page.  You should see it has a title and an abstract.  Previously the import would fail completely and not import or it would import partially and the title would be truncated and the abstract missing.

This is such a simple PR and it has already been tested by @bradfordcondon with issue #438 that I think this only needs on reviewer. 

## Additional Notes (if any):

### DIFF
--- a/tripal_feature/includes/tripal_feature.gff_loader.inc
+++ b/tripal_feature/includes/tripal_feature.gff_loader.inc
@@ -650,7 +650,12 @@ function tripal_feature_load_gff3($gff_file, $organism_id, $analysis_id,
         $strand = -1;
       }
       if (strcmp($phase, '.') == 0) {
-        $phase = '';
+        if ($type == 'CDS') {
+          $phase = '0';
+        }
+        else {
+          $phase = '';
+        }
       }
       if (array_key_exists($type, $cvterm_lookup)) {
         $cvterm = $cvterm_lookup[$type];
--- a/tripal_pub/includes/importers/tripal_pub.PMID.inc
+++ b/tripal_pub/includes/importers/tripal_pub.PMID.inc
@@ -513,9 +513,7 @@ function tripal_pub_PMID_parse_article($xml, &$pub) {
           tripal_pub_PMID_parse_journal($xml, $pub);
           break;
         case 'ArticleTitle':
-          $xml->read();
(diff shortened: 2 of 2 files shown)


tripal_feature/includes/tripal_feature.gff_loader.inc
```diff
@@ -650,7 +650,12 @@ function tripal_feature_load_gff3($gff_file, $organism_id, $analysis_id,
         $strand = -1;
       }
       if (strcmp($phase, '.') == 0) {
-        $phase = '';
+        if ($type == 'CDS') {
```
bradfordcondon, line 653: This looks like code from another issue.  If you remove it I will approve and merge

spficklin: Oh there was no other issue. It was a fix that I copied from the Tv3 code that was missing in Tv2.   I didn't mean for that to get into this PR, but given that's it's the same as Tv3 code can we leave it in. so it gets merged.  Here's the same code in Tv3:

https://github.com/tripal/tripal/blob/44ba26f4bd17b5f52f6c042d0e9e58543db190fd/tripal_chado/includes/TripalImporter/GFF3Importer.inc#L565-L573

bradfordcondon: oh shoot im sorry, looks good then, lets merge.

spficklin: Oh, good call.  I'm glad you saw it.

spficklin:
Thanks @bradfordcondon

Issue #449, opened 2018-06-17
Annoyances


spficklin:
This is just a place to catch small annoyances that need fixing before the release but may or may not be bugs and bugs are minor.

spficklin:
Annoyance 1:  The content types are not sorted alphabetically

spficklin:
Annoyance 2:  The code for autocomplete for specifying a cvterm is duplicated in multiple places. It really should be one function so we can make fixes (if need be) in one place.

spficklin:
Annoyance 3: When changing the field layout for a content type the page loading is too slow.

bradfordcondon:
Annoyance 4:  I often want to look up the bundle and entity id given a chado record and table, but i dont think theres an API to do so.

edit: saw after i posted that this is "needs fixing before release" ... dont know if i feel that way about this.

spficklin:
I'm marking those that have been resolved with a horray emoji

bradfordcondon:
teh cvterm lookup box returns results that aren't valid for term creation.
For example:

<img width="742" alt="screen shot 2018-06-18 at 4 07 26 pm" src="https://user-images.githubusercontent.com/7063154/41559364-d78f163c-7311-11e8-81ed-aa08ac538a85.png">



<img width="673" alt="screen shot 2018-06-18 at 4 07 32 pm" src="https://user-images.githubusercontent.com/7063154/41559362-d3f97fd0-7311-11e8-8606-72b85cbdbfbd.png">

spficklin:
Oh, this is the same problem as the third comment above.   It's because that autocomplete term is a synonym but the lookup code isn't using synonyms.  It got fixed in one place but not in other places.

spficklin:
Annoyance 5: Data Searches should not search by default, but it should be set to require a button press before results are shown.

spficklin:
Annoyance 6:  The Pub Importer still has a link to the old help pages.

spficklin:
Annoyance 7: importing pubs results in warning messages if the pub already exists.

spficklin:
Annoyance 8: non Tripal fields aren't showing up in the web services.

spficklin:
Annoyance 9:  Tripal Bundles don't have the uid integrated with Views and there are no relationships for the user content.

spficklin:
Annoyance 10:  when importing pubs using the Tripal importer the publications are not automatically published like they were in Tripal v2.

bradfordcondon:
[...]

PR #471, opened 2018-06-23
Pub Importer Annoyances

spficklin:
Issue #449

## Type(s) of Change(s)

## Description

This PR fixes some of the annoyances described in issue #449 and one bug.  Here's the list
1)  The old link to the legacy help page is removed.
2)  The publications are now automatically published after importing.
3)  The AGL importer info has been commented out as the AGL importer is broken and no one has had to to work with AGL to try and fix it.  The problem seems to be with AGL.  So, it's been removed as an option for now.  
4)  Previously when adding a publication importer you could not mix AND and OR on a single criteria line, but if you used AND and a word with the letters 'or' in it (e.g. normalization) then it thought you were trying to use both AND and OR and would give an error message.  It not longer does that.
## Testing?

1) Before pulling this branch, go to Admin > Tripal > Data Loaders and click on  'Chado Publication Importers'.  Click 'New Importer' and enter the in the first line of criteria the search phrase 'gene expression AND normalization'.  Click the 'Test Importer' button.  You should see an error message about not using AND and OR in the same clause.  Next pull the branch, click 'Test Importer' button again and it should work.
2)  After pulling the branch you should no longer have 'USDA Agricultural library' and an option for Remote Database.  
3)  Before pulling the branch you should see on the page that lists all of the importers a link to the help page which when clicked goes to a non-existent page.   After pulling the sentence with the link as been removed.
4)  Create a brand new importer searching for whatever you want, but preferably something with just a few pubs.  For example, if you pick 'WGCNA' as the criteria and set the number of days to 14 you should only get 7 publications that match.   Save the importer, then click the "Import Pubs" link next to your new importer.  It will add a job.  Run that job and you should see at the end that all of the new publications that were imported are now also published as well.
## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/loaders/tripal_chado.pub_importers.inc
+++ b/tripal_chado/includes/loaders/tripal_chado.pub_importers.inc
@@ -74,10 +74,8 @@ function tripal_pub_importers_list() {
       <li>Create a new importer by clicking the 'New Importer' link above, and after saving it should appear in the list below.  Click the
           link labeled 'Import Pubs' to schedule a job to import the publications</li>
       <li>The first method only performs the import once.  However, you can schedule the
-          importer to run peridically by adding a cron job. See the " .
-     l("Pub Module help instructions", "admin/tripal/legacy/tripal_pub/help") . " to learn how to
-     set the importers to run automatically.") . '</li>
-     </ol><br>';
+          importer to run peridically by adding a cron job. </li> 
+     </ol><br>");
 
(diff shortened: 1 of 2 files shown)


laceysanderson:
I've tested this and confirmed it all works as described. 👍 Furthermore, I think the changes are minor enough to only need one review.

spficklin:
Thanks @laceysanderson

Issue #474, opened 2018-06-24
Permissions for importers


spficklin:
### System information

* Tripal Version: 3
* Drupal Version: 7.x
* PostgreSQL Version:  N/A
* PHP Version:  N/A

### Issue description

There is currently no way to set specific permissions for which users can use which TripalImporter importers.  Currently, the default permission is "load tripal data" but that is probably not specific enough especially if some trusted users are allowed to import genomic featuers (i.e. GFF files) while others should only import things like publications. 

An easy solution would be to automatically add a new permission for every TripalImporter instance.  Then Importer developers don't need to worry about providing permissions, Tripal will do it automatically and the site admin can easily set permissions on a per-loader basis.  This should not take too much code to implement.

bradfordcondon:
thanks to both of you.

PR #475, opened 2018-06-25
Importer permissions

spficklin:
Issue #474

## Type(s) of Change(s)

## Description

It was not possible to give specific users permissions to specific Importers.  There was only one permission to use them all.  Now each TripalImporter gets is own permission and a site admin can control who can use each one.

## Testing?

To test this you'll need two user accounts, one with admin permission and one without.   
1) First pull the branch and then clear the cache.   
2) Now go to the Permissions page and you should now see a set of permissions for all of the Tripal Importers.  They are grouped together with the prefix "Importer:". Set the permission for your non-admin user any importer you want to test, preferably something you have a small data set for.  Also, be sure the "Upload Files" permission is also set, otherwise user's won't be able to upload a file for importing.   Save, these settings.
3)  Log in to your site with the non-admin user and go to the page for the importer.  You will can just cut and paste the URL bar in the browser.  The page should show up.   
4)  As the non admin user try to use the loader.  It should successfully submit a job.
5)  Now go back to the site as the admin user, disable the permissions.  
6)  Return the the site as the non-admin user and the page for the importer should give an 'access denied error'.

## Additional Notes (if any):


laceysanderson:
This change is **not backwards compatible**... On update users who previously were able to load data via the "Load data into the Tripal site" permission now cannot. 

Instead perhaps the "Load data into the Tripal site" permission should provide permission to all loaders. If you want to restrict a user to a specific loader then give them only the specific loader permission (e.g. "Importer: Use the Chado GFF3 File Loader") and not "Load data into the Tripal site". Thoughts?

spficklin:
It's only not backwards compatible with previous versions of Tripal v3 which is still technically "beta".  So, I'm not sure it is that critical.  But if you feel it's necessary or you feel strongly that you need a quick "switch" to enable use of all loaders at once, then I can write a function to make it work.

laceysanderson:
I think currently it's confusing/non-intuitive. Either the Load data into the Tripal site" permission should be removed all together or act as a "quick 'switch' to enable use of all loaders at once". 

Since the second maintains backwards compatibility and saves effort for early adopters, I lean towards that.

However, an argument could be made that both specific permissions and an overriding general permission is confusing so :shrugs:. Either way works, it's just currently confusing so needs a change.

spficklin:
Some of the non TripalImporters (e.g. Tripal Pub Importer) still use that permission.  Would you be happy if we removed that permission and set individual ones for non TripalImporer so that we have consistency?

laceysanderson:
I'm always happy with consistency 👍

spficklin:
Okay, I'll make the change!

spficklin:
Okay, I have fixed this PR so that the 'load tripal data' permission is now gone.  It turns out the only data importer that did not use the new TripalImporter class was the bulk publication importer.  So, I wrote a class for it.  However, that code is a bit too complex for the TripalImporter class, so I added a new $callback, $callback_module and $callback_path static variables to the TripalImporter.  This way you can still take advantage of using the TripalImporter (i.e. consistency in permissions and menu placement).  Without this, the importer permission shows up in the 'tripal_chado' module section of the permissions page while all others show up in the 'tripal' module section. That's not very consistent.  With this, the importer now shows up with all other importers.

You'll need to clear you cache before trying this out.

laceysanderson (changes requested):
Ok, I did a full functional test on the permissions and they work as expected!
I created a new test user and gave it a test role, then I did the following:
- [x] I checked that I couldn't access any of the loaders with this test user with none of the loader-specific permissions checked.
- [x] I gave the user one permission at a time and checked that they now could access each loader in succession.
- [x] I also checked that they could only access loaders they had permission to and not ones they didn't during the above test.
- [ ] Checked that all the menu items were consistent
- [x] Checked that permission names were consistent and in the same category 
- [x] Checked that the bulk pub importer still worked. Specifically I checked I could add, edit, test and submit a bulk pub importer. I also checked the job submitted ran correctly ;-)
- [x] Checked that my test user could submit a loading job

### Menu items consistent
The names of the menu items are consistent 👍 However, they are all spilled out into the main Tripal menu :-( This is due to the removal of the containing "Data Loaders" menu item:
```diff
-
-  $items['admin/tripal/loaders'] = array(
-    'title' => 'Data Loaders',
-    'description' => t('Tools facilitating data import.'),
-    'access arguments' => array('load tripal data'),
-    'type' => MENU_NORMAL_ITEM,
-    'weight' => 6
-  );
-
```
Ideally I think the access arguments would include every single individual importer permission. That way if you have permission for at least one importer then you can see a page listing the importers you have access to. However, I've never had more then one permission in the 'access arguments' so we would need to confirm it does what I think it would (eval to true if you have any one of them).

laceysanderson (approved):
My commit fixes the issue I found in my function + code review above.

laceysanderson:
Now this PR just needs a code review 👍

tripal_chado/includes/TripalImporter/PubBulkImporter.inc
```diff
@@ -0,0 +1,132 @@
...
+   * If the $use_analysis value is set above then this value indicates if the
+   * analysis should be required.
+   */
+  public static $require_analysis = FALSE;
+
+  /**
+   * Text that should appear on the button at the bottom of the importer
+   * form.
+   */
+  public static $button_text = 'Import from NCBI Taxonomy';
```
bradfordcondon: I think this is wrong... maybe just Import?  

I'm submitting a commit with mostly spelling changes, I'll include this change with it.

bradfordcondon (approved):
Functionality confirmed, code looks good.

Thanks!

spficklin:
Thanks @laceysanderson and @bradfordcondon

Issue #481, opened 2018-06-26
Unwarranted warnings when checking for new fields


laceysanderson:
### System information

* Tripal Version: 3.x-dev
* Drupal Version: 7.x

### Issue description
I'm receiving a large number of warning on my Manage Fields page when checking for new fields.

##### Case 1: System tries to add a Chado Propery field even though it already exists.
For example, when checking for new fields on my publication content type, I see the following warnings:
```
A field of type tpub__publisher already exists, yet a property wants to use the same term. The property cannot be added.
A field of type tpub__year already exists, yet a property wants to use the same term. The property cannot be added.
```
This is because those fields already exist as shown in the following screenshot:
![screen shot 2018-06-26 at 12 22 00 pm](https://user-images.githubusercontent.com/1566301/41931530-9699d52e-793b-11e8-9d2f-a7fdbedc5299.png)
Note: I have done no customizing to this content type. All of the fields added here were added automatically by Tripal earlier and have not been changed.

##### Case 2: missing the "term_vocabulary" setting
For example, when checking for new fields on the gene content type, I see the following warnings:
```
    The field instance, data__accession, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, obi__organism, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, data__sequence, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, data__sequence_length, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, data__sequence_checksum, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, data__sequence_coordinates, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, ogi__location_on_map, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
    The field instance, schema__alternate_name, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.
```
As you can see in the term column of the following screenshot, the term is set for each of these fields:
![screen shot 2018-06-26 at 12 29 39 pm](https://user-images.githubusercontent.com/1566301/41931880-9ec2e348-793c-11e8-9481-1504ca3c398d.png)
The error is being generated here: 
[...]

PR #483, opened 2018-06-27
Fix for "field instance is missing the 'term_vocabulary' setting"

laceysanderson:
Issue #481 

## Type(s) of Change(s)

## Description

When checking for new fields we would see warnings that the term was not set even though it was. This was due to the vocab not being set in the $info array when the field was added to the bundle. This PR adds the vocab to the $info array for all the fields that were missing it.

## Testing?

1. Navigate to the "Manage Fields" page for a Tripal Content Type. For example, Admin >Structure > Tripal Content Type > Gene > Manage Fields.
2. Click on " + Check for new fields" link near the top of the page.
3. Confirm that there are no errors of the form `The field instance, data__accession, is missing the "term_vocabulary" setting. The field instance cannot be added. Please check the field settings.`

**Note: Make sure to test Gene, mRNA, and publication content types, at a minimum. There should be no warnings on any content type.**

### DIFF
--- a/tripal_chado/includes/tripal_chado.fields.inc
+++ b/tripal_chado/includes/tripal_chado.fields.inc
@@ -1458,6 +1458,9 @@ function tripal_chado_bundle_instances_info_custom(&$info, $entity_type, $bundle
         'chado_table' => $table_name,
         'chado_column' => $table_column,
         'base_table' => $table_name,
+        'term_accession' => '0100026',
+        'term_vocabulary' => 'OBI',
+        'term_name' => 'organism',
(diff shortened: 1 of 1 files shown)


laceysanderson:
Although this is more then 10 lines of code, I believe it only needs one review due to being a simple addition of term definitions.

spficklin:
I tested that I get the errors before loading the branch and after and indeed it does remove the warnings!  Thanks for fixing those!

laceysanderson:
Thanks @spficklin!

PR #470, opened 2018-06-22
Support for Views Tripal Entity Relationship

spficklin:
Issue #467 

## Type(s) of Change(s)

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] API-specific change (fix or addition to an API function)
- [ ] Updates documentation (inline or markdown files)

## Description

Currently, it's not possible to create relationships between a Tripal entity and other Drupal related tables such as Taxonomy.  So, for example, if you add the "Taxonomy Term Reference" field to any of your entities you will be able to tag them,  but you can never filter your entities by those tags.  This is because our code for entities doesn't support it in the TripalFieldQuery nor in Views.   This PR fixes both.  It also fixes some major problems with integrating results should we ever have other storage backends... which in fact is the Taxonomy example is one. 

I'm marking this as a release blocker because it fixes some major issues with queries.

## Testing?

To test do the following:

NOTE:  please run this test on a non critical machine.  This is because it has a very important update function that  adds a new setting to all Tripal fields to help distinguish them from non tripal fields.  I have tested the update on my machine and it worked fine but I would hate for the test to not work on a production machine.

1)  After pulling this branch you will need to run ```drush updatedb```.
2)  Go the Admin > Structure > Taxonomy and create a new vocabulary named "Publication Tags".
3)  Add some new terms to it.  You can pick anything that makes sense to you to tag publications (e.g. "Software", "Genome", etc.).
4)  Go to your publication content type:  Admin > Structure > Tripal Content Types.  Click the 'manage fields' for the Publication type.  Add a new field of type 'Term Reference'.  Name it 'field_pub_tags'. If you give it that exact name then the view I attach to this PR should work.  The following two screen shots show the settings I think are best for testing but it really shouldn't matter what you choose. tag_field_settings tag_field_settings2
Note: if you enable the term_reference module you get the nice tree view of terms.
5) Click on the 'Manage Display' field to move your new tags field out of the Disabled section and somewhere that you can see it on the page and save.
6) Now search for a publication entity on your site and add a new tag to by editing the publication and setting the tag.
7) Next create a new view that uses the Publication content type.  For a shortcut you can use the exported view here: [pub_tags-view.txt](https://github.com/tripal/tripal/files/2125795/pub_tags-view.txt).  Here's a screenshot of the settings for the view:  pub_tags-view
[...]

PR #488, opened 2018-06-28
fix warnings for pr 470

bradfordcondon:
This PR sets some defaults to prevent warnings if there are no results in the view.

Note this PR is to branch 467( pr #470 ) and is therefore not subject to standard PR guidelines.

### Issue

Pr #470  Adds support for views tripal entity relationship. However, if no results are returned because of the tripal field query, we get warnings.  This PR fixes that by setting some defaults.

### Testing

* Specify a tag in the view search, and also specify a publication year that will yield no results.  There will be no warnings with this PR.
 

Please see my [comments in the PR for images](https://github.com/tripal/tripal/pull/470#pullrequestreview-132831612).


spficklin:
Danke!

Issue #421, opened 2018-06-08
cv browser irregularities


bradfordcondon:
### System information

* Tripal Version: 3

### Issue description
im sorting an issue we had on [HWG](https://github.com/statonlab/hardwoods_site/issues/251) where I had duplicated terms in the tree for GO.

note molecular function regulator:  

![image](https://user-images.githubusercontent.com/7063154/41175152-25ce52e4-6b2a-11e8-82a8-4104c6b4e5df.png)

i went on the demo site to look for this term but didnt see it at all!  http://demo.tripal.info/3.x/cv/lookup/GO

heres the EBI entry:  https://www.ebi.ac.uk/ols/ontologies/go/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FGO_0098772

![screen shot 2018-06-08 at 2 44 07 pm](https://user-images.githubusercontent.com/7063154/41175236-6d39ab42-6b2a-11e8-9c84-a8da0794f642.png)

it should actually show up in tWO places: once under molecular function (direct child) and once under biological process ( ->b. regulation -> regulation of mo. function -> mo. function regulator).

So: does hte live site have the full GO loaded, or a subset?

Should the tree browser make sure it doesnt add duplicate terms?  How do the dupe terms get added? Is it a problem with our Chado records and if so what table?  If not, how can browser be corrected?

bradfordcondon:
first things first i guess we need to know if hte demo site has the full GO vocab loaded or not: @spficklin any ideas?

bradfordcondon:
took another look at this: easy fix.

`tripal_chado_vocab_get_term_children` did not select distinct in this query:

```php
 // Get the children.
  $sql = "
    SELECT subject_id
    FROM {cvterm_relationship} CVTR
    WHERE object_id = :object_id
  ";
```

when there were multiple relationship *types* in cvterm_relationship (ie the same subject/object ID's but distint type_ids) they got counted twice.

spficklin:
This looks like it's fixed from PR #492

PR #492, opened 2018-06-28
make child term lookup distinct  for chado cvterm get children

bradfordcondon:
Issue #421 

## Type(s) of Change(s)

## Description
* The cv browser can have duplicates, if the same subject/object relationship exists in cvterm_relationship with different type_ids.
* This simple PR will prevent duplicates from being returned.

## Testing?
Testing is a bit tricky because it depends on the OBO files you have loaded.  I dont find duplicates in an easy OBO file (IE goslim) so i dont have a quick way to test for you.  The demo site also doesnt have this issue for some reason- I'm assuming its because the same OBO was loaded with different versions of the OBO loader.

Before:
 (note duplication of the term molecular function regulator)

image

After:
screen shot 2018-06-28 at 4 20 47 pm

## Additional Notes (if any):


spficklin:
Easy enough fix and simple to review in terms of just looking at the code. I see no problems with it.

Issue #496, opened 2018-06-30
Publication page layout is a mess after first installation


spficklin:
### System information

* Tripal Version:  3.x
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: 7.2

### Issue description

On a fresh in stall of Tripal, when following the online User's Guide it instructions the user to add a publication.  After saving the publication many of the fields are not in a Tripal Pane and it looks like a mess:

![screenshot from 2018-06-29 20-11-14](https://user-images.githubusercontent.com/1719352/42120919-a2300346-7bd8-11e8-87df-26cc6da0a7a8.png)

If I go back to the Publication content type and re-apply the default layout and click "Save" then it fixes the problem.  Those fields that are showing up without being in a Tripal Pane look like they are disabled but they still show up.

shawnawsu:
@spficklin  

Publication is the only content type with custom handling in the tripal_ds module. It looks like a bunch of field names changed, did that happen? If it did, can we connect to go over the new preferred layout?

spficklin:
I don't know how field names changed.  Do you know the before and after names of the fields?   Let's chat when I'm back in town.

shawnawsu:
Apparently, there was a problem with my local, when I delete the publication content type and rebuilt it half the fields disappeared which is why I thought fields were renamed. They are not renamed, they just disappeared.

Given that I had another look and found the problem. I pushed a fix in pull request #508.

laceysanderson:
Merged PR fixes the publication page so it's no longer a mess 👍 

I wonder if we shouldn't discuss a better default though. I think it would be pretty generic to change the settings as follows and **it would show a lot more information by default**. Currently it's annoying for an admin to drag every single chado property into a pane ;-) ([screenshot of settings](https://user-images.githubusercontent.com/1566301/42963500-a493617c-8b51-11e8-80a8-eac311d79cc7.png))

- pull the abstract out of the summary table but within the summary pane
- show all other chado property fields in an "Metadata" pane
- also show the following fields in the "Metadata" pane: publisher, publisher location, page numbers, year, issue, series name, volume, title of the volume.

![firefox_screenshot_2018-07-19t18-44-20 909z](https://user-images.githubusercontent.com/1566301/42963488-935e6bd6-8b51-11e8-8502-a2fc0479816b.png)

PR #508, opened 2018-07-18
Publication layout fix

shawnawsu:
Issue #496

## Type(s) of Change(s)
- [X] Bug fix (non-breaking change which fixes an issue)

## Description
The publication ds layout wasn't working, ie the fields were not coming through as disabled and therefore on initial content type build they all went into the right side and didn't get menu links. When the tripal_ds module was initially built the publications received a different and specific layout which was not being applied but is in this branch. 

The problem was being caused by a very small typo, so I corrected that and everything appears to be working on my local.

## Testing?
To test:

1. Checkout 496-tv3-pub_layout_fix, pull, drush cc all
2. Delete the publication content type here admin/structure/bio_data.
3. Add the publication content type back admin/structure/bio_data/add. In the autocomplete field type pub, choose the first publication in the drop down. In the 'chado table' select list choose 'pub' and then choose 'yes' under 'Are all records in the "pub" table of type "publication". Like this:
screen shot 2018-07-18 at 9 40 08 am
4. Click 'create content type'.
5. Now create a test publication and ensure there are no random fields in the display, it should NOT looks like this: 
42120919-a2300346-7bd8-11e8-87df-26cc6da0a7a8
It should look like this:
screen shot 2018-07-18 at 10 37 35 am

I also tested through a new site creation to make sure the layout was correct when a new site is stood up.

## Additional Notes (if any):
- [X] My code follows the code style of this project.

### DIFF
--- a/tripal_ds/includes/tripal_ds.ds.inc
+++ b/tripal_ds/includes/tripal_ds.ds.inc
@@ -304,7 +304,7 @@ function _ds_layout_pub_settings_info($bundle_name, $instances) {
         field_update_instance($instance);
       }
       elseif($instance_name == 'schema__additional_type' || $instance_name == 'tpub__doi'
-        || $instance_name == 'tpub__publication_date' || 'tpub__abstract' ||
+        || $instance_name == 'tpub__publication_date' || $instance_name == 'tpub__abstract' ||
         $instance_name == 'tpub__citation') {
         array_push($properties, $instance_name);
       }


laceysanderson (approved):
Code ✅ 
Works as described ✅

laceysanderson:
I tested this PR as Shawna suggested and ended up with the simplified layout:
firefox_screenshot_2018-07-19t18-25-46 847z

Since this PR is such a small type, I agree it only needs one review and am merging it :-) Thanks @shawnawsu for tracking this one down!

PR #510, opened 2018-07-19
Added more error checking to the Tripal Bundle create function.

laceysanderson:
Issue n/a

## Type(s) of Change(s)

## Description

Added more error checking to the create tripal bundle function to make it easier to debug when something goes awry. Most of the changes are simply adding checks that variables were populated and if not, reporting a tripal error. 

The only change that's a little more involved relates to `unset($transaction);` which formally commits the transaction. I added this in when experiencing some issues with a postcreate hook which was causing my entire content type to be rolled back. Since postcreate happens after the bundle is already created, I don't feel it should uncreate the bundle if an error is encountered here. Thus this PR commits the transaction once the content type is created successfully and then executes any postcreate hooks.

## Testing?

A. Check that you can add a Tripal Content Type through the interface without problems :-)
     - Go to Admin > Structure > Tripal Content Types > Add Tripal Content Type
     - Fill out the form as you normally would and click "Create"
B. Code review :-) This will be the most valuable on this PR as the changes arn't seen during normal operation and there are no known bugs with content type creation :-)

## Additional Notes (if any):

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -442,9 +442,9 @@ function tripal_create_bundle($args, &$error = '') {
         $vocab->save();
       }
       else {
+        $transaction->rollback();
         tripal_report_error('tripal_entities', TRIPAL_ERROR, 
           'Unable to create TripalVocab :vocab', array(':vocab' => $vocabulary));
-        $transaction->rollback();
         return FALSE;
       }
     }
@@ -461,9 +461,9 @@ function tripal_create_bundle($args, &$error = '') {
         $term = $term->save();
       }
       else {
+        $transaction->rollback();
         tripal_report_error('tripal_entities', TRIPAL_ERROR,
           'Unable to create TripalTerm :term', array(':term' => $term_name));
-        $transaction->rollback();
         return FALSE;
       }
     }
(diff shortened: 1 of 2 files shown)


spficklin (approved):

spficklin:
All of the changes that you proposed look reasonable and implemented well, so I approve the code changes.   I was able to create the new content type.  However, my site timed out and I got the WSOD.   It took longer than 30 seconds for the page to return.  I don't think that's related to these changes so I approve.  But we may want to watch out for that behavior.

laceysanderson:
I have noticed the timeout quite a few times: both before and after implementing these changes so I can confirm it's not related to the error checking. I agree we should definitely add it to our list to look into though -I simply increased my page timeout...

almasaeed2010:
Hello @laceysanderson 

I tested the branch by creating a new Polypeptide bundle and it worked perfectly fine. 

I can also confirm that the delay/timeout happens with and without these changes. I had to resort to increasing php's max script execution time to deal with this issue.

Thanks!

almasaeed2010 (approved):

bradfordcondon:
2 approvals, thanks @laceysanderson @spficklin @almasaeed2010

laceysanderson:
Awesome! Thanks @almasaeed2010 and @spficklin :-D

PR #520, opened 2018-07-24
Some empty fields rendered when Tripal pane not hidden on display.

laceysanderson:
Issue: none open

## Type(s) of Change(s)

## Description

Empty fields are not supposed to be rendered even when the Tripal pane containing them is set to show by default. This allows admin to indicate they want a pane shown by default if there is data for that particular page but not otherwise. However, I noticed a few fields on feature-based content types that were not respecting this: specifically ogi__location_on_map, sio__annotation, and schema__publication.

## Testing?

1. Ensure you have a feature-based content type (e.g. gene) with the above fields added to the display.
2. On the "Manage Display" UI for that content type, click the settings gear for the Tripal pane containing each field and uncheck the "Hide panel on page load" checkbox then click "Update" for each pane. Finally, remember to click "Save" at the bottom of the page or your changes will not take effect.
3. Clear the cache :-)
4. Create or navigate to a page where there is no data for those three fields. 
    - on 7.x-3.x (left screenshot): you should see an empty table for ogi__location_on_map, sio__annotation and an empty pane for schema__publication. Note: the Tripal panes will appear on the page despite not being included in the table of contents.
    - on empty_fields branch (right screenshot): all three fields should not be rendered including the Tripal pane they are in (assuming they are the only field in the tripal pane).

### Left: 7.x-3.x; Right: this branch
- Panes to notice: "Location on Map", "Annotations" and "Publication". These panes should not be shown because they have no data for this page.
screen shot 2018-07-24 at 3 16 59 pm

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/TripalFields/ogi__location_on_map/ogi__location_on_map.inc
+++ b/tripal_chado/includes/TripalFields/ogi__location_on_map/ogi__location_on_map.inc
@@ -321,6 +321,12 @@ class ogi__location_on_map extends ChadoField {
         $i++;
       }
     }
+
+    // If there are no map positions expanded above then remove the stub.
+    // This is needed to ensure this field isn't displayed when there are no locations.
+    if (!isset($feature->featurepos->feature_id) OR (sizeof($feature->featurepos->feature_id) == 0)) {
+      unset($entity->{$field_name});
+    }
   }
 
 }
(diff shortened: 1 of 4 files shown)


shawnawsu:
I tested this and it works as described.

laceysanderson:
Awesome! Thanks @shawnawsu :-)

bradfordcondon:
thanks for this @laceysanderson .  elegant solution.  👍 sorry for the delay.

laceysanderson:
Thanks :-)

PR #522, opened 2018-07-26
Removed old views patch and update the new patch with the same file as in Tripal v3

chunhuaicheng:
The old Views patch (i.e. views-sql-compliant-three-tier-naming-1971160-22.patch) no longer works. It was removed in Tripal v3 and should also be removed from v2 since it confuses user when applying the patch.
The new patch (i.e. views-sql-compliant-three-tier-naming-1971160-30.patch) was also updated with the same file provided by Tripal v3 which had only space difference.

We only need one review of the PR because it's not a major change, just a small bug fix.


spficklin (approved):

spficklin:
This is just a simple replacement of a badly formatted patch for views that worked it's way into Tv2 and no one noticed. I confirmed there is no difference with this patch and the same one offered with Tv3.  This is simple PR so I'm marking as only needing one review and since I did the review I'm going to merge.

Issue #523, opened 2018-07-29
Alternate Relationship field formatters


laceysanderson:
## Feature Request/Discussion

- [ ] does this feature attempt to solve an existing problem with Tripal?
- [x] are you open to developing or collaborating on an extension module if this is not a good fit for Tripal Core (no pressure here -just good to know upfront :-) )
- [x] do you **Need** this feature ASAP?

### Description

Currently the relationship formatter for Tripal 3 states all relationships in sentence form in a single-columned table. This is quite different from the format for relationships used in Tripal 2 with a multi-columned table per relationship type.

![relationship_formatter_screenshot_from_demo](https://user-images.githubusercontent.com/1566301/43369694-86674b6a-932f-11e8-8512-303575db6675.png)

#### Table per Relationship Type Formatter
This formatter would mimic the Tripal 2 format providing one table per remationship type with the columns shown in the above screenshot (name, uniquename, species, type of related item; eg. stock). I think this one belongs in core since it allows admin to mimic the old pages during upgrade.

~~Comma-separated List for a single relationship type.~~

~~This formatter would allow the admin to choose a single relationship type and either object or subject to display. Then it would simply show a list with the `name (uniquename)` for each relationship of the type. Choosing either subject or object in the formatter makes it flexible enough to handle any relationship type. Furthermore, it would link to the related page the same way the default formatter does.~~

~~Using this formatter the admin can display related items in the summary table (for example). I think this formatter would be generic enough and broadly useful to be included in Tripal Core.~~

### Your Specific Use Case

I need these two formatters as part of the upgrade of KnowPulse to Tripal3. I'm trying to make as few UI changes for the user as possible during this upgrade and thus would like to mimic the old relationship format. 

~~Additionally, I had modified templates in the past to include a list of related items in the summary table for quick jumping between types. For example, my Variant pages have a row for "Markers" in the summary table with a simple list of the features related to the current one via the "is a marker of" relationship. I need a way to duplicate this in Tripal 3 and thus will be developing the formatter regardless.~~

### Generally Applicable

I feel a "Table per Relationship Type Formatter" is generally applicable since it mimics the Tripal2 format. As such, some sites may prefer this format over the new one during upgrade in order to minimize the UI changes during upgrade.

~~I feel a "Comma-separated List for a single relationship type formatter" is generally applicable since it allows condensing of related items into the summary table and is a really simple, comma-separated format.~~

### Additional information/screenshots

[...]

PR #528, opened 2018-08-03
Additional Relationship Formatter

laceysanderson:
Issue #523

## Type(s) of Change(s)

## Description

Currently the relationship formatter for Tripal 3 states all relationships in sentence form in a single-columned table. This is quite different from the format for relationships used in Tripal 2 with a multi-columned table per relationship type.

This formatter mimics the Tripal 2 format providing one table per relationship type with the columns name, uniquename, species, type of related item; eg. stock. I think this one belongs in core since it allows admin to mimic the old pages during upgrade.

Things of interest:
- Supports related tables where there is no uniquename and/or organism_id by simply not including those tables.
- Required backwards compatible update to the sbo__relationship field to load the organism information.
- Should be similar performance as the original formatter (both loop through all relationships to add to a table)
- Doesn't support paging since Tripal 3 `TripalFormatter->ajaxifyPager()` currently only supports a single pager per field.

## Testing?

1. Pick a content type with relationships to test and navigate to Admin > Structure > Tripal Content Types > [your content type] > Manage Fields.
2. Clear the cache and click find new fields.
3. Navigate to Manage Display, you should now see two items in the format drop-down for the relationship field: "Relationship Statements" (original formatter) and "Relationship Table" (this formatter).
4. Switch the formatter to "Relationship Table" and click "Save" at the bottom of the page.
5. Navigate to an entity page of that type with relationships and confirm the field now looks similar to the screenshot below.

The following screenshot shows an example of the relationship formatter in use on a stock-based content type. As you can see, it follows the relationships in both directions (current entity can be subject or object) and provides the same readable sentence describing the relationship above the table the way it was done in Tripal 2.
screen shot 2018-08-03 at 8 57 08 am

The following screenshot shows the configuration of this formatter. It provides default "sentences" for use at the top of each table as shown in the screenshot above. The admin can change these sentences to make them more intuitive to their user base. Supports use of a few tokens for embedding type information in the sentence and different sentences based on whether the current entity is the subject or object in the relationship.
screen shot 2018-08-03 at 8 58 42 am

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
@@ -304,10 +304,18 @@ class sbo__relationship extends ChadoField {
 
     // If the subject or object have a unqiuename then add that in for refernce.
     if (property_exists($relationship->$subject_id_key, 'uniquename')) {
-      $entity->{$field_name}['und'][$delta]['value']['local:relationship_subject']['data:0842'] = $relationship->$subject_id_key->uniquename;;
+      $entity->{$field_name}['und'][$delta]['value']['local:relationship_subject']['data:0842'] = $relationship->$subject_id_key->uniquename;
     }
     if (property_exists($relationship->$object_id_key, 'uniquename')) {
       $entity->{$field_name}['und'][$delta]['value']['local:relationship_object']['data:0842'] = $relationship->$object_id_key->uniquename;
+    }
+
+    // If the subject or object have an organism then add that in for reference.
+    if (property_exists($relationship->$subject_id_key, 'organism_id')) {
+      $entity->{$field_name}['und'][$delta]['value']['local:relationship_subject']['OBI:0100026'] = $relationship->$subject_id_key->organism_id->genus . ' ' . $relationship->$subject_id_key->organism_id->species;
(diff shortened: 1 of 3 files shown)


spficklin (approved):

spficklin:
Works great!  I approve, although there is one oddity.  Notice in the relationship table in the screenshot below that it says "The following Genes(s) are Part Of of this Gene".   Shouldn't it read: "The following mRNA(s) are Part of this gene".... The text in the screenshot has "Gene" instead of "mRNA" and there are two of's.

image

spficklin:
I'm marking this as approved as I think a fix for the text I pointed out doesn't need my input again.

laceysanderson:
I've fixed the wording issues you pointed out -Thanks @spficklin!

**NOTE:** To test the new defaults, you need to copy the following into the configuration form on the "Manage Display" page. **This is only needed if you have tested before this change was made.**
 - Subject Caption: `This @content_type is <em>@rel_type</em> the following:`
 - Object Caption: `The following are <em>@rel_type</em> this @content_type:`

chunhuaicheng:
It works great on my site. Thank you for the new feature.

laceysanderson:
Awesome :-) Thank you @chunhuaicheng!

Issue #77, opened 2017-04-01
Creating of proteins by GFF3 loader should be optional


spficklin:
Copied from the Drupal issue queue:  https://www.drupal.org/node/2550979

Creation of protein features when importing a GFF3 file was recently added. This was done to make the Tripal GFF loader similar to the perl Chado loader. However, this functionality should be optional as those protein features are technically not in the GFF3 file and may confuse folks when they unexpectedly appear.

bradfordcondon:
Do we still want this feauture?  Should the default still be to ignore them?  I would think the default should be to create, with the option to ignore....

bradfordcondon:
assigning myself because i implemented and wrote some tests because why not.

bradfordcondon:
Can I get some suggestion for the text?

>      '#description' => t('Check the box if you would like to create the proteins described in the GFF file.'),

Also @laceysanderson  we decided to add a category as "loader" for test locations right?  I need to update the README as well...

spficklin:
Issue fixed with PR -- long time overdue.

PR #529, opened 2018-08-06
add checkbox for proteins to be created GFF loader

bradfordcondon:
Issue #77 

## Type(s) of Change(s)

## Description

#77 details request to make creating proteins optional, by deafult not creating them.  This accomplishes that.

screen shot 2018-08-06 at 8 57 05 am
  

## Testing?

I add unit tests for a) the GFF loader itself, and b) this specific change, where we run the loader with defaults and find no protein, and with the new "create proteins" option checked.

If you would like to test, go to the GFF3 Tripalimporter and run the importer without the box checked: proteins should not be created.  Rerun the importer with teh box checked, and proteins should be created.

you can use the GFF3 file in the test (Tripal DevSeed) or your own.

## Parting thoughts

* I think tests belong in `tripal_chado/loaders/` ?  Previous discussion with lacey, i cant find it for sure though.
* Any requested changes to the text/UI?
* I opted to use a remote GFF file.  I pledge that the location of these devSeed files for F excelsior will not change since I'm using them as my "remote development seed data solution".  However if we prefer we can include it with the tripal repo instead....


laceysanderson:
Response to **Parting Thoughts**:
- Yes, we decided tests for loaders should be in `tripal_chado/loaders/`
- I'm content with the remote GFF3 since this causes less size increase to Tripal Core. I accept your pledge that the path will remain stable and will hold you to it ;-p 
- As far as UI goes, we need to make sure it's clear that we will always create proteins if they are specified in the GFF3 (have their own line). The checkbox should make it optional for us to not infer proteins from transcripts and create them. **or at least I think that was the intent? @spficklin**

Looks like the travis build is failing due to a path issue?
> The command "composer update" exited with 0.
> 0.42s$ ./vendor/bin/phpunit
> PHP Fatal error:  require_once(): Failed opening required '/home/travis/build/tripal/drupal/sites/all/modules/tripal/tests/tripal_chado/loaders/../../tripal_chado/includes/TripalImporter/TaxonomyImporter.inc' (include_path='.:/home/travis/.phpenv/versions/5.6.32/share/pear') in /home/travis/build/tripal/drupal/sites/all/modules/tripal/tests/tripal_chado/loaders/TaxonomyImporterTest.php on line 8

bradfordcondon:
>As far as UI goes, we need to make sure it's clear that we will always create proteins if they are specified in the GFF3 (have their own line). The checkbox should make it optional for us to not infer proteins from transcripts and create them. or at least I think that was the intent? @spficklin



* i agree thats the intent.  And I don't think thats how i implemented it, i just made proteins optional or not, oops.  Are proteins ever directly specified rather than inferred?

* it sounded to me like the issue was written that users should opt **in** to creating proteins, not opt out.

* yeah i made trivial changes and the build failed, i'll have another look when i have a second.  Once we talk about how th form should look il ping and say ready for review...

bradfordcondon:
oops i see the problem: i moved the taxonomy importer test, which pointed to the loader with a relative path.  I should add to the  instructions to use `module_load_include()` to prevent this.

spficklin:
@bradfordcondon. I think we should keep the original behavor that the protein is created automatically. This checkbox should turn that off rather than turn it on (as the current wording implies).  Can we change the wording of that checkbox to something like

> "Skip automatic protein creation"

And then for the description state

> The GFF loader will automatically create a protein feature for each transcript in the GFF file if a protein feature is missing in the GFF file.  Check this box to disable this functionality. Protein features that are specifically present in the GFF will always be created.

bradfordcondon:
thanks y'all!  

>. I think we should keep the original behavor that the protein is created automatically. This checkbox should turn that off rather than turn it on (as the current wording implies)

I agree with this, and ill do it this way.

bradfordcondon:
@spficklin do you have an example GFF that explicitly defines a protein to test that it is still created?  I cant find one.  Is it like this?

```
##gff-version 3
Contig0	FRAEX38873_v2	gene	16315	44054	.	+	.	ID=FRAEX38873_v2_000000010;Name=FRAEX38873_v2_000000010;biotype=protein_coding
Contig0	FRAEX38873_v2	mRNA	16315	44054	.	+	.	ID=FRAEX38873_v2_000000010.1;Parent=FRAEX38873_v2_000000010;Name=FRAEX38873_v2_000000010.1;biotype=protein_coding;AED=0.05
Contig0	FRAEX38873_v2	polypeptide	16315	44054	.	+	.	ID=FRAEX38873_v2_000000010.1.3_test_protein;Parent=FRAEX38873_v2_000000010.1

```

bradfordcondon:
confirmed that polypeptides as described above are created if the `ignore_proteins` parameter is set to 1.

This PR is ready for testing.

bradfordcondon:
resolved conflicts with @almasaeed2010 PR, which was doing the same thing (updating test suite).

almasaeed2010:
Tested and it worked for me.

PHPUnit tests pass successfully

Other tests:
1. Loaded GFF file while checking the `skip proteins` checkbox.
1. Checked the DB for proteins , none were found.
1. Load GFF file without checking the checkbox.
1. Checked the DB for proteins, proteins exist.

Thanks @bradfordcondon

almasaeed2010 (approved):

spficklin:
Sorry this took so long for a second review.  I can verify it works.  I loaded the same GFF3 file each time with the skip protein creation box checked or unchecked and protein features were created/not created correctly.

PR #532, opened 2018-08-08
Alternate Tripal2-like Sequence Coordinate Formatter

laceysanderson:
Issue n/a

## Type(s) of Change(s)

## Description

Currently the sequence coordinate formatter for Tripal 3 lists all coordinates in compact form. This is quite different from the format for the featureloc alignments used in Tripal 2 with a multi-columned table breaking down the location.

This formatter mimics the Tripal 2 format providing a table with the name, location and strand of the features the current entity is aligned to. I think this one belongs in core since it allows admin to mimic the old pages during upgrade.

**NOTE: This PR also includes a small bug fix to the sequence coordinate field. Originally, it was assuming the strand was set in featureloc so if it wasn't, the strand was reported as -. I just expanded the if-else to be more specific and catch the case where strand is not set.**

## Testing?

1. Pick a content type with sequence coordinates (featureloc records) to test and navigate to Admin > Structure > Tripal Content Types > [your content type] > Manage Fields.
2. Clear the cache and click find new fields.
3. Navigate to Manage Display, you should now see two items in the format drop-down for the sequence coordinate field: "Sequence Coordinate List" (original formatter) and "Sequence Coordinate Table" (this formatter).
4. Switch the formatter to "Sequence Coordinate Table" and click "Save" at the bottom of the page.
5. Navigate to an entity page of that type with sequence coordinates and confirm the field now looks similar to the screenshot below.

screen shot 2018-08-08 at 11 42 33 am

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/TripalFields/data__sequence_coordinates/data__sequence_coordinates.inc
+++ b/tripal_chado/includes/TripalFields/data__sequence_coordinates/data__sequence_coordinates.inc
@@ -243,8 +243,11 @@ class data__sequence_coordinates extends ChadoField {
         if ($featureloc->strand == 1) {
           $strand = '+';
         }
-        else {
+        elseif ($featureloc->strand == -1) {
           $strand = '-';
+        }
+        else {
+          $strand = '';
         }
         $fmin = $featureloc->fmin + 1;
         $fmax = $featureloc->fmax;
--- a/tripal_chado/includes/TripalFields/data__sequence_coordinates/data__sequence_coordinates_formatter.inc
+++ b/tripal_chado/includes/TripalFields/data__sequence_coordinates/data__sequence_coordinates_formatter.inc
@@ -2,7 +2,7 @@
 
 class data__sequence_coordinates_formatter extends ChadoFieldFormatter {
   // The default lable for this field.
-  public static $default_label = 'Sequence Coordinates';
+  public static $default_label = 'Sequence Coordinate List';
 
   // The list of field types for which this formatter is appropriate.
   public static $field_types = array('data__sequence_coordinates');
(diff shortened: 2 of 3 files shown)


laceysanderson:
Added configuration to allow admin to expand the strand to full words (default):
screen shot 2018-08-08 at 1 54 10 pm

spficklin (approved):

spficklin:
Works great!  Given that you've already submitted a similar PR for the relationship field and that was tested I think this one only needs 1 review.  But I'll not merge in case you think it needs two.

spficklin:
Oh, one thought.  I see this text above the table:  

```This Gene is aligned to the following:```

The word 'aligned' might be confusing.   An alignment implies an analysis was performed to align a sequence.  But often gene models are not aligned but predicted.  Would it make sense to change the wording from "aligned to" to "located at", or "found at"?

bradfordcondon:
I was about to review this but see @spficklin has approved and tagged as 1 reviewer.  I agree with his suggestion that aligned is maybe not the best wording.  

Associated with seems to be the safest wording, to me (albeit less accurate since it doesnt communicate location).  


  @laceysanderson thoughts?

laceysanderson:
I went with `aligned` trying to be generic -specifically to avoid using the type of the parent in the sentence since I didn't have access to that and didn't want to increase load time. The sentence can be configured per content type to whatever the admin wants... 

That said, I'm Very open to changing it if either of you have a more generic suggestion? Perhaps, `This @content_type is located on the following:`?

bradfordcondon:
> The sentence can be configured per content type to whatever the admin wants...
this is important to keep in mind.

How about instead: has the following sequence coordinates

since thats the most direct use of the cvterm for hte field?

laceysanderson:
There, I switched it to `This @content_type has the following sequence coordinates:` for maximum generic-ness **since it can be configured by the admin** and this will always be correct/make sense. Thanks for the suggestion @bradfordcondon and thank you both for the review/feedback!

I'm leaving the PR unmerged so @spficklin has the chance to look at the change and give feedback on it. Can you merge if you agree @spficklin ?

spficklin:
Thanks for merging. I've been a bit distracted on other projects. I like the updated change.

Issue #530, opened 2018-08-07
Publishing Entities causes  serious performance issues


laceysanderson:
### System information

* Tripal Version: 7.x-3.x-dev
* Drupal Version: 7.x
* PostgreSQL Version: n/a
* PHP Version: n/a

### Issue description

Currently publishing entities can cause a site to become unusable due to excessive slowness -it can even cause timeouts. Furthermore, when publishing a large number of entities it can cause continuous timeouts making the site completely unavailable. For example, I recently tried to publish my SNP pages in KnowPulse (~426K) and after running for 15hrs the site become completely unresponsive for >2hrs requiring me to kill the job rolling back all progress.

#### Steps to reproduce

1. Go to Admin > Structure > Tripal Content Types and click "Publish Tripal Content"
2. Fill out the drop-down to select a content type for which you have a large amount of data and submit the job.

#### Error messages and screenshots

There are no error messages or screenshots for this issue since it's performance rather then errors that are the problem.

laceysanderson:
My first step on this is going to be to allow the admin to specify in the publish form to only publish a given number of entities. We had this feature in Tripal2 and it can be used to effectively chunk the publishing process into more reasonable length transactions.

Other **potential optimizations** based on looking at the code:
- Optimize, estimate or remove the count query used to report progress. We can't show the progress through the UI anyway due to the transaction so how useful is this?
- Make a way to create multiple entities at once to take advantage of bulk inserts rather then one at a time.
- Attempt to optimize query built to select chado records to publish

I also need to evaluate the transaction more. Long running transactions can have serious performance issues in PostgreSQL due to the ACID-compliance keeping those records cordoned off until the transaction is committed --you should have seen the christmas lights in htop when I had to cancel my publish job ;-p). One option is to move the transaction within the loop so that each one is not very long but then we end up with more overhead creating transactions. A better albeit more complicated approach might be to break the job into chunks where each one is a single transaction. I'll do some timing on the various options as far as this goes.

laceysanderson:
So interestingly, I didn't actually do any of the potential optimizations above. Instead, once I started profiling where the time was spent, I found a lot of re-loading of the same objects -removing this sped things up drastically. Furthermore, there was a  good speed increase once I broke the job into chunks where each one is a single transaction. Combined these two optimizations were enough for my purposes and thus are submitted in the above PR.

spficklin:
PR with fix merged.  Closing.

PR #536, opened 2018-08-09
Optimization of Publishing Entities

laceysanderson:
Issue #530 

## Type(s) of Change(s)

## Description

This PR optimizes the performance of Publishing Tripal Content. Specifically, I see a performance increase from 2.6 minutes to 1.17 minutes to publish 500 entities. Furthermore, publishing no longer grinds my site to a halt but instead there seems to be barely an impact to users.

Optimizations done:
- The bundle  and term objects were being loaded (4 and 1 time(s) respectively) per entity published. This is now cached by loading it once at the beginning of the publish job and then passing it to $entity->create(), $entity->save(), $entity->setAlias(), and $entity->setTitle() through a generic $cache variable. In all cases, I still load the bundle/term if it's not set in the cache making this a backwards compatible change.
- ~~There were nested transactions (publish and $entity->save) which caused unnecessary overhead. Instead I used the same $cache variable to pass the transaction through the save process.~~
- The Drupal Field cache is cleared for each entity saved to ensure new values can be loaded. However, during publish we don't attach fields so I added a conditional to ensure this cache isn't cleared during publish.
- I chunked the publishing process into 500 entity chunks to drastically speed up the select query, limit transaction time and allow reporting of progress to the jobs UI.

Additional Note: There was a comment in $entity->setAlias() implying `SELECT DISTINCT SUBSTRING_INDEX(source, '/', 1) AS path FROM url_alias` was causing performance issues during publish. I tracked this query down to drupal_path_alias_whitelist_rebuild() but didn't see it called during publish personally. We may want to look into this further.

## Testing?

1. Navigate to Admin > Tripal Content > Publish Tripal Content
2. Select a Tripal Content Type you have unpublished content for and submit the publishing job.
3. Check that the runtime reported is reasonable. NOTE: It may not be exactly the same as mine since runtime is obviously hardware dependent. Ideally you would have published the same set and timed it without the PR to allow you to compare the numbers.
4. Check that your site doesn't experience performance issues during the publish job.
5. Check that the entities published were created correctly by viewing one from Admin > Tripal Content.

## Additional Notes (if any):

### DIFF
--- a/tripal/includes/TripalEntity.inc
+++ b/tripal/includes/TripalEntity.inc
@@ -15,4 +15,20 @@ class TripalEntity extends Entity {
     return array('path' => 'TripalEntity/' . $this->id);
   }
 
+  /**
+   * Permanently saves the entity.
+   *
+   * @param DatabaseTransaction $transaction
+   *    (OPTIONAL) Allows you to pass in your current transaction to save nested transaction overhead.
+   * @param boolean $clear_cached_fields
+   *    (OPTIONAL) Clearing cached fields is NECESSARY. IF you choose to set this to false then YOU
+   *    must clear the cache yourself using cache_clear_all('field:TripalEntity:[entity_id]', 'cache_field', TRUE).
+   *    The only known reason to set this to FALSE is to clear the cache in bulk for perfomance reasons.
+   *
+   * @see entity_save()
+   */
+  public function save(DatabaseTransaction $transaction = NULL, $clear_cached_fields = TRUE) {
+    return entity_get_controller($this->entityType)->save($this, $transaction, $clear_cached_fields);
+  }
+
 }
--- a/tripal/includes/TripalEntity.inc
+++ b/tripal/includes/TripalEntity.inc
@@ -18,17 +18,23 @@ class TripalEntity extends Entity {
   /**
    * Permanently saves the entity.
    *
-   * @param DatabaseTransaction $transaction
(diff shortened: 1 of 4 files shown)


laceysanderson:
Hold off testing for now :-( There may be a problem with this PR -looking it over again with a fine-toothed comb.

laceysanderson:
**Ok, ready for testing/review again!**

The weirdness was related to caching of transactions. Further reading on my part has shown caching of transactions to be unnecessary since Drupal handles nested transactions by simply ignoring the existence of the innermost calls. This was confirmed when my most recent commit removed them, fixing the weirdness and not significantly increasing the runtime.

See https://www.drupal.org/docs/7/api/database-api/transactions

> If one transactional function calls another, our transaction abstraction layer nests them by performing no transactional operations (as far as the database sees) within the inner nesting layers.

almasaeed2010 (approved):

almasaeed2010:
The code looks great. I tested the performance of this using `tripal_dev_seed` data.

|Branch|Time|Type|Number of records|
|-------|-----|-----|-------------------|
|`7.x-3.x`|5.860s|mRNA|200|
|`530-publish`|1.066s|mRNA|200|

spficklin:
Works good.  Publishing 4000 gene records took just under 3 minutes prior to the updated and just under 1 minute afterwards.

spficklin (approved):

spficklin:
Thanks for the fix @laceysanderson

laceysanderson:
Awesome! Thanks for the review :-)

PR #537, opened 2018-08-09
Fixed a bug for the API function chado_query()

chunhuaicheng:
Fixed a bug that when an SQL statement spans multiple lines, chado_query() messed up the SQL by removing the carriage returns and caused error. The carriage return should be replaced by at least one space when joining lines.

          

Issue #

## Type(s) of Change(s)

## Additional Notes (if any):


laceysanderson:
Thanks for the fix @chunhuaicheng! Since this change is so small, I'm marking this as only needing a single review. I have looked at the code and it's 👍 and confirmed that it fixes the issue you described. As such, I'm merging this in :-)

Issue #538, opened 2018-08-09
Slow static pages (i.e. Drupal nodes) caused by adding Tripal Content to the menu


chunhuaicheng:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.4
* PHP Version: 7.0.30

### Issue description

This problem will probably affect sites with expandable menu links that show on all pages. The pages are slow to load when you have Tripal Content added to the menu as link items. The more links you add, the slower the performance of the site will be. We have 11 'Organisms' added to the menu for quick access and it drags the performance for our site. 

The slowness is caused by checking permission on each link so Drupal can decide whether to show it in the menu or not. The permission checking process includes loading the entities and it's slow because the organism entities are trying to attach all its fields.

#### Steps to reproduce
1. Go to a page that shows the 'main menu'
2. Add some 'Organisms' to the 'main menu' as quick links, the more items you add the slower the page will load.

#### Error messages and screenshots

```
error message
```

almasaeed2010:
I did a quick profiling and here are my results for a brand new tripal site with 200 mRNA, 1 Organism and 1 Analysis

|Number of items in menu|Time to load the home page|
|-------------------------|---------------------------|
|2|328 ms|
|3|493 ms|
|4|545 ms|
|5|689 ms|
|6|728 ms|
|7|878 ms|

The items in the menu point to `bio_data/#` links. There is an obvious increase in time and it would probably be more if the entities had more fields.

almasaeed2010:
So changing the following lines to not use `$wildcard` and instead pass the entity id fixes the issue. This requires reimplementing `tripal_view_entity()` and `tripal_entity_access()` to handle the id instead of expecting an object. With that, we can load only a single field using `tripal_load_entity()` instead of loading all the fields. Or if there is a better function to pull the bundle name without having to load the entity at all, it'd be preferable.

https://github.com/tripal/tripal/blob/188a9f3eb6b03d5b21e63ec95915153478773e5d/tripal/includes/TripalEntityUIController.inc#L62-L82

almasaeed2010:
PR submitted #542

spficklin:
PR merged.. closing this issue.

PR #542, opened 2018-08-10
Stop loading the entire entity to check user access

almasaeed2010:
Issue #538

## Type(s) of Change(s)

## Description

See issue #538 

This fix preserves backwards compatibility by applying the fix only when an integer is passed to the function instead of an object as expected by the old behavior.

## Testing?

1. add `bio_data/#` links to your main menu (structure -> menus -> main menu -> add link)
1. measure the speed of the page (you can use chrome dev tools. The network tab can show waterfall chart detailing how long a page takes to load)
1. the more links added, the slower the page.

## Additional Notes (if any):

Using this fix we managed to speed up our site loading by 500%. Cache rebuilding used to take 1.2 minutes while with this fix it only takes 13.35 seconds. Visiting the home page before this fix takes 5.55 seconds after the cache has been repopulated, now it takes 813 milliseconds. This data was collected from testing the fix on our [live site](https://hardwoodgenomics.org).

### DIFF
--- a/tripal/includes/TripalEntityUIController.inc
+++ b/tripal/includes/TripalEntityUIController.inc
@@ -60,7 +60,7 @@ class TripalEntityUIController extends EntityDefaultUIController {
     }
 
     // Link for viewing a tripal data type.
-    $items['bio_data/' . $wildcard] = array(
+    $items['bio_data/%'] = array(
       'title callback' => 'tripal_entity_title',
       'title arguments' => array(1),
       'page callback' => 'tripal_view_entity',
@@ -71,7 +71,7 @@ class TripalEntityUIController extends EntityDefaultUIController {
     );
 
     // 'View' tab for an individual entity page.
-    $items['bio_data/' . $wildcard . '/view'] = array(
+    $items['bio_data/%/view'] = array(
       'title' => 'View',
       'page callback' => 'tripal_view_entity',
       'page arguments' => array(1),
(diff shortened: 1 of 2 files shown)


tripal/includes/TripalEntityUIController.inc
```diff
@@ -115,7 +115,20 @@ class TripalEntityUIController extends EntityDefaultUIController {
  * @see hook_entity_view_alter()
  */
 function tripal_view_entity($entity, $view_mode = 'full') {
-   $content = '';
+   if(!is_object($entity)) {
+     $id = intval($entity);
+     if($id === 0) {
+       return drupal_not_found();
```
almasaeed2010, line 121: Should we return `drupal_not_found()` here or simply return NULL?

spficklin:
I can confirm that this fix works!  I just want to do a look over of the code and then I'll mark my review.

chunhuaicheng:
Thank you @almasaeed2010 for providing a fix so quickly. I can confirm the fix works for me too.

spficklin (approved):

tripal/includes/TripalEntityUIController.inc
spficklin: This seems reasonable to me.

spficklin:
Thanks @almasaeed2010 for the very quick fix on this one!

Issue #551, opened 2018-08-14
Error in Web Services Module


almasaeed2010:
### System information
* Tripal Version: 3
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

A fatal error is showing up for my remote data field.

#### Steps to reproduce

I am not sure when or why this started happening.

#### Error messages and screenshots

<img width="1440" alt="screen shot 2018-08-14 at 10 07 32 am" src="https://user-images.githubusercontent.com/1512664/44096691-e2f0cc4a-9fa9-11e8-9b0a-8991b20ffbed.png">

```
Notice: Trying to get property of non-object in remote__data->makeRemoteRequest() (line 255 of /Users/Almsaeed/Work/DevSites/Tripal/sites/all/modules/tripal/tripal_ws/includes/TripalFields/remote__data/remote__data.inc).
Exception: Please provide a numeric site ID for the tripal_get_remote_content function. in tripal_get_remote_content() (line 247 of /Users/Almsaeed/Work/DevSites/Tripal/sites/all/modules/tripal/tripal_ws/api/tripal_ws.api.inc).
```

almasaeed2010:
PR submitted #552

spficklin:
Seems to be fixed... closing.

PR #552, opened 2018-08-14
Fix remote field

almasaeed2010:
Issue #551

## Type(s) of Change(s)

## Description

The remote data field needed some error checking to avoid usage of undefined variables.

## Testing?

Unfortunately I was not able to replicate the issue from scratch so I am not sure how to test this. I do think though that a quick look at the code should suffice since all we are adding is simple error checking and provide a fallback when a variable or a value is missing.

## Additional Notes (if any):

### DIFF
--- a/tripal_ws/includes/TripalFields/remote__data/remote__data.inc
+++ b/tripal_ws/includes/TripalFields/remote__data/remote__data.inc
@@ -105,7 +105,7 @@ class remote__data extends WebServicesField {
     // We don't want remote content to be available in web services.  There
     // is an if statement to not show this field in the web services but the
     // entity_load function doesn't know this field shouldn't be loaded so
-    // we need to short-circuit that.  
+    // we need to short-circuit that.
     if (preg_match('/web-services/', $_SERVER['REQUEST_URI'])) {
       $this->loaded_via_ws = TRUE;
       return;
@@ -127,7 +127,7 @@ class remote__data extends WebServicesField {
    * @see WebServicesField::load()
    */
   public function load($entity) {
-    
+
     // If this field is being loaded via web services then just return.
     if ($this->loaded_via_ws == TRUE) {
       return;
@@ -159,7 +159,7 @@ class remote__data extends WebServicesField {
       $entity->{$field_name}['und'][0]['remote_entity'] = NULL;
       $entity->{$field_name}['und'][0]['error'] = TRUE;
       $entity->{$field_name}['und'][0]['warning'] = FALSE;
-      $entity->{$field_name}['und'][0]['query_str'] = $this->buildRemoteURL($this->remote_site, $query_str);      
+      $entity->{$field_name}['und'][0]['query_str'] = $this->buildRemoteURL($this->remote_site, $query_str);
       return;
     }
     // Make sure we didn't have a problem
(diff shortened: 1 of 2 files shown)


bradfordcondon:
these changes look really no-nonsense to me, just checking if things are set. Approved.

note: the whitespace changes make this look like more than it is: they can be disabled n the PR review

tripal_ws/includes/TripalFields/remote__data/remote__data.inc
```diff
@@ -237,6 +237,11 @@ class remote__data extends WebServicesField {
     if (preg_match('/\?/', $query)) {
       list($path, $q) = explode('?', $query);
     }
+
+    if(empty($remote_site)) {
+      return null;
```
laceysanderson: I would add a tripal_report_error() before returning NULL. That way an admin would see there's a problem if looking in the recent messages log rather then left scratching their head. :-)

tripal_ws/includes/TripalFields/remote__data/remote__data.inc
```diff
@@ -245,14 +250,24 @@ class remote__data extends WebServicesField {
...
    private function makeRemoteRequest($query) {
      $path = $query;
      $q = '';
      if (preg_match('/\?/', $query)) {
        list($path, $q) = explode('?', $query);
      }
-     $data = tripal_get_remote_content($this->remote_site->id, $path, $q);
+
+    if(empty($this->remote_site)) {
+      return null;
```
laceysanderson: Always use tripal_report_error() when returning null.

tripal_ws/includes/TripalFields/remote__data/remote__data.inc
```diff
@@ -245,14 +250,24 @@ class remote__data extends WebServicesField {
...
+    }
+
+     try {
+       $data = tripal_get_remote_content($this->remote_site->id, $path, $q);
+     } catch (Exception $exception) {
+       return null;
```
laceysanderson: Always use tripal_report_error() when returning null.

tripal_ws/includes/TripalFields/remote__data/remote__data_formatter.inc
```diff
@@ -139,19 +139,23 @@ class remote__data_formatter extends WebServicesFieldFormatter {
...
     $content .= '<p>';
 
-    $content .=  t('This content provided by !site.',
-      array('!site' => l($site->name, $site->url, array('attributes' => array("target" => '_blank')))));
+    if($site) {
+      $content .= t('This content provided by !site.', [
+          '!site' => l($site->name, $site->url,
+            ['attributes' => ["target" => '_blank']])
+        ]);
+    }
```
laceysanderson, line 150: I would also add an else with a tripal_report_error() here.

laceysanderson (commented):
I agree the code looks pretty straightforward :-) My only comment is that it could benifit from usage of `tripal_report_error()` so that admin running into these edge cases would see what the problem is when looking at the logs. :-)

Thanks for the PR! :-)

almasaeed2010:
Hello @laceysanderson 

Thanks for the suggestions!

I added error reporting although I am not sure the messages are very helpful since I am not entirely positive which situations we'd have a missing site and still get to that part of the code. In my opinion, we should figure out a way to null the field in the loader before encountering this part of the code or report the error only once at the top of the loader then exiting since there is no need to continue if the site is not available.

Any thoughts? I could reimplement this PR to issue 1 warning instead 3 in the loader if you agree it's a better way.

Thanks!

laceysanderson (approved):
I’m content with this solution 🙂 Thanks @almasaeed2010 for the PR and the fixes!

Issue #562, opened 2018-08-20
Bundle creation permission error due to admin username


almasaeed2010:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

Creating a tripal content type for `Library` results in an error on our site. We used the term [nucleic acid library](https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0000160) to create the bundle.

#### Steps to reproduce

- Create a new content type
- Use the term: [nucleic acid library](https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0000160)
- If the term doesn't exist in your DB, insert it first.
- Choose library as the chado table
- Submit without changing the type_id column

#### Error messages and screenshots

```
PDOException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "" LINE 4: WHERE ( (rid = '') AND (permission = 'create bio_data_27') ) ^: SELECT 1 AS expression FROM {role_permission} role_permission WHERE ( (rid = :db_condition_placeholder_0) AND (permission = :db_condition_placeholder_1) ); Array ( [:db_condition_placeholder_0] => [:db_condition_placeholder_1] => create bio_data_27 ) in user_role_grant_permissions() (line 3174 of /home/www/modules/user/user.module).
```

<img width="582" alt="screen shot 2018-08-20 at 9 39 11 am" src="https://user-images.githubusercontent.com/1512664/44343803-ecb48480-a45c-11e8-840f-2233601d20e1.png">

### Notes

The content got actually created and seems to function properly even though the error above showed up. It seems as if tripal simply failed to edit the permissions for the content type but didn't fail anywhere else.

spficklin:
I am not able to recreate this problem. I added the "nucleic acid library" term to my database, created the type using the chado.library table (with type_id differentiating the type) and it created just fine.  

I think though, that the problem may be caused by these lines of code:

https://github.com/tripal/tripal/blob/85b3c2988e5ebc8083c67b34ed8fa1ef7efd253e/tripal/includes/TripalBundleUIController.inc#L882-L895

[...]

PR #564, opened 2018-08-21
Setting admin permissions on content creation

spficklin:
Issue #562

## Type(s) of Change(s)

## Description

See Issue #562 for a full description of the problem.  But in short, if your admin role is not named 'administrator' then you got errors when trying to create a new content type.  This fixes that problem.

## Testing?

There is only two lines of code changes for this fix.  It retrieves the admin role from a Drupal variable rather than using a hardcoded 'administrator' role name.  To test it, make sure your administrator role is named something other than 'administrator', then try to create a content type.  It should fail.  Then pull this branch, try to create another content type and it should succeed with no errors.

As this is a small code change I think only 1 review is required.  As @almasaeed2010 reported the problem I request that he review it as I think he can check it quickly.

## Additional Notes (if any):

### DIFF
--- a/tripal/includes/TripalBundleUIController.inc
+++ b/tripal/includes/TripalBundleUIController.inc
@@ -878,10 +878,29 @@ function tripal_admin_access($entity) {
     return FALSE;
   }
 
-  // Identify the administrative user roles.
-  $admin_rid = variable_get('user_admin_role');  
-  $admin_role = user_role_load($admin_rid);
-  $roles = array($admin_role->rid => $admin_role->name);
+  // Get the administrative user roles.
+  $admin_role = NULL;
+  $admin_rid = variable_get('user_admin_role'); 
+  if (!$admin_rid) {
+    // If we couldn't identify a single role from the 'user_admin_role' variable
+    // then let's get the role that is currently set to administer tripal. If
+    // there is more than one then we don't really know which to choose unless
+    // the default rid of '3' is present.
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Hello @spficklin 

Unfortunately, we are still experiencing the same issue with the following error:

```
PDOException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "" LINE 4: WHERE ( (rid = '') AND (permission = 'create bio_data_30') ) ^: SELECT 1 AS expression FROM {role_permission} role_permission WHERE ( (rid = :db_condition_placeholder_0) AND (permission = :db_condition_placeholder_1) ); Array ( [:db_condition_placeholder_0] => [:db_condition_placeholder_1] => create bio_data_30 ) in user_role_grant_permissions() (line 3174 of /home/www/modules/user/user.module).
```

However, I think you are right about the cause of the problem. Our admin user is not named administrator and the issue does not happen on my local site which has administrator as the admin role name.

Thanks!

spficklin:
Okay @almasaeed2010 . I have updated the fix to hopefully catch your case.  Can you try again?

almasaeed2010 (approved):

almasaeed2010:
Works for us.

Thanks @spficklin

spficklin:
As @almasaeed2010 indicated it worked and I think this only needs 1 review I've merged.

Issue #569, opened 2018-08-22
chado_delete_mview doesnt unset global variable for table


bradfordcondon:
im trying to delete and recreate an mview.  However im getting this insane pattern where the job will fail the first run, pass the second run.

here's the update
```php

/**
 * Don't count relationship cvterms as ontology roots.
 */
function tripal_chado_update_7328() {
    $mv_name = 'cv_root_mview';
    // Remove the old mview.

  $mview_id = chado_get_mview_id($mv_name);
    if ($mview_id){
      chado_delete_mview($mview_id);
    }


  var_dump($GLOBALS["chado_tables"]['drupal']['chado']) ;


    module_load_include('inc', 'tripal_chado', 'includes/setup/tripal_chado.chado_vx_x');
    // Re-add the mview.

    tripal_chado_add_cv_root_mview_mview();

    $mview_id = chado_get_mview_id($mv_name);
    chado_populate_mview($mview_id);

}

```

I've traced it to `chado_add_mview` being told the mview exists by `chado_table_exists`, but the table has been dropped.

I'm pretty sure that if `chado_delete_mview` updated the global variable properly this would resolve.

Errors:

```
WD tripal_chado: PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  table "cv_root_mview" does not exist: DROP TABLE     [error]
chado.cv_root_mview; Array
(
)
 in chado_query() (line 1715 of /Users/chet/UTK/tripal/sites/all/modules/tripal/tripal_chado/api/tripal_chado.query.api.inc).
WD tripal_mviews: PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "chado.cv_root_mview" does not exist       [error]
LINE 1: DELETE FROM chado.cv_root_mview
                    ^: DELETE FROM chado.cv_root_mview; Array
(
)
 in chado_query() (line 1715 of /Users/chet/UTK/tripal/sites/all/modules/tripal/tripal_chado/api/tripal_chado.query.api.inc).
Could not add custom table 'cv_root_mview': SQLSTATE[42P01]: Undefined table: 7 ERROR:  table &quot;cv_root_mview&quot; does not[error]
exist.
Materialized view, cv_root_mview, deleted.                                                                                      [status]
Materialized view 'cv_root_mview' created                                                                                       [status]
```

run it again, with success:

```
drush eval "module_load_install('tripal_chado'); tripal_chado_update_7328();"
Done.
[...]

PR #570, opened 2018-08-22
unset global mview variable when deleting if set

bradfordcondon:
Issue #569 

## Type(s) of Change(s)

## Description

The `chado_delete_mview` function does not unset the global variable that the table exists.  This means that trying to delete, and then re-add an mview, causes problems because tripal will think a table exists via the global variable even after its been deleted.

## Testing?

the below update function is what i was using to test: you can add to your chado.install and run it manually like so: `drush eval "module_load_install('tripal_chado'); tripal_chado_update_7328();"`

```php

/**
 * Don't count relationship cvterms as ontology roots.
 */
function tripal_chado_update_7328() {
  try {
    $mv_name = 'cv_root_mview';
    // Remove the old mview.
    $mview_id = chado_get_mview_id($mv_name);
      chado_delete_mview($mview_id);

    module_load_include('inc', 'tripal_chado', 'includes/setup/tripal_chado.chado_vx_x');
    // Re-add the mview.

    tripal_chado_add_cv_root_mview_mview();

    $mview_id = chado_get_mview_id($mv_name);
    chado_populate_mview($mview_id);
  } catch (\PDOException $e) {
    $error = $e->getMessage();
    throw new DrupalUpdateException('Could not perform update: ' . $error);
  }

}

```

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/api/tripal_chado.mviews.api.inc
+++ b/tripal_chado/api/tripal_chado.mviews.api.inc
@@ -355,6 +355,13 @@ function chado_delete_mview($mview_id) {
       $sql = "DROP TABLE {" . $mview->mv_table . "}";
       $success = chado_query($sql);
       if ($success) {
+        //unset the variable
+        global $databases;
+        $default_db = $databases['default']['default']['database'];
+        $chado_schema = chado_get_schema_name('chado');
+        if (isset($GLOBALS["chado_tables"][$default_db][$chado_schema][$mview->mv_table])){
+          unset($GLOBALS["chado_tables"][$default_db][$chado_schema][$mview->mv_table]);
+        }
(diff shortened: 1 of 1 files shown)


spficklin:
Yes, it works!  Thanks for the fix.  I did borrow your update function, but it confuses me a bit, did you just use that for testing?

bradfordcondon:
It is part of PR #571 .  if it should be improved let's discuss it there.  In short I was trying to drop, then re-create, an mview, something I roughly copied from another chado update function.

almasaeed2010:
Tested. works for me.

Thanks @bradfordcondon

almasaeed2010 (approved):

bradfordcondon:
thats reviewing @spficklin @almasaeed2010

Issue #572, opened 2018-08-22
Pub title field should not  be of 'Chado Property' type


chunhuaicheng:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.54
* PHP Version: 7.0

### Issue description

The Pub Title field is stored as a 'Chado Property' which does not use the 'title' column in the 'pub' table. This has caused our site to have empty title for newly imported pubs.

#### Steps to reproduce
List fields for the Publication Content Type e.g. Go to /admin/structure/bio_data/manage/bio_data_5/fields
The 'tpub__title' field has 'Text' field type and 'Chado Property' widget in its setting.

spficklin:
So this problem is caused by the following (after offline discussion with @chunhuaicheng).  If a site has a 'Title' as an entry in the pubprop table of Chado it conflicts with the pub.title field.  Tripal v3 prefers to create a field for the property rather than for the table column.   The reason that @chunhuaicheng was having this problem with the Title is because the Tripal Pub Importer at some time created properties for all data.  The site he is upgrading has some of those pubs that were loaded at a previous time.  The Pub Importer no longer does this.  It doesn't duplicate the title as a property.  So the Title property didn't exist for new records, but did for older ones.   When the site was upgraded to Tv3 and the Publication content type was created it added a "Title" field but linked it to the property rather than to the pub.title column.  So, when Tripal published the new pubs it tried to look for a title using the non-existent 'Title' record from the pubprop table... Therefore all of the titles were blank for the new pubs.

I've submitted a fix.

spficklin:
I've added this as 'URGENT" as @chunhuaicheng is hoping to put a new Tv3 site live ASAP.  

I will submit the PR for the fix if @chunhuaicheng confirms it works.

chunhuaicheng:
Thank you Stephen. I just tried it by removing and recreating the field. It works. The title field is now created as a Text type with TextField widget.

spficklin:
Great Thanks. I'll add the PR.  When I do since you've already tested it can you go in and just comment that you tested it and mark it as approved (like you just did here)?

spficklin:
I went ahead and marked it for you and referred to this issue.

chunhuaicheng:
Before you close this issue. I also found that if you 'Edit' a pub, the title text field isn't populated with the value. Don't know if this is caused by the same bug or not.

chunhuaicheng:
[...]

PR #574, opened 2018-08-23
Publication Properties Mismatch

spficklin:
Issue #572 

## Type(s) of Change(s)

## Description

When adding fields to a content type, Tripal has a preference for properties over table columns.  Normally this is not a problem except for the pub table.  The problem is that the Tripal Publication Importer in the past duplicated some of the Chado pub table column values in the pubprop table.   Newer version of the Importer does not do this.  However, for backwards compatibility, if someone has publication records created using this old method then Tripal will add Property Fields in place of fields for the table columns.  And, for newer records if a Title field is mapped to the pubprop table and there is no Title in the puprop table the publication records never get a title.  Similarly any column from the pub table with a property will have issues.  This pull request fixes this. 

## Testing?

You need to have a puprop record that uses a CV term that is also mapped to one of the columns of the pub table.  To see these terms go to Administration > Tripal > Data Storage > Chado > Semantic Web. Then select the pub table and you'll see the terms.    
1. To test this, add a property to any publication that uses the same term assigned to the pub.title columns.   **update** (see comments below). You should add this using an SQL statement or via phpPgAdmin.
2.  Go to the Publication Content Type, click "manage fields" and remove the tpub__title field.  The click the link at the top to re-add the field.  This will create a new tpub__title field but it uses the ChadoProperty field.
3.   Next, update your code to use the branch for this PR.  Repeat step 2 above.  This time the field will not be a ChadoProperty but a TextField even though the title exists as a pubprop.  

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/tripal_chado.fields.inc
+++ b/tripal_chado/includes/tripal_chado.fields.inc
@@ -712,22 +712,10 @@ function tripal_chado_bundle_fields_info_linker(&$info, $details, $entity_type, 
   // PROPERTIES
   $prop_table = $table_name . 'prop';
   if (chado_table_exists($prop_table)) {
-    // Get the list of existing property types for this table.
-    $sql = 'SELECT DISTINCT type_id FROM {' . $prop_table . '}';
-    $props = chado_query($sql);
-    while ($prop = $props->fetchObject()) {
-      $term = chado_generate_var('cvterm', array('cvterm_id' => $prop->type_id));
-
-      // The tripal_analysis_KEGG, tripal_analysis_blast, and
-      // tripal_analysis_interpro modules store results in the analysisprop
-      // table which is probably not the best place, but we don't want to
-      // create a ton of fields for this, so skip them.
-      if ($prop_table == 'analysisprop' and
-          ($term->dbxref_id->db_id->name == 'KEGG_BRITE' or
-           $term->dbxref_id->db_id->name == 'tripal')) {
-        continue;
-      }
-
+    
+    $props = tripal_chado_bundle_get_properties($table_name, $prop_table, $type_table, $type_column);
+    foreach ($props as $term) {
+          
       $field_name = strtolower(preg_replace('/[^\w]/','_', $term->dbxref_id->db_id->name . '__' . $term->name));
 
       // The field name can only be 32 chars, but if our name is longer we need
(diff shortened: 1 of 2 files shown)


spficklin:
I'm marking this as urgent because @chunhuaicheng needs it for a Tv3 update of a new Tripal site going live next week.

spficklin:
I'm also marking this as approved by @chunhuaicheng because he tested it before I posted the PR, you can see his comment about the successful test on the issue.

bradfordcondon:
I can't test this as written because tripal won't allow fields to have the same term.  When re-adding the fields i'll get this warning.

>The field, tpub__title, uses a term, TPUB:0000039, that is already in use on this content type. The field instance cannot be added.

spficklin:
@bradfordcondon That error message confuses me.  Did you delete the tpub__title field first and the re-add it and then you get this message?

bradfordcondon:
* Delete title field
* create property field with same term
* re-add title field, and am prevented from doing so because the property field exists using the same term.

i guess I need to use a different term for one of them, then switch the term?  Just not sure how to get both fields attached at the saem time using the same term because we actively prevent that from happening.

spficklin:
So for this test you only need one tpub__title field.  By default (before pulling this PR branch) if you don't have a property Tripal will add the TextField field that will be associated to the pub.title column.  If you have the "Title" property in the pubprop table then it will preferentially select the property and add a ChadoProperty field. It should not try to add both.   I don't understand why you would have two tpub__title fields that it's trying to add.

bradfordcondon:
i created a chadoprop field with the tpub title term in the GUI.

Sounds like instead, I should have inserted a property via sql of type title and then checked for fields.

spficklin:
Oh, you manually added the property via the GUI?  Yeah, you have to add it to the table via SQL... sorry I didn't make that clear.

bradfordcondon:
thank you for the clarification.  I've recreated hte issue and tested the fix on the branch.  Let me see if i have any code suggestions and we'll be good to merge.

tripal_chado/includes/tripal_chado.fields.inc
```diff
@@ -2828,6 +2735,128 @@ function tripal_chado_bundle_instances_info_linker(&$info, $entity_type, $bundle
...
+      ";
+    $args[':cvterm_id'] = $cvterm_id;
+    $props = chado_query($sql, $args);
+  }
+  // Third, if this is the case where a content type is uniquely identified
+  // via a term/value pair in the prop table.
+  else if ($type_column and $type_table == $prop_table and !empty($type_value)) {
+    $sql = "
+        SELECT DISTINCT P2.type_id
+        FROM {" . $prop_table . "} P1
```
bradfordcondon: should this use `db_escape_table`?   https://api.drupal.org/api/drupal/includes%21database%21database.inc/function/db_escape_table/7.x

bradfordcondon (approved):
Looks good to me, I'm not sure is this a case where we should b using the `escape_table` function when dynamically building the queries?

tripal_chado/includes/tripal_chado.fields.inc
spficklin: yeah, agreed. I'll fix it

spficklin: Okay, I've added the updates to the code and tested it again to make sure I didn't flub anything up. Still works for me.  If you approve then please feel free to merge!

bradfordcondon: ok thanks!

bradfordcondon:
hi @spficklin i now have this error when publishing phylotrees.  the location makes me suspect this PR is the cause (ie, its in `tripal_chado_bundle_get_properties`)

```php

Remapping Chado Controlled vocabularies to Tripal Terms...
Done.
WD php: Error: Call to a member function fetchObject() on null in                          [error]
tripal_chado_bundle_get_properties() (line 2820 of
/Users/chet/UTK/tripal/sites/all/modules/tripal/tripal_chado/includes/tripal_chado.fields.inc).
Cannot modify header information - headers already sent by (output started at              [warning]
phar:///usr/local/bin/drush/includes/output.inc:41) bootstrap.inc:1486
Error: Call to a member function fetchObject() on null in tripal_chado_bundle_get_properties() (line 2820 of /Users/chet/UTK/tripal/sites/all/modules/tripal/tripal_chado/includes/tripal_chado.fields.inc).
Drush command terminated abnormally due to an unrecoverable error.
```

spficklin:
This PR also fixed use case 1 for issue #481

Issue #290, opened 2018-02-14
newick tree loader db issue


bradfordcondon:
### System information

* Tripal Version: 3 
* Drupal Version: 7.56
* PostgreSQL Version: 9.6.1 
* PHP Version: 7.1.7

### Issue description

Two issues:

#### Loader doesnt allow missing accessions despite saying it does

The loader reads:

>Enter a database cross-reference of the form [DB name]:[accession]. The database name must already exist in the database. If the accession does not exist it is automatically added.

However, attempting to insert a tree with an accession not in the db fails validation:
`The dbxref provided does not exist in the database: .`

I think this is because of the below code (line 193, `tripal_chado.phylotree.api.inc`:

```
 $db_name = $matches[1];
    $accession = $matches[2];
    $values = array(
     'accession' => $accession,
      'db_id' => array(
        'name' => $db_name
      ),
    );
    $dbxref = chado_generate_var('dbxref', $values);
    
    if (!$dbxref) {
      $errors['dbxref'] = t('The dbxref provided does not exist in the database: %dbxref.', array('%dbxref' => $dbxref));
      return FALSE;
    }
    $options['dbxref_id'] = $dbxref->dbxref_id;
  }
```
I get confused by chado_generate_var but i can fix this... inserting the term beforehand resolves the issue.

#### Loader can't bulk load

This is the bigger issue.  It really needs to be able to load in mutliple newick trees.  It also needs to be able to assign those trees to accessions in bulk: that part is trickier, unless the name of the tree is the accession.

Anthony has developed this feature for his chado_python module so I'll likely try to adopt that, but i imagine this will be a desired feature for other adopters.

bradfordcondon:
shoot: inserting the term beforehand passes validation, but fails in the load step with 

```

SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "dbxref_id" violates not-null constraint
DETAIL:  Failing row contains (4, null, , 531, null, null).
WD tripal_job: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "dbxref_id" violates not-null constraint     [error]
DETAIL:  Failing row contains (4, null, , 531, null, null).
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "dbxref_id" violates not-null constraintDETAIL:  Failing row contains (4, null, , 531, null, null).

```

laceysanderson:
[...]

PR #582, opened 2018-08-23
resolve Phylotree TripalImporter errors

bradfordcondon:
Issue #290 

## Type(s) of Change(s)

## Description
* The Newick Importer cvterm lookup gave ajax errors if you did not have the legacy tripal_cv module enabled
* the Newick Importer ended up nulling all arguments in the form by accident.  This means that trees were created with no name, and with the default type only (polypeptide)  We fix that.
* Importer incorrectly validated file itself.  The base class handles that.
* add note to importer that we expect one tree per file.

additionally
* The Database Cross-Reference says it will create an accession for an entry if it doesnt exist but the db doesnt.  However, supply a db:accession that doesnt exist results in a validation error.

## Testing?
Run the importer to test.  If you have a newick file, thats great.  If not, you can use devseed:

https://raw.githubusercontent.com/statonlab/tripal_dev_seed/master/Fexcel_mini/sequences/mrna_mini.fasta.tree
and the corresponding fasta file:

https://github.com/statonlab/tripal_dev_seed/blob/master/Fexcel_mini/sequences/mrna_mini.fasta

(edit: replaced above link with link of the "raw" fasta not the aligned)

on master, look for the imported trees not having the SO type you selected (it will always be polypeptide) and publishing with no name.  This branch will resolve that.

Edit: you can also test the dbxref error.  Try creating a dbxref that isnt in your database  but where the db is(local:pigstailsarecurly for example).  On master you will get a validation error, whereas this branch it will insert the accession.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
@@ -198,11 +198,29 @@ function chado_validate_phylotree($val_type, &$options, &$errors, &$warnings) {
       ),
     );
     $dbxref = chado_generate_var('dbxref', $values);
+
     if (!$dbxref) {
-      $errors['dbxref'] = t('The dbxref provided does not exist in the database: %dbxref.', array('%dbxref' => $dbxref));
-      return FALSE;
+
+      $db = chado_generate_var('db', array( 'name' => $db_name));
+
+      if (!$db){
+              $errors['dbxref'] = t('
+              dbxref could not be created for db: %dbxref.', array('%dbxref' => $dbxref));
+              return FALSE;
+      }
+    $dbxref =  chado_insert_record('dbxref', $values);
+
+    if (!$dbxref){
+            $errors['dbxref'] = t('
+            dbxref could not be created for db: %dbxref.', array('%dbxref' => $dbxref));
+            return FALSE;
     }
+  }
+  if (is_array($dbxref)){
+    $options['dbxref_id'] = $dbxref['dbxref_id'];
+    }else {
     $options['dbxref_id'] = $dbxref->dbxref_id;
+  }
   }
 
   // Make sure the tree name is unique.
--- a/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.phylotree.api.inc
@@ -3,7 +3,7 @@
 /**
  * @file
  * Provides API functions specificially for managing phylogenetic and taxonomic
- * tree records in Chado.  
+ * tree records in Chado.
  */
 
 /**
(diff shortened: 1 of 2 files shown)


bradfordcondon:
I forgot to fix the actual issue #290  was originally about: db:accession's were supposed to be created if the accession didnt exist but teh db did.  I added a fix for this.

spficklin:
So, for testing, I need to preload the FASTA file to create the features.  One question.... the devseed FASTA file has a lot of gaps (- characters).  Is it supposed to be that way?  That looks like a FASTA sequence from an alignment. Do I really want to import that as a features' sequence?

bradfordcondon:
did i specify the wrong fasta file?  there are two on devseed one with the gaps (created by the phylotree program) and one without.

i did, sorry!

use this one instead

https://github.com/statonlab/tripal_dev_seed/blob/master/Fexcel_mini/sequences/mrna_mini.fasta

bradfordcondon:
>So, for testing, I need to preload the FASTA file to create the features. One question.... the devseed FASTA file has a lot of gaps (- characters). Is it supposed to be that way? That looks like a FASTA sequence from an alignment. Do I really want to import that as a features' sequence?

I also have a bunch of functional tests so you can take a look at that and ensure that `phpunit` runs on your dev site if you'd rather not test via the GUI

spficklin (approved):

spficklin:
I can confirm this works.  Without this branch items get loaded as 'polypeptide', and I can change the type with this PR.  Although, I think that field that lets you set the type isn't being used appropriately.  Shouldn't it be used to help lookup the features?   I don't have the time at this moment to dig into the loader, and we have another issue to deal with the loader issues related to that questions so I approve the fix.

almasaeed2010:
Works 😃 

screen shot 2018-08-29 at 2 25 00 pm

almasaeed2010 (approved):

bradfordcondon:
>I can confirm this works. Without this branch items get loaded as 'polypeptide', and I can change the type with this PR. Although, I think that field that lets you set the type isn't being used appropriately. Shouldn't it be used to help lookup the features? I don't have the time at this moment to dig into the loader, and we have another issue to deal with the loader issues related to that questions so I approve the fix.

this is the `leaf_type` argument.    I think it is used correctly: it gets passed down into line 711 of `tripal_chado.phylotree.api.inc`, see below.  There, it is used to specify the `$sel_values` variable of `chado_select_record`.  IE to help look up the right feature record.

https://github.com/tripal/tripal/blob/7c06ef26d3985abfc1d44b17edc4267ba48eaf28/tripal_chado/api/modules/tripal_chado.phylotree.api.inc#L711-L727

Because we have 2 approvals, and to my eye it looks like we use the passed leaf_type value correctly, I'm going to merge. 

@spficklin please feel free to create an issue if you still don't agree.

Issue #578, opened 2018-08-23
Default Tv3 Content Type views provide results by default


spficklin:
### System information

* Tripal Version: 3.x
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: N/A

### Issue description

In the past we've always created default search views for Tripal and those views typically only show results when the search button is clicked.  Currently, for Tv3 the search views show results by default which slows down display of the search form.

laceysanderson:
PR Merged 👍

PR #589, opened 2018-08-25
Auto search on default views.

spficklin:
Issue #578 

## Type(s) of Change(s)

## Description

Default search views for each Tripal content type were searching automatically.  This is not what the searches did in Tv2.  They should not search unless the user pushes the 'Search' button.  

## Testing?

Pull the branch, clear the cache and navigate to any of the default Tripal content type search views.  They should come up relatively quickly and with no search results.  The button should say 'Search' rather than 'Apply.  

If you have already customized any of these views then the PR won't fix it. It only works for non customized views.

## Additional Notes (if any):

### DIFF
--- a/tripal/tripal.views_default.inc
+++ b/tripal/tripal.views_default.inc
@@ -61,7 +61,9 @@ function tripal_bundle_default_views(&$views) {
     $handler->display->display_options['access']['perm'] = 'view ' . $bundle->name;
     $handler->display->display_options['cache']['type'] = 'none';
     $handler->display->display_options['query']['type'] = 'views_query';
-    $handler->display->display_options['exposed_form']['type'] = 'basic';
+    $handler->display->display_options['exposed_form']['type'] = 'input_required';
+    $handler->display->display_options['exposed_form']['options']['submit_button'] = 'Search';
+    $handler->display->display_options['exposed_form']['options']['text_input_required_format'] = 'filtered_html';
     $handler->display->display_options['pager']['type'] = 'full';
     $handler->display->display_options['pager']['options']['items_per_page'] = '10';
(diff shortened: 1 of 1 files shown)


spficklin:
This change is only a few lines... only 1 review needed.

almasaeed2010:
Hello @spficklin 

First of all, the PR works as expected. However, I feel like we've come to expect the behavior of listing a first set of results before searching for those views and we are using that functionality to list some of our content on our tripal sites. It is not hard to create custom views and migrate to them to maintain this functionality, but I wanted to note that this may affect others as well.

Note that this PR does make initial loading of the pages much faster!

@spficklin  @laceysanderson @bradfordcondon any thoughts?

Thanks!

spficklin:
Backwards compatibility for Tv2 was that views did not show results. If we keep this fix and if you want to go back to having the search results listed automatically, it's easy to do by editing the view in the web GUI and changing one parameter.

I can see the value in both ways.  @laceysanderson can you weigh in?

laceysanderson:
As the size of your chado tables increase, the sort on these views will slow them down drastically. We originally changed the views to require input before showing results after some tripal sites began experiencing timeouts loading the searches. It was so bad that Tripal sites with large datasets couldn't even load the searches in order to filter them.

With this in mind, **I think we should keep backwards compatibility with T2 and ensure the default searches require input**. While I prefer a subset of the results being shown, I also want to ensure no Tripal sites experience timeouts on their searches and am unwilling to remove the sort criteria.

almasaeed2010:
I agree. The default Tripal behavior shouldn't cause sites to slow down. 👍 

Thanks!

spficklin:
Thanks @almasaeed2010 and @laceysanderson .  I'm merging.

Issue #533, opened 2018-08-08
Creation of Content Type times out


spficklin:
### System information

* Tripal Version: 3.x
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version:  N/A

### Issue description

Sometimes creation of content types results in a WSOD.  This seems to happen when there are too many properties in a table (e.g. feature table).

The solution is to create content types using a Tripal job.  So, we need to change the creation form so that it adds a job rather than try to create the type via the GUI.

#### Steps to reproduce
To reproduce this you have to a large site with lots of properties for a table.

#### Error messages and screenshots

Error log in Apache:

```
[Wed Aug 08 11:44:50.328442 2018] [php7:error] [pid 5003] [client 127.0.0.1:48422] PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /var/www/html/includes/file.inc on line 2169, referer: http://localhost/admin/structure/bio_data/add
```

bradfordcondon:
resolved.  Thank you.


@laceysanderson if you'd like to create a separate issue to allow job-free creation of bundles, but then job-launched attaching of fields (or another solution) please do so.

PR #593, opened 2018-08-25
Creation of content types times out

spficklin:
Issue #533 

## Type(s) of Change(s)

## Description

Sometimes you can get a WSOD when creating new content types. This is because in some cases it must look through all of the properties in a table to create the prop tables.   This PR makes content type creation (via the Drupal interface) a job submission.

One Note, I changed the tripal_create_bundle() function.  It had an $error argument that was meant to be passed in by reference and was set to the error message. Apparently that was an older argument and is no longer used.   We don't normally change API functions but this is so new and I'd rather not lock in an argument that is not used so I thought it was okay to remove that.

## Testing?

1).  Pull the PR
2)  Create a new content type. It should submit a job.
3)  Run the job,
4).  Return to the Content Types page and see that your new content type is available.

## Additional Notes (if any):

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -417,20 +417,31 @@ function tripal_add_notification($title, $details, $type, $actions, $submitter_i
  *     - term_name: A human-readable name for this term.  This will became
  *       the name that appears for the content type.  In practice, this
  *       should be the name of the term. (E.g. the name for SO:0000704 is gene).
- * @param $error
- *  A string, passed by reference, that is filled with the error message
- *  if the function fails.
+ * @param $job
+ *  The job ID if this is launched via a job.
  *
  * @return
  *   The bundle object or FALSE if failure.
  *
  * @ingroup tripal_entities_api
  */
-function tripal_create_bundle($args, &$error = '') {
+function tripal_create_bundle($args, $job = NULL) {
   $vocabulary = $args['vocabulary'];
   $accession = $args['accession'];
   $term_name = $args['term_name'];
   $storage_args = $args['storage_args'];
+  
+  $message_args = [
+    'job' => $job,
+    'print' => TRUE,
+    'watchdog' => TRUE,
+  ];
+  
+  tripal_report_error('tripal_entities', TRIPAL_INFO, 
+   "Creation of a content type is performed using a database transaction. " .
+    "If it fails or is terminated prematurely then all insertions and " .
+    "updates are rolled back and will not be found in the database", 
+    [], $message_args);
 
   $transaction = db_transaction();
   try {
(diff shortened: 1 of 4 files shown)


bradfordcondon:
functionality verified 👍 submitted and run a content creation job.

bradfordcondon (approved):
looks good to me, thanks @spficklin

laceysanderson:
I hesitate on this one as it feels like it degrades the user experience. I would expect this to be quick and not need a job. Is there any way we can optimize the process rather then put it in a Tripal Job? Or separate the content type creation and the field addition process? The type could be created immediately and the prop fields added via a tripal job?

spficklin:
I get your complaint. I wasn't too thrilled to move it to a job either but I couldn't see another way around it.  I think separating content type creation from field creation would complicate the setup too.  So, either way we're adding an extra step.... :-(

laceysanderson:
My idea wasn't to make the user initiate the field creation step... I figured they would go to `Add Tripal Content Type`, fill out the form, click create. At this point the user would be redirected to their new content type which would have a message at the top saying "Tripal Job submitted to automatically create appropriate fields for this content type based on the data in your site. It is recommended you wait until the job completes before configuring this content type. Check on the progress of this job here." This way they get immediate gratification and we don't risk a WSOD. There's no extra step for them, especially if they setup the daemon. It just causes a hopefully short delay for configuring the content type. Thoughts?

spficklin:
I think in either case the user ends up waiting and the experience is the same... they can't use the content type until a job runs.   I'm not opposed to the suggestion. I'm just not sure the experience is any better.

laceysanderson:
I still really hesitate here. I would like to take a stab at optimizing this process before separating it out into a job but I doubt I would have the time before Friday as I have KnowPulse deadlines to meet this week :-( 

How many properties do you need to have before this becomes a problem? Could it be put off until the next point release? If we move it into a tripal job for 3.0, and I find a way to optimize it later, can we move it back?

spficklin:
So how about this... As part of a stable release I don't think we should have WSOD if we know about it.  So, how about we leave it as a job submission and merge this in.  In the end, if we are able to improve performance we can move it back to not requiring a job and that ends up being an improvement for the end-user.  If we leave as is, and we determine we must use a job, then a future update would have this fix and it would seem to the end-user as a step-down.

almasaeed2010 (approved):
Confirmed. A job is submitted and succeeds.

Thanks @spficklin

bradfordcondon:
we have 2 approvals (thanks @almasaeed2010 ) and it sounds like @laceysanderson is fine with this approach with improvement down the road.

Merging, thanks @spficklin

spficklin:
Thanks @almasaeed2010 @bradfordcondon and @laceysanderson !

Issue #605, opened 2018-08-28
Creating entities produces a PHP notice


almasaeed2010:
### System information

* Tripal Version: 3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

When creating any new entity, we get the following error:
```
Notice: Undefined variable: created_date in TripalEntityController->save() (line 397 of /Users/Almsaeed/Work/DevSites/Tripal/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
```

#### Steps to reproduce

1. create a new organism
1. the error will appear on the newly created organism page

almasaeed2010:
PR submitted

laceysanderson:
PR Merged

PR #606, opened 2018-08-28
Fix php notice when creating an entity

almasaeed2010:
Issue #605 

## Type(s) of Change(s)

## Description

This PR removes the undefined `$created_date` variable and replaces it with `time();`

## Testing?

1. once the branch is checked-out, you shouldn't see the errors described in issue #605

## Additional Notes (if any):

### DIFF
--- a/tripal/includes/TripalEntityController.inc
+++ b/tripal/includes/TripalEntityController.inc
@@ -394,7 +394,7 @@ class TripalEntityController extends EntityAPIController {
       // If our entity has no id, then we need to give it a
       // time of creation.
       if (empty($entity->id)) {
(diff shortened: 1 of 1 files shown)


spficklin (commented):
I think the fix should be this:

```
$entity->created = create_date;
```
Note the code above this line that initializes the $create_date variable and then does some checks to make changes if need be.

almasaeed2010:
Thanks @spficklin I didn't notice the variable. I updated accordingly.

spficklin:
This is just a one liner and it looks good so only 1 review needed. I'll merge.

spficklin:
Thanks @almasaeed2010

Issue #550, opened 2018-08-13
Select List Views Exposed Filter for Tripal Fields


laceysanderson:
## Feature Request/Discussion

- [x] does this feature attempt to solve an existing problem with Tripal?
- [x] are you open to developing or collaborating on an extension module if this is not a good fit for Tripal Core (no pressure here -just good to know upfront :-) )
- [x] do you **Need** this feature ASAP?

### Description

I need to be able to choose select list and/or autocomplete as the exposed filter form element for various Tripal Fields when designing my search views. Ideally there would be a very generic views handler that would select from the table the field is attached to finding the currently used values and using them to populate the list. I could also allow it to be configured to just provide all values from the linked table as a faster option. Autocomplete would query on the fly.

I would generically implement it for all foreign key fields and use tokens from the table the foreign key references to make the human-readable option generic. For example, for the feature table, you could use a select list for the organism_id, type_id and dbxref_id; for the organism_id select list the tokens available would be all columns from the organism table.

This is handled in Drupal by having two fields you can add in the views UI: `[field name]` for the straight textfield and `[field name] (Allowed Values)` for the select list. For Drupal Fields the allowed values are set in the instance settings and thus no querying is needed.

Are we content with two options in the views UI for each field attached to a chado foreign key? Also are we content querying for the options? Or should I require options be set in the views ui?

### Your Specific Use Case

Specifically I need:
- organism_id => scientific names
- type_id => cvterm name
- project/analysis => name

### Generally Applicable

I feel select lists in exposed filters are more friendly to users and general expected in web forms.

laceysanderson:
After discussion with @spficklin:

**Clarification:** This would allow admin to add a select list or autocomplete exposed filter for many Tripal/Chado Fields where the values in the list match those already used in the chado table. For example, you could have a select list for a field connected to the feature.type_id where the options are the unique set of feature.type_ids (e.g. sequence_variant, genetic_marker, gene if those are the only 3 unique type_id in the feature table).

**My plan** would be to add a `TripalField::getValueList($keyword = NULL)` method to both TripalField and ChadoField classes. This method would retrieve a list of options for a select list when `$keyword = NULL` or a list of autocomplete options when `$keyword` is set. This allows the query retrieving the distinct list to be customized for custom fields. For non-Tripal/Chado fields I would default to a generic API function (`chado_get_field_value_options`) to retrieve the options.

[...]

PR #613, opened 2018-08-29
Tripal Field getValueList Method

laceysanderson:
Issue #550 (this PR does not yet fix the marked issue but is needed to fix it)

## Type(s) of Change(s)

- [x] Includes Tests

## Description

This PR adds a method (getValueList) to the ChadoField class which returns a list of the unique values already used for a given field. This method was developed with select list functionality in mind for filtering of search results but may have many other applications. It was suggested to use a Field class method to allow overriding by custom fields allowing developers to more easily tweak select list values.

## Testing?

This PR includes automated testing for this method on all bundle/field combinations. It is independent of entity instances making it an ideal first test case for field testing in Tripal. To test this, run the specific test suite as follows (in your tripal directory):
```
composer up
./vendor/bin/phpunit --group getValueList
```

Additionally, if you would like to test this manually, the following code snippet can be used either in Developer Tools > Execute PHP Code or in a custom module. Simply set `$field_name` and `$bundle_name` to the combination you would like to test.
```php
      // In this example we want the values for the obi__organism field
      // attached to the Tripal Content Type with a machine name of bio_data_17:
      $field_name = 'obi__organism';
      $bundle_name = 'bio_data_17';

      // The following two calls get information about the field we want the values for.
      $field_info = field_info_field($field_name);
      $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);
      // Construct the Field instance we want the values for.
      $instance = new ChadoField($field_info, $instance_info);

      // Retrieve the values.
      // $values will be an array containing the distinct set of values for this field instance.
      $values = $instance->getValueList();
```
You then want to inspect `$values` to ensure:
- each value only occurs once
- all values for that chado table column are represented (e.g. for the obi__organism on a feature based bundle, the value list should contain all organisms you have features for). You can check this by running `SELECT organism_id FROM chado.feature GROUP BY organism_id` and then checking all returned values exist as keys in the values array returned above.

You can also test the parameters as follows:
- limit parameter (e.g. `$values = $instance->getValueList(['limit' => 5]);`) and ensure the values list contains at most the indicated number of values
- label_string parameter (e.g. `$values = $instance->getValueList(['label_string' => '[genus] [species] ([common_name])']);`) and ensure the values returned have the specified format. This parameter only applies to foreign key fields.

## Additional Notes (if any):

- [x] My change requires a change to the documentation. (ideally this would be documented with the rest of the ChadoField class but I don't know how to update the docs yet)


bradfordcondon:
@laceysanderson for future reference if you have a fork of tripal you can configure its CI to run, so you could push to lacey'sfork/Tripal and work out the CI stuff before making the PR to the main repo.

likewise as you can see in hte commit history, your individual commits run *before* you make the PR, so you can check that way too, on teh travis website.

laceysanderson:
I think I would want to add the following documentation to https://tripal.readthedocs.io/en/latest/dev_guide/custom_field/manual_field_creation.html but am unsure how to do so... If someone in the know would like to add it to this PR I'd appreciate it!

## getValueList() function

The getValueList() function can be used in custom forms to retrieve a list of the unique values already used for a given field instance (field/bundle combination). This is great for select lists, checkboxes or radios and can provide much friendlier options then the typical empty textfield.

The following code example shows you how to pull the value list for a specific instance of a field.
```php
      // In this example we want the values for the obi__organism field
      // attached to the Tripal Content Type with a machine name of bio_data_17:
      $field_name = 'obi__organism';
      $bundle_name = 'bio_data_17';

      // The following two calls get information about the field we want the values for.
      $field_info = field_info_field($field_name);
      $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);
      // Construct the Field instance we want the values for.
      $instance = new ChadoField($field_info, $instance_info);

      // Retrieve the values.
      // $values will be an array containing the distinct set of values for this field instance.
      $values = $instance->getValueList();
```

bradfordcondon:
i can do it for you but if you want to try, go to `docs/` and find the corresponding file (`manual_file_creation`).  Add your text in the new markup language [Refstructured Text](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)

from `docs/` you can run `make html` if you have sphinx installed: http://www.sphinx-doc.org/en/master/usage/installation.html

it will appear in `_build/html`


Once the changes are pushed to master, they will appear at  https://tripal.readthedocs.io automatically

laceysanderson:
Adding docs and additional assertions this morning.

spficklin:
Hi @laceysanderson how do I test the new getValueList function?

laceysanderson:
Ok, I got all set up to add documentation and then started questioning whether this method should be added in that section. Originally I assumed that section was simply documenting all the chadofield methods but it turns out it's instructing developers on which methods they should implement. Very few developers should need to implement ChadoField->getValueList() as it's default implementation works for all tables described by the Tripal Schema API. The main time to implement it is to override the current functionality with a faster, specific query. **So perhaps the inline documentation is sufficient?**

Also, @spficklin I've updated the testing section above to show how you could test this method manually rather then relying on the automated testing. The code snippet is also how you would use this method in development (e.g. if you were creating a select list in a form).

bradfordcondon:
> So perhaps the inline documentation is sufficient?

I think that a guide not on this specific api but on "important tripal methods for working with bundles" would be helpful.  We have in depth docs on the anatomy of fields but less on using the API.  

Maybe write it up as a piece for such a guide and mark the rest as "under construction" ?


In this case the most important bit is the in-code documentation.  

I'll note that RTD has the possibility of integrating with doxygen so you can write some very cool guides that automatically reference methods/sections of the doxygen-generated API, but i ran into some problems trying to set it up....

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
```diff
@@ -0,0 +1,442 @@
...
+      }
+      elseif (($column_type == 'varchar') OR ($column_type == 'text')) {
+        $fake_value = $faker->words(2,TRUE);
+      }
+
+      if ($fake_value !== NULL) {
+        factory('chado.'.$chado_table)->create(array(
+          $fixed_column => $fake_value,
+        ));
+        return fake_value;
```
almasaeed2010: I think this should be `$fake_value` instead of `fake_value`

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
```diff
@@ -0,0 +1,442 @@
...
+    // Attempt to create a fixed fake value.
+    // This needs to match the column type in the chado table and if the column is a
+    // foreign key, this value should match a fake record in the related table.
+    $fake_value = NULL;
+
+    // If we weren't told the related table then we assume this is a simple column (not a foreign key).
+    if ($fk_table === FALSE) {
+      $column_type = $schema[$fixed_column]['type'];
+      if (($column_type == 'int')) {
+        $fake_value = $faker->randomNumber();
```
almasaeed2010, line 405: I don't see `$faker` being created here so that's probably an undefined variable?

almasaeed2010 (changes requested):

bradfordcondon:
### Evaluation

This is awesome.  It works really well.  It makes it really clear to others how to conjure fields and bundles for testing purposes.  A+.

### Suggestions

I ran your tests and all passed but one:  `ChadoFieldGetValuesListTest::testBaseTableColumns`.  Looking at the output, its very clear this test failed because i have a field attached that is misconfigured.  

(i changed the base_table/chado_table parameters in the fields.info file, and now your tests all pass)

I actually like this because it detects misconfigured fields.  However, this is where we have a case where the test will pass for someoen but fail for someone else, which isn't ideal.  Thoughts?


also, testing your snippet in drush php:

```php
 $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);```

```
this returns null.
`$field_info` returns an array as expected.  any idea why?

Edit: oops, its because bio_data_x needs to match a bundle that has organism attached.  Maybe instead you should get the bundle ID by the bundle label so its clear whats going on there in the example?

almasaeed2010:
Hello @laceysanderson 

I tested this and it's working great for me. The unit tests pass (after fixing the couple of typos) and the returned values are as expected (match the values in the bundle, are unique and executes fast!).

Thanks!

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
```diff
@@ -0,0 +1,442 @@
...
+
+      // Re-generate the values...
+      $values = $instance->getValueList(array('limit' => 200));
+
+      // And ensure our fake value is in the returned list.
+      // We can only check this if all the results are returned.
+      // As such, we set the limit quite high above and if we have
+      // less then the limit, we will go ahead with the test.
+      // @note: this tests all fields on TravisCI since there is no pre-existing data.
+      if (sizeof($values) < 200) {
```
bradfordcondon, line 74: so this test wont execute if i have over 200 records.  Is that as intended?

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
```diff
@@ -0,0 +1,442 @@
...
+   * @group getValueList
+   */
+  public function testBaseTableForeignKey($field_name, $bundle_name, $info) {
+    include_once(drupal_get_path('tripal_chado', 'module') . '/includes/TripalFields/ChadoField.inc');
+
+    // Construct the Field instance we want the values for.
+    // Specifying "ChadoField" here ensures we are only testing our
+    // implementation of getValueList() and not the custom version for any
+    // given field.
+    // YOU SHOULD TEST CUSTOM FIELD IMPLEMENTATIONS SEPARATELY.
```
bradfordcondon, line 118: im seeing this method as a super helpful way to test custom fields.  Maybe a new section in the "creating custom fields" guide?

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
almasaeed2010: Looks like this should be an assertion `$this->assertLessThanOrEqual(200, sizeof($values))`?

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
```diff
@@ -0,0 +1,442 @@
...
+   * Generate fake data for a given bundle.
+   *
+   * If only the first parameter is provided this function adds fake data to the indicated
+   * chado table. If the third parameter is provided the generated fake data will
+   * have a fixed value for the indicated column.
+   *
+   * @return
+   *   Returns FALSE if it was unable to create fake data.
+   */
+  function generateFakeData($chado_table, $schema, $fixed_column = FALSE, $fk_table = FALSE) {
```
bradfordcondon: Probably want this to be a private function?

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
laceysanderson: Yes it should be, thanks!

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
laceysanderson: It's intended for the nested assertions to not run if there are more then 200 records. 

This is due to the fact that I can't guarantee my fixed fake values will be in the subset returned. We need this because I was unable to create fake bundles and as such there may be existing data in the database. We don't want the test to fail since it doesn't actually indicate the method is not working, it's simply that we can't test as well as I would like at this point.

tests/tripal_chado/fields/ChadoFieldGetValuesListTest.php
laceysanderson: *facepalm*

laceysanderson:
Awesome! I've made the fixes you both suggested -Thanks @almasaeed2010 and @bradfordcondon!

I think these tests can be made more independent of the testing environment by creating fake bundles. This way only the data I insert is being used in the test (no need for the > 200 check) and, since all fields are added automagically, it should get around the misconfigured fields issue. However, I ran into difficulties implementing a bundle databaseSeeder and wanted to get this feature out before the release...

bradfordcondon:
>I think these tests can be made more independent of the testing environment by creating fake bundles. 

yes but then you'd also need to attach fake fields... it's definitely a really large-scale goal.

bradfordcondon (approved):

laceysanderson:
I'm moving discussion of documentation to #623 since I feel the page suggested by @bradfordcondon would be best-done more tutorial-like rather then method documentation and I just don't have time for it at the moment.

laceysanderson:
> yes but then you'd also need to attach fake fields... it's definitely a really large-scale goal.

Exactly, that's why I scaled back. I thought it better we get some good tests and this new functionality rather then me getting lost down a never ending rabbit hole 😝

bradfordcondon:
waiting on approval from @almasaeed2010   i believe?

almasaeed2010:
I didn’t have any problems testing this. I think we can merge if the work is complete:)

Thanks!

bradfordcondon:
ok @laceysanderson are you finished on this branch?  youre clear to merge if so....

laceysanderson:
Merged. Thanks @bradfordcondon and @almasaeed2010!!! :-D 

I'll keep the branch though to continue to the select lists for views.

PR #615, opened 2018-08-29
edit to documentation: added common path to modules directory

srobb1:
## Type(s) of Change(s)

## Description

adding some detail to the instructions as I go through them. Just in case someone does not know where modules are stored, I added the path.

### DIFF
--- a/docs/user_guide/install_tripal/upgrade_from_tripal2.rst
+++ b/docs/user_guide/install_tripal/upgrade_from_tripal2.rst
@@ -44,7 +44,7 @@ Step 1: Upgrade Tripal
 
     drush pm-disable tripal_core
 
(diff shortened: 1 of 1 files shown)


bradfordcondon:
thanks @srobb1 
I really appreciate this, please keep creating issues or PRs as you go through the documentation!  It will make it so much better.


how would you feel about this instead?

>The Tripal modules must also be downloaded and updated. To do this, delete the old Tripal v2 modules directories, located in `sites/all/modules` from your Drupal root:  for example `/var/www/html/sites/all/modules `(be sure you have a backup before removing). The following command will retrieve the Tripal 3 version:

Your suggestion is just as corect, but i want to convey that a) tripal is by default in `/sites/all/modules/` from wherever the drupal root is, and, b) the "default" drupal root is `/var/www/html`

spficklin:
Thanks @srobb1 this is great!

srobb1:
@bradfordcondon That is great! Thanks!

bradfordcondon:
ok thanks, i made the suggestion and will merge in 👍 thanks again

Issue #618, opened 2018-08-30
PHP Undefined Property Notice in TripalEntityController


almasaeed2010:
### System information

* Tripal Version: 3.x
* Drupal Version: 7
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

Our logs are cluttered with PHP notices for undefined properties. I am not sure why the props doen't exist but I think a simple check will fix this. Example error:

```
PHP Notice: Undefined property: TripalEntity::$sio__annotation in /var/www/html/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc on line 678
```

#### Steps to reproduce

I am not sure how to reproduce this. It is happening only our live site but not my local site.

#### Error messages and screenshots

```
PHP Notice: Undefined property: TripalEntity::$sio__annotation in /var/www/html/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc on line 678
```

almasaeed2010:
The issue seems to be with this line:

https://github.com/tripal/tripal/blob/e60511585bfb577e781b09e168d235887c2b89df/tripal/includes/TripalEntityController.inc#L649

It adds the field as an array to the entities array but does not consider the `$entity` itself.

Will submit a PR to fix this.

spficklin:
PR Merged

PR #621, opened 2018-08-30
PHP Undefined Property Notice in TripalEntityController

almasaeed2010:
Issue #618

## Type(s) of Change(s)

## Description

PHP is throwing notices because some fields are not defined on the entity.
This fix simply initializes the field in the (not passed by reference) `$entity` as an empty array.

## Testing?

1. Clear the cache
2. Make sure entities load without the php notice described in #618 

## Additional Notes (if any):

### DIFF
--- a/tripal/includes/TripalEntityController.inc
+++ b/tripal/includes/TripalEntityController.inc
@@ -647,6 +647,7 @@ class TripalEntityController extends EntityAPIController {
           // Add this field to the entity with default value.
           if (!isset($queried_entities[$id]->{$field_name})) {
             $queried_entities[$id]->{$field_name} = array();
+            $entity->{$field_name} = array();
           }
 
           // Options used for the field_attach_load function.
--- a/tripal/includes/TripalEntityController.inc
+++ b/tripal/includes/TripalEntityController.inc
@@ -645,8 +645,8 @@ class TripalEntityController extends EntityAPIController {
           $field_id = $field['id'];
 
           // Add this field to the entity with default value.
-          if (!isset($queried_entities[$id]->$field_name)) {
-            $queried_entities[$id]->$field_name = array();
+          if (!isset($queried_entities[$id]->{$field_name})) {
+            $queried_entities[$id]->{$field_name} = array();
           }
 
           // Options used for the field_attach_load function.
@@ -661,7 +661,7 @@ class TripalEntityController extends EntityAPIController {
           if ($field_cache) {
             $cache_data = cache_get($cfid, 'cache_field');
             if (!empty($cache_data)) {
-              $queried_entities[$id]->$field_name = $cache_data->data;
+              $queried_entities[$id]->{$field_name} = $cache_data->data;
               $queried_entities[$id]->{$field_name}['#processed'] = TRUE;
               continue;
             }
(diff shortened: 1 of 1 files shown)


tripal/includes/TripalEntityController.inc
```diff
@@ -645,8 +645,9 @@ class TripalEntityController extends EntityAPIController {
           $field_id = $field['id'];
 
           // Add this field to the entity with default value.
-          if (!isset($queried_entities[$id]->$field_name)) {
-            $queried_entities[$id]->$field_name = array();
+          if (!isset($queried_entities[$id]->{$field_name})) {
+            $queried_entities[$id]->{$field_name} = array();
+            $entity->{$field_name} = array();
```
spficklin: I'm a bit leery of this change. It's my understanding the if a field is attached it should be "all the way" attached and have the language, the delta the values.  I can see how this fixes the problem but I'm worried it may cause other sideffects.

spficklin:
So, the problem seems to be on line 678 of the TripalEntityController.  On line 674 the field is supposed to be attached. Afterwards on line 678 if tries to then cache the field.  If you are getting an "undefined property" message for the field name then it means something went wrong with attaching the field and that needs to be explored more.

almasaeed2010:
Hello @spficklin 

From what I can see in the code, line 649 (on 7.x-3.x branch): 
https://github.com/tripal/tripal/blob/85fa3601e86b0164ceeb98d2836958ec4b62cdcb/tripal/includes/TripalEntityController.inc#L649

is initializing the field as an empty array if it does not already exist. However, it's being added to the array that we are iterating over instead of the `$entity` variable we are using to cache the field. If we instead pass the `$entity` variable by reference, the initialization would've also worked and no errors would be thrown:

```php
// Pass $entity by reference
foreach ($queried_entities as $id => &$entity) {
```

That would also keep the entity in sync with the array. 

From your comment though, it sounds like the first initialization is not supposed to happen? 

Thanks!

tripal/includes/TripalEntityController.inc
laceysanderson: I second @spficklin concerns... I would instead look into the sio__annotation field/formatter. Is the PHP notice only for that field?

laceysanderson:
Ok, a much closer reading of the code triggered by your comment above has led me to agree with @almasaeed2010. `$entity` in the scope of this function is simple the value of the array being iterated over (`$queried_entities`) and thus it should be fine to set a default.

That said, I would be more inclined to change the usage of `$entity->$field_name` to `$queried_entities[$id]->{$field_name}` for consistency since that's how the field is being loaded/processed. That should also remove your PHP notice and not instantiate anything other then what @spficklin already has :-)

almasaeed2010:
Hello @laceysanderson 

Thanks for taking the time to examine this! 

To answer your first question, we get that error on multiple fields so it's less likely to be a field formatter problem. 

I'll resubmit the PR with `$entity` switched to `$queried_entities[$id]`.

Thanks!

almasaeed2010:
Ok I just pushed a commit that addresses the consistency issue.

Thanks!

laceysanderson (approved):
The new changes look perfect to me :-) Did you test and find this solved the errors you were seeing?

almasaeed2010:
I have tested this on our live server where the errors where showing up and they are no longer an issue using this branch. 

Thanks!

laceysanderson:
@spficklin does this address your concerns?

spficklin (approved):

spficklin:
Thanks @laceysanderson for looking deeper at that.  Good catch!   Yes, I'm happy with these new changes.

Issue #541, opened 2018-08-10
Enabling Ajax for Fields


almasaeed2010:
I noticed today that on my local dev site, entities are loading their fields using Ajax instead of preloading them on the server. However, this behavior is not happening on our live site and we'd like to know if it is possible to activate it?

almasaeed2010:
I realized that this is the default behavior but enabling tripal_ds disables it.

spficklin:
If you edit a content type and change the Field Display from "hide empty fields" to "show empty fields" then it loads the non "auto-attach" fields via AJAX.   Originally Tripal v3 loaded all fields via AJAX so that the pages would load quickly.  But many folks did not like fields that were empty to show up.

spficklin:
But as @chunhuaicheng pointed out in issue #538 we probably should try to find a way to make AJAX work either way.

almasaeed2010:
Since we know the pane where data is supposed to go to, then we have enough info to delete the pane if the returned data is not available. This however may create an unwanted effect of things disappearing after a few seconds so we could possibly load the panes as hidden by default and do a little css animations as we get data back from the server?

I'll try to look more into this to see if we can come up with a good solution.

laceysanderson:
Listening in on this. I would Love a solution to have AJAX loaded fields while still hiding empty ones from my users!

spficklin:
Just curious if you were able to figure anything out on this one?

almasaeed2010:
Sorry I haven't gotten to this yet. Will work on it and update here soon.

Thanks!

almasaeed2010:
I started working on this and it looks to me that it's easily doable.

tripal.js loads the fields and it can tell if there is content so it can easily broadcast via events to tripal_ds that a field has completed loading. From there, tripal_ds can check if the entire pane is empty and remove it if it is. We can also take this a step further and have tripal.js broadcast the start of loading which will allow tripal_ds to not hide the pane until it's ready (when ajax has completed).

I can implement this and push a branch up for testing if it sounds good?

almasaeed2010:
The only concern is fields that are placed inside anything other than a tripal_pane such as tables.
Take the `description` field in an `organism` page for example, it's placed in a table by default but it can be moved to a list or something else so JS won't be able to predict how to remove it.

Maybe we can add more information to the returned response from the Ajax call to tell us the type of the parent element.

spficklin:
[...]

PR #628, opened 2018-09-04
Remove empty fields from view when loaded via ajax

almasaeed2010:
Issue #541 

## Type(s) of Change(s)

## Description

**NOTE:** this PR is not ready for merging.

This addresses issue #541 where fields are not removed when they are loaded via ajax and have no value.

## Testing?

**1.** Before switching to this branch, edit a content type (e,g. Organism) to not hide fields when empty:
  - Go to structure > tripal content types > edit (organism)
  - Find the dropdown titled `Field Display` and choose `Show empty fields`
  - Now the fields that are not marked as `auto attach` will be loaded via ajax

**2.** Test that the organism page works and all fields are loaded. Check which panes/fields have empty values and display nothing
**3.** Now switch to this branch and clear the cache on the same page.
**4.** Verify that panes that contain empty fields are gone

## Additional Notes (if any):

There is still an outstanding edge case that we have to deal with before fully considering this solution and that's when a pane has multiple ajax fields. We'll need to make sure all fields are loaded before removing the pane.

Thanks!


almasaeed2010:
Note that fields that return values such "there are 0 items available" will not be considered empty.

spficklin:
It works beautifully!  I tested it on both an Organism and mRNA content page.  The page loads quickly and items loading by ajax appear.  The sidebar items disappear from view if there is no data, so that might be a bit disorienting, Any chance that could be reversed and the items appear if there is data?

spficklin:
Oh, but I should say. If we would like to get this merged into v3 stable (which I like that idea) then my suggestion to reverse the changes to the TOC can wait.  The speed improvement is well worth it.

spficklin:
I'm listing as release blocker, but if @laceysanderson or @bradfordcondon or you disagree we can take it off.

almasaeed2010:
Reversing it is pretty simple. We can hide the fields upon page load and make them appear when ready. However, I would suggest instead reducing the opacity and adding a spinner to show that the page is still getting constructed so that when panes disappear it’s a bit more obvious why.

What do you think?

laceysanderson:
I personally prefer reversing it so panes appear as they have data :shrugs:

shawnawsu:
So, now there is no way for a site admin to show empty fields? Say, if someone wanted site visitors to know the field is there but it's empty because they don't have the data.

laceysanderson:
There is still a way. As @almasaeed2010 said,
> Note that fields that return values such "there are 0 items available" will not be considered empty.

We just need to ensure that fields check the bundle setting: 

- if they are supposed to show empty fields then they print something like "there are 0 items available" 
- if not, they don't print anything and this PR makes the field disappear.

That's actually the way its supposed to work currently but I'm not sure if it's consistently implemented since I never want to show my empty fields and thus haven't tested it.

almasaeed2010:
@laceysanderson we can do that 👍 

@shawnawsu we can modify the bundle form to send more  settings to JS to make that possible. Currently, if the field does render a message (something like there are 0 results), the field will still appear.

shawnawsu:
I prefer Lacey's suggestion, I worry about too much JS making the page jumpy. @spficklin thoughts?

spficklin:
In regards to @shawnawsu comment and the subsequent suggestion to have formatters return `there are 0 items available`... To make that happen for empty fields we would would have to ensure that all formatters behaved this way.  That may be too much to ask.  Especially for custom formatters that others make.  I think the best option would be to adjust our drop down on the Content Type Edit page so that users have multple options:  

-  Hide empty fields with AJAX
-  Hide empty fields no AJAX
-  Show empty fields

These options can solve these problems:
1) It allows site admins to show all empty fields as @shawnawsu mentions.
2)  It allows folks who don't like the "jumpiness" of AJAX loaded fields to turn that off and just allow the pages to wait as they currently do now until data is loaded.
3)  It allows for quick page loads via AJAX loaded fields.

edit:  Or alternatively, maybe separate out hiding fields and AJAX into two different options.  So we have one drop down with these options:

- Hide empty fields
- Show empty fields

And another with

- Load via AJAX
- No Ajax loading.

Between those two we have all the combinations we need.

spficklin:
@laceysanderson You wrote:

> I personally prefer reversing it so panes appear as they have data :shrugs:

That's my preference too.

almasaeed2010:
Hello all,

I added the new code to reverse the showing/hiding of fields. 

Regarding the new bundle settings, I am not sure how to go about it since it sets Drupal variables that are used all around Tripal. Some help with that would be appreciated.

spficklin:
@almasaeed2010 we can help with the Drupal settings if other's agree that's the way they would prefer it be managed.

laceysanderson:
> Or alternatively, maybe separate out hiding fields and AJAX into two different options. So we have one drop down with these options:
> 
>     Hide empty fields
>     Show empty fields
> 
> And another with
> 
>     Load via AJAX
>     No Ajax loading.

I'm all for this idea 👍 

So for clarification currently
- `Show Empty Fields` has `Load via AJAX`
- `Hide Empty Fields` has `No AJAX Loading`

This PR currently adds `Hide Empty Fields` with `Load via AJAX`. So we will also need to implement  `Show Empty Fields` with `No AJAX Loading`?

spficklin:
@laceysanderson I believe this PR adds AJAX to the 'Hide Empty Fields' to hide the empty fields and load the non "auto attach" fields..  The 'Show Empty Fields' has always used ajax to load the non "auto attach" fields.

almasaeed2010:
@laceysanderson we would simply disable the `AjaxField` class that's in this commit and run only the loader. 

The only thing we need now is a way to pass the settings down from drupal to JS, which is really simple. Any idea where this should be done?

Thanks!

almasaeed2010:
Oh I didn't notice your comment @spficklin. I didn't change how the PHP handles adding ajax fields. So currently, this activates ajax fields when `Show empty fields` is selected. But Since we adding a new setting to choose when to use Ajax, we can pass it down and activate when `hide empty + load via ajax` are selected.

spficklin:
@almasaeed2010 I'll take a quick look to see where we can adjust the code to allow for the multiple options and get back to you.

shawnawsu:
Hi @almasaeed2010,

I have a branch that creates two variables one for ajax fields and one for hiden fields that are available to the javascript layer. I think this is what you need to finish this up. I've just run into a small snag, I can't push the branch out because the origin is the stantonlab github and I don't have write permissions. 

Given that, what's the best way to make that branch available to you?

Thanks,
Shawna

almasaeed2010:
Hello @shawnawsu 

Thanks for working on this! I think the best way to do this is to push your branch to this repo where I can pull then merge to this PR.

Thanks!

bradfordcondon:
I. Suggest pushing to a remote you do have access to (personal fork or even tripal master) and pring from there to staton if you can

almasaeed2010:
I don't think you can PR to a fork? But it's fine to push to this repo 👍

bradfordcondon:
to clarify, you can indeed PR to a fork even from the master fork- just select "compare to a different fork" when creating the PR.

shawnawsu:
@almasaeed2010 Ok, I just pushed up the branch 541-tv3-ajax_fields-ss for you to have a look at. It exposes two variables, 'tripal_field_settings_hide' and 'tripal_field_settings_ajax' to javascript in Drupal.settings.tripal_ds. And now if you go into tripal content types you will see two checkboxes, one that says 'Hide empty fields' and one that says 'Load field using AJAX '. When they are checked they return true. 

@spficklin The hide_empty_fields variables were used in several places in web services and because I changed the variable I had to go in and change those references, can you check this branch doesn't cause any problems? I just changed the default from 'hide' to TRUE to be consistent.

I forgot! This requires an update so that new ajax_field variable goes into the tripal_variables table.

almasaeed2010:
@shawnawsu thanks for the quick updates! After pulling the branch, updating db and clearing cache, the ajax fields are not showing up anymore. I tried all combinations using both new options "Load field using AJAX" and "Hide empty fields," but the fields are not being sent down with the class `tripal-entity-unattached` that allows them to be loaded via AJAX. Is there a step I am missing?

Thanks!

almasaeed2010:
I just pushed a commit that has @shawnamspoor's changes and mine merged with the addition of JS code to respond to the new settings.

shawnawsu:
@almasaeed2010 Can you check the tripal_bundle_variables table and make sure the correct variable information is being stored?

shawnawsu:
@almasaeed2010 Ignore that last message. I realized it's a mistake I fixed in some places but not in others, the function tripal_get_bundle_variable takes three variables, the last of which is a default value if the variable has no value. The variable in this case is a boolean so even though 'false' is being returned tripal_get_bundle_variable is interpreting that as empty and taking the default value. So I removed the default value from the calls. 

I also needed to change one thing to tripal_entity_view because I was referencing the hide_empty_field and not the ajax_field variable. 

I hope this fixes the issue, let me know!

I forgot, I pushed this all to my 541-tv3-ajax_fields-ss branch.

almasaeed2010:
Ok @shawnamspoor and I got this working. Please take the time to test all combinations of the new settings.

If all works as you'd expect, then the PR is ready to be merged 🎉

spficklin:
Note: in addition to the instructions above, if you are testing you have to run a ```drush updatedb```

spficklin:
@almasaeed2010 and @shawnawsu I tested this just now and  when I have both check-boxes checked the fields are not hidden as expected.   Can you take a look?

@almasaeed2010  can you consider merging in branch 541-tv3-ajax_fields on the Tripal repository into this branch on the statonlab fork?  I have added a new tab to the top of each entity page titled 'Reload'.  It's purpose is to clear the cache for just the entity you're viewing.  You no longer need to clear the entire cache for the site and allows you to test this code quickly.  I've been wanting this for awhile on other things (very handy for testing new fields).  Only users with 'edit' permission have this tab.  @laceysanderson and @bradfordcondon can you give your thumbs up to this addition?

laceysanderson:
@spficklin, I give my thumbs up if we can change the permission to the `admin role` rather then `edit permission`. My reasoning is I don't want my content managers thinking that's the best way to refresh the page and then complaining about speed as the cache is rebuilt ;-p. This is an amazing feature for developers though!

almasaeed2010:
Hello @spficklin 

I merged the branch from this repo to this PR.

For the issue you are experiencing, any idea which fields aren't hiding? Currently, it's working for me for all the fields on the page.

Thanks!

bradfordcondon:
hi all,

the commit history of this PR got messed up and includes all commits to 7.x-3.x in hte past two weeks i think.  Any ideas on how to fix?

almasaeed2010:
I think after merging `541-tv3-ajax_fields` from this repo into this fork brought a few other branches with it. Possibly it had other branches merged into it? I could attempt reverting and repushing.

bradfordcondon:
functionality looks good to me 👍 we just need to fix the commit history, I should pretend to look at the code before I approve :D

Issue #631, opened 2018-09-04
Updated API Functions


spficklin:
## Feature Request/Discussion

- [x] does this feature attempt to solve an existing problem with Tripal?
- [ ] are you open to developing or collaborating on an extension module if this is not a good fit for Tripal Core (no pressure here -just good to know upfront :-) )
- [x] do you **Need** this feature ASAP?

### Description

Some of the Tripal API functions are going on 7-8 years old.  Some of them are very long and their structure can create slowness with loading large datasets in loaders.  They are also hard to write unit tests for.    A revamp of some of these functions (e.g. chado_select_record, chado_insert_record, chado_insert_cvterm) might be a good idea.

spficklin:
This will get addressed in the Tripal v5 design so I'm closing.

PR #627, opened 2018-09-04
New OBO Importer

spficklin:
Issue #525, #120, #303,  #76, #647 

## Type(s) of Change(s)

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] API-specific change (fix or addition to an API function)
- [ ] Updates documentation (inline or markdown files)

## Description

This PR fixes many of the outstanding issues with the OBO loader (see issues above).   In summary:
- Issue #525.  The problems with weird CVs and backwards accessions is fixed.
- Issue #120.  The Plant Trait Ontology now handles "borrowed" (i.e nameless) terms with grace. Although there are still outstanding issues due to Chado design for storing these borrowed terms.
- Issue #76.  The loader now appears to support the 1.4 specification--at least as far as we can go with it into Chado.
- Issue #303. Loading of OBOs is lightning fast now in comparison to how long it used to take.  GO and CHEBI can now both be loaded in less than 15 minutes.  But, populating of the cvermpath may still take a bit of time so it may still take some time but much less time than before.

This PR also includes a new API class which needs review at least by @laceysanderson.  Part of the slowness with loading ontologies was our chado_insert_cvterm, chado_select_record, chado_insert_record, chado_update_record functions and all of the checks that they do.   When those are included in long-running loaders like the OBO loader they become inefficient because they may not need to perform every check and they are often repeatedly used on the same records.   The speed issue was resolved by:
1)  Creation of a new ChadoRecord class that is more lightweight, easier to use, and is much easier to follow the  code.  
2)  The OBO loader makes extensive use of caching in memory.   The loader was repeatedly looking up the same cv's, db's, cvterm's, dbxref's etc.  Thanks to caching of some of these queries are now very limited.  GO uses about 500MB of RAM, CHEBI uses about 1GB of RAM.

I would like to propose this ChadoRecord as a model for replacing some of our older API functions.  We can discuss that in another issue.  But if there is any concern about this ChadoRecord class holding up this PR, I'd suggest we allow the merge for now to allow for a timely v3 stable release and the code can be reworked if need be where these class functions are called.

**Note:**  If this PR is approved we should backport this loader to Tripal v2 to resolve the problems described by these issues #97 and #186.

## Testing?

1) You will need to pull this branch and run the update functions.  
2)  Next ontologies and load them.
[...]

PR #632, opened 2018-09-04
Added ChadoRecord class as a proposal for improved API functions

spficklin:
Issue #631 

## Type(s) of Change(s)

## Description

This PR adds a new ChadoRecord class for review. It is a proposal for a future replacement of our existing chado_insert_record, chado_upate_record, chado_select_record, etc.  Here's how it works:

1) Any chado record is an instance of this class with the table the record belongs to specified in the constructor
   ``` $dbxref = new ChadoRecord('dbxref');```
2) Before you can search, insert or update you need to set values.
   ``` 
     $dbxref->setValues([
        'db_id' => $db->db_id,
        'accession' => $accession
      ]);
   ```
3) Suppose you want to insert.  Before inserting you can check if the record already exists:
   ```$dbxref->find()```
   If only one record exists then the record is automatically updated with any additional values that were not specified in the setValues() function.  If more matches are found then no values are set. The function returns the number of matches records.
4)  If the find() function returns 0 then you can insert:
   ```$dbxref->insert();```
5) Updates and deletes can be performed in a similar way.
6) If you are not sure if the record exists and you don't want to repeatedly perform "finds" within if statements you can simplify with the following:
    ```$dbxref->save();```
   The save function will perform an update if the record exists and an insert if it doesn't.  By having this save function separate it allows us to maintain functionality of the chado_insert_record (with the 'update_if_exists' flag set), but not require it (can just call insert() or update() directly) if a long-running loader doesn't need the extra overhead.
7)  You can retrieve the values of the record in this way:
     - Get all values as an array:
        ```$values = $dbxref->getValues()```
     - Get a single value:
        ```$dbxref_id = $dbxref->getValues('dbxref_id)```
    - Get all values as an object, similar to a fetchObject() on a database query.
       ```$values = (object) $dbxref->getValues()```
8) Some other interesting functions
     - Get the table from the record:
        ```$dbxref->getTable()```
     - Get the ID (primary key value):
        ```$dbxref->getID()```
     - Get the table schema:
        ```$dbxref->getSchema()```

Unlike the current chado_select_record, the values cannot be nested and no nested recursive searching occurs, but that can be easily added.  

## Testing?

This class is not used anywhere and cannot be tested with the current code.  Unit tests are needed to fully test it.

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/api/ChadoRecord.inc
+++ b/tripal_chado/api/ChadoRecord.inc
@@ -125,7 +125,7 @@ class ChadoRecord {
    *   The name of the table that the record belongs to.
    */
   public function getTable() {
-    return $this->table;
+    return $this->table_name;
   }
   
   /**
--- a/tripal_chado/api/ChadoRecord.inc
+++ b/tripal_chado/api/ChadoRecord.inc
@@ -49,6 +49,9 @@ class ChadoRecord {
    * 
    * @param string $table_name
    *   The name of the table that the record belongs to.
+   *
+   * @param string $record_id
+   *  An optional record ID if this record is already present in Chado.
    */
   public function __construct($table_name, $record_id = NULL) {
     
@@ -140,7 +143,7 @@ class ChadoRecord {
    * 
    * If the record already exists it will be updated. If the record does not
    * exist it will be inserted.  This function adds a bit more overhead by
-   * checking for the existance of the record and performing the appropriate
+   * checking for the existence of the record and performing the appropriate
    * action. You can save time by using the insert or update functions directly
    * if you only need to do one of those actions specifically.
    * 
--- a/tripal_chado/api/ChadoRecord.inc
+++ b/tripal_chado/api/ChadoRecord.inc
@@ -1,0 +1,444 @@
+<?php
+
+class ChadoRecord {
+  /**
+   * @var string
+   *   Holds the name of the table that this record belogns to.
+   */
+  protected $table_name = '';
+  /**
+   * @var array
+   *   Holds the Drupal schema definition for this table.
+   */
+  protected $schema = [];
+  /**
+   * @var array
+   *   Holds the values for the columns of the record
+   */
+  protected $values = [];
+  
+  /**
+   * @var array
+   *   An array of required columns.
+   */
+  protected $required_cols = [];
+  
+  /**
+   * @var integer
+   *   The numeric Id for this record.
+   */
+  protected $record_id = NULL;
+  
+  
+  /**
+   * @var string
+   *   The column name for the primary key.
+   */
+  protected $pkey = '';
+  
+  
+  /**
+   * The list of column names in the table.
+   * @var array
+   */
+  protected $column_names = [];
+  
+  
+  /**
+   * The ChadoRecord constructor
+   * 
+   * @param string $table_name
+   *   The name of the table that the record belongs to.
+   */
+  public function __construct($table_name, $record_id = NULL) {
+    
+    if (!$table_name) {
+      $message = t('ChadoRecord::_construct(). The $table_name argument is required for a ChadoRecord instance.');
+      throw new Exception($message);
+    }
+    
+    // Set the table name and schema.
+    $this->table_name = $table_name;
+    $this->schema = chado_get_schema($this->table_name);
+    if (!$this->schema) {
+      $message = t('ChadoRecord::_construct(). Could not find a matching table schema in Chado for the table: !table.',
+        ['!table' => $this->table_name]);
+      throw new Exception($message);
+    }
+    
+    // Chado tables never have more than one column as a primary key so
+    // we are good just getting the first element.
+    $this->pkey = $this->schema['primary key'][0];
+    
+    // Save the column names.
+    foreach ($this->schema['fields'] as $column_name => $col_details) {
+      $this->column_names[] = $column_name;
+    }
+    
+    // Get the required columns.
+    foreach ($this->schema['fields'] as $column => $col_schema) {
+      foreach ($col_schema as $param => $val) {
+        if (preg_match('/not null/i', $param) and $col_schema[$param]) {
+          $this->required_cols[] = $column;
+        }
+      }
+    }
+    
+    // If a record_id was provided then lookup the record and set the values.
+    if ($record_id) {
+      try {
+        $sql = 'SELECT * FROM {' . $this->table_name . '} WHERE ' . $this->pkey . ' = :record_id';
+        $result = chado_query($sql, [':record_id' => $record_id]);
+        $values = $result->fetchAssoc();
+        if (empty($values)) {
+          $message = t('ChadoRecord::_construct(). Could not find a record in table, !table, with the given !pkey: !record_id.',
+            ['!pkey' => $this->pkey, '!record_id' => $record_id, '!table' => $this->table_name]);
+          throw new Exception($message);
+        }
+        $this->record_id = $record_id;
+        $this->values = $values;
+      }
+      catch (Exception $e) {
+        $message = t('ChadoRecord::_construct(). Could not find a record in table, !table, with the given !pkey: !record_id. ERROR: !error',
+          ['!pkey' => $this->pkey, '!record_id' => $record_id, '!table' => $this->table_name, '!error' => $e->getMessage()]);
+        throw new Exception($message);
+      }
+    }
+  }
+  
+  /**
+   * Retrieves the record ID.
+   * 
+   * @return number
+   */
+  public function getID() {
+    return $this->record_id;
+  }
+  
+  /**
+   * Retrieves the table name.
+   * 
+   * @return string
+   *   The name of the table that the record belongs to.
+   */
+  public function getTable() {
+    return $this->table;
+  }
+  
+  /**
+   * Retrieves the table schema.
+   * 
+   * @return array
+   *   The Drupal schema array for the table.
+   */
+  public function getSchema() {
+    return $this->schema;
+  }
+
+  /**
+   * Performs either an update or insert into the table using the values.
+   * 
+   * If the record already exists it will be updated. If the record does not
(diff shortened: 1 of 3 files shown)


almasaeed2010:
This looks great! 

My only suggestion would be to change the name of the `save()` method to something that explains its functionality a bit more. For example, `findOrInsert()`.

I am willing to write unit tests for this if you'd like. Wouldn't take me long to test all of the methods. We can push to another branch and merge it into this branch when ready.

spficklin:
Note this class was used in PR #627 for the new OBO loader and is an important part for why performance increased dramatically.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,444 @@
...
+      $insert_cols[] = $column;
+      $insert_vals[] = ':' . $column;
+      $insert_args[':' . $column] = $value;
+    }
+    $sql = 'INSERT INTO {' . $this->table_name . '} (' . 
+      implode(", ", $insert_cols) . ') VALUES (' . 
+      implode(", ", $insert_vals) . ')';
+    try {
+      chado_query($sql, $insert_args);
+      // TODO: we can speed up inserts if we can find a way to not have to
```
almasaeed2010: If we can use `db_insert()` instead of raw SQL, it'll return the new id: https://api.drupal.org/api/drupal/includes%21database%21query.inc/function/InsertQuery%3A%3Aexecute/7.x

spficklin: we would need an equivalent chado_insert() function, just like we have a chado_query for db_query. But, yes that would do it.

spficklin:
@almasaeed2010, The save() function also does a find() and either an update() or an insert() depending if the record exists or not.  I do like save, but if @laceysanderson and others agree we can change it.

spficklin:
Oh and Thanks for offering to do unit tests @almasaeed2010 I think @laceysanderson indicated she might do some as well so maybe a divide and conquer approach?  I can chip in too.

almasaeed2010:
Oh I wasn't aware that it would also update if the record exists 👍

tripal_chado/api/ChadoRecord.inc
laceysanderson: We can use PostgreSQLs `RETURNING` keyword: https://www.postgresql.org/docs/8.4/static/sql-insert.html

```sql
INSERT INTO distributors (did, dname) VALUES (DEFAULT, 'XYZ Widgets') RETURNING did;
```

almasaeed2010:
Since I can't commit directly to this branch, I think it's better to have you and @laceysanderson on it 👍 But please let me know if I can be of any help.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,444 @@
...
+         '!record_id' => $this->record_id,
+         '!error' => $e->getMessage()]);
+        throw new Exception($message);
+    }
+  }
+  /**
+   * A general-purpose setter function to set the column values for the record.
+   * 
+   * This function should be used prior to insert or update of a record. For
+   * an update, be sure to include the record ID in the list of values passed
```
bradfordcondon: Why do i need to include the record ID in the update?  Didnt I include it in the construct?

spficklin: No, the construct only provides the table name.  You have to set the record ID if you are doing an update, or the record ID gets set when you do a find() using the values provided.

bradfordcondon: but the construct takes an optional `record_id`  parameter

`  public function __construct($table_name, $record_id = NULL) {`

if i construct with a record id, why would i have to explicitly put it into the $values for an update?  Let's just check if the record_id is set and, if so, add it to the $values array?

spficklin: Oh... I forgot I added that.   Yes, if you included it in the constructor then you are good to go.  Maybe that description can be reworded to make better sense.  It is meant to say that you must have a record ID set before you can run the update.  If that ID gets set in the constructor or via the setValues() or setValue() it doesn't matter.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,444 @@
...
+  }
+  
+  /**
+   * Retrieves the table name.
+   * 
+   * @return string
+   *   The name of the table that the record belongs to.
+   */
+  public function getTable() {
+    return $this->table;
```
bradfordcondon: oops, this needs to be `$this->table_name`

tripal_chado/api/ChadoRecord.inc
bradfordcondon: ok, yes thats how i would expect it to work, thanks for clarifying.  Let's say the record_id must be included in `$values` if it was not already provided when the class instance was constructed.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,530 @@
...
+    
+    // Make sure we have values for this record before updating.
+    if (empty($this->values)) {
+      $message = t('ChadoRecord::update(). Could not update a record into the table, !table, without any values.',
+        ['!table' => $this->table_name]);
+      throw new Exception($message);
+    }
+    
+    // We have to have a record ID for the record to update.
+    if (!$this->record_id) {
```
laceysanderson: Perhaps we could make this more fault tolerant by executing a find here and then throw an exception if we still don't have a record_id.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,530 @@
...
+   * 
+   * A record ID must be part of the current values.
+   * 
+   * @throws Exception
+   */
+  public function delete() {
+    
+    // We have to have a record ID for the record to be deleted.
+    if (!$this->record_id) {
+      $message = t('ChadoRecord::delete(). Could not delete a record in the table, !table, without a record ID.',
```
laceysanderson: Same as with update, perhaps we should try to find if we don't have a record_id -> don't give up so easily ;-p

tripal_chado/api/ChadoRecord.inc
laceysanderson: Keep in mind, the record_id could have been set via find() before the update was executed. As such, perhaps we should say that the record_id needs to be present in the values, whether it's set during construction, set using setValues or retrieved using find.

tripal_chado/api/ChadoRecord.inc
```diff
@@ -0,0 +1,530 @@
...
+      else {
+        $message = t('ChadoRecord::setValues(). The column named, "!column", does not exist in table: "!table". Values: !values".', 
+          ['!column' => $column, '!table' => $this->table_name, '!values' => print_r($values, TRUE)]);
+        throw new Exception($message);
+      }
+    }
+  
+    // Make sure that the user did not miss any required columns or has
+    // set a column to be NULL when it doesn't allow NULLs.
+    foreach ($this->required_cols as $rcol) {
```
laceysanderson: If I'm intending to find, I should be aloud to not set all required values. I may know that the combination is sufficiently unique.

laceysanderson (changes requested):
So far, I'm enjoying using this class :-) However, there are a few things I think are needed to make it more friendly to users:
- We need to be able to specify values through the foreign key:
```php
             $record->setValues([
               'name' => 'My Favourite Plant',
               'uniquename' => 'Plectranthus scutellarioides Trailing Plum Brocade',
               'organism_id' => [ 'genus' => 'Tripalus', 'species' => 'databasica' ],
               'type_id' => [ 'name' => 'sample', 'cv_id' => [ 'name' => 'Sample processing and separation techniques' ] ],
             ]);
```
 - We need to follow foreign keys when returning values.
 - It needs to be a little more fault tolerant: you don't have a record_id, maybe we should try finding it first?
 - We don't have support for many of the options we did using the api functions directly... We should at least add todo's to support them.

Also, we can only select a single record? This is a barrier to replacing chado_select_record()...

spficklin:
Hi @laceysanderson I agree with you 100%.  I think we need to work in those changes.   For the sake of timing can we merge this in now, so the OBO loader updates can be merged?   We do not have to advertise use of this class in the API until we get these fixes that you suggest.

bradfordcondon:
we're all in agreement that this class should ultimately cover all the functionality that the previous API did.  However as that API is still available I don't think its a make or break for merging and getting it started.

I therefore approve this.

tripal_chado/api/ChadoRecord.inc
bradfordcondon: Isn't this  like saying the PDO `->fetchAll()` command should work even if you forget to `->execute()` beforehand?

What if your setValues matches multiple records, but you are assuming it will only match 1?  I feel like those sorts of mistakes will be easier to make if we auto-set record_id if the user forgot to.

laceysanderson:
Ok, I hear you :-) I've created #634 to keep track of what needs to be done.

I would like to fix a few more things before merging. Specifically, 
- the required values check needs to move into the update/insert. For performance, I'm going to make the check in setValues but not emit an error -simply save the result to be checked quickly in insert/update.
- If you set a key to an array you receive an SQL error as it tries to implode multiple values.
- Look through the old api functions to see if there's any major validation we're missing.

I'll do the above right now due to the timeline mentioned above 🙂  👍

laceysanderson (approved):

laceysanderson:
There. My main concerns are fixed and I've added inline documentation for outstanding issues using the `@todo` tag. I now approve this PR. @spficklin you can merge if you agree with my changes 🙂

spficklin:
Thanks @laceysanderson .  It looks good!  I made some adjustments to comments (can't go past the 80 char line for everything (except code example lines perhaps)), and in the find() function I set the missing_reqiured_cols array to be empty as the find will import the record and there will no longer be missing values.

I'm going to let the Travis CI run and verify nothing is broken and then merge.  Thanks for your quick review!

Issue #633, opened 2018-09-04
DRUPAL_HOME env variable for User's Guide


spficklin:
What do folks think if we alter the User's Guide to include setting an environment variable of $DRUPAL_HOME.  This way, we can use that in the command-line instructions and folks can easily just cut-and-paste rather than cut-and-pasting and then editing if their setup is different?  

We can add notes to the top of major section pages reminding folks that they need the DRUPAL_HOME env set to run the commands on the page.

bradfordcondon:
yep, good idea.

bradfordcondon:
started this.  did one page but ill do the rest: https://github.com/tripal/tripal/tree/633_drupal_home_env

bradfordcondon:
merged, thanks!

Issue #387, opened 2018-05-25
Docs: Tripal Daemon Documentation


spficklin:
We need better documentation for how to install and use the Tripal Daemon on the Tripal v3 User's Guide. Right now new users have to traverse multiple sites to figure it out.

laceysanderson:
Added to the upgrade Docs (`http://tripal.info/node/269`):

> (Optional but Recommended) Tripal Daemon (provides automatic job execution) is now part of the main Tripal. To use it, first enable the module,
>
>  `drush pm-enable tripal_daemon`
>    - NOTE: If you had Tripal Daemon installed with Tripal 2, first disable the module and remove the module directory.
>    - The Tripal Daemon runs in the background, checking periodically for Tripal Jobs and running them once found.
>    - Start the daemon using the following command. For more information about using the Daemon, see the [Jobs Management section](http://tripal.info/tutorials/v3.x/tripal-daemon).
>
> `drush trpjob-daemon start`

Added the contents of the README to http://tripal.info/tutorials/v3.x/tripal-daemon. This is still not ideal but it's a start...

laceysanderson:
Also need to add information here: http://tripal.info/tutorials/v3.x/installation/job-automation.

laceysanderson:
Also need to add it to the user tutorial so no one needs to run the tripal jobs command manually

kyrenya:
hi! I'm trying to use the tripal_blast module, and actually got stuck in the tripal_daemon documentation. 
I get this message 
The Drush Daemon API module requires the PHP-Daemon library v2.0 (Download) Please install it as per the Libraries API Instructions before trying again. (Currently using PHP Daemon Not Installed)

but I cannot figure out how to install the library needed. 
when I do 
l$ drush pm-download Libraries
No release history was found for the requested project (Libraries).                   [error]
Could not download requested project(s).                                              [error]

Any help? 
Thanks in advance!

laceysanderson:
Hi @kyrenya,

I think it's a matter of case sensitivity for your drush command as the following works for me: 
```
drush pm-download libraries
drush pm-enable libraries
```

kyrenya:
yep, oops, I was copy-pasting and I didn't pay attention to that, thanks!

laceysanderson:
Glad it works for you now! I updated the docs with the fix -thank you for bringing it  to our attention!

eglabmsu:
I had trouble with the line
`mv v2.0.tar.gz PHP-Daemon`
[...]

PR #637, opened 2018-09-10
633 implement drupal home env in install documentation

bradfordcondon:
Issue #633 

## Type(s) of Change(s)

## Description

Documentation improvement.  We use an environmental variable for `/var/www/html` to make copy-pasting the commands easier for those installing elsewhere. 

I also made some typo correction and formatting changes as i spotted them (ensuring subsections were underlined, trying to use the code markup for paths, etc)

## Testing?
To build the site locally and look:

```
cd docs
make html
```
open `_build/index.html`: all changes are in the installation portion of the user's guide.

Here's an example of the notice and variable.

screen shot 2018-09-09 at 10 13 28 pm


docs/user_guide/install_tripal/manual_install/install_drupal.rst
```diff
@@ -78,11 +85,11 @@ If an index.html file is present (as is the case with Ubuntu installations) you
 
 .. note::
 
-  It is extremely important the the hidden file .htaccess is also moved (note the second 'mv' command above. Check to make sure this file is there
+  It is extremely important the the hidden file ``.htaccess`` is also moved (note the second ``mv`` command above. Check to make sure this file is there:
 
-.. code-block:: bash
+  .. code-block:: bash
```
bradfordcondon, line 86: Note this isn't simply whitespace changing: this moves the codeblock *INSIDE* the note.  I think this is more appropriate.

laceysanderson (commented):
Your changes look good 👍 

We also mention `/var/www/html` in other places though. For example, whenever we execute a tripal job:
```
drush trp-run-jobs --username=administrator --root=/var/www/html
```

We should likely change it throughout the user guide 😬

bradfordcondon:
yeah i left it out of that command- adding the `$DRUPAL_HOME` warning to every page that runs a job seems really cumbersome.    Is there another way?

laceysanderson:
I'm approving this PR. On further reflection -there is another way: We actually use the Tripal Daemon in the tutorial so no one has to manually run the tripal jobs command... As such, I've added that to #387 to remind me.

spficklin:
I approve this pull request but I made some changes and I think someone else should double check it one more time.   I thought the "note" text was a bit too long for the pages on which it appeared.  So, I created a new page at the very beginning of the guide that describes the DRUPAL_HOME environment variable.  Then everywhere the variable is used on a page I altered the "note" to just be a very short reminder with a link back to that new page.  

I also agree with @laceysanderson that we should change `/var/www/html` everhwhere it is used to be `$DRUPAL_HOME` so I made that change and added the note to the top of every page that uses it.

spficklin:
Here's what the new DRUPAL_HOME Variable page looks like:
screenshot from 2018-09-21 23-29-32


Here's a page with the note at the top:
screenshot from 2018-09-21 23-22-08

laceysanderson:
I re-tested after @spficklin's changes and think it's 👍 My review stands. Are you content with this change @bradfordcondon? Specifically, the $DRUPAL_HOME page added at the beginning.

bradfordcondon:
Can we maybe move the DRUPAL_HOME variable page one layer in, maybe into install_tripal?  It looks weird top-level, and it shouldnt be a problem that its hidden because it'll still be one of the first pages when clicking through and all guides that utlize the variable will link to it.

spficklin:
I thought about putting it in the install portion but we use that variable all throughout the User's Guide not just in the installation.  What if we renamed the page to something like "Introduction" and add a bit of an explanation for what the User's Guide is and the DRUPAL_HOME setup is part of that page?

bradfordcondon:
I'd be happy with putting it as a subheading in a less technically named section (ie introduction).  That said, because you are linking to the guide in a note at the top of each page, i dont think its a problem to put it in the install guide.  I'll leave that decision up to you.

I looked over the other changes you made and i like it 👍 thank you

laceysanderson:
My inclination would be to nest it in with install. We do use it throughout but since we are linking to it, I think it’s ok for it not to be on the top level :shrugs:

spficklin:
Okay, I've moved the page into the "Install Tripal" section and updated all the links:

screenshot from 2018-09-24 09-01-31

I think it's ready for merging if you two approve.

bradfordcondon:
approved.  thats 3/3 so lets merge.

Issue #603, opened 2018-08-28
Docs: How to create a custom Data Loader


laceysanderson:
We need documentation on how to use the importer class to make a custom data loader.
https://tripal.readthedocs.io/en/latest/dev_guide/custom_data_loader.html

bradfordcondon:
@laceysanderson have you started this?  I can draft it if not.

laceysanderson:
No I haven't started it yet. Go for it!

bradfordcondon:
First draft posted as PR.  would welcome any and all suggestions.

bradfordcondon:
merged. live at https://tripal.readthedocs.io/en/latest/dev_guide/custom_data_loader.html

PR #643, opened 2018-09-12
503 custom data loader

bradfordcondon:
Issue #603 

## Type(s) of Change(s)

## Description

I would consider this my first draft of the TripalImporter guide.  any and all feedback is welcome.  I went ahead and recorded the whole thing as a youtube video too which I'll post in two parts, with "unit testing" as its own video.

## Testing?
```
cd docs
make html
```
open the index.html in `_build/` and visit the "custom data loader" page.

screen shot 2018-09-12 at 12 46 39 pm


laceysanderson (changes requested):
First off, Thank you! Overall, this is Great :-D

Suggestions/Problems:
- I see these errors on build:
```
/Users/laceysanderson/tripal/docs/dev_guide/custom_data_loader.rst:82: WARNING: Mismatch: both interpreted text role prefix and reference suffix.
/Users/laceysanderson/tripal/docs/dev_guide/custom_data_loader.rst:201: WARNING: Inline literal start-string without end-string.
```
- "You should place this class in its own .inc file at includes/TripalImporter/". It would be good to specifically say: "in your own custom module" here so that no one kills kittens.
- "There is no need to **include** the importer in your module" sounds a bit like you don't need to even have the file ;-p It would be enough to make `include` in-line code.
- "The variables that are private static should not be changed." This would be more obvious in a warning block :-)
- I see "The global settings above ::ref:`<Static Variables>`_ provide some elements “out of the box”." the RST syntax in the rendered doc.
- "A totally empty TripalImporter class can provide the tow below components" Small typo for "tow" => "two"
- In your third code block: `$options[0] = '--please select an option--';` It's better to set `#empty_option` rather then add to the options array.
- Also in your third code block: the indenting is off:
screen shot 2018-09-13 at 3 23 00 pm
- Your example should mark the field as `#required` rather then check that in validate. I know you were looking for something simple to check but I worry people will then think it's always their responsibility to check required. The cvterm exists wouldn't be too much if you use chado_select_record, right?
- "on my site its: ``drush trp-run-jobs –username=admin –root=/Users/chet/UTK/tripal `) to run " the in-line code sample is missing the second backtick.
- Something went awry here:
screen shot 2018-09-13 at 3 53 49 pm

Other comments:
- I Love the inclusion of testing and best practices!

bradfordcondon:
thank you for the proof checking!  I'll make the changes next week.

spficklin:
@bradfordcondon this is wonderful!  Thank you for taking time to work on this.   I went through it and made some edits.  I hope you don't mind.   Please check it over and make sure I didn't mess anything up.  Also, I think you can remove the sections titled "Test Data" and "Writing the Tests".  The tutorial is meant to be about using the TripalImporter and other than giving the reader setup instructions for how to initialize their loader for unit testing I don't think you need to go into how to test it as writing unit tests is generally the same no matter what your testing and I think we can assume that if the reader has no experience with unit testing that we can refer them to another document to learn about it.  

If you like my edits please mark this as approved by me.

laceysanderson:
I like having the tests in the tutorial. I think importers are the most important place for unit tests to be and including it in the base tutorial shows that and makes it more likely extension developers will do it...

bradfordcondon:
I’m happy to link to documentation hosted instead in tribal test suite.  However I do think at least understanding that the class isnt available without explicitly loading the file is a major stumbling block and I’d like to include it here.

spficklin:
@bradfordcondon yes I agree.  I think we should keep the intro for the unit testing so that folks know how to set up the importer for testing.  It was those two sections I specifically mentioned that I think should come out.  @laceysanderson, so if we just take out those two sections we still have instructions for setting up an importer class for testing, it just doesn't go into further details specifically for testing the ExampleImporter.   I think that would just be duplication of documentation.  Writing unit tests is documented elsewhere.

bradfordcondon:
thanks for the clarification, sorry responded on my phone initially so misunderstood.  I'll take care of this today, thank you both for the feedback.

bradfordcondon:
@laceysanderson 


Looks to me like stephen has addressed almost all your concerns.  Thanks stephen!

>Your example should mark the field as #required rather then check that in validate. I know you were looking for something simple to check but I worry people will then think it's always their responsibility to check required. The cvterm exists wouldn't be too much if you use chado_select_record, right?


I don't understand "The cvterm exists wouldn't be too much if you use chado_select_record, right?"


How would you feel about adding a ``..note`` directive in the Validation section that says "we could have used `#required` instead of using formValidate"?  For some reason I'm struggling to think of a simple example to stick in for validation that wouldn't throw off the rest of the document...  Maybe a type_id for the feature and we can double check that features exist with that type_id?  


@spficklin 

Do you think the logging section should be its own document that we link to? I'm kind of torn.  

I removed the two sections you suggested. I'll reproduce the whole testing section in Tripal Test Suite and link to it here, I think.  https://github.com/statonlab/TripalTestSuite/issues/113

laceysanderson:
I just though an example of testing a cvterm exists would also be super helpful since it's something common. For the example, I would just do something like (untested without re-building your tutorial -sorry time-crunch):
```php
// Here we ask for the cvterm in chado associated with our users choice.
// This ensures their choice exists in the database, protects you from bugs
// caused by assumptions in the submit, hacker-proofs your submit from SQL injection.
$cvterm = chado_select_record('cvterm',array('cvterm_id'), array('cvterm_id' => $form_state['values']['cvterm_select']));
if (!$cvterm) {
  form_set_error('cvterm_select', 'the cvterm must exists');
}
```

That said, the note would likely be sufficient :-) I hadn't thought of that option. Just make sure to say using `#required` is Best practice.

laceysanderson (approved):

laceysanderson:
You have my approval after the addition of note or updated validation -whichever you decide. You can either add the tag yourself or, ping me if you'd like someone else to re-build for one last look at it.

spficklin:
I think the logging is an integral part of the TripalImport class so it needs to be in the document.  Tripal itself does not have any consistent error or logging infrastructure, so the TripalImporter class is the only one that has a specific function to handle it.

bradfordcondon:
I'm going with a notice in the Validator section, simply because its doesn't require me to rethink the example.  The thing about the example you propose is that it isn't something that needs to be validated, as we provide the cvterms from chado_get_cvterm--- we could allow free-text cvterm selection but i think that is too "advanced".

  That said I wouldn't be offended if it was redone.

screen shot 2018-09-17 at 4 36 47 pm

bradfordcondon:
@spficklin waiting on your approval to merge

I see you've marked it as approved with your edits.  I'm considering this PR approved then.  Thank you @laceysanderson and @spficklin for your work on this with me!

laceysanderson:
Awesome, Thank you for the tutorial @bradfordcondon!!!

Yosra3:
Hello 
Thank you for the tutorial it is so clear but i have a problem. Now my file ( with 2 columns in your example ) is imported on tripal but i can't find the data (test_gene_1 and test_gene_2) in my database.
there is a warning with the drush command : 

fgets() expects parameter 1 to be resource, string given ExampleImporter.inc:152                                                [warning]
The feature, , does not exist in the database

Thank you !!

bradfordcondon:
hi @Yosra3 .  Looks like `  while ($line = fgets($file_path)) {` failed for you, so it tried to load afeature named `` and couldn't.  Looking at the example on the guide, `fgets` is being passed a string and it instead needs to be passed an open file.  Something like this:


```
          $in_fh = fopen($file_path, "r");

  while ($line = fgets($in_fh)) {


```

I'm going to create an issue so that we can continue the discussion there if necessary and i can update the guide and tie the fix to the discussion.

Thanks

PR #644, opened 2018-09-12
remove php tag and fix site build to use php syntax highlighting prop…

bradfordcondon:
…erly

          

Issue N/A

## Type(s) of Change(s)

## Description

This is a minor pR only requiring one reviewer.

It's a big Quality of Life fix for the documentation, though.  What I did was update the build `conf.py` to fetch the correct PHP Lexer.  Previously, if you wanted the PHP syntax highlighting to show up right, you needed to add `<?php` to the head of EVERY PHP block.  Not only was this annoying, but it also messed up my IDE, making it think the whole document was in PHP.

By changing the conf.py we can just use the ``.. code-block php`` directive and the highlighting will be correct (albeit ugly: fixing the theme is on my wish list).

This PR also removes the unnecessary `<?php` tags from the codeblocks.

### Before
screen shot 2018-09-12 at 2 24 41 pm
### After
screen shot 2018-09-12 at 2 24 25 pm

please notice in the first code block, there is no syntax highlighting.  In the second, we have the `<?php` tag which is unnecessary and can mess up your IDE.

In the after, both blocks have syntax highlighting without a `<?php` tag.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,6 +6,16 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+
+#PHP stuff
+#see: https://www.sitepoint.com/using-sphinx-for-php-project-documentation/
+
+from sphinx.highlighting import lexers
+from pygments.lexers.web import PhpLexer
+lexers["php"] = PhpLexer(startinline=True, linenos=1)
+lexers["php-annotations"] = PhpLexer(startinline=True, linenos=1)
+
+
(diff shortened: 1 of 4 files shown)


laceysanderson (approved):
I've tested this and confirmed the syntax highlighting throughout the docs 👍

laceysanderson:
Since it only requires one review and I've approved it, I'm also merging :-) Thanks @bradfordcondon !!!

Issue #635, opened 2018-09-06
Importer jobs fail if web directory is mounted to a different location


chunhuaicheng:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.59
* PostgreSQL Version: 10.5.1
* PHP Version: 7.2.7

### Issue description

The importer jobs (e.g. FASTA importer or GFF3 importer) will fail if the Drupal web directory is mounted on a different location and Drush command is launched in that mounting point. We have a login node that everyone works on and it has all web directories mounted as NFS shares. Tripal 2 does not have any problem with this setup but Tripal 3 does.

#### Steps to reproduce
1. mount your Drupal web directory on a different location, (e.g mount /local/sites/tripal3 to /mnt):
sudo mount --bind /local/sites/tripal3 /mnt

2. Submit an importer job (e.g: admin/tripal/loaders/chado_gff3_loader or admin/tripal/loaders/chado_fasta_loader)

3. Launch the job from the mounting point and the job will fail
cd /mnt
drush trp-run-jobs --username=admin

#### Error messages and screenshots

```
2018-09-06 11:16:40: There are 1 jobs queued.
2018-09-06 11:16:40: Job ID 5350.
2018-09-06 11:16:40: Calling: tripal_run_importer(7)
PHP Fatal error:  Cannot redeclare hook_handle_uploaded_file() (previously declared in /mnt/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc:36) in /local/sites/tripal3/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc on line 38
Drush command terminated abnormally due to an unrecoverable error.                                                                                                                          [error]
Error: Cannot redeclare hook_handle_uploaded_file() (previously declared in /mnt/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc:36) in
/local/sites/tripal3/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc, line 38

```

bradfordcondon:
thank you for reporting and approving @chunhuaicheng

PR #648, opened 2018-09-13
fixed issue #635 that importer jobs fail when launching in an NFS mount

chunhuaicheng:
Remove a redundant inclusion of a library file i.e. 'tripal/api/tripal.importer.api'.

          

Issue [[#635](https://github.com/tripal/tripal/issues/635)]

## Type(s) of Change(s)

## Description

'tripal/api/tripal.importer.api' was loaded redundantly in the 'TripalImporter.inc'. Removing the inclusion of this file solved the problem. Said file was already loaded when the the tripal module loads (See line 26 of 'tripal.module')

## Testing?

Before pulling this branch, the following steps will fail to launch the importer job.   

1.  mount your Drupal web directory on a different location, (e.g mount /local/sites/tripal3 to /mnt):
    sudo mount --bind /local/sites/tripal3 /mnt

 2. Submit an importer job (e.g: admin/tripal/loaders/chado_gff3_loader or admin/tripal/loaders/chado_fasta_loader)

 3. Launch the job from the mounting point and the job will fail
    cd /mnt
    drush trp-run-jobs --username=admin

Now pull this branch and submit and launch a new job as in step 2 & 3. The importer job should work.

## Additional Notes (if any):

### DIFF
--- a/tripal/includes/TripalImporter.inc
+++ b/tripal/includes/TripalImporter.inc
@@ -427,9 +427,7 @@ class TripalImporter {
     // Add a job to run the importer.
     try {
       $args = array($this->import_id);
-      $includes = array(
-        module_load_include('inc', 'tripal', 'api/tripal.importer.api'),
-      );
+      $includes = array();
       $job_id = tripal_add_job($class::$button_text, 'tripal',
           'tripal_run_importer', $args, $user->uid, 10, $includes);
 


almasaeed2010:
Hello @chunhuaicheng,

Just for reference when others come to review this PR. This is the line that loads the importer api:

https://github.com/tripal/tripal/blob/b83136aec50b012ef54f90553ccc9d2443a4fa29/tripal/tripal.module#L976

Since the line is in `tripal.module`, then it is getting called twice and one of the calls should be removed. However, the `module_load_inc` uses `require_one` to load the file and keeps a static cache of the requested files itself so that it doesn't load the same file twice. So I am not sure how this solves the NFS mount problem.

Thanks!

chunhuaicheng:
Thank you @almasaeed2010 for reviewing the PR. I think require_once did not work because the mounting point uses a different path from the original Drupal root. From the error message, this file is loaded twice because of the path difference. i.e. in my case:
/mnt/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc (mount point)
and 
/local/sites/tripal3/sites/all/modules/tripal/tripal/api/tripal.importer.api.inc (drupal root)

bradfordcondon:
jobs continue to run fine on this branch for me.  The change itself is pretty logical: the API file is included in `tripal.module` already so no need for the Importer to import it.

👍 thanks for the submission @chunhuaicheng

Issue #652, opened 2018-09-14
Deleting an entity does not delete its record from the chado_bio_data_X tables


almasaeed2010:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

When deleting an entity using the GUI, the record is removed from `tripal_entity` but is not removed from `chado_bio_data_X`.

#### Steps to reproduce

1. Visit any entity that you'd like to delete
1. Find its `entity_id` and make sure it exists in one of the `chado_bio_data_X` tables.
1. Delete the entity using the GUI: click `Edit` -> `Delete`
1. Check the chado_bio_data_X table once more to make sure it was deleted.
1. In my case, the record still exists.

#### Error messages and screenshots

No error message for this issue.

Thanks!

almasaeed2010:
PR submitted #653

bradfordcondon:
merged by @laceysanderson thank you for the fix @almasaeed2010

PR #653, opened 2018-09-14
Delete removed entity from chado_bio_data_x

almasaeed2010:
Issue #652 

## Type(s) of Change(s)

## Description

As detailed in issue #652, deleting an entity using the GUI (or the `entity_delete` function), does not delete the corresponding record in the `chado_bio_data_x` table. This PR uses the `hook_entity_delete` hook to act on the deletion and remove the record from the other  table as well.

## Testing?

1. Before pulling this branch, create an organism entity and find the new record in `chado_bio_data_x`, where `x` is the `bundle_id` for `organism`
1. Delete the newly created organism using the GUI and you'll notice that the record still exists in the `chado_bio_data_x` table but does not exist in `tripal_entity`.
1. Now pull this branch and repeat the steps.
1. The record should no longer exist in the bundle table.

## Additional Notes (if any):

### DIFF
--- a/tripal_chado/includes/tripal_chado.entity.inc
+++ b/tripal_chado/includes/tripal_chado.entity.inc
@@ -148,7 +148,7 @@ function tripal_chado_entity_delete($entity, $type) {
 
   // Delete the removed entity from the corresponding chado_bio_data_x table
   $bundle = $entity->bundle;
-  db_delete("chado_{$bundle}")->condition('entity_id', $entity->id);
+  db_delete("chado_{$bundle}")->condition('entity_id', $entity->id)->execute();
 }
 
 /**
--- a/tripal_chado/includes/tripal_chado.entity.inc
+++ b/tripal_chado/includes/tripal_chado.entity.inc
@@ -142,9 +142,13 @@ function tripal_chado_entity_update($entity, $type) {
  * Implements hook_entity_delete().
  */
 function tripal_chado_entity_delete($entity, $type) {
-  if ($type == 'TripalEntity') {
-
+  if ($type !== 'TripalEntity') {
+    return;
   }
+
+  // Delete the removed entity from the corresponding chado_bio_data_x table
+  $bundle = $entity->bundle;
+  db_delete("chado_{$bundle}")->condition('entity_id', $entity->id);
 }
 
 /**


tripal_chado/includes/tripal_chado.entity.inc
```diff
@@ -142,9 +142,13 @@ function tripal_chado_entity_update($entity, $type) {
...
 function tripal_chado_entity_delete($entity, $type) {
-  if ($type == 'TripalEntity') {
-
+  if ($type !== 'TripalEntity') {
+    return;
   }
+
+  // Delete the removed entity from the corresponding chado_bio_data_x table
+  $bundle = $entity->bundle;
+  db_delete("chado_{$bundle}")->condition('entity_id', $entity->id)->execute();
```
almasaeed2010, line 151: Do we need a check here to verify the table exists? Or can we safely assume that it will exist?

chunhuaicheng:
I tested the code and it did delete the entry from 'chado_bio_data_x'. However, I think it should delete the corresponding chado record as well. Here is our use case:

Our curator uses the web interface to delete unwanted pubs that were imported by the Tripal pub importer. After she deletes the pubs, they still show up in the Tripal Pub Search (/find/publications) results, even after applying you fix.

almasaeed2010:
Hello @chunhuaicheng, 
Thanks for testing!

I think this is something the PMC (@spficklin, @laceysanderson, @bradfordcondon) should decide on. I do agree though that we need a method to delete from chado but I am not sure if there are cases where we don't want that behavior when deleting entities.

Thanks!

bradfordcondon:
Deleting a tripal entity, intentionally, does not delete the chado record (thats my recollection of this discussion i've had with stephen).

@spficklin @laceysanderson I think this is not always the desired behavior, for example in the case @chunhuaicheng describes.  I would go so far as to say that deleting the chado record should be the **default behavior**.  That said, maybe we need to support both?  Deleting the record brings up a dialogue that lets you specify you want to **also delete the associated chado record** ?

We probably need a separate issue for this discussion: I'll create one.

laceysanderson (approved):
I confirmed that the bud described did exist and that this PR fixed it. Thanks @almasaeed2010!

 - [x] Checked that there was a record in chado_bio_data_1 for all my entities
```sql
tripaldb=# SELECT * FROM chado_bio_data_1;
 mapping_id | entity_id | record_id | nid 
------------+-----------+-----------+-----
          1 |         1 |         1 |    
          2 |         2 |         2 |    
          7 |         7 |       309 |    
(3 rows)
```

- [x] showed that after deleting one, I still have the `chado_bio_data_1` record but no record in `tripal_entity`

```sql
tripaldb=# SELECT * FROM chado_bio_data_1;
 mapping_id | entity_id | record_id | nid 
------------+-----------+-----------+-----
          1 |         1 |         1 |    
          2 |         2 |         2 |    
          7 |         7 |       309 |    
(3 rows)

tripaldb=# SELECT * FROM tripal_entity;
 id |     type     |   bundle   | term_id |      title      | uid | status |  created   |  changed   
----+--------------+------------+---------+-----------------+-----+--------+------------+------------
  1 | TripalEntity | bio_data_1 |       1 | Lens culinaris  |   1 |      1 | 1537117504 | 1537117517
  2 | TripalEntity | bio_data_1 |       1 | Lens ervoides   |   1 |      1 | 1537117530 | 1537117539
(2 rows)

tripaldb=# \q
```

- Checked out the branch

```bash
$ git checkout 652-tv3-entity_delete_fixBranch 652-tv3-entity_delete_fix set up to track remote branch 652-tv3-entity_delete_fix from origin.
Switched to a new branch '652-tv3-entity_delete_fix'

$ drush cc all
'all' cache was cleared in /var/www/dev/tripal#default                                         [success]
```

- [x] showed that deleting now, deletes both connections.

```sql
tripaldb=# SELECT * FROM tripal_entity;
 id |     type     |   bundle   | term_id |      title      | uid | status |  created   |  changed   
----+--------------+------------+---------+-----------------+-----+--------+------------+------------
  1 | TripalEntity | bio_data_1 |       1 | Lens culinaris  |   1 |      1 | 1537117504 | 1537117517
(1 row)

tripaldb=# SELECT * FROM chado_bio_data_1;
 mapping_id | entity_id | record_id | nid 
------------+-----------+-----------+-----
          1 |         1 |         1 |    
          7 |         7 |       309 |    
(2 rows)
```

laceysanderson:
This PR is less than two lines and easily tested. As such, I've marked it needing **only 1 review**. Since I have provided that review, I'm merging this PR. Thank you @almasaeed2010!!!

PR #655, opened 2018-09-15
T3 chado compliance checking

bradfordcondon:
This PR is to another BRANCH not to `7.x-3.x` and is therefore only subject to @laceysanderson 's approval.

Suggestions as pointed out in my review of your PR.  Please feel free to reject and use what you like.


laceysanderson (approved):
These look awesome! Thanks @bradfordcondon!!!

Issue #657, opened 2018-09-17
chado update property fails


bradfordcondon:
## BUG/ERROR report

`chado_update_property` fails for me.

This line of code appears to be the culprit:

https://github.com/tripal/tripal/blob/b83136aec50b012ef54f90553ccc9d2443a4fa29/tripal_chado/api/tripal_chado.property.api.inc#L337-L338

if chado_get_property gets a single property, it returns an object, not an array.  Therefore the statement 
 `!is_array($prop)` will fail, and the function returns FALSE, as if it couldnt find the property to update.

chado_get_property returns:

```
 An array in the same format as that generated by the function
 *   chado_generate_var().  If only one record is returned it
 *   is a single object.  If more than one record is returned then it is an 
 *   array of objects
```

ie, if chado_get_property returns only a single result, it will return an object and the update function fails here.

Should we instead write this?

```
  if (empty($prop) or count($prop) == 0) {
```

e - that suggestion is redundant.  it can just be if `(empty($prop))`

bradfordcondon:
I'm finding other bugs in the property API.  I'm confirming each with a test and will submit a PR that addresses all methods in the file.

https://github.com/tripal/tripal/blob/b83136aec50b012ef54f90553ccc9d2443a4fa29/tripal_chado/api/tripal_chado.property.api.inc#L519-L523

needs to be `$base_id` not `$record_id`


similarly deleting by propkey fails at     `$property_pkey = $table_desc['primary key'][0];` because:

```
chado_delete_record; There is no table description for prop
```

The API says you dont need to supply base table if you give the prop_id, but i don't think thats true.  I'm going to resolve this one by changing the documentation.

bradfordcondon:
thank you to @laceysanderson @almasaeed2010 for the reviews.

PR #659, opened 2018-09-18
657 fix broken chado property API calls and add unit tests for each

bradfordcondon:
Issue #657 

## Type(s) of Change(s)

## Description
There were several bugs with the chado properties API.  Although issue #657 was created for a specific one, I made a simple unit test for each base function: this PR includes those tests and also the necessary fixes to get them to pass.

The following issues were addressed:

* `chado_get_property` and `chado_update_property` say that if `prop_id` is supplied, then table and id keys are not needed. However, it still needs the table key.
* `chado_get_property` fails if a prop_id and table are supplied.  This is because the base key was added regardless of if it was passed in or not.
* `chado_update_property` failed.  This was because it expected an array returned from `chado_get_property`.  However, that function returns an object if theres only one match (which one would generally expect for this call.)
* `chado_delete_property` used an unitialized variable, causing it to always fail.

## Testing?

The easiest way I can think of to test this is to pull this branch, then run `phpunit --group api`.  All tests will pass.  Next, revert the `tripal_chado/api/tripal_chado.property.api.inc` file to the copy on `7.x-3.x`, and rerun the tests.  You should now have 3 failing tests: `chado_get_property`, `chado_update_property`, and `chado_delete_property`.  So:

```
phpunit --group api
git checkout 7.x-3.x tripal_chado/api/tripal_chado.property.api.inc
phpunit --group api
```

Theres no way I'm aware of to test these changes via the GUI.

### Test Cases

Because test are the main way to confirm this functionality, I describe my test cases below:

#### test_chado_insert_property

Create a feature and cvterm: use chado_insert_property to add them.  Confirm via SQL that the property was added.

#### test_chado_get_property

Create a feature and cvterm: use chado_insert_property to add them.  Then use `chado_get_property` to retrieve them.  Test cases where the `$record` provided is table/id, and table/prop_id.

#### test_chado_update_property

Insert a feature property, then provide a new value and run `chado_update_property`.  Confirm that the property value was updated but that other values stayed the same.

#### test_chado_delete_property

Insert a featureprop, and use chado_delete_property to remove.

### test_chado_get_record_with_property 
insert a single featureprop, and use chado_get_record_with property to retrieve it, given the property type_id and value.
Then, insert a second feature and featureprop with the same type_id and value, confirm that both featureprops are returned.

## Additional Notes (if any):

- [x] I have updated the documentation accordingly.

### DIFF
--- a/tests/tripal_chado/api/TripalChadoPropertyAPITest.php
+++ b/tests/tripal_chado/api/TripalChadoPropertyAPITest.php
@@ -4,25 +4,31 @@ namespace Tests;
 
 use StatonLab\TripalTestSuite\DBTransaction;
 use StatonLab\TripalTestSuite\TripalTestCase;
+use StatonLab\TripalTestSuite\Database\Factory;
 
 class TripalChadoPropertyAPITest extends TripalTestCase {
 
   use DBTransaction;
 
   /**
+   * Tests chado_insert_property() with all prop tables.
+   *
+   * @dataProvider propTableProvider
+   *
    * @group chado
    * @group api
-   *
+   * @group chado-property
    */
-  public function test_chado_insert_property() {
+  public function test_chado_insert_property($prop_table, $base_table) {
 
-    $feature = factory('chado.feature')->create();
+    $base_record = factory('chado.'.$base_table)->create();
+    $base_pkey = $base_table.'_id';
     $term = factory('chado.cvterm')->create();
 
     $value = 'chado_API_test_value';
 
     // Linker column
-    $record = ['table' => 'feature', 'id' => $feature->feature_id];
+    $record = ['table' => $base_table, 'id' => $base_record->{$base_pkey}];
     $property = [
       'type_id' => $term->cvterm_id,
       'value' => $value,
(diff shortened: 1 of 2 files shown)


bradfordcondon:
At @almasaeed2010 's advice I added a "test case" section that describes what each test does.  If people find that helpful please say so and maybe we cna discuss adding to template/encouraging for features that cannot be tested via the GUI...

almasaeed2010:
The changes look good to me and the tests run as described.

Thanks!

almasaeed2010 (approved):

laceysanderson (approved):
I tested this by running the tests and did a code review. I think these changes are awesome and give my approval.

**I did submit a PR to extend these tests to all property tables and would love to see it merged :-)** However, I **give my approval either way** since this does exactly what it claims and is a great addition. Thank you @bradfordcondon!

bradfordcondon:
Thank you @laceysanderson for the review and also for the suggestions.

At first I was hesitant to test all chado prop tables. We want to be careful that testing all tables is something that adds value to the test suite.  In this case you raise the point that not all prop tables necessarily have cvalue_ids, for example, and I think thats a good point and worth ensuring it works on all tables for.  

For future tests we've agreed that erring on the side of *overtesting is ok*, unless its a particularly laborious and time consuming test and will therefore slow down the whole test process significantly without adding any value.

PR #641, opened 2018-09-11
ChadoSchema class and chado compliance checking

laceysanderson:
## Type(s) of Change(s)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] API-specific change (fix or addition to an API function)
- [ ] Updates documentation (inline or markdown files)

## Description

Sometimes various Tripal site chado instance can become out-of-sync with the expected chado schema. This can happen on purpose when trying to fit data into chado (not recommended) or accidentally (failed upgrade, etc.). Since it can cause tests and Tripal functionality in general to fail, it was deemed helpful to have tests checking chado schema compliance. That way, as developers we can diagnose if the problem is due to chado compliance or not and ,as administrators, we can check for accidental loss of compliance and backwards compatibility of any changes we've made on purpose.

Since we may want a drush and/or UI interface to chado compliance testing, it was decided that it would be best to create a ChadoSchema class to do the checking. Furthermore, Drupal8 is moving towards classes making this an ideal choice over functions.

### Automated Tests
This PR provides two types of automated tests: `chado-compliance` for checking if the current database is compliant with the Tripal Chado Schema API and `chado-schema` for checking the new ChadoSchema class. You can run all automated tests provided by this PR by using the following commands:
```bash
composer up
./vendor/bin/phpunit --group chado-schema
./vendor/bin/phpunit --group chado-compliance
```

### Manual Testing
You can also manually test this PR by interacting with the ChadoSchema class directly. The following code snippets show some examples:
```php
// Retrieve the schema array for the feature table.
$chado_schema = new \ChadoSchema();
$table_schema = $chado_schema->getTableSchema('feature');

// Retrieve the schema array for the organism table in chado 1.2
$chado_schema = new \ChadoSchema('1.2');
$table_schema = $chado_schema->getTableSchema('organism');

// Retrieve all chado tables.
$chado_schema = new \ChadoSchema();
$tables = $chado_schema->getTableNames();
$base_tables = $chado_schema->getbaseTables();

// Check the feature.type_id foreign key constraint
$chado_schema = new \ChadoSchema();
$exists = $chado_schema ->checkFKConstraintExists('feature','type_id');

// Check Sequence exists
$chado_schema = new \ChadoSchema();
$exists = $chado_schema->checkSequenceExists('organism','organism_id');
// Or just check the primary key directly
[...]

PR #661, opened 2018-09-19
For Bradford: Update Property Tests PR

laceysanderson:
## Not a Typical PR as it's for @bradfordcondon branch. As such it is only subject to his review.

PR #659 

The main thing this PR does is extend your unit tests to all property tables in Chado. I did this by using the chado_get_base_tables() API function (so this is NOT dependent on #641). For each base table, I check if there is an associated prop table and that there is a TripalTestSuite Factory available.

Feel free to modify as you see fit :-)


bradfordcondon:
thank you @laceysanderson !!

PR #664, opened 2018-09-20
Remove empty fields from view when loaded via ajax REDUX (revert revert)

bradfordcondon:
This is a clone of PR #628 

I accidentally merged, so i reverted, and this PR is reverting that reversion.  It sounds complicated, but actually it fixed the commit history problem in hte previous PR.

Here is the issue submission provided by @almasaeed2010 

          

Issue #541 

## Type(s) of Change(s)

## Description

**NOTE:** this PR is not ready for merging.

This addresses issue #541 where fields are not removed when they are loaded via ajax and have no value.

## Testing?

**1.** Before switching to this branch, edit a content type (e,g. Organism) to not hide fields when empty:
  - Go to structure > tripal content types > edit (organism)
  - Find the dropdown titled `Field Display` and choose `Show empty fields`
  - Now the fields that are not marked as `auto attach` will be loaded via ajax

**2.** Test that the organism page works and all fields are loaded. Check which panes/fields have empty values and display nothing
**3.** Now switch to this branch and clear the cache on the same page.
**4.** Verify that panes that contain empty fields are gone

## Additional Notes (if any):

There is still an outstanding edge case that we have to deal with before fully considering this solution and that's when a pane has multiple ajax fields. We'll need to make sure all fields are loaded before removing the pane.

Thanks!


bradfordcondon:
https://github.com/tripal/tripal/compare/7.x-3.x...cf3dcf15943797f2894b868944abe0e23e02f47f

this shows that this whole revert/revert thing has caused no issues on 7.x-3.x (thanks Lacey)

laceysanderson (changes requested):
My Test Case: a clean tripal3 site with two organism pages.

## No Tripal DS

Without Tripal DS I still see both options on Structure>Organism>Edit but none of the fields hide on the entity page. **This functionality should work without Tripal DS.**
### Show, No AJAX Loading
nods_defaultsettings
### Hide, No AJAX Loading
nods_hidenojs

## Tripal DS Enabled

I'm still noticing a lot of inconsistencies and weird behavior:
- I never see the empty publication, annotation panes, even with it set to `show empty fields`.
- In Hide *without* AJAX, 
     - I still see empty fields but only those set in field configuration to auto-attach (e.g. Common Name). 
     - The same field hides perfectly when in it's own field or when it's the only field in a pane.
- Hide *with* AJAX looks the same as "Show with/without AJAX". In order words, it doesn't seem to hide anything for me?

### Show, AJAX
ds-show-js-allpanes
### Show, No AJAX
ds-show-nojs-allpanes
### Hide, AJAX
ds-hide-js-allpanes
### Hide, No AJAX
ds-hide-nojs-allpanes

almasaeed2010:
Thanks for the review @laceysanderson!

>Without Tripal DS I still see both options on Structure>Organism>Edit but none of the fields hide on the entity page. This functionality should work without Tripal DS.

I don't think we can support this functionality without Tripal DS since it provides the structure needed to detect what should be done. For example, we can't predict where the `pane title` is and therefore can't remove it. If we want to support this without tripal ds, we'd need to set a few rules to how you can theme your page which I don't think is appropriate. Instead, I suggest clearly stating that you need to override this functionality if you build your own DS module. 

Do think there is a better way? 

> I never see the empty publication, annotation panes, even with it set to show empty fields.

I'll look into that. I am not sure if it's a PHP or JS issue.

> - In Hide without AJAX,
>   - I still see empty fields but only those set in field configuration to auto-attach (e.g. Common Name).
>   - The same field hides perfectly when in it's own field or when it's the only field in a pane.

This is not happening on my end which makes me think it might be a cache issue? In order for anything to take effect, I think the entity cache should be reset.

The description field in the screenshot below shows/hides (without ajax) although it's in a table and not on its own: (I also attempted with common name). Could there be a special case here that we are missing?

screen shot 2018-09-21 at 10 07 06 am

screen shot 2018-09-21 at 10 07 24 am

> - Hide with AJAX looks the same as "Show with/without AJAX". In order words, it doesn't seem to hide anything for me?

Hmmm, it looks like it's hiding everything no matter what the setting is once AJAX loading is enabled. It could be reading the variables wrong.

I'll look into those today and see what I can come up with. Thanks!

laceysanderson:
> Without Tripal DS I still see both options on Structure>Organism>Edit but none of the fields hide on the entity page. This functionality should work without Tripal DS.

I think there should be a way to do this in the future but there just isn't time now. In the meantime, if it's only supported with Tripal DS turned on then the options should only be enabled if Tripal DS is too. Ideally there would be a simple check for Tripal DS, if not available, disable the options and show an in-place notice that the options are dependant on Tripal DS.

> In Hide without AJAX: I still see empty fields but only those set in field configuration to auto-attach (e.g. Common Name).
>> This is not happening on my end which makes me think it might be a cache issue? In order for anything to take effect, I think the entity cache should be reset.

Not a caching issue as I cleared all caches multiple times to make sure :-). It works for me in a table as well (I had a typo above 😝 ) The problem arises when it's outside of a table and not the only field in the pane.

almasaeed2010:
@laceysanderson I made a simple change in the bundle form to make sure tripal ds is enabled before offering the ajax options.

In order for us to fix the rest of the issues though we need to fix this one first: #674 

Thanks!

almasaeed2010:
@laceysanderson the PR is ready for testing again. @shawnamspoor and I pushed a few changes that address the following:

- If the tripal ds module is not enabled, the ajax options are removed from the form
- A single empty field in a pane does not cause the pane to disappear if the other fields have data
- When load with ajax is selected but not hide fields, all fields stay visible even when empty.

Thanks!

spficklin:
@almasaeed2010 do the changes fix the issue I was having as well?

almasaeed2010:
I believe it does @spficklin. When checking both options, it works fine for me.

Thanks!

laceysanderson (changes requested):
I expected the following to be true:
- [x] Disable options when no Tripal DS. I added an extra commit to disable the elements rather then hide them so users know the functionality exists.
- [x] show, with or without ajax, should result in the same markup
- [ ] hide, with or without ajax, should result in the same markup.

## hide with or without ajax should result in the same markup.
Checking or unchecking `Ajax Loading` should not change the end markup but only change how it's loaded. This is the case when the admin chooses to "show fields" but not when they choose to "Hide fields". Tripal DS alone seems to be able to hide more fields then AJAX Field can. Specifically, in the following screenshots, Tripal DS manages to hide contact, cross reference and synonyms correctly whereas, AJAXField does not.

## No Ajax 👍 
ds-hide-noajax

## AJAX 😞 
ds-hide-ajax

almasaeed2010:
I would call that a field issue. They are not designed consistently in that some would print “no results” message and some wouldn’t. The Ajax/JS can only determine if data exists by looking at the content whereas the PHP implementation only looks at the data from the load method of the field class. Maybe we can change the implementation such that the JS receives more information? I’ll look into that later today 👍

spficklin:
Just a clarification.  The field should be considered "empty" if there is no 'value' for it.  All fields should consistently provide an empty 'value' if there is not a value.  The "no results" would come from a formatted which acts on the empty field and decides to deal with it in it's own custom way... which should be okay.

So the AJAX should not be looking at what comes out of the formatter but rather if the field is empty.

spficklin:
Given that the AJAX code seems to be looking at the formatter output rather than the field value, perhaps we should postpone this fix to Tripal v3.1 release to give us time to work out this problem?

laceysanderson:
I'm hoping this is a "simple" matter of Tripal DS adding a JS setting to indicate if the value (from the field->load) is empty. This would allow the AJAX field to use that rather then the formatter output and would fix the problem.

**Since I think this is what @almasaeed2010 was implying he would look into later today, lets reserve comment about it's release blocker status until we hear how big of change it's going to be.**

laceysanderson:
What about if we add a Drupal js setting indicating the field name an/or css class of the field to be removed? I think we would add it here: https://github.com/tripal/tripal/pull/664/files#diff-af393657f01c51e9be5040b9e62d2cf0R1533. Then AjaxField could check for that markup and remove it?

laceysanderson:
Ok, my recent commit adds an array of the fields which are empty according to their load at `Drupal.settings.tripal_ds.tripal_field_settings_empty_fields`:
screen shot 2018-09-26 at 10 07 47 pm

However, I didn't get the chance to go through the JS code to see where I could use this setting to hide the field.

bradfordcondon:
>empty according to their load 

@laceysanderson doesn't this defeat the purpose of loading the fields via AJAX?  We are using AJAX to display them as they load rather than after they load?

laceysanderson:
>> empty according to their load

> @laceysanderson doesn't this defeat the purpose of loading the fields via AJAX? We are using AJAX to display them as they load rather than after they load?

So, `field_get_items()`doesn't load the items meaning it won't defeat the purpose of loading the fields via AJAX. That said, your question made me inspect my work more closely and I realized the list of fields was not those that were empty due to load but rather those that were unattached -giving us no additional information... *facepalm* **As such I've reverted my commit.**

*Side note: Apologizes everyone and Thanks @bradfordcondon for pointing out my mistake! I fell into old (bad) habits last night and should have made a PR rather then committing directly to this branch...*

bradfordcondon:
>So, field_get_items()doesn't load the items meaning it won't defeat the purpose of loading the fields via AJAX. 

It may not load the items but doesnt the field's `load` function have to run first so that field_get_items can return the values array?  It looks to me like this function bypasses the `formatter` but not the `load` of the field, and the real speed issue occurs at `load`.  Does that sound right to you?

edit: assuming they arent cached.

I dont see how field_get_items can return the  value without first running `load` which is what we're trying to handle via AJAX, looks like a chicken and the egg thing to me.

bradfordcondon:
oh shoot, i think im the one getting lost here: the AJAX can hide the field **after** it loads and knows if its empty or not.  Before that it will hide it until it loads.

laceysanderson:
Well [field_get_items()](https://api.drupal.org/api/drupal/modules%21field%21field.module/function/field_get_items/7.x) just pulls the correct array elements from the already loaded entity. To my understanding, Tripal already ensured that the unattached fields were not loaded when the entity was (supported by the list I received). I think that's why my list includes empty fields and all unattached fields. If `field_get_items()` inadvertently triggered the load then I wouldn't have all unattached fields in my list: see `sbo__relationship` in my list which is not empty but is unattached.

I'm now thinking a better place to add more information would be in the JSON AjaxField receives? Perhaps returning in a object with both the formatted output and a flag?

laceysanderson:
The recent commit fixes the inconsistencies I was noticing on initial load after cache clear 🎉 👍 Now we just need to get it working with the cache -So Close!

almasaeed2010:
Ok so the latest changes fixes the inconsistencies for me. I tested this with custom fields and on new tripal installs. Everything worked as expected.

I'd like to point out though that using this branch, a new tripal install does **not** set the `hide empty fields` to true like it used to. Should we keep it like that or do we want a specific default (maybe enable ajax by default?). Let me know.

Please give this a test when you get a chance. Thanks!

laceysanderson (approved):
Recent changes fix my concerns and the inconsistencies I've noticed 👍 Thanks @almasaeed2010!!! 🎉

spficklin:
For my benefit, can someone summarize what the latest fix is?   When does the JS variable `.tripal_field_settings_empty_fields` get set?  Does it get set on initial page load or is it returned as part of the AJAX calls when loading unattached fields?

laceysanderson:
Oh sorry, @spficklin. That was my original misguided "fix" which didn't end up working the way we wanted. That commit has been revert and the variable to mentioned is no more.

@almasaeed2010's much better fix is to check when the field is loaded via AJAX and if it's empty, set the content returned to empty as well (see here: https://github.com/tripal/tripal/pull/664/files#diff-a97903891992d9d660963bf192748a68R570). This makes hide with and without ajax consistent.

spficklin:
Oh perfect.  I'll give this a review now.

spficklin:
I think we're getting closer, but there are still issues.  Here are my tests for loading an organism page with tripal_ds turned on and the organism  has no relationships:

| Status | Hide Empty | Load w/ AJAX | Time to Display | Result |
|--------|------|-----------------|--------|--------|
| Success &#x2714; | | | 2.18 seconds| Works, all fields are present No AJAX loading performed.  Relationship field shows formatted value. |
| Success &#x2714; | &#x2717; | | 2.19 seconds |  Works, empty fields are missing, including the relationship field. No AJAX loading performed. |
| Fails &#x274C; | | &#x2717; | 1.551 seconds | All empty fields are present, which is good, **but the relationship field appears blank**. It should have result from the field formatted value indicating there are no relationships.  AJAX was performed as expected. |
| Fails &#x274C; | &#x2717; | &#x2717; | 1.51 seconds |  All empty fields are removed **except the relationship field**, which appears blank. It should not be present at all.  AJAX was performed as expected.

laceysanderson:
Hi @spficklin, I've added a recent commit that fixes:
> All empty fields are present, which is good, but the relationship field appears blank. It should have result from the field formatted value indicating there are no relationships.  AJAX was performed as expected.

Now the relationship pane is shown with the "no relationships" msg when show is indicated with ajax loading. 

However, I think the second one is a Tripal DS bug. I can only get this to happen when the Relationship Tripal pane is configured NOT to "hide panel on page load". In this case, I see the following:

Status | Hide Empty | Load w/ AJAX | hide panel on page load | Relationship in TOC | Relationship Panel
-- | -- | -- | -- | -- | --
Success ✔ |   |   | ✗ | Works, In TOC | Works. Pane shows with "no relationships" msg as expected.
Fails ❌ | ✗ |  | ✗ | Works, not in TOC | Relationship Pane shows with "no relationships" msg **when there should be no pane.**
Success ✔ |   | ✗ | ✗ | Works, In TOC | Works. Pane shows with "no relationships" msg as expected.
Fails ❌ | ✗ |  | ✗ | Works, not in TOC | Relationship Pane **but is empty when it should not show at all**.
Success ✔ |   |   |   | Works, In TOC | Works, all fields are present No AJAX loading performed.  Relationship field shows formatted value.
Success ✔ | ✗ |   |   | Works, not in TOC | Works, empty fields are missing, including the relationship field. No AJAX loading performed.
Success ✔ |   | ✗ |   | Works, In TOC | Works, relationship field shows with "no relationships" msg
Success ✔ | ✗ | ✗ |   | Works, not in TOC | Works. All empty fields are removed including the relationship field.  AJAX was performed as expected.

laceysanderson:
To summarize a slack conversation: Both me and @spficklin were noticing a number inconsistencies:

Case | Hide Empty | Load via Ajax | Relationship Field Location | Other settings | Expected | Observed after reload | Observed when cached
-- | -- | -- | -- | -- | -- | -- | --
1 | ✗ |   | Relationship Pane | "hide panel on page   load" unchecked | Relationship pane hidden | Relationship pane visible and   empty message visible. | Relationship pane visible and   empty message visible.
2 | ✗ | ✗ | Relationship Pane | "hide panel on page   load" unchecked | Relationship pane hidden | Relationship pane visible but   empty | Relationship pane visible and   empty message visible.
3 | ✗ | ✗ | Summary Pane in table |   | Relationship row removed from   table | Relationship row present but empty | Relationship row present and field   empty message visible.
4 | ✗ | ✗ | Summary Pane below table |   | Relationship field removed | Works | Relationship field present with   empty message

* In all cases above, the Relationships link was not in the table of contents (TOC) as expected.

laceysanderson:
Commit 7b90c9b fixed case 4 above.

spficklin:
Okay, I think I have all of the kinks worked out of this PR.  The problems were actually very deep. I had to touch just about every piece of code that dealt with empty fields and ajax loading.  We just weren't catching every case.  Part of the problem was that we were mixing code that was in the tripal_ds module that should have been in tripal and some code in tripal that should have been in tripal_ds.  I cleaned most of that with except to some of the code in the `tripal.js` file that referenced panes.  But I didn't feel like trying to split that apart.   Good news is hiding of empty fields and AJAX loading now works regardless if you are using the tripal_ds module or not.   

I have tested all of the cases that @laceysanderson lists in the table above and I think it's all working.  So, @laceysanderson if you can test it again and @almasaeed2010 if you could as well then if you are happy and it works then I think we can merge and this should be our last milestone issue/PR before stable release!

spficklin:
Oh also, I added updates to the  User's Guide.  I added a new subsection titled `Working with Content Types` and moved all of the pages related to content types and their display to that subsection. I also added a new page about hiding empty fields and turning on/off ajax.

almasaeed2010:
Tested this. All the cases seem to be fixed for me. I think this is ready to be merged.

Thanks @spficklin and @laceysanderson!

spficklin:
Thanks @almasaeed2010 for retesting . @laceysanderson will you have time in the short term to review this again?

laceysanderson (approved):
I tested all combinations of hide & ajax loading for:
- Without Tripal DS enabled but tables with DS
- With Tripal DS enabled but no panes in use
- With Tripal DS and the default layout
- With Tripal DS "hide on page load" checked/unchecked

and they were all consistent and looked as I expected 🎉 🎉 🎉

laceysanderson:
I think with mine and @almasaeed2010's recent approval, this PR is ready to be merged!

PR #666, opened 2018-09-20
Add reload entity button to Remove empty fields when using ajax

almasaeed2010:
This PR adds the functionality that Stephen wanted to refresh the field.

Only @spficklin and @bradfordcondon need to review this since it affects the messed up branch.

### DIFF
--- a/tripal/includes/TripalEntityUIController.inc
+++ b/tripal/includes/TripalEntityUIController.inc
@@ -91,6 +91,15 @@ class TripalEntityUIController extends EntityDefaultUIController {
       'type' => MENU_LOCAL_TASK,
       'weight' => -8,
     );
+    $items['bio_data/' . $wildcard . '/reload'] = array(
+      'title'  => 'Reload',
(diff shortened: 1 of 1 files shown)


bradfordcondon:
i renamed PR since its confusing with all these reverts i made :'(

looks good to me 👍 you want us both to approve?

screen shot 2018-09-20 at 3 08 58 pm

laceysanderson:
I think the permission should be for admin role only! My content managers are going to so confused by this :-( See my previous comment to this effect here: https://github.com/tripal/tripal/pull/628#issuecomment-421803327

almasaeed2010:
Would a `administer tripal` permission be better @laceysanderson ?

almasaeed2010:
@laceysanderson the code now uses `administer tripal` :)

laceysanderson:
I think this is ready to merge. @almasaeed2010 do you still want to wait on @spficklin ?

spficklin:
I tested and it works.  Thanks for submitting this.

Issue #478, opened 2018-06-26
Collection files are not generated if an entry is missing a field


bradfordcondon:
### System information

* Tripal Version: 3 
* Site: HWG live

### Issue description

Creating a feature collection fails to generate fails if it cant retrieve a value for a field. Instead this error is returned:
```
ParseError: syntax error, unexpected &#039;;&#039; in module_load_include() (line 57 of /var/www/html/sites/all/modules/tripal/tripal/includes/TripalFieldDownloaders/TripalCSVDownloader.inc).
```

https://github.com/tripal/tripal/blob/3302486540c12042fb26e29acb167176e8c1149c/tripal/includes/TripalFieldDownloaders/TripalTabDownloader.inc#L43-L60

I'm able to fix the error by simply checking if the property exists and adding a blank element/continuing if it doesnt.

Note that the field *ISNT BLANK* for these failed records, so im not sure why the problem is occurring.  What I can say is that there are 3 organisms in my collection: one has 100% sequences, one has 0% sequences, and one has 70% sequences.

After fixing the below error by continuing if the property doesnt exist, the FASTA, csv, and tsv files are generated.  While the CSV and FASTA are able to "carry on" and produce the file, with blank sequences for the failed entries, the TSV file is not and only consists of the failed line.  

#### Steps to reproduce

I'm not positive what the source of the problem is.  I create a collection of features and attempt select the name and protein sequence fields.

* [problematic entity](https://www.hardwoodgenomics.org/bio_data/1633342)
* [succesful entity](https://www.hardwoodgenomics.org/bio_data/1909091)

p serotina might be more helpful since it contains records that both fail/succeed:

* [failure](https://www.hardwoodgenomics.org/bio_data/1914787)

* [success](https://www.hardwoodgenomics.org/bio_data/1367387)

Features are the same type, and both have protein sequences (the field that is causing issues for me).

here is a look at the CSV generated:

<img width="513" alt="screen shot 2018-06-26 at 10 19 31 am" src="https://user-images.githubusercontent.com/7063154/41919001-7d437c2a-792b-11e8-9a4e-ba9f7936a4ef.png">

#### error

the below error occurs for both the CSV and tab downloader

```
WD php: ParseError: syntax error, unexpected ';' in module_load_include() (line 57 of                                           [error]
/var/www/html/sites/all/modules/tripal/tripal/includes/TripalFieldDownloaders/TripalCSVDownloader.inc).
[...]

PR #675, opened 2018-09-24
Check that the field name exists before continuing

almasaeed2010:
Issue #478 

## Type(s) of Change(s)

## Description

As described in issue #478, Tripal's CSV and TSV downloaders fail in some cases. This PR mitigates that issue by checking that `$field_name` is not null/undefined. It also reports the error instead of failing fatally.

## Testing?

I am not sure which field is triggering this error yet. It's currently happening on our live server and this simple fix works for us.

## Additional Notes (if any):


laceysanderson (approved):
CSV and TSV downloaders work without error and the file created is as expected.

laceysanderson:
The changes in this PR are intuitive and only a few lines: as such I'm marking this needing only one review, which I have provided.

Issue #649, opened 2018-09-13
Need a function to 'unpublish' content


chunhuaicheng:
## Feature Request/Discussion

- [N] does this feature attempt to solve an existing problem with Tripal?
- [x] are you open to developing or collaborating on an extension module if this is not a good fit for Tripal Core (no pressure here -just good to know upfront :-) )
- [N] do you **Need** this feature ASAP?

### Description

We need a function to 'unpublish' content, similar to the 'cleanup orphaned feature' function in Tripal 2

### Your Specific Use Case

From time to time, our curators may request to delete data in bulk from underlying chado database. We need a 'unpublish content' function to remove the entities associated with those chado records.

bradfordcondon:
We (well, @almasaeed2010 ) have added this functionality to our tripal entity helper tool [tripal_alchemist](https://github.com/statonlab/tripal_alchemist) for now.  If you delete the chado records, then alchemist has a function to "delete orphaned entities" which will remove them.

chunhuaicheng:
This a great. Thanks @bradfordcondon 
Any chance this will be incorporated into Tripal core since this was part of the core functions in T2?

bradfordcondon:
I will wait to see what other community members/PMC members think.  Seems like a good case for including it in the core.

bradfordcondon:
@spficklin @laceysanderson what are your thoughts on porting this feature into core.  This constitutes a new feature and therefore a PMC vote.

laceysanderson:
I think we should definitely have unpublish functionality in core and with it already written in Tripal Alchemist I think that is an amazing place to start!

So consider this **a vote in favor for me** 👍

spficklin:
I agree that it makes sense to have this functionality in core as we had it in Tv2.

almasaeed2010:
I'll start a new PR to port the functionality over from alchemist.

almasaeed2010:
PR submitted https://github.com/tripal/tripal/pull/678

laceysanderson:
Merged so I believe this can be closed 🎉

PR #678, opened 2018-09-24
Add unpublish orphaned entities functionality

almasaeed2010:
Issue #649 

## Type(s) of Change(s)

## Description

This is an implementation of the unpublish orphaned entities functionality.

## Testing?

1. Pull the branch and clear menu cache
1. Visit Content -> Tripal Content -> Delete Orphaned Entities (should be next to the publish link)
1. Select a bundle to inspect for orphaned entities.
1. If the bundle has orphaned entities, It'll provide you with the number of entities that should be removed and an example consisting of 10 entities that will get deleted.

If you need to manually create an orphaned entity you can:
1. create an organism entity using the GUI
1. delete the organism from chado using SQL.

screen shot 2018-09-24 at 1 03 36 pm

## Additional Notes (if any):

Note that the queries used are optimized to handle millions of records. This was tested on our live site and managed to analyze 5 million+ entities in under 2 seconds. However, the deleting job will be slow since we decided that it's best to trigger all `entity_delete` hooks which required that we process each entity individually.

Thanks!


laceysanderson (approved):
Functionality works great! Very fast, correctly detected orphaned entities. I appreciated the preview table to make it easier to confirm that I really wanted to delete the specified entities.

Code review also looked good 👍

spficklin:
I see quite a few problems with this PR.   

1.  The biggest one is that the unpublished code for orphaned entities should not be in the `tripal_chado` module:  
    - The `TripalEntityController` is the class meant to handle changes to the database related to entities. 
    - Its really easy to always think about data as being stored in Chado.  For now it is.  But Tripal v3 is meant to be database agnostic so the `tripal` module should really initiate this.
2.  The code warned the user after they submitted the unpublish form to ignore any Chado errors.  
    -  I don't think we should do that.  The code should not give errors.  User's won't know when a real problem occurs vs errors that get printed as a result of cleaning.  
    -  I believe errors were occurring because the entities were being removed by calling the `TripalEnttiyController->delete()` function which would then try to remove attached fields.  I believe the Drupal code was trying to load all of those fields before deleting and without a record in the database it caused problems.  
3.  There were some minor adjustments that I thought were needed:
    - There was no title on the unpublished form page. 
    - There were HTML tags in translation `t()` function calls (e.g. `<p>`).  That is not allowed 
    - Some of the terminology used may not be familiar to a user who was not a Tripal developer (e.g. bundle, entity, etc.).
    -  I submitted a commit that I think addresses these minor issues.
4.  Unpublishing entities is extremely slow!  Again I think this is because Drupal tries to load all of the fields prior to deleting the entity.  

To address these issues, I propose we add a `findOrphans` and a `deleteOrphans` function to the `TripalEntityController` class.  We should also add a `hook_field_storage_find_orphans`  and a `hook_field_storage_delete_orphan` function to allow the storage backend for the content type to find and then clean up it's association.   With these additions it doesn't matter if the data is stored in Chado or elsewhere.  

When implementing the `deleteOrphans` function, it should be different from the `delete` function. Instead, it should first call the `hook_field_storage_delete_orphan` hook.  For Chado data, that should call the  `tripal_chado_field_storage_delete_orphan` function (found in the `tripal_chado/includes/tripal_chado.field_storage.inc` file)  which will remove the record from the appropriate `chado_bio_data_x` table.  Then the `deleteOrphans` function should directly remove the record from the `tripal_entity` table and finally remove any cached fields from the `cache_field` table.   

That's it. It will bypass all of the Drupal code that would normally work but does not because the storage back-end no longer has the data.  it should dramatically increase performance too.

spficklin (changes requested):
See my comments in the conversation thread for proposed adjustments

spficklin:
Also, one other idea. Rather than have this as a form, what if instead there was a cron that would go through all of the bundles and create a listing (either in a table or in a chart) that would be added to the Dashboard.  That way, a user doesn't have to go looking for orphaned entities, If they check the dashboard they would see them and how many.  A simple button on the dashboard could allow a user to remove *all* orphaned entities.  Right now with so many content types, it's a bit unwieldy to have to go through each one in the form dropdown just to see if there are any oprhans.

laceysanderson:
My thoughts:
- I really liked the listing of a subset of the orphans so we need to find a way to keep that. It makes it much easier as an admin to diagnose why you might have orphaned entities. 
- I also like the idea of the dashboard showing when you have orphaned entities. Perhaps we have the dashboard table and a link in each row to the current form. Deleting orphans should not be taken lightly so I don't think we want to make it too easy (i.e. I don't think we should allow deleting in bulk of all orphans for all types with no preview).
- I like the idea of a `findOrphans` and `deleteOrphans` function. Not only will this have the advantages @spficklin mentioned, it will also make unit testing easier.
- Since this has morphed into quite a re-write, don't feel like you need to spearhead it @almasaeed2010 (unless you want to of course :-) )

almasaeed2010:
Thanks for the review all! 

I think all of that sounds very reasonable. I hope this at least gives us a starting point. 

Maybe we can split tasks and tackle this today? Doesn't seem like a lot.

Please let me know if I can start on a specific part.

Thanks!

spficklin:
@almasaeed2010 I won't have time today.  My priority is to get this PR merged #664 and see if we can get the functional analysis tools working so we can do a Tripal v3 release this week.  I think we could work this into the 3.1 release.

almasaeed2010:
That sounds good to me. I'd like to point out for anyone who is looking for a temporary solution, you can use [Tripal Alchemist](https://github.com/statonlab/tripal_alchemist) which is a module that helps in managing entities.

Thanks!

spficklin:
Okay, I have reworked the code to reflect the suggestions I have made.  This is now ready for testing again.

In short I moved the concept of an orphaned entity into the TripalBundleController class. It now has the `findOrphans()` and `deleteOrphans()` functions.  Those functions have hooks that they call and each module that creates entities can implement those hooks to then find and do the cleanup.  This approach allows all storage backends to handle their own orphaned entities.   A side effect is that I had to modify the table that listed orphaned entities to exclude Chado tables and just list titles.

spficklin:
Also since I changed the code I shouldn't be the one to test. Perhaps @laceysanderson you can do a quick re-test and @almasaeed2010 you can make sure it works for you?

spficklin (approved):

spficklin:
One more thing. I just added a new page to the User's Guide with instructions for removing orphaned entities.

almasaeed2010:
Approved. This works perfectly for me.

On en empty, new tripal site:

1. create an organism
1. run `delete from chado.organism`
1. clear orphaned entities from `/admin/content/bio_data/unpublish`

The organism page should not exist anymore.

I also pushed a simple commit to remove an undefined variable.

Thanks @spficklin!

spficklin:
@bradfordcondon would you mind doing a code review since @laceysanderson is out?

spficklin:
Thanks @almasaeed2010 for the test!

docs/user_guide/content_types/cleaning_orphans.rst
```diff
@@ -0,0 +1,16 @@
+Handling Orphaned Entities
+==========================
+It is common for site developers to work directly with Chado, especially as they become more familiar with it. And sometimes, there may purpsefully or accidentally remove data that is published in Tripal. This results in an entity being "orphaned".  The entity will still have a page on the website but the data about that entity is no longer available.  If this occurs you can easily remove published entities that are "orphaned" by navigating to **Administer > Tripal Content Types** and clicking the **Unpublish Orphaned Content** link.  The following page appears.
```
bradfordcondon: typos
Also:  
> accidentally remove data  that is published in Tripal

i would (and do) say:  accidentally remove data from Chado utilized by published Tripal entities.


```suggestion
It is common for site developers to work directly with Chado, especially as they become more familiar with it. And sometimes, they may purposefully or accidentally remove data from Chado that is utilized by published Tripal entities. This results in an entity being "orphaned".  The entity will still have a page on the website but the data about that entity is no longer available.  If this occurs you can easily remove published entities that are "orphaned" by navigating to **Administer > Tripal Content Types** and clicking the **Unpublish Orphaned Content** link.  The following page appears.
```

docs/user_guide/content_types/cleaning_orphans.rst
```diff
@@ -0,0 +1,16 @@
...
+
+.. image:: cleaning_orphans.1.png
+
+Next, select the content type that has missing data in the **Content Type** select box.  For example, suppose an organism was removed:
+
+.. image:: cleaning_orphans.2.png
+
+A list of at most 10 entities that are orphaned are shown. If the content type has no orphaned entities then nothing is shown.  You can submit a job to clean the orphaned entities by clicking the
```
bradfordcondon: please clarify: are all orphans deleted or just the ones in the list.

tripal/api/tripal.entities.api.inc
```diff
@@ -1649,3 +1649,94 @@ function tripal_update_entity($bundle_name, $values) {
...
+ * A hook for modules to delete details for orphaned entities.
+ * 
+ * This hook is called by the TripalBundleController.  Modules that create
+ * entities should use this hook to clean up entities that are orphaned. The
+ * list of $ids passed should be entities who are already known to
+ * be orphaned. These IDs are found by the TripalBundleController using the
+ * results from the hook_bundle_find_orphans() function. 
+ * 
+ * An implementation of this hook should not try to clean up the entity itself,
+ * but rather it should only clean up it's own records used to manage the
```
bradfordcondon:
```suggestion
 * but rather it should only clean up its own records used to manage the 
```

bradfordcondon:
minor suggestion to the documentation, otherwise 👍

docs/user_guide/content_types/cleaning_orphans.rst
spficklin: Good point. I've updated the code.  Thanks for reading it carefully.

spficklin:
Thanks all for the reviews!

PR #573, opened 2018-08-23
Add tripal web services http tests

almasaeed2010:
Issue #567

## Type(s) of Change(s)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] API-specific change (fix or addition to an API function)
- [ ] Updates documentation (inline or markdown files)
- [x] Unit Tests

## Description

Adds unit tests to verify that http calls to web services behave as expected.

## Testing?

1. Be sure to run `composer update` to get the latest version of TripalTestSuite
2. Run `./vendor/bin/phpunit tests/tripal_ws` (there are no side effects to running tests since they are wrapped in a transaction)
3. Verify all assertions are passing (green)

## Additional Notes (if any):

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

I encountered an issue when testing pulling a newly published feature. When sending a GET request to `/web-services/content/v0.1/mRNA/$entity_id`, where `$entity_id` is a newly created mRNA entity, I get an error indicating that the entity is not available. However, using `tripal_load_entity()` finds and loads the entity perfectly fine. Any idea why this test doesn't pass? (see method `testGettingAFeatureResource()` in this PR).

Thanks!

PR #679, opened 2018-09-24
remove extra checks from tripal ws file loader

almasaeed2010:
## Type(s) of Change(s)

This is an simple enhancement to the code.

## Description

The `tripal_load_include_web_service_class` function performs a check to verify that the file exists before loading it. However, it uses `module_load_include` to load the file which in turn does its own similar checks. Also, the checks in this function does not always succeed. In my docker environment, this function failed to detect that the file exists. After removing the extra checks, it succeeded.

## Testing?

1. visit `/web-services/content/v0.1` and verify that the page loads successfully.
1. Navigate through the pages to perform additional checks.

## Additional Notes (if any):

Note that this change will allow Tripal Test Suite to access web services using phpunit and thus would also fix issue #573  👍

### DIFF
--- a/tripal_ws/api/tripal_ws.api.inc
+++ b/tripal_ws/api/tripal_ws.api.inc
@@ -60,7 +60,7 @@ function hook_tripal_ws_value(&$items, $field, $instance) {
  *
  * @return
  *   A list of TripalWebService names.
- * 
+ *
  * @ingroup tripal_ws_api
  */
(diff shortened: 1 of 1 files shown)


spficklin:
I think this only requires 1 review because the code change is just a few lines and it's a simple, intuitive fix.  I have tested it and it works just fine.

Issue #672, opened 2018-09-22
Documentation Guidelines


laceysanderson:
We need a README added to our docs directory to give contributing developers a quick guide and some guidelines for contributing documentation to Tripal.

**Remember: when this gets done, we need to add the link to the PR template.**

laceysanderson:
# Tripal Documentation

Our documentation is build using **Read the Docs**. All of the Tripal documentation is in this 
directory (`docs/`) and is split into two guides:

1. User Guide: focuses on installation and configuration of Tripal from a content administrator point 
of view.
2. Developers Guide: focuses on developers extending the functionality of Tripal through 
development of extension modules.

## Building and Contributing to Documentation

You can build the docs by installing Sphinx and running:

```bash
# in the docs directory
make html
```

You can open `_build/index.html` in your browser and navigate to the page with your changes to 
check they compiled as expected.

## Guidelines

Please follow these guidelines when updating our docs. Let us know if you have any questions or 
something isn’t clear.

We use the following syntax:

```
Title of File (using title case)
=====================

Introduction text.

Other Heading
---------------

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut 
labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco 
laboris nisi ut aliquip ex ea commodo consequat.

We use backticks to indicate `inline-code` including file names, function and method names, 
paths, etc.

Longer code-blocks should begin with the `.. code-block:: [type]` directive and should 
be indented at least one level. There should also be a blank line before and after it as 
shown below.

.. code-block:: sql
  if ($needs_documentation) {
      use $these_guidelines;
      $contribute_docs = $appreciated;
  }

Sub-section
^^^^^^^^^^^^

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut 
labore et dolore magna aliqua.
```

laceysanderson:
Above is my draft of documentation but I would love discussion, thoughts, corrections before posting a PR.

bradfordcondon:
code styling is double ticks ``
Our use of headings is kind of bad in general.. the "right" way is

```
================
 Document Title
================
----------
 Subtitle
----------

Section Title
=============
Section 1.1 Title
-----------------

Subsection 1.1.1 Title
~~~~~~~~~~~~~~~~~~~~~~

Section 1.2 Title
-----------------
```

[...]

PR #686, opened 2018-09-26
Added our current syntax to the guide.

laceysanderson:
## This PR Extends @bradfordcondon's existing PR thus only requires his review

I added our current syntax to the guide... As per #672, we may want to change our current guidelines to match ReadtheDocs conventions. **I'm including them here anyway to give people guidance in the meantime and ensure our documentation remains consistent.**

Note: I know the code example could be found elsewhere but none of the documentation I found made it clear that you had to indent the code block. This is something I struggled with when first contributing documentation. The importance of example code blocks makes me feel the couple of lines to make this clear is worth it.


bradfordcondon:
works for me, thanks @laceysanderson

Issue #625, opened 2018-09-01
Remote Field "Field to Display" expects label rather then token.


laceysanderson:
### System information

* Tripal Version:
* Drupal Version:
* PostgreSQL Version:
* PHP Version:

### Issue description

I noticed when configuring my remote field, the "Field to Display" is expected to be the label rather then the token from the "Available Tokens" list. Specifically, I set up my remote organism.species field as in #491. When I set "Field to Display" = "[taxrank__species]" I received the following error in my remote field:

![organism_ws_error](https://user-images.githubusercontent.com/1566301/44941861-9bad6600-ad63-11e8-8e56-fbc305089b42.png)

However, when I set the "Field to Display" to "species", the remote field displayed "databasica" as I expected. **This was the case regardless of the label of the content type.** 

For full field configuration, reference the following screenshot:
![organism_ws_config](https://user-images.githubusercontent.com/1566301/44941877-ca2b4100-ad63-11e8-9b49-e427775c172c.png)

#### Steps to reproduce

1. Add a new Tripal remote site that really points to the local site. Do this by going to Tripal > Data Storage > Remote Tripal Sites > Configuration. For my local test machine I used the URL http://localhost and version 'v0.1'.
2. Go to the Organism content type: Structure > Tripal Content Types and click manage fields on the Organism content type and create a new field of type 'Remote Tripal Site'. 
   - Set the Remote Tripal Site to the site you created in step 1; 
   - set the query to "Organism?genus=[taxrank__genus]&species=[taxrank__species]"; 
   - set the Field to display to '[taxrank__species]' or any of the other tokens under "Available Tokens". 
3. Make sure the new field shows up on the page by clicking on the 'Manage Display' tab at the top and moving it out of the Disabled section.
4. Go to any organism page that already exists on the site or create a new one. You should see the error shown in the screenshot above.
5. Edit the field and change the "Field to Display" to "species" and then refresh the organism page. Now the remote field works.

spficklin:
Hi @laceysanderson.  So, I think the problem you are having is that the 'Field to Display' does not take a token.  It takes the name of the term or the accession.  So, if you want to show the species you would enter `species` or `TAXRANK:0000006`

Can you try making that change and see if that fixes the problem for you?  If so then perhaps the fix is to improve the description to indicate what is wanted.

laceysanderson:
See 
> However, when I set the "Field to Display" to "species", the remote field displayed "databasica" as I expected. This was the case regardless of the label of the content type.

[...]

PR #691, opened 2018-09-29
Fixed instructions for the field

spficklin:
Issue #625 and discovered bug.

## Type(s) of Change(s)

## Description

This addresses issue #625 where the instructions for setting field values for the remote__data field were not clear enough.  The problem was that the `Available Tokens` fieldset appeared below the field named `Field to Display` and it implied that tokens could be used.  This is not the case.  To fix this I

1) Moved the `Available Tokens` fieldset below the `Query to Execute` field.
2) The instructions needed more details and were becoming too long so I added fieldsets to contain the instructions for both fields.  

Also, there was a 1-line bug fix in web services that I included in this PR. You can see it in the code review. It was so minor I decided to include it here rather than create it's on PR.  It fixes the problem when browsing web services.  For fields that are not auto attached the web services gives a link.  If the user follows that link it takes them to another web service result that then provides the values.  This result was not complete.  This fixes it.

## Testing?

There is no need to test the fixes to the remote__data field unless you want to see the changes yourself. Otherwise a code review and looking at the screenshots should suffice.  

To test the web services fix, pull the branch, navigate through web services to any entity that has relationships and follow the relationships link. You should see all of the relationships.

Screenshot of new remote__data form page with described fixes:

image

Screenshot of web services for a gene with non-attached fields shown as links:

image

Screenshot of web services when looking at relationships, before the fix.

image

And after the fix:

image

## Additional Notes (if any):

### DIFF
--- a/tripal_ws/includes/TripalFields/remote__data/remote__data.inc
+++ b/tripal_ws/includes/TripalFields/remote__data/remote__data.inc
@@ -320,7 +320,7 @@ class remote__data extends WebServicesField {
       ->execute()->fetchAll();
 
     foreach ($sites as $site) {
-      $rows[$site->id] =$site->name;
+      $rows[$site->id] = $site->name;
     }
 
     $element['data_info']['remote_site'] = array(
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
I changed the wording a little bit on "Query to Execute" instructions:
> If the full URL to the remote tripal content web service is "https://[tripal_site]/web-services/content/v0.1/". Then this field should contain the text immediately after the "content/v0.1" portion of the URL. For information about building web services queries see the online documentation at The Tripal v3 User's Guide. For example, suppose this field is attached to an Organism content type on the local site, and you want to retrieve a field for the same organism on a remote Tripal site. To retrieve the matching record, you will want to query on the genus and species, since it is unique and, you want them to match the organism for each specific local organism page. You can use tokens to do this (see the "Available Tokens" fieldset below). For this example, the full remote web service endpoint would be "https://[tripal_site]/web-services/content/v0.1/Organism" and the query text should be "Organism?genus=[taxrank__genus]&species=[taxrank__species]".
screen shot 2018-09-28 at 8 55 58 pm

And switched the example on the "Field to Display" to common name so that it was visible in the JSON without scrolling.
screen shot 2018-09-28 at 8 57 34 pm

With those changes, I approve this PR as fixing my original concerns :-) Thanks @spficklin!

laceysanderson:
I think this change is small enough (not many lines of code) that it only requires one review. As I've tested and approved it, it can be merged once @spficklin approves my small changes :-)

spficklin:
Thanks @laceysanderson

PR #695, opened 2018-10-01
Cache terms to improve the performance of page load

chunhuaicheng:
#

Issue # n/a

## Description

Cache terms to improve the performance of page load. Since Tripal 3 is now ontology based, it sends a lot of queries to retrieve the definition of terms. This PR caches the terms to reduce the number of SQL sent to the server and thus improves the performance. 

The cached terms will be cleared when an OBO lmporter job finishes, or a cvterm gets updated using the GUI interface.
## Testing?

Time the page load before and after the fix. You should notice the difference. This PR improves the performance the second time a page of same type (not just the exact same page) is loaded. It also improves the single page load when a page tries to load the same term multiple times.

### DIFF
--- a/tripal_chado/includes/TripalImporter/OBOImporter.inc
+++ b/tripal_chado/includes/TripalImporter/OBOImporter.inc
@@ -575,8 +575,11 @@ class OBOImporter extends TripalImporter {
    *
    */
   public function postRun() {
+    
+    // Clear the cached terms
+    cache_clear_all('tripal_chado:term:*', 'cache', TRUE);
 
-    // Update the cv_root_mview materiailzed view.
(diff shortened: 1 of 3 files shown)


spficklin:
I have tested this on my local site, on the Tripal v3 demo site and the GeneNetEngine site and it makes a tremendous speed improvement.  I approve.

almasaeed2010:
I second @spficklin's approval. This PR increases the speed by a pretty good amount 👍 

Thanks @chunhuaicheng!

laceysanderson:
Awesome! Thanks for the quick fix @chunhuaicheng 🎉 I can confirm it works for me 👍

PR #696, opened 2018-10-02
Added fix for TOC auto hide

spficklin:
# Bug Fix

Issue # NA

## Description

This is a very simple 1 line of code change to the JavaScript that fixes a bug when removing the row from the sidebar TOC when a panel is empty.  Previously it just removed the `<a>` tag but left the surrounding `<span>` and that left uneven spacing between links.  This fixes that problem.

## Testing?

It's hard to test depending on your CSS and the spacing between your links.  Just pull the branch and load a page to makes sure nothing breaks. You have to have 'Hide Empty fields' checked for the content type.

This is just a simple code fix and only needs one review. It's so simple you can probably just review the code itself. I'd like to get this into the Tv3 release if possible.

Before:
image

After:
image

### DIFF
--- a/tripal/theme/js/tripal.js
+++ b/tripal/theme/js/tripal.js
@@ -73,7 +73,7 @@
     pane.remove();
     
     // Remove the pane's title from the TOC.
-    $('#' + this.id).hide(0);
+    $('#' + this.id).parents('.views-row').remove();
   }
   
   /**


almasaeed2010:
Works for me. 👍

laceysanderson (approved):
This looks good to me 👍 And I've confirmed it doesn't break my site ;-p

almasaeed2010 (approved):

spficklin:
Thanks for the quick review!

PR #698, opened 2018-10-02
interproscan guide

bradfordcondon:
This PR is not to the main branch and is therefore only subject to @spficklin 's approval (although anyone's feedback is welcome).

Write out the interproscan document.  I didnt add screenshots because I haven't loaded this dataset yet.

I had some questions/ideas...

* what to do about shared document snippets (ie data upload)?  Can we factor out and reuse?  Is that overkill?
* I noticed you dont have a field configuration section on blast.  Do you think it would be a good idea to add one for each of these stubs?  IE visit mRNA and check for new fields....


bradfordcondon:
resolved

edit grrrrr sorry wrong github window, i need to be more careful about this...

spficklin:
Hey @bradfordcondon I did have adding fields and adding panels, etc for blast.... I'll check to make sure that didn't get lost.

bradfordcondon:
I looked at your latest on github and there is definitely the adding fields guide, so let me re-pull and add that section to this guide...

bradfordcondon:
I see the problem: I didnt see `setup.rst` which has all the fields in one place (smart).

bradfordcondon:
I finally revisited this!  IPR is a draft.  I'm working on KEGG next: I need to replace the heirarchy files with the new expected KEGG files.


Honestly my IPR file doesnt look great after loading:  (using the Citrus files)

image

* what are all those "seg" annotations?
* All the annotations are atthe start of the mRNA only.  I'm pretty sure this is because the coordinates from the protein/mRNA are mixed up.  ie the feature start/end are for the protein, but they get mapped on to the mRNA and therefore always appear at the first 1/3 of the sequence.

Issue #702, opened 2018-10-03
Undefined property ogi__location_on_map


abretaud:
### System information

* Tripal Version: 3.0
* Drupal Version: 7.59
* PostgreSQL Version: 9.5
* PHP Version: 7.1

### Issue description

I saw this notice while displaying an mRNA (with data from the Citrus demo from tripal 2.x):
`Notice: Undefined property: TripalEntity::$ogi__location_on_map in TripalEntityController->attachLoad() (line 681 of /var/www/html/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).`
I have no idea if it has any impact, reporting just in case

#### Error messages and screenshots

Launch docker-tripal, and loaded data with python-tripal with something similar to this: https://github.com/abretaud/docker-tripal-demo/blob/master/data/load.sh
Published mRNA records, and went to an mRNA page

```
Notice: Undefined property: TripalEntity::$ogi__location_on_map in TripalEntityController->attachLoad() (line 681 of /var/www/html/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
```

laceysanderson:
I have seen this as well. I'll submit a PR to fix -Thanks for reporting!

laceysanderson:
PR Submitted :-) Can you test that it fixes the issue for you @abretaud?

spficklin:
PR to fix this was approved. I'm closing this.  But comment if there's still issues.

PR #705, opened 2018-10-03
Issue #702: Fix Location on Map field undefined error.

laceysanderson:
# Bug Fix

Issue #702 

## Description

@abretaud saw this notice while displaying an mRNA (with data from the Citrus demo from tripal 2.x) and I say it on gene pages created using Tripal DevSeed:

```
Notice: Undefined property: TripalEntity::$ogi__location_on_map in TripalEntityController->attachLoad() (line 681 of /var/www/html/sites/all/modules/tripal/tripal/includes/TripalEntityController.inc).
```

It was caused because the ogi__location_on_map field unset itself when it was empty rather then just returning a non-true value. This PR fixes this behavior.

### Warning is gone
1. Go to an existing Gene page (before pulling this branch) and click the "Reload" tab. You should see the above warning.
2. Pull the PR, go to the same page and click the "Reload" tab. The message should be gone :-)

### feature location field still works.
Ideally also test a page with a location on a map to ensure the field still works as it's supposed to. To test this, 
1. create a fake map feature (or use an existing feature),
2. insert a fake featuremap record 
3. add a `featurepos` pointing to the map feature, fake featuremap and your gene. 
4. Check that the on "Reload" the featuremap field now displays the record you just inserted.

The following screenshot shows the field still works with this PR ;-p
screen shot 2018-10-03 at 10 20 14 am

### DIFF
--- a/tripal_chado/includes/TripalFields/ogi__location_on_map/ogi__location_on_map.inc
+++ b/tripal_chado/includes/TripalFields/ogi__location_on_map/ogi__location_on_map.inc
@@ -322,10 +322,11 @@ class ogi__location_on_map extends ChadoField {
       }
     }
 
+
     // If there are no map positions expanded above then remove the stub.
     // This is needed to ensure this field isn't displayed when there are no locations.
     if (!isset($feature->featurepos->feature_id) OR (sizeof($feature->featurepos->feature_id) == 0)) {
-      unset($entity->{$field_name});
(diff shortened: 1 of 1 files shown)


laceysanderson:
I'm marking this as only needing one review since it's a one-line change :-)

spficklin:
I've made quite a few of these fixes so I can tell just by looking at the code that this should do it.

abretaud:
Thanks, it works!

Issue #680, opened 2018-09-25
Terms not found found on EBI OLS cause fatal errors when the problem is the OBO


bradfordcondon:
### Issue description
Loading the plant trait ontology (TO) fails because of a "homeless" term. 

The term is looked up in  PO, which is where the accession is.  But the term is only defined in TO, with a PO accession!
lookup the OBO loader uses:
http://www.ebi.ac.uk/ols/api/ontologies/po/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FPO_0001108

correct lookup:
http://www.ebi.ac.uk/ols/api/ontologies/to/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FPO_0001108

#### Steps to reproduce

Load the plant trait ontology, at http://purl.obolibrary.org/obo/to.obo

#### Error messages and screenshots

```
Cannot find the term via an EBI OLS lookup: PO:0001108. EBI Reported: Resource not found.Consider finding the OBO file for this ontology and manually loading it first.
```

bradfordcondon:
Issue submitted to the planteome here:

https://github.com/Planteome/plant-trait-ontology/issues/439

bradfordcondon:
Here's the summary: The TO defines this term as being in the PO, but the PO doesnt define the term.  Then, teh EBI OLS doesn't mark this term as belonging to any ontology, so we are left unsure what to do with the term.

So plenty of things to think about here.  We can figure this out in the future by searching for terms that 404 in the current namespace as i did above (ie swap PO with TO).  if we find the term, its still unclear waht to do though.

As for if the loader should fail, or warn... we want to address the problem but ultimately it isn't Tripal's responsibility to resolve these issues, and what happens if a community is unresponsive: we're effectively blocked from using the ontology at all.  If only a warning is logged, then its possible no one will ever read hte logs, and we wont discover the problem...

bradfordcondon:
https://github.com/Planteome/plant-trait-ontology/commit/fa64c9983636b135b22e551022901175f44796f4

looks like they've resolved the issue and given the term a TO, instead of PO, accession.

`"is_defining_ontology": false,` still but i imagine this will change when the new ontology is pushed.

This means we just need to solidify how to deal with this in hte future and implement it.

bradfordcondon:
ok, updated.  New error:

```

Performing EBI OLS Lookup for: RO:0002310,568 bytes.
[...]

Issue #120, opened 2017-06-06
Error loading plant trait ontology (nameless terms)


bradfordcondon:
I was trying to load the plant trait ontology [OBO file](http://www.obofoundry.org/ontology/to.html) and  encountered the below error:
```
ERROR:  null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (101429, 61, null, , 195954, 0, 0).:
INSERT INTO chado.cvterm (cv_id, name, definition, dbxref_id,
is_obsolete, is_relationshiptype) VALUES (:cv_id, :name, :definition,
:dbxref_id, :is_obsolete, :is_relationshiptype); Array
(
    [:cv_id] => 61
    [:name] =>
    [:definition] =>
    [:dbxref_id] => 195954
    [:is_obsolete] => 0
    [:is_relationshiptype] => 0
```

My belief is that the nameless terms at the top of this OBO file throw the error:

>[Term]
id: CHEBI:15841 ! polypeptide
is_a: CHEBI:16670 ! peptide
is_a: CHEBI:33839 ! macromolecule

I'm able to load other ontologies that lack these termless definitions at the head of the OBO.

bradfordcondon:
Hi all,

Looks like this issue persists (along with my OBO memory issue #140 ).  See the error below.

I can try my hand at resolving these myself this week if it has not been fixed already.

Thanks!


```

-------------------
2017-10-11 20:22:44: There are 1 jobs queued.
2017-10-11 20:22:44: Calling: tripal_run_importer(22)

NOTE: Loading of this file is performed using a database transaction.
If the load fails or is terminated prematurely then the entire set of
insertions/updates is rolled back and will not be found in the database

Downloading URL http://purl.obolibrary.org/obo/to.obo, saving to /tmp/obo_OCNWQZ
Step 1: Preloading File /tmp/obo_OCNWQZ...
Step 2: Loading type defs...emory: 45,524,352 bytes.
Step 3: Loading terms...%. Memory: 45,524,272 bytes.
SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (3400, 80, null, , 4010, 0, 0).
WD tripal_job: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "name" violates not-null constraint          [error]
DETAIL:  Failing row contains (3400, 80, null, , 4010, 0, 0).
[...]

PR #707, opened 2018-10-04
dont exit if we get an error looking up the term, but log a warning

### LINKED ISSUE #665: Issues Loading Trait Ontology (TO)
### System information

* Tripal Version: most recent dev including #627
* Drupal Version: 7.x

### Issue description
Some versions of the Trait Ontology (TO) will not load even with the new OBO loader re-write. We are still having troubles with reproducibility so this issue is meant as a place for discussion as well as an eventual PR.

Personally, I have tested the following versions of TO and all fail with a similar error message (see below).

 - http://purl.obolibrary.org/obo/to.obo
 - https://raw.githubusercontent.com/Planteome/plant-trait-ontology/master/to.obo

Error Message:
```
Running 'OBO Vocabulary Loader' importer
NOTE: Loading of file is performed using a database transaction. 
If it fails or is terminated prematurely then all insertions and 
updates are rolled back and will not be found in the database

Downloading URL http://purl.obolibrary.org/obo/to.obo, saving to /tmp/obo_EAwOxp
Step 1: Preloading File /tmp/obo_EAwOxp...
A term that belongs to another ontology is used within this vocabulary.  Therefore a lookup will be performed with the EBI Ontology Lookup Service to retrieve the information for this term. Please note, that vocabularies with many non-local terms require remote lookups and these lookups can dramatically decrease loading time. 
Performing EBI OLS Lookup for: BFO:0000002
Found BFO:0000002 in EBI OLS.
array_key_exists() expects parameter 2 to be array, null given OBOImporter.inc:1046                                [warning]
array_key_exists() expects parameter 2 to be array, null given OBOImporter.inc:1056                                [warning]
ChadoRecord::insert(). Could not insert a record into the table, cv, with the following values: Array
(
    [name] => 
)
. ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (62, null, null).
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] ChadoRecord::insert(). Could not insert a record into the table, cv, with the following values: Array(    [name] => ). ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "name" violates not-null constraintDETAIL:  Failing row contains (62, null, null).

real    0m4.167s
user    0m0.352s
sys     0m0.100s
```

### Other pertinent messages from original OBO Loader thread (#627):
 - [@almasaeed2010 reported the same error as me on different term using file1 above](https://github.com/tripal/tripal/pull/627#issuecomment-423229866)
- [@spficklin loading stops but doesn't fatally error using file2 above ](https://github.com/tripal/tripal/pull/627#issuecomment-423256753)
- [@almasaeed2010 reported the same error as @spficklin using file2](https://github.com/tripal/tripal/pull/627#issuecomment-423273677)

bradfordcondon:
# Bux Fix

Issue #680 

This PR allows loading the plant trait ontology (TO) so we should be able to close #665 and #120  .  

It fixes this error:

```
Cannot find the term via an EBI OLS lookup: PO:0001108. EBI Reported: Resource not found.Consider finding the OBO file for this ontology and manually loading it first.

```

## Description

When the EBI OLS lookup fails, i throws a fatal error, blocking loading the ontology.  @spficklin and I discussed whether this is correct or not, and we agreed that the loader should WARN and not ERROR.  This is primarily because, as we found out in the case of #680 , the OBO file might not actually be correct.  Its good that it errors because it prompts us to contact the ontology so the error can be fixed, but it isnt our responsibility to ensure every ontology is perfect before it can be loaded into Chado. 

## Testing?
Load the plant trait ontology.  (http://purl.obolibrary.org/obo/to.obo)
On master branch you will not be able to because of two malformed EBI OLS calls and get this error: (ive dumped the full error for your convenience.  the error you'll actually get will just be the message "Cannot find the term via an EBI OLS lookup: RO:0002310. EBI Reported: Resource not found.Consider finding the OBO file for this ontology and manually loading it first.") 

```
Performing EBI OLS Lookup for: RO:0002310,568 bytes.
string(118) "http://www.ebi.ac.uk/ols/api/ontologies/ro/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_00000500002310"
array(6) {
  ["timestamp"]=>
  int(1537984394315)
  ["status"]=>
  int(404)
  ["error"]=>
  string(9) "Not Found"
  ["exception"]=>
  string(62) "org.springframework.data.rest.webmvc.ResourceNotFoundException"
  ["message"]=>
  string(18) "Resource not found"
  ["path"]=>
  string(98) "/ols/api/ontologies/ro/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_00000500002310"
}
Cannot find the term via an EBI OLS lookup: RO:0002310. EBI Reported: Resource not found.Consider finding the OBO file for this ontology and manually loading it first.
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] Cannot find the term via an EBI OLS lookup: RO:0002310. EBI Reported: Resource not found.Consider finding the OBO file for this ontology and manually loading it first.
```
(note: this error is due to us forming the API call wrong.  there were more which i have fixed in the original ontologies.  We also need to fix this malformed call, and i've made a seperate issue for it here #708.). 

If you switch to this branch, you should be able to load the ontology with no problems.

the happily loaded TO!  🎨 

screen shot 2018-10-04 at 12 28 18 pm

## Additional Notes (if any):

I know some users ie @laceysanderson reported a **different error** when trying to load this ontology, not related to the EBI OLS lookup.  This PR will not help you ,sorry :(

### DIFF
--- a/tests/tripal_chado/loaders/OBOImporterTest.php
+++ b/tests/tripal_chado/loaders/OBOImporterTest.php
@@ -19,38 +19,38 @@ class OBOImporterTest extends TripalTestCase {
    * @throws \Exception
    */
   private function loadOBO($name, $path) {
-    
+
     $obo_id = db_select('public.tripal_cv_obo', 't')
-    ->fields('t', ['obo_id'])
-    ->condition('t.name', $name)
-    ->execute()
-    ->fetchField();
-    
+      ->fields('t', ['obo_id'])
+      ->condition('t.name', $name)
+      ->execute()
+      ->fetchField();
+
     if (!$obo_id) {
-      
+
       $obo_id = db_insert('public.tripal_cv_obo')
-      ->fields(['name' => $name, 'path' => $path])
-      ->execute();
+        ->fields(['name' => $name, 'path' => $path])
+        ->execute();
     }
-    
+
     $run_args = ['obo_id' => $obo_id];
-    
+
     module_load_include('inc', 'tripal_chado', 'includes/TripalImporter/OBOImporter');
     $importer = new \OBOImporter();
     $importer->create($run_args);
     $importer->prepareFiles();
     $importer->run();
   }
-  
-  
+
+
   /**
-   * Tests that an OBO from a remote URL can be loaded.  
-   * 
+   * Tests that an OBO from a remote URL can be loaded.
+   *
    * For this test we will use the GO Plant Slim.
    *
    * @group obo
    */
-   public function testRemoteOBO() {
+  public function testRemoteOBO() {
 
     $name = 'core_test_goslim_plant';
     $path = 'http://www.geneontology.org/ontology/subsets/goslim_plant.obo';
@@ -64,7 +64,7 @@ class OBOImporterTest extends TripalTestCase {
       ->condition('name', 'biological_process')
       ->execute()
       ->fetchField();
-    $this->assertNotFalse($bp_cv_id, 
+    $this->assertNotFalse($bp_cv_id,
       "Missing the 'biological_process' cv record after loading the GO plant slim.");
 
     $cc_cv_id = db_select('chado.cv', 'c')
@@ -82,7 +82,7 @@ class OBOImporterTest extends TripalTestCase {
       ->fetchField();
     $this->assertNotFalse($mf_cv_id,
       "Missing the 'molecular_function' cv record after loading the GO plant slim.");
-    
+
     // Make sure we have a proper database record.
     $go_db_id = db_select('chado.db', 'd')
       ->fields('d', ['db_id'])
@@ -91,8 +91,8 @@ class OBOImporterTest extends TripalTestCase {
       ->fetchField();
     $this->assertNotFalse($go_db_id,
       "Missing the 'GO' database record after loading the GO plant slim.");
-  } 
-  
+  }
+
   /**
    * Tests that an OBO from a local path can be loaded.
    *
@@ -103,9 +103,9 @@ class OBOImporterTest extends TripalTestCase {
   public function testLocalOBO() {
     $name = 'tripal_obo_test';
     $path = __DIR__ . '/../example_files/test.obo';
-    
+
     $this->loadOBO($name, $path);
-    
+
     // Make sure we have a proper vocabulary record.
     $tot_cv_id = db_select('chado.cv', 'c')
       ->fields('c', ['cv_id'])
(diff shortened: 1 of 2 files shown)


almasaeed2010 (approved):
The changes look sensible to me. Approved. Thanks @bradfordcondon

almasaeed2010:
After the new updates to this branch, the loader finishes successfully for me 👍 Thanks @bradfordcondon

laceysanderson (approved):
I finally found the time to setup my cloud9 devsite to test this! On that devsite this PR works perfectly and allows the trait ontology to complete loading 🎉 

As you can see from the following, I did run into the edge case fixed by this PR.
```
Downloading URL http://purl.obolibrary.org/obo/to.obo, saving to /tmp/obo_ONhd1U
Step 1: Preloading File /tmp/obo_ONhd1U...
A term that belongs to another ontology is used within this vocabulary.  Therefore a lookup will be performed with the EBI Ontology Lookup Service to retrieve the information for this term. Please note, that vocabularies with many non-local terms require remote lookups and these lookups can dramatically decrease loading time. 
The term, PATO:0001237, is replaced by, PATO:0001241
Cannot find the term via an EBI OLS lookup: PO:0028006. We tried to access: http://www.ebi.ac.uk/ols/api/ontologies/po/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FPO_0028006EBI Reported: Resource not found. Consider finding the OBO file for this ontology and manually loading it first.
Cannot find the term via an EBI OLS lookup: RO:0002310. We tried to access: http://www.ebi.ac.uk/ols/api/ontologies/ro/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_00000500002310EBI Reported: Resource not found. Consider finding the OBO file for this ontology and manually loading it first.
Cannot find the term via an EBI OLS lookup: RO:0002577. We tried to access: http://www.ebi.ac.uk/ols/api/ontologies/ro/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_00000500002577EBI Reported: Resource not found. Consider finding the OBO file for this ontology and manually loading it first.
Found the following namespaces: plant_trait_ontology, bfo, chebi, envo, foodon, go, pato, peco, po.
Step 2: Examining relationships...
Step 3: Loading type defs...: 81,285,560  bytes.
```

Thanks to the warning versus exiting, I now have a loaded TO according to
screen shot 2018-11-16 at 1 32 16 pm

bradfordcondon:
what do we think, do we want to merge like this?  Or figure out why those malformed urls are being created first?

bradfordcondon:
no response.  The issue is tracked here (#708) separately and i think this is a clear improvement so im going to merge.

Issue #709, opened 2018-10-05
undefined legacy function tripal_stock_add_organism_count_mview


bradfordcondon:
`tripal_stock_add_organism_count_mview` is called in legacy tripal_stock but is not defined.

reported here:  https://github.com/erasche/docker-tripal/issues/27 by @RichardBruskiewich

bradfordcondon:
Here's the mview definition in 7.x-2.x

https://github.com/tripal/tripal/blob/3b3ddebf8fb420a0cc77cc6572a9dcbba4377a7c/tripal_stock/tripal_stock.install#L110-L178


do we need this mview?  edit: yes we do, its at least used by 
`tripal_stock_load_organism_stock_counts()`

RichardBruskiewich:
The function is in v2.x but not v3.x
http://api.tripal.info/api/tripal/search/current?keyword=tripal_stock_add_organism_count_mview

Who has the power to fix this quickly?  @laceysanderson are you in a position to help with this?

bradfordcondon:
I made a PR to add it back in, just need some eyes to confirm its a good idea.  I think it is though since we've still got functions in the tripal_stock legacy module calling that mview.  @laceysanderson you might be best qualified to double-check me here.

laceysanderson:
It should be added back in 👍 I'll test the PR next week -thanks!

bradfordcondon:
thanks @laceysanderson 

I realize this is a hole in our CI.  we test upgrading from tripal 2 to tripal 3, but we dont do **any testing** on the legacy modules.

I might add a final check at the end and enable them all.

RichardBruskiewich:
@bradfordcondon, I might cherry pick your branch being submitted to the pull request so that I can take one more run at the hill today. I'm on Pacific time so I still have 3 hours in my regular working day!

laceysanderson:
PR Merged 👍

PR #710, opened 2018-10-05
Add undefined function back to  tripal_stock.install

bradfordcondon:
# Bug Fix

Issue #709 

## Description
When enabling the legacy module tripal_stock, a function is called `tripal_stock_add_organism_count_mview` but not defined, erroring out the install.

  This PR adds it back from the 7.x-2.x branch.  I verified that this mview is called by the legacy module, so it should likely remain.

## Testing?

on a tripal-3 site, drush pm-enable tripal_stock should fail on currnt master, and succeed with this PR.

## Additional Notes (if any):
heres the warning reported here: https://github.com/erasche/docker-tripal/issues/27
```
WD php: Error: Call to undefined function [error]
tripal_stock_add_organism_count_mview() in tripal_stock_install()
(line 56 of
/var/www/html/sites/all/modules/tripal/legacy/tripal_stock/tripal_stock.install).
Cannot modify header information - headers already sent by (output [warning]
started at /root/.composer/vendor/drush/drush/includes/output.inc:40)
bootstrap.inc:1486
Error: Call to undefined function tripal_stock_add_organism_count_mview() in tripal_stock_install() (line 56 of /var/www/html/sites/all/modules/tripal/legacy/tripal_stock/tripal_stock.install).
Drush command terminated abnormally due to an unrecoverable error. [error]
A non well formed numeric value encountered drush.inc:1643 [notice]
```

### DIFF
--- a/legacy/tripal_stock/tripal_stock.install
+++ b/legacy/tripal_stock/tripal_stock.install
@@ -101,3 +101,74 @@ function tripal_stock_schema() {
 
   return $schema;
 }
+
+/**
+ * Creates a materialized view that stores the type & number of stocks per organism
+ *
+ * @ingroup tripal_stock
+ */
+function tripal_stock_add_organism_count_mview() {
+  $view_name = 'organism_stock_count';
+  $comment = 'Stores the type and number of stocks per organism';
+
+  $schema = array(
+    'description' => $comment,
+    'table' => $view_name,
+    'fields' => array(
+      'organism_id' => array(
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Tested and confirmed via code review. Looks perfect!

laceysanderson:
Since this pertains to the legacy stock module and is a simple matter of adding back in a missing function, I think this only needs a single review.

Thanks @bradfordcondon!

PR #711, opened 2018-10-09
Fix relationship widget unset variable errors

almasaeed2010:
# Bux Fix

## Description

Simple fix to check if a field exists in the form_state array before attempting to access it.

## Testing?

Let me know if the code isn't clear enough. **Note** that the changes are much clearer if you select `Diff Settings -> hide white space` under `Files changed`.

Thanks!

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
@@ -388,7 +388,7 @@ class sbo__relationship extends ChadoField {
     if (!$record) {
       return;
     }
-    
+
     // Not all tables have the columns named 'subject_id' and 'object_id'.
     // some have variations on that name and we need to determine what they are.
     $fkeys = $schema['foreign keys'];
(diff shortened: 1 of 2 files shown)


tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
```diff
@@ -856,8 +856,8 @@ class sbo__relationship extends ChadoField {
       $type_id = $item['chado-' . $field_table . '__type_id'];
       $type_id = isset($item['type_id']) ? $item['chado-' . $field_table . '__type_id'] : $type_id;
       $type_name = isset($item['type_name']) ? $item['type_name'] : '';
-      $subject_name = $item['subject_name'];
-      $object_name = $item['object_name'];
+      $subject_name = isset($item['subject_name']) ? $items['subject_name'] : null;
+      $object_name = isset($item['object_name']) ? $item['object_name'] : null;
```
laceysanderson: I think it's safer to return `''` if the item is not set. This would also be consistent with the `$type_name` above.

laceysanderson (changes requested):
Other then my one comment, looks good

almasaeed2010:
Hello @laceysanderson thanks for the review! I submitted a new commit to address your suggestion.

Thanks!

laceysanderson (approved):

bradfordcondon:
seems very straight forward to me.  thank you @almasaeed2010

Issue #640, opened 2018-09-10
Undefined callback for menu item in tripal_chado


almasaeed2010:
### System information

* Tripal Version: 3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

There is a menu item in the `tripal_chado` module that defined an undefined `page callback`

https://github.com/tripal/tripal/blob/4b321e67e876dba323553a49210a7b2cee98976b/tripal_chado/tripal_chado.module#L452-L459

In the above code, the method `tripal_pub_importer_setup_page_update_remotedb` does not exist but no errors are thrown due to that when visiting `/admin/tripal/loaders/pub/changedb`. 

I am not sure what should we do here. Any thoughts?

Thanks!

spficklin:
Thanks for noticing.  This problem also exists in Tripal v2.  I think it's a left over bit of code that maybe was used in some previous iteration when developing that loader and I suspect it's no longer needed and can just come out.

laceysanderson:
Fixed by @almasaeed2010 in both Tripal3 and Tripal2! Thanks @almasaeed2010

PR #715, opened 2018-10-12
Remove undefined menu callback

almasaeed2010:
# Bux Fix

Issue #640 

## Description

As discussed in the issue, this PR simply removes some leftover code.


laceysanderson (approved):
I confirmed that the callback no longer exists in Tripal3 and reviewed the code.

laceysanderson:
This change qualifies for only needing one review due to it's size and the extremely low likelihood of negative effects. Since I approved it, I'm going to merge.

PR #735, opened 2018-11-06
Fixes node sync error issue due to missing $transaction variable.

guignonv:
In function chado_node_sync_records(), the $transaction variable has been commented line 914 but not line 973 which prevents error from being displayed correctly.

#

Issue #

### DIFF
--- a/tripal_core/api/tripal_core.chado_nodes.api.inc
+++ b/tripal_core/api/tripal_core.chado_nodes.api.inc
@@ -911,7 +911,7 @@ function chado_node_sync_records($base_table, $max_sync = FALSE,
   print "\n$count $base_table records found.\n";
 
   $i = 0;
-  //$transaction = db_transaction();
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Hi @guignonv 

I think we should bring back the transaction instead of commenting out the rollback. The message does say that the syncing will happen in a transaction so I think we need to honor that.

Thanks for the contribution!

laceysanderson:
@guignonv would you like to add another commit bringing the transaction back? I want to be sure you end up with the commit credit :-) Good catch btw!

guignonv:
Done!

laceysanderson (approved):
Contribution looks great :-) Thanks @guignonv!

laceysanderson:
Additionally, I believe this change is small enough to only warrant a single review. As such, I'm merging :-)

Issue #729, opened 2018-10-24
References to rc2 in final 3.0 release


abretaud:
### System information

* Tripal Version: 3.0
* Drupal Version: 7.59
* PostgreSQL Version: 9.5
* PHP Version: 7.1

### Issue description

In most of the tripal modules info files (like [here](https://github.com/tripal/tripal/blob/7.x-3.x/tripal/tripal.info#L6)), there are still references to the version 7.x-3.0-rc2 instead if 7.x-3.0
I don't know if it could have any impact, but it could be confusing I guess.
ping @mboudet

laceysanderson:
Thank you for reporting this @abretaud. The fix submitted by @bradfordcondon has been merged so this should be fixed in the most recent dev version.

PR #736, opened 2018-11-07
set all submodule versions to correct version

bradfordcondon:
# Bux Fix

Issue #729

## Description
Users were confused by rc2 mentions when latest release is 7.x-3.0

This PR requires only one review.

## Testing?
None required


laceysanderson (approved):
👍 Looks perfect and I agree it only needs a single review.

Issue #712, opened 2018-10-09
Error generated during tripal_export_views_integration + Temporary Fix


risharde:
Tripal Version: 7.x-3.0-rc3

Initial Page Error:
```
PDOException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "views-integration" LINE 1: SELECT * FROM tripal_views WHERE setup_id='views-integration... ^: SELECT * FROM {tripal_views} WHERE setup_id=:setup; Array ( [:setup] => views-integration ) in tripal_export_views_integration() (line 726 of /var/www/Drupal/sites/all/modules/tripal/tripal_chado_views/api/tripal_chado_views.api.inc).
The website encountered an unexpected error. Please try again later.
````

URL location: 
admin/tripal/storage/chado/views-integration/export/143

File location: 
sites/all/modules/tripal/tripal_chado_views/api/tripal_chado_views.api.inc

Temporary (nasty) fix:
function tripal_export_views_integration($setup_id)
`$c_path = current_path(); $c_path_parts = explode('/',$c_path); $c_path_parts_len = count($c_path_parts); $setup_id = $c_path_parts[$c_path_parts_len - 1];`
// Main setup details

**Comment by bradfordcondon** [comment]
ok, confirmed at http://tripal.test/admin/tripal/storage/chado/views-integration/export as well.  Thank you for reporting.

**Comment by bradfordcondon** [comment]
`tripal_export_views_integration` is being called with the string `views-integration` in the query 
```
  $r = db_query("SELECT * FROM {tripal_views} WHERE setup_id=:setup", array(':setup' => $setup_id));

```
I get this at 

http://tripal.test/admin/tripal/storage/chado/views-integration/export

and any int for http://tripal.test/admin/tripal/storage/chado/views-integration/export/2, 3, 4, whatever you pass in.

I dont have any background info whatsoever here so i cant make sense of your fix, but ill look into it...


edit: oh duh, the menu item is probably passing in the wrong variable.  lets fix that.

bradfordcondon:
so to fix the problem @risharde has reported I think the answer is to update the `page argument` for the variable from .4 to 6 here:
https://github.com/tripal/tripal/blob/a2f3c1e3414011adfced99e98b15365326904792/tripal_chado_views/tripal_chado_views.module#L117-L125

but what about when no variable is passed:

[...]

PR #737, opened 2018-11-07
fix variable passed in hook menu for views integration

bradfordcondon:
# Bug Fix

Issue #712

## Description
The tripal views integration module passes the wrong part of the url as the 
`access arguments` probably left over from changing the menu structure.

As a result, from `admin/tripal/storage/chado/views-integration` clicking on export or delete will give a PDO error and crash.  Edit was not effected.

thank you @risharde for identifying the issue.

## Testing?
Visit `admin/tripal/storage/chado/views-integration` and click export/delete on any view.  it will crash.

Pull this branch, and clear your cache, the links will now work.

## Additional Notes (if any):

I deleted teh menu callback for `admin/tripal/storage/chado/views-integration/export` since i dont see it used: `admin/tripal/storage/chado/views-integration/export/%` exports a specific view, and the menu is at `admin/tripal/storage/chado/views-integration`, so `admin/tripal/storage/chado/views-integration/export` doesnt make any sense, and also would pass null in the code which isnt handled, so im pretty sure its safe to delete.

### DIFF
--- a/tripal_chado_views/tripal_chado_views.module
+++ b/tripal_chado_views/tripal_chado_views.module
@@ -82,7 +82,7 @@ function tripal_chado_views_menu() {
   $items['admin/tripal/storage/chado/views-integration/delete/%'] = array(
     'title' => 'Delete Views Integration',
     'page callback' => 'tripal_chado_views_integration_delete',
-    'page arguments' => array(4),
+    'page arguments' => array(6),
     'access arguments' => array('manage tripal_views_integration'),
     'type' => MENU_CALLBACK,
   );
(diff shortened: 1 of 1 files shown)


tripal_chado_views/tripal_chado_views.module
```diff
@@ -105,21 +105,11 @@ function tripal_chado_views_menu() {
     'weight' => 2,
   );
 
-  $items['admin/tripal/storage/chado/views-integration/export'] = array(
```
laceysanderson, line 108: At one point you could access this page and choose the views integration to export but that solution was clunky and not ideal. I think this should be removed and was simply cruft.

laceysanderson (approved):
Perfect 👍

laceysanderson:
Thank you for the fix @bradfordcondon!

Issue #676, opened 2018-09-24
FASTA Importer requires a regexp if associating with a parent


bradfordcondon:
### System information

* Tripal Version: 3

## description

The Tripal FASTA importer requires a regular expression if you choose to associate the features with a parent.  However, this parameter should be optional.  It is optional for other importers (For example the InterPro importer).  
Why?

If I am associating interproscan annotations for a set of proteins, where the protein names === the mRNA names, i should just be able to specify the type (mRNA).  As it is, I also have to provide a regexp ie `>(.*)`

Heres where validation fails:

https://github.com/tripal/tripal/blob/9541d27c4843d2cbfedf01226ef048a4c23daacb/tripal_chado/includes/TripalImporter/FASTAImporter.inc#L280-L289

If we allow this, then we also need to update the importer here:

https://github.com/tripal/tripal/blob/9541d27c4843d2cbfedf01226ef048a4c23daacb/tripal_chado/includes/TripalImporter/FASTAImporter.inc#L538-L543

and set `$subject` to `$accession` (defined right above) instead of `''` by default.

bradfordcondon:
any input from @tripal/project-managent-commitee-pmc on if this change should be made or not?

laceysanderson:
I’ll vote yes for making the regex optional for the fasts loader 🙋‍♀️

bradfordcondon:
ok good enough for me.  simple fix, i can take care of it when i have the chance.

PR #738, opened 2018-11-08
676 allow no regexp for parent in FASTA importer

bradfordcondon:
# New Feature

Issue #676 

## Description
The FASTA importer adds a relationship to another feature if you specify the type and a regexp.  however a regexp is not necessary if the parent feature name is identical to the feature you are importing.

This simple PR allows a form to be submitted with a parent feature type and relationship type but no regexp.  And, if this is so, it will search for the parent with the same uniquename.  This feature is found in other importers as discussed in #676.

## Testing?
As a bonus i wrote a test case for this feature.  We insert two mRNAs.  then we load polypeptides with the same names and verify that the importer links them (that the mRNA has a feature_relationship record)


bradfordcondon:
ill look into the failed test today, it passes locally...


oops, i added a join and forgot you cant chain joins.  should be fixed.

tripal_chado/includes/TripalImporter/FASTAImporter.inc
```diff
@@ -536,7 +535,7 @@ class FASTAImporter extends TripalImporter {
         }
 
         // Get the relationship subject
-         $subject = "";
+         $subject = $name ? $name : $uname;
```
bradfordcondon: should this actually be $uname or null? or instead of $name or $uname?

laceysanderson: I think it should be 
```suggestion
         $subject = $uname ? $uname : "";
```
based on https://github.com/tripal/tripal/pull/738/files#diff-8006ac719a97e82d92833b60886802a2L831 where the query looking up the parent/subject for the relationship checks the uniquename column.

laceysanderson (commented):
Test passes locally for me 👍

laceysanderson (approved):

bradfordcondon:
the second review for this PR does not need to be functional.

spficklin (approved):

spficklin:
Non-functional code review looks good. I merged it.

Issue #703, opened 2018-10-03
Fields with disabled widgets


bradfordcondon:
## Feature Request/Discussion

* This feature [does ] attempt to solve an existing problem with Tripal
* I am] open to developing or collaborating on an extension module if this is not a good fit for Tripal Core
* This feature is [Not Urgent]

### Description

Site admins should be able to disable widgets for a field.  I think being able to disable them based on **permisions** would be extra awesome.

### Your Specific Use Case
Some fields we use sparingly, and only by an admin.    We woudl like undergrad curators to not get bogged down and distracted by these fields when creating content.  Note- without fieldsets, these forms can get scary fast!

Additionally, whati f you like a field to display, and you utilize programmatic ways of populatng it, but you dont want to ever manually set it, but the field provides a widget for it?  An example is the **Annotation** field.  We set that with blast/interpro importers, never manually.

laceysanderson:
I have also come across this need. For example, we have properties we no longer use on our germplasm pages. I do not want to delete the field since I want to display it if it's already set but I do not want people to set this field on edit since it's deprecated.

I also just personally prefer hiding fields rather then deleting them. That way I remember they are there and can easily only show one additional field if I change my mind. Currently I have to delete ant used fields to keep them out of the edit form and then when I next click "Find new fields" they all come back again. This is particularly tiresome on the publication content type which has a lot of property fields that are not necessarily used.

My preferred solution would be able to just select "Hidden" as the widget. This would be as easy as creating a widget with no form that can be used for any field.

bradfordcondon:
i've kind of conflated two issues here:

* Hiding fields

* Hiding based on permissions.

@laceysanderson you sounded your support for the first but not for permissions-based hiding: what do you think?  I see this as critical for tripal HQ.

edit: its possible that we will want to implement the permission-based hiding on tripal_hq's end which is fine.

almasaeed2010:
I think I'd be ok with `hidden` as a widget option as @laceysanderson suggested since I have never felt the need to add data to an auto-populated field. Is there a specific case @bradfordcondon where admins would need to edit some field and at the same time won't let the user (submitter) edit it?

bradfordcondon:
@almasaeed2010 :
>Is there a specific case @bradfordcondon where admins would need to edit some field and at the same time won't let the user (submitter) edit it?


I think all of the lookup fields fall under this category.  Publication, annotation, dbxref.  They are too complicated for an end user and/or require content creation, inserting terms, etc.

[...]

PR #742, opened 2018-11-09
703 field permission guide

bradfordcondon:
# Documentation

Issue #703

## Description

Provides documentation for setting individual field permissions using the field permissions module.

## Testing

Ideally you will build the documentation locally (`make html` in the docs folder, open the build in `_build`) and then follow the guide to configure a field permission.  If you havent built our docs before you'll need to [install sphinx first](http://www.sphinx-doc.org/en/master/usage/installation.html)

## images

screen shot 2018-11-09 at 4 10 51 pm


bradfordcondon:
question to ask yourself if you review: is this document in the right place :P


edit: now that im thinking about this, it belongs in the user guide instead, iun the "set up example genomic site" for sure.

docs/dev_guide/field_permissions.rst
```diff
@@ -0,0 +1,50 @@
...
+Setting Field-specific Permissions
+--------------------------------------------
+
+
+
+Let's assume I want to hide the Cross-Reference field from my users submitting Genome Assembly data, but still want it available for my administrators.
+
+.. image:: ./field_permissions.1.cross_ref_GA.png
+
+First, navigate to the bundle field configuration page via **Admin --> Structure --> Tripal Content --> Genome Assembly**.  For each field we want to hide, we must configure the field instance settings individually.  Click **Edit** for the Cross Reference field, and scroll down to **CROSS REFERENCE FIELD SETTINGS**.
```
almasaeed2010: Admin --> Structure --> Tripal Content --> Genome Assembly --> Manage Fields

Missing `Manage Fields`

almasaeed2010 (approved):
Followed this guide on our live site to customize field permissions. Clear and very helpful! Thanks @bradfordcondon

bradfordcondon:
thanks for reviewing.  Doyou agree this belongs in user guide instead of dev guide?

almasaeed2010:
I do. There is no programming at all in this. It's part of using Tripal rather than developing for Tripal.

bradfordcondon:
agreed. ill commit to move it.

bradfordcondon:
ok i stuck it in the user guide instead.  i still am not sure if i like where it is, but its an improvement.

spficklin:
Some suggestions for this document.
1)  Can we move it under the header 'Working with Content Types"
2)  I think since this is meant for end-users (not developers) we need to adjust some vocabulary.
    a)  The word bundle should be renamed 'content types'.  I don't think we use the word bundle for users..
    b)  It refers to the HQ module, but that's not been introduced in our tutorial prior to this document.  Perhaps we need an introduction here to the HQ module and how this field_permission module integrates with it? Or we can just leave out the HQ sentences?
    c)  So this sentence bothers me:  "Some fields are just not intuitive without some Chado experience: for example, the Cross-Reference field."    The point of fields was to take Chado out of the picture. A cross reference should be understandable without knowing anything about Chado... right?

So, one last thing to consider.... The web services do not support field_permissions I don't think.  So, if a field is hidden on a page it should also be hidden in web services, right?  I know this is just documentation and not code, but we should probably test it to see if web services is also affected.

bradfordcondon:
Thanks a lot for your feedback.  I wrote it on my blog thinking of developers but you are right i should re-think it with end users more in mind now that i've moved it there.


I moved it where you suggested, renamed bundles to content types.


>Or we can just leave out the HQ sentences?

I tried removing HQ all-together.  What do you think?  Should we add a `..note` that says if you are interested in having users submitting content, check out Tripal HQ?
 
>A cross reference should be understandable without knowing anything about Chado... right?

The field might be understandable, but not the **Widget** i would say?  Really what i mean is you need to add stuff to another table first... How about: "Some fields require prior insertion of data elsewhere: for example, the Cross-Reference field."

(side note, i just tried creating an analysis with a cross ref to the db AGL of the nonexistent accession "waffle".  the content got created, but without a dbxref.  No warning, no validation failure. This is what im getting at, a user might get confounded that they need to insert the accession first, and maybe they dont even have permission to do so.  Thats why we want to hide this widget from them.)


>The web services do not support field_permissions I don't think. So, if a field is hidden on a page it should also be hidden in web services, right? I know this is just documentation and not code, but we should probably test it to see if web services is also affected.

This is interesting.  The desired behavior would be to display the field if its set to display to an anonymous user (logged in user?)?  And the other purpose of field permissions, setting it for the widget only, doesnt effect web services...

bradfordcondon:
Looks like setting field permissions has no effect on web services: the field shows up  no matter what.

The main purpose of field permissions is (in my case) to hide confusing widgets from curators submitting content. Hiding content for non-admins was a bonus.  I would say we could keep as is and add a warning that content wont be hidden from web-services in the guide, but then we still have that warning to deal with....

Here's the warning i get:

```php
Notice: Undefined offset: 0 in TripalContentService_v0_1->addEntityField() (line 383 of /Users/bc/tripal/sites/all/modules/tripal/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc).
Notice: Undefined offset: 0 in TripalContentService_v0_1->addEntityField() (line 396 of /Users/bc/tripal/sites/all/modules/tripal/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc).

```

bradfordcondon:
easy to resolve the warning by adding this to the web services class just prior to here: 

https://github.com/tripal/tripal/blob/bfca9765a5ef7aece7b4355c886d91513a89cadb/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc#L383


```php
  if (empty($values)){
        return;
      }

      // If the value is an array and this is the field page then all of those
      // key/value pairs should be added directly to the response.
      if (is_array($values[0])) {
```

adding this also causes the ws to work how i'd expect: it isnt returned since we didnt have the permission enabled for an anonymous user.

probably also need to add a case for where cardinality != 1....  what do you think?  This would make the ws behave how i expect (field isnt shown on web services), and it resolves the php warning.

spficklin:
I like the changes!  I think we can merge, but I think we should add a new issue that requests support for the field_permissions within the web services.  If we recommend the module we should be sure to support it.

bradfordcondon:
awesome, thanks!  Ill make a separate issue for supporting it in web services so we can take more care in applying the fix, as you suggest.

bradfordcondon:
im going to add in a note referring users to tripal HQ for more information about users submitting content, then merge.  thanks for hte reviews both of you.

Issue #734, opened 2018-11-05
Issue with entity search


mboudet:
### System information

Tripal: 7.x-3.0
Drupal : 7.59 
Postgres: 9.5.8 
Php: 7.1.23

### Issue description
When using one of the entity search form (organism in this case), if a query with no results is selected, a warning pops up.

#### Steps to reproduce
Search for query with no results.

#### Error messages and screenshots
```
Notice: Undefined index: TripalEntity in TripalFieldQuery->execute() (line 127 of /var/www/html/sites/all/modules/tripal/tripal/includes/TripalFieldQuery.inc).
```
When we do a query, we generate two queries : one normal, and one with the 'count' option 
[here](https://github.com/tripal/tripal/blob/5bcbff466be1d6a530df7246cdf3c17ed6b5188d/tripal/tripal_views_query.inc#L264)
(I'm not exactly sure why.. since in this case the queries are exactly the same. Maybe there are some cases where the count query is faster?)
In any case, when the execute() method is called on the count query, we use this code : 

```
if (is_numeric($results)) {
  return $results;
}
return count($results['TripalEntity']);
```
I suppose that in some cases (non entity search maybe?), the query directly return the value. In entities, it returns all the results in $results['TripalEntity'].
However, if there are no results, $results = {}, so $results['TripalEntity'] crashes.
Testing for an empty array should be enough (returning 0 if that's the case)

The normal query (executed after this one) works fine.

bradfordcondon:
>(I'm not exactly sure why.. since in this case the queries are exactly the same. Maybe there are some cases where the count query is faster?)

Hmmm... so what I typically do is have the count query fetch ALL and the regular query fetch whats in the pagination.  This is indeed a much faster way to run the query than fetching all results and then counting them (since you only fetch the fields for whats in your pager).


>Testing for an empty array should be enough (returning 0 if that's the case)

Sure.  Or count being 0, i would think.  Which is what we attempt to do:

https://github.com/tripal/tripal/blob/5bcbff466be1d6a530df7246cdf3c17ed6b5188d/tripal/includes/TripalFieldQuery.inc#L123-L129

maybe count of 0 isnt false, and we should check if count == (int) 0 instead.

### the search tool


https://tripal.readthedocs.io/en/latest/user_guide/searching/default_pages.html?highlight=search#tripal-content-specific-search-tools

[...]

PR #745, opened 2018-11-13
Fix undefined index in search tool

mboudet:
# Bug Fix

Issue #734 

## Description
Added a test for a key existence in array. This key was not set when there was no results, such bringing up a warning. If there are no results, it just return 0.

## Testing?
Use any of the search tool (ex: /data_search/mrna) and use any query that will not return any results.
(Might need to be logged as admin to see the warning without the fix)

### DIFF
--- a/tripal/includes/TripalFieldQuery.inc
+++ b/tripal/includes/TripalFieldQuery.inc
@@ -124,7 +124,10 @@ class TripalFieldQuery extends EntityFieldQuery {
       if (is_numeric($results)) {
         return $results;
       }
-      return count($results['TripalEntity']);
+      if (array_key_exists('TripalEntity', $results)) {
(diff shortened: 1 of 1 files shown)


bradfordcondon:
I think this is a really minor and obvious fix so I'm marking it 1 review only.  Thanks for the contribution, hopefully I or someone else will get to this soon 😃 FYI I edited your post because you didnt delete the close of the comment so it looked blank.

bradfordcondon:
heres the error:

```
Notice: Undefined index: TripalEntity in TripalFieldQuery->execute() (line 127 of /Users/bc/tripal/sites/all/modules/tripal/tripal/includes/TripalFieldQuery.inc).
```

The patch does fix it.  And I can't think of any unintended consequences.  Thanks for the PR!

Issue #402, opened 2018-06-01
module collaboration marketplace and showcase


bradfordcondon:
In [today's meeting](https://docs.google.com/document/d/1sLJm2ugd5JVAI3_Ggbc15JJm9zjgnZV1wnUSjBWWukM/edit?ts=5b115cbb#) we talked about ways to improve collaboration.   We'll discuss more next month about what makes collaborations successful (thanks @ekcannon !)

One of the offshoots of the initial discussion is the idea of a curated module list.  This already exists on tripal.info, to some extent, but it requires people keep it up to date, which is hard.  Think of it as a community engagement portal, a module showcase/marketplace.    Some ideas...

* what sites use what modules?
* what modules are actively seeking collaboration?
* what modules deal with what sort of data?
* Filter could include the lead on the project
* tag modules that want collaborators, users, new maintainer/projecet lead...
* can we use github api to keep a list of modules tagged with tripal for autodiscovery/maintenance?
* It would be nice to have a listing of the Tripal sites and which modules each site uses. 

What do we think of this idea?

Where would we host it?

How can we maximize its utility while minimizing the work that goes into it?

bradfordcondon:
I'm thinking of doing a writeup for each of my modules.  2 sentences and a screenshot: what does it do, a shot of it in action on HWG.  Then I'll say I welcome collaboration, input, users, feature requests, whatever.

This is what i imagine in an ideal world: something not unlike tripal.info where each module is an entity thats searchable/filterable, maybe a random module is pulled and highlighted on the front page (like how agbiodata has the site cards you can flip through)... but i recognize that's maybe ad odds with the idea of a curated list.

spficklin:
What if we go more simple and just have a table of modules with the module name, brief description, Tripal compatibility (e.g. Tv2 or Tv3 or both), and a link to the repository?  That way it's not too much work for us to maintain on the tripal.info site.

bradfordcondon:
That could work, but is it so much better than someone searching github tags for Tripal?  We could have a direct link to this search on tripal.info and the repo...

I do think a picture is worth a thousand words in terms of showing others what functionality the module brings to the site.

I'd also love to keep it somewhere that anyone can contribute so others can update as they like when we inevitably fall behind...


[I made a showcase repo for my modules](https://github.com/bradfordcondon/condon_tripal_showcase).  I still want to add a picture per repo...

bradfordcondon:
I threw together a github pages for the repo.

[...]

PR #754, opened 2018-11-16
Extension Module list on ReadtheDocs

laceysanderson:
# Documentation
Issue #402 

## Description
This is meant to replace the extension modules list on tripal.info with the hopes of being easier to maintain and contribute to. Additionally it has the benefit of directly linking modules to the Tripal version they are compatible with and giving extension module developers commit credit in Tripal via documentation :-)

## Testing?
Build the docs (`make html` in the dcs directory. See http://www.sphinx-doc.org/en/master/usage/installation.html if you dont have Sphinx installed yet). 

**New documentation available at "Extension Modules"**: check for correct descriptions and missing modules.

**Alternatively you can look at the documentation here: https://tripal.readthedocs.io/en/402_rtd_module_list/extensions.html**

screen shot 2018-11-16 at 12 33 05 pm

## Questions to ask yourself
- Are these good categories for extension modules?
- Would it be easy to contribute to and maintain?
- Is it easy to find a module you know exists?
- Are any Tripal3-compatible modules missing?


bradfordcondon:
Some thoughts:

>Make sure to follow alphabetical order when choosing where on the category page to add your module.

If we used a TOC and loaded all modules in the folder, thsi would happen automatically.  Downside, each module would need to be its own file.

>Include links to both the documentation (even if it’s your README) and the repository (e.g. Github, Gitlab)

can we clarify that in order for the PR to be accepted, both links are **required**

relatedly, its kind of interesting.  we totally steamroll the drupal module distribution option via drupal, drush, etc).  i have some of the blame here because i dont like it, but we should maybe allow for it?


I would also say we should define some standard of "what is a tripal 3 module" that makes it elligible for this list.

docs/extensions.rst
```diff
@@ -0,0 +1,48 @@
...
+==================
+
+The below modules are Tripal 3 compatible and grouped roughly by category:
+
+.. toctree::
+   :maxdepth: 2
+
+   extensions/administrative
+   extensions/analysis
+   extensions/datainput
```
bradfordcondon: in the rest of the guide, we use snake case.  let's do so here as well for consistency.

docs/extensions/developer.rst
```diff
@@ -0,0 +1,53 @@
+
+Developer Tools
+===============
+
+The following modules provide support to both core and extension Tripal developers.
+
+Tripal Docker (Centos)
```
bradfordcondon: tehse are deprecated.  i made an issue about it in the module.  I'm going to replace it with our tripaldock tool.

bradfordcondon (approved):

spficklin:
First off. This is Fantastic!  Lots of work went into this, so thanks!

> Are these good categories for extension modules?

I like these categories.  I don't think I would change any.

> Would it be easy to contribute to and maintain?

Yes. I think anyone who develops modules could easily edit the doc and provide a pull request.

> Is it easy to find a module you know exists?

I'm not sure it matters. If you know the name of the module I think a Google search is where you'd go first.  So, I don't think that's as a big a need as advertising to new users what's out there.

> Are any Tripal3-compatible modules missing?

I'd like to add my Tripal Network module.  But that raises a question.  What modules do we list?  Do we require a full release of a module or component before they are listed?  We don't want to send folks to a module that is still in development and not likely to perform as expected.  Maybe we could have a section titled 'Modules in Development'?  And add a comment above them that developers are open to collaboration on these modules?

edit: I just re-read @bradfordcondon 's post above and saw he brought up the same question....

spficklin:
Oh the Tripal Map module is missing https://tripal.info/extensions/modules/tripalmap

spficklin:
@bradfordcondon wrote:
> relatedly, its kind of interesting. we totally steamroll the drupal module distribution option via drupal, drush, etc). i have some of the blame here because i dont like it, but we should maybe allow for it?

Yes, I think we should allow for it.  But we should encourage folks to put their modules up on Drupal.org as full projects so that the updating is more smooth.

bradfordcondon:
An in development section is a great idea. 👍

bradfordcondon:
I added an in development section... but do we have any modules to put there?  My Tripal Apollo module, maybe.  TripalMap until its t3 compatible.  Anything else?

spficklin:
I'm adding the On Hold label so that we can sort out or rating scheme before we push this. If you disagree just say so.

spficklin (approved):

spficklin:
I took the "On Hold" off of this and merged.  I was thinking about this again and thought it was silly to hold it while we waited on issue #774 .  My approval adds to @bradfordcondon so I've merged.

Issue #717, opened 2018-10-15
Unable to Create Content with Optional Contact Field


laceysanderson:
### System information

* Tripal Version: most recent dev
* Drupal Version: 7.x
* PostgreSQL Version: 9.4
* PHP Version: 5.6.7

### Issue description
When trying to create a project I receive 

>  Could not save the TripalEntity: SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "contact_id" violates not-null constraint DETAIL: Failing row contains (1, 1, null).
>    Cannot save entity

and the form is reset without creating my project.

#### Steps to reproduce
1. Go to "Add Tripal Content" > "Project"
2. Enter a name only and click Save.

*You can also test this with "DNA Library" by only entering the required information and NOT A CONTACT.*

laceysanderson:
There is a temporary fix of removing the contact field from the tripal content type using manage fields.

bradfordcondon:
is anyone working on this issue?  I can look into it tonight...
edit sorry not getting to it tonight, maybe tomorrow...

spficklin:
I am not

laceysanderson:
Feel free if you end up having time! I haven’t even had the chance to look at it yet

bradfordcondon:
oops- this looks pretty straight forward.  The linker contact field should be attached, not the base table contact field.

The fix- less straight forward, since the project fields arent directly added but inferred somewhere along the way...

bradfordcondon:
heres where the contact field gets added. I'm going to sleep on it.  It looks to me like the **base** contact field gets added if the linker table exists? ie instead of `local__contact_widget`, attach `chado_linker__contact_widget` ???  But that can't be right, otherwise, all tables with a contact linker would be broken?


https://github.com/tripal/tripal/blob/be494de1cb223bcb8c4a4d556410fe702b9622cd/tripal_chado/includes/tripal_chado.fields.inc#L2215-L2257


Yep changing the widget and formatter to `chado_linker__contact_widget` resolves the error.... just a simple mistake then?  since the `local__contact` field is manually added to the tables that use it.  I'll submit a PR but let's think on it before labeling it a simple fix...

also need to provide an updatedb that will remove the faulty field and add the new one, i guess?

bradfordcondon:
[...]

PR #764, opened 2018-11-26
resolve content creation bug w/ contact field: attach linker contact field instead of base field 

bradfordcondon:
# bug fix 

Issue #717

## Description
per @laceysanderson 's report:

When trying to create a project:

```
Could not save the TripalEntity: SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "contact_id" violates not-null constraint DETAIL: Failing row contains (1, 1, null).
Cannot save entity
```
and the form is reset without creating the project.

after examining the bug, the same should be true for other content with a contact linker field instead of a base contact field, for example, library.

## Testing?
create a project and/or library on 7.x-3.x with NULL for contact, it should fail.  

switch to this branch.  **clear the cache**.  remove the contact field, and then press the check for new fields button.  You should now be able to create the content.

(note: how do I delete and re-add the field as an `updatedb`?)

### DIFF
--- a/tripal_chado/includes/tripal_chado.fields.inc
+++ b/tripal_chado/includes/tripal_chado.fields.inc
@@ -2241,7 +2241,7 @@ function tripal_chado_bundle_instances_info_linker(&$info, $entity_type, $bundle
         'term_name' => 'contact',
       ),
       'widget' => array(
-        'type' => 'local__contact_widget',
+        'type' => 'chado_linker__contact_widget',
         'settings' => array(
           'display_label' => 1,
         ),
@@ -2249,7 +2249,7 @@ function tripal_chado_bundle_instances_info_linker(&$info, $entity_type, $bundle
       'display' => array(
         'default' => array(
           'label' => 'hidden',
-          'type' => 'local__contact_formatter',
+          'type' => 'chado_linker__contact_formatter',
           'settings' => array(),
         ),
       ),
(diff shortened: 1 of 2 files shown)


laceysanderson (commented):
Confirmed that this fixes the issue! However, I am also not sure on how to programmatically remove and re-add the field.

laceysanderson:
@spficklin can you point us in the right direction for programatically removing fields?

bradfordcondon:
i added `tripal_chado_update_7336` which updates the field instance info to use the correct formatter and widget!!!!   i tested this with the library table and confirmed that running the `updatedb` resolves the issue.

friendly reminder you can run the update explicitly: `drush eval "module_load_install('tripal_chado'); tripal_chado_update_7336();"`

spficklin (approved):
I'm providing just a quick code review, not a functional review for now.  It's looks great.

bradfordcondon:
regarding the failed test, @almasaeed2010 has seen this elsewhere as well.  im pretty sure it's unrelated, its the test environment running out of memory because the OBO loader test loads the full GOslim, it can't handle it when the image resources get squeezed down from having all the environments running.

laceysanderson (approved):
Functionally tested and it works flawlessly! 👍 😀

laceysanderson:
Re-ran and now the tests passed 👍

laceysanderson:
Merged! Thank you for the fix @bradfordcondon!!

Issue #669, opened 2018-09-21
Updating a non-required field to an empty value does not clear it


almasaeed2010:
### System information

* Tripal Version: latest dev (3.x)
* Drupal Version: 7.59
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

When updating an entity, we can't remove a value from field.

#### Steps to reproduce

1. Create an organism page and make sure you add a value to the `common name` field.
1. Edit the entity and remove the value from the `common name` field.
1. Save and check that the `common name` field.

bradfordcondon:
I have confirmed this bug.  It seems to occur for TEXT FIELD type fields.

bradfordcondon:
I think this is resolved but want to verify.

bradfordcondon:
confirmed **unresolved**.  once set, you cannot unsett the abbreviation or the common name field of an organism, for example.


Heres what I think is going on:

we assign fields and widgets to the base columns first by looking up the semantic web term, and then adding the appropriate widget based on the value type:

https://github.com/tripal/tripal/blob/c3b0ea1914b4a80aeeaf5b62fe638d3987a3a695/tripal_chado/includes/tripal_chado.fields.inc#L1000-L1012

So i think we need to hook into the base text__textfield and text__textarea field widget validate method and account for when users unset them.  It would be no problem for field storage, but, we specifically deal with unsetting the value in our custom fields.  see for example the sequence field's validate method:

https://github.com/tripal/tripal/blob/c3b0ea1914b4a80aeeaf5b62fe638d3987a3a695/tripal_chado/includes/TripalFields/data__sequence/data__sequence_widget.inc#L77-L82


if hooking doesnt work, we need to extend the text__textifeld widget and overwrite the validate.

bradfordcondon:
ok my plan is to replace the base text_textfield and text_textarea widgets with widgets that extend ChadoFieldWidget.

I plan on calling them `chado_base__textfield_widget` (to mesh with the `chado_linker` named fields.) . No need to define a new field, just the widget.

Replacing the common name textfield widget with my new `chado_base__textfield_widget` allows me to null out the common name, so I'll start working on a PR.  I need to replace the widgets for *only the fields connected to chado* that are textfield and textarea.  I think I can do this by checking if the field has a `['chado_table']` key.

spficklin:
@bradfordcondon I think there may be an easier way.  Before you get too deep into this let me spend some time on it.

bradfordcondon:
[...]

PR #775, opened 2018-12-06
Removing non-required fields from an Entity

spficklin:
# Bux Fix

Issue #669

## Description

Copying @almasaeed2010 's first comment on the issue:

> When updating an entity, we can't remove a value from field.

This fix corrects that problem.  It's just a few lines of code, so I think it only needs one functional review.

## Testing?

Copying @almasaeed2010 's testing steps from his first comment on the issue:

> Create an organism page and make sure you add a value to the common name field.
> Edit the entity and remove the value from the common name field.
> Save and check that the common name field.


almasaeed2010 (approved):
Works for me. Thanks @spficklin

Issue #165, opened 2017-09-18
Feature Request: Tripal Pane Browser History


almasaeed2010:
Hi,

Recently we encountered a situation with Tripal Panes where we needed a specific pane to be auto displayed if the pane had paginated content and the user requested a page other than the first. We are not sure if a solution for this has already been implemented, but we'd be happy to help create it if it hasn't been created yet.

I'd propose using [browser history](https://developer.mozilla.org/en-US/docs/Web/API/History_API), which is now supported by all major browsers, along with simple url query parsing to implement this. If you agree this solution would work, I wouldn't mind implementing it.

Thanks!

spficklin:
Ah... thanks for noticing this issue.  I'm looping in @shawnawsu so we can talk on this one.

bradfordcondon:
Hi @spficklin @shawnawsu  @almasaeed2010 

Bumping this issue, another situation arose on HWG where we wish we had this feature.  When Elasticsearch fails, we want to reload the page and have the elasticsearch box open with a "no results found" message. Very disorienting for the entity to reload in its starting configuration instead of with the elasticsearch pane open.

almasaeed2010:
Hello all,

I think the best solution for this is if we can somehow pass the tripal pane information (html ids and classes of the container pane) to the field formatter one way or another. Is this possible?

Thanks!

bradfordcondon:
@almasaeed2010 I *think* this is resolved as we can keep panes open.  Is it?

almasaeed2010:
Hi! I don't think this has been resolved yet on core. We do implement our own workaround but it works only on our site. Also, the fields that we developed work but the ones we didn't won't work (for us) so we still need an implementation on the core side that makes it consistent across sites.

Thanks!

bradfordcondon:
oops thanks for the clarification.  Moving this out of Q&A and into the backlog.  I think the need for this is self-evident without PMC voting as a new feature, its more of a bug fix (ah which stephen already labeled as).

bradfordcondon:
Discussing with @spficklin 

"If you use AJAX this isnt a problem"

So the tripal field has coding that allows you to do paging via AJAX.  Its in the base field and/or the formatter.


We're going to add documentation.  Then we'll check with @almasaeed2010 and see if thats good enough or if it still should be added somehow to core.

almasaeed2010:
[...]

PR #780, opened 2018-12-08
New developer docs:  TripalFieldFormatter, pagers and creating field instances.

spficklin:
# Documentation

Issue #165

## Description

One solution for issue #165  was to provide documentation for how to use the AJAX pager for fields that had too much data to show it all.  But this opened up a can of worms in that there was no documenation for how to setup a formatter class and there was no page for how to create fields and field instances and allow users to add fields via the web interface.  This PR adds all of the documentation.

## Testing?

Just pull the branch, and rebuild the docs.


docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+        "local:abbreviation" =>  "C. sinensis",
+        "TAXRANK:0000005" => "Citrus",
+        "TAXRANK:0000006" => "sinensis",
+        "entity" => "TripalEntity:3",
+      ],
+      "chado-feature__organism_id" => 12,
+    ],    
+  ];
+  
+You may recall that the ``$items`` array structure is the same as that created by the ``load()`` function described in the :doc:`manual_field_creation` page. Note that each key in the ``value`` array is an accession for a controlled vocabulary term.  These accessions are used to unambiguously describe the value. To display the organism on a page we need the element named ``rdfs:label``.  Therefore, on line 4 of the code above we set the ``$content`` variable to contain this value.
```
laceysanderson: > Therefore, on line 4 of the code above we set the ``$content`` variable to contain this value.

This line is a bit confusing since it doesn't apply to the code block directly above the sentence but the one before that.

```suggestion
You may recall that the ``$items`` array structure is the same as that created by the ``load()`` function described in the :doc:`manual_field_creation` page. Note that each key in the ``value`` array is an accession for a controlled vocabulary term.  These accessions are used to unambiguously describe the value. To display the organism on a page we need the element named ``rdfs:label``.  Thus, we set the ``$content`` variable to contain this value as shown on line 4 of the ``view()`` function above.
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Because our organisms are also published entities we want to link to their respective pages each time an organism is displayed.  Because the ``value`` array has an element named ``entity`` we know that this item is published.  Lines 5-6 above use this information to create a clickable link to the organism page.   Finally, the ``$element`` argument is set to provide content of type ``markup``.  This ``$element`` array is a `Drupal renderable array <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_.
```
laceysanderson: Same concerns with the following sentence:

>Lines 5-6 above use this information to create a clickable link to the organism page.

My suggestion simply specifies that it's the `view()` function code block.

```suggestion
Because our organisms are also published entities we want to link to their respective pages each time an organism is displayed.  Because the ``value`` array has an element named ``entity`` we know that this item is published.  Lines 5-6 of the ``view()`` function shown above use this information to create a clickable link to the organism page.   Finally, the ``$element`` argument is set to provide content of type ``markup``.  This ``$element`` array is a `Drupal renderable array <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_.
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Lastly, notice the element named ``chado-feature__organsim_id``.  This element is at the same level as the ``value`` element.  This data is meant to be used internally by the field. It maps this fields values to the appropriate table in Chado where the data is stored.  
+
+.. warning:: 
+
+  You should never show to the user any data that is outside of ``value`` element.
```
laceysanderson:
```suggestion
  You should never show the user any data that is outside of ``value`` element.  
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+Lastly, notice the element named ``chado-feature__organsim_id``.  This element is at the same level as the ``value`` element.  This data is meant to be used internally by the field. It maps this fields values to the appropriate table in Chado where the data is stored.  
+
+.. warning:: 
+
+  You should never show to the user any data that is outside of ``value`` element.  
+
+In summary, the following should be observed when processing the ``$items`` array for viewing:
+
+  - A field with only one value (a cardinality of 1) will always have only one element in the ``$items`` array and can use the index 0. This is what has been done in this example code. 
+  - A field with more than one value can have any number of elements in the ``$items`` array.  You should therefore iterate through all of them.
```
laceysanderson, line 76: Just adding a few more points here that might trip people up.

```suggestion
  - A field with more than one value can have any number of elements in the ``$items`` array.  You should therefore iterate through all of them.
  - For every index in ``$item`` you should create a matching index in ``$element``to display the data found in that ``$item``.
  - If there are no items, then nothing you return will be displayed.
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+
+To create a pager we first need to calculate the number of items we want to display per page and the total number of pages required to display all of the data.  
+
+.. code-block:: php
+  
+  $items_per_page = 10;
+  $total_records = count($items);
+  $total_pages = (int) ($total_records / $items_per_page) + 1;
+  
+Next, we must initialize the pager by calling the ``pager_default_initialize`` function.  We pass it the total number of records, the number of items per page and the index (i.e. ``$pelement``) for this pager on the page.
```
laceysanderson, line 100: This sentence still leaves me wondering how to set the index properly...

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+  $total_pages = (int) ($total_records / $items_per_page) + 1;
+  
+Next, we must initialize the pager by calling the ``pager_default_initialize`` function.  We pass it the total number of records, the number of items per page and the index (i.e. ``$pelement``) for this pager on the page.  
+
+.. code-block:: php
+
+  $pelement = 0; 
+  $current_page = pager_default_initialize($total_records, $items_per_page, $pelement);
+  
+The call to ``pager_default_initialize`` will return the current page. Each time the user navigates to other pages the ``view()`` function is called and the current page is always provided via this function call. Next, we must theme the pager so that it follows the look-and-feel prescribed for the site. For this we use the Drupal ``theme()`` function.
```
laceysanderson: It could be a bit more clear what the ``$current_page`` variable holds. Is it simple the index (e.g. `3`) of the current page or is it all the items for the current page? An example of how I use this variable to process my items array would be helpful.

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,203 @@
...
+      'default' => array(
+        'label' => 'hidden',
+        'type' => 'sbo__relationship_formatter',
+        'settings' => array(
+          'title' => 'Relationships',
+          'empty' => 'There are no relationships'
+        ),
+      ),
+    ),
+
```
laceysanderson, line 180: This has often tripped me up!

```suggestion

.. warning::

    In order for the ``settingsForm()`` implemented to be available on the "Manage Display" page, you must also implement ``settingsSummary()`` as described below.
```

laceysanderson (approved):
I built the documentation and it looks as expected :-) 

Great documentation --this is So Needed! There are a few minor suggestions but overall, I approve these docs!

docs/dev_guide/custom_field/custom_formatter.rst
spficklin: Unfortunately, this is a problem.   The real answer is that the index has to be a unique number for the page.  But you don't know how many fields there will be on the page that have pagers.  So, the unfortunately reality is you don't know what to set it as.  We need to adjust the code so that this index gets set automatically for the user and they don't have to worry about it.

spficklin: I added a new issue to address this problem  #783

docs/dev_guide/custom_field/custom_formatter.rst
spficklin: I've adjusted the text to make this more clear.

docs/dev_guide/custom_field/custom_formatter.rst
spficklin: So as not to hold up this documentation and to resolve issue #165 I'm going to merge this change. But issue #783 should address this concern for you @laceysanderson

spficklin:
@bradfordcondon and @almasaeed2010 would either of you care to take a look at this before I merge?

docs/dev_guide/custom_field/create_instance.rst
```diff
@@ -0,0 +1,220 @@
+Attach Fields to Content Types
+==============================
+In summary, creation of a new field requires creation of three classes that inherit from the ``TripalField``, ``TripalFieldWidget`` and ``TripalFieldFormatter`` base classes.  If the fields are created correctly and placed in the ``includes/TripalFields`` directory of your module then Tripal will automatically find them.  However, the field is not yet attached to any content type. They must be attached.  Fields can be attached programmatically or via the online Drupal interface by a site admin. 
+
+The hook_bundle_field_info() function
+-------------------------------------
+ The three TripalField classes simply define how the field will function, but Drupal does not yet know about the field.  The ``hook_bundle_field_info`` function tells Drupal about your field. It must be implemented in a custom Drupal module,and provides an array that tells Drupal about the fields and the classes to use for the field.  Suppose we were creating a field named ``obi__genus`` which displays the Genus for a species and we have a custom module named ``tripal_org2``.  The hook function would be named ``tripal_org2_bundle_field_info()``:
```
bradfordcondon:
```suggestion
 The three TripalField classes simply define how the field will function, but Drupal does not yet know about the field.  The ``hook_bundle_field_info`` function tells Drupal about your field. It must be implemented in a custom Drupal module, and provides an array that tells Drupal about the fields and the classes to use for the field.  Suppose we were creating a field named ``obi__genus`` which displays the Genus for a species and we have a custom module named ``tripal_org2``.  The hook function would be named ``tripal_org2_bundle_field_info()``:
```

typo (module,and).

I also notice we dont bring up the libraries option...

docs/dev_guide/custom_field/create_instance.rst
```diff
@@ -0,0 +1,220 @@
...
+          'type' => 'field_chado_storage',
+        ],
+        'settings' => [],
+      ];
+   }
+    
+    return $info
+  }
+  
+This function receives as it's second argument the ``$bundle``. This is the bundle that Drupal is requesting new fields for.  For this example we only want to attach the field if the content type is the organism content type.  The format of the returned ``$info`` array should have the field name as the key and an array that follows the instructions provided by Drupal's `field_create_field() <https://api.drupal.org/api/drupal/modules%21field%21field.crud.inc/function/field_create_field/7.x>`_ function.
```
bradfordcondon:
```suggestion
This function receives as its second argument the ``$bundle`` object. This is the bundle that Drupal is requesting new fields for.  For this example we only want to attach the field if the content type is the organism content type.  The format of the returned ``$info`` array should have the field name as the key and an array that follows the instructions provided by Drupal's `field_create_field() <https://api.drupal.org/api/drupal/modules%21field%21field.crud.inc/function/field_create_field/7.x>`_ function. 
```

#### typos
it's
#### suggestions

bundle -> bundle object

docs/dev_guide/custom_field/create_instance.rst
```diff
@@ -0,0 +1,220 @@
...
+    }
+    return $info;
+  }
+  
+The format of the returned ``$info`` array should have the field name as the key and an array that follows the instructions provided by Drupal's `field_create_instance() <https://api.drupal.org/api/drupal/modules%21field%21field.crud.inc/function/field_create_instance/7.x>`_ function. 
+
+Unique to this info array are the settings related to Chado.  Because we expect our data to be loaded from Chado we must specify these settings:
+
+ - ``base_table``: the name of the base table to which the record will be associated. In our case the ``organism`` table of Chado is the base table.
+ - ``chado_table``: the name of the actual table form which the value of the field will be loaded or saved to.  In our case the ``organism`` table is also the ``chado_table``.
```
bradfordcondon, line 104: I ALWAYS found this confusing.  Can we please elaborate on the case where base table != chado_table ?

Maybe a ..note directieve after this:

```rst
.. note

   When does the chado_table not match the base_table?  Consider the contact linker field when attached to an organism.  In this case, the base table is the  contact field, and the chado table is the organism_contact table
```
I'm actually not even 100% sure about that since i still find it confusing hahah

docs/dev_guide/custom_field/create_instance.rst
```diff
@@ -0,0 +1,220 @@
...
... we must let Drupal know that our field exists.  We do this by adding an entry to the ``$info`` array of in the ``hook_bundle_field_info()`` function described above.  This let's Drupal know about our field. However, because we are not programmatically creating an instance of the field on a content type, but allowing the user to create them we do not need to implement the ``hook_bundle_instance_info()`` function. Instead, we must implement the ``hook_bundle_create_user_field()``.  This function is called when the user attempts to add our new field to a bundle.  One field that comes with Tripal is the ``chado_linker__prop`` field.  Most Chado base tables have an associated property table (e.g. ``organismprop``, ``featureprop``, ``stockprop``, etc). By default, the ``tripal_chado`` module automatically adds this field to all bundles that have existing properties. It adds a new instance for every propert type.  However, new properties can be added to bundle, and the site admin may want to add those properties via the user interface rather. Therefore, this field has the `$no_ui` set to TRUE and uses the  ``hook_bundle_create_user_field()`` to create the new field instnace for the user.
```
bradfordcondon:
```suggestion
Next, we must let Drupal know that our field exists.  We do this by adding an entry to the ``$info`` array of in the ``hook_bundle_field_info()`` function described above.  This lets Drupal know about our field. However, because we are not programmatically creating an instance of the field on a content type, but allowing the user to create them we do not need to implement the ``hook_bundle_instance_info()`` function. Instead, we must implement the ``hook_bundle_create_user_field()``.  This function is called when the user attempts to add our new field to a bundle.  One field that comes with Tripal is the ``chado_linker__prop`` field.  Most Chado base tables have an associated property table (e.g. ``organismprop``, ``featureprop``, ``stockprop``, etc). By default, the ``tripal_chado`` module automatically adds this field to all bundles that have existing properties. It adds a new instance for every property type.  However, new properties can be added to bundle, and the site admin may want to add those properties via the user interface rather. Therefore, this field has the ``$no_ui`` set to TRUE and uses the  ``hook_bundle_create_user_field()`` to create the new field instance for the user.
```

#### typos

let's -> lets
propert
instnace
``$no_ui`` needs double `

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
+Creating a Custom Formatter
+===========================
+The third component of a field is the formatter.  Thus far we have introduced how to create a field class and a widget class for a field.  The field class is responsible for describing the field, loading data into it, and providing search support.  The widget class provided a Drupal form for online editing of the field.  Finally, the formatter is responsible for display of the field on a Tripal site.  
+ 
+.. note::
+  This guide assumes you already have your formatter class file created. For more information, see :doc:`manual_field_creation` or, :doc:`tripal_field_generator`. 
+  
+The formatter class is the most simple of all the Tripal field classes.  Here we will again use the **obi__organism** field that come swith the ``tripal_chado`` module.
```
bradfordcondon:
```suggestion
The formatter class is the simplest of all the Tripal field classes.  Here we will again use the **obi__organism** field that comes with the ``tripal_chado`` module.  
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+        '#markup' => $content,
+      );
+    }
+  }
+  
+In the code above the input arguments have the following meaning:
+  - ``$element`` is the first argument. It is an array into which you should set the contents to be displayed.  
+  - ``$entity_type`` will always have the value ``Tripal Entity``.
+  - ``$entity`` is the entity object which contains all information about the entity including the loaded data values.
+  -  ``$langcode`` is the language. This is used by Drupal to provide translations of data into other spoken languages. By default, Tripal does not use a language.
```
bradfordcondon: maybe explain why
```suggestion
  -  ``$langcode`` is the language. This is used by Drupal to provide translations of data into other spoken languages. By default, Tripal does not use a language, as biological data is generally language agnostic.  Consider for example a gene sequence or a feature coordinate.
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+
+Because our organisms are also published entities we want to link to their respective pages each time an organism is displayed.  Because the ``value`` array has an element named ``entity`` we know that this item is published.  Lines 5-6 of the ``view()`` function shown above use this information to create a clickable link to the organism page.   Finally, the ``$element`` argument is set to provide content of type ``markup``.  This ``$element`` array is a `Drupal renderable array <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_.
+
+Lastly, notice the element named ``chado-feature__organsim_id``.  This element is at the same level as the ``value`` element.  This data is meant to be used internally by the field. It maps this fields values to the appropriate table in Chado where the data is stored.  
+
+.. warning:: 
+
+  You should never show the user any data that is outside of ``value`` element.
```
bradfordcondon: explain why?  because webservices right?

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+
+.. warning:: 
+
+  You should never show the user any data that is outside of ``value`` element.  
+
+In summary, the following should be observed when processing the ``$items`` array for viewing:
+
+  - A field with only one value (a cardinality of 1) will always have only one element in the ``$items`` array and can use the index 0. This is what has been done in this example code. 
+  - A field with more than one value can have any number of elements in the ``$items`` array.  You should therefore iterate through all of them.
+  - For every index in ``$item`` you should create a matching index in ``$element``to display the data found in that ``$item``.
```
bradfordcondon:
```suggestion
  - For every index in ``$item`` you should create a matching index in ``$element`` to display the data found in that ``$item``.
```
note formatting:

screen shot 2018-12-13 at 10 51 44 am

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+  - If there are no items, then nothing you return will be displayed.
+  - For each element in the ``$items`` array there is a ``value`` key.  Only the data in the ``value`` key should be shown to the user.
+  - Each element in the ``$items`` array may have more than a ``value`` key.  These values are meant to help manage the data. 
+
+.. warning::
+
+  You should never have SQL statments or any API calls that retreive data in the foramter ``view()`` function. The formatter should strictly format data for viewing.
```
bradfordcondon, line 84: again i find these sort of directives easier to follow, and to remember, if i know why im being told to do it.

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+.. code-block:: php
+
+  $pager = theme('pager', array(
+    'tags' => array(),
+    'element' => $pelement,
+    'parameters' => array(),
+    'quantity' => $total_pages,
+  ));
+  
+By default, all links in the pager cause the page to reload.  We do not want the page to reload, rather we only want to update the contents of the field.  The TripalFieldFormatter class provides a function named ``axaifyPager`` to convert a pager into an AJAX pager:
```
bradfordcondon:
```suggestion
By default, all links in the pager cause the page to reload.  We do not want the page to reload, rather we only want to update the contents of the field.  The TripalFieldFormatter class provides a function named ``ajaxifyPager`` to convert a pager into an AJAX pager:
```

docs/dev_guide/custom_field/custom_formatter.rst
```diff
@@ -0,0 +1,208 @@
...
+.. warning::
+
+    In order for the ``settingsForm()`` implemented to be available on the "Manage Display" page, you must also implement ``settingsSummary()`` as described below.
+The settingsSummary() Function.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``settingsSummary()`` function provides a summary of the current settings values for a field on the **manage display** page.  The following shows the same relationship field from the previous section, but with the settings form closed, and a summary of the current values shown:
+
+.. image:: custom_formatter.settings.2.png
+
+An example of the ``sesttingsSummary()`` function that generates the summary in the image above is as follows:
```
bradfordcondon:
```suggestion
An example of the ``settingsSummary()`` function that generates the summary in the image above is as follows:
```

bradfordcondon (changes requested):
looks good.  Thank you very much.  These guides were much needed and I'll get use out of referring to them.  i think you'll find most of my requested changes obvious typo fixes, but one or two are requests for elaboration on the "why" of things

spficklin:
Thanks @bradfordcondon I merged in your suggestions and I updated the docs to respond to your requsted changes.   Thanks for the review!

Issue #769, opened 2018-12-03
Can't remove publication from organism page


patricksis:
### System information

* Tripal Version: 3.0
* Drupal Version: 7.61
* PostgreSQL Version: 9.6.11
* PHP Version: 7.2.11

### Issue description
Publication was link to the organism page instead of reference genome and now the publication cannot be remove from organism page. 

#### Steps to reproduce
Create organism page and link the publication. Attempt to edit and remove publication. Try to save organism. You should get an error

#### Error messages and screenshots

```
Could not save the TripalEntity: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "" LINE 1: ...nism_pub_id = '3', organism_id = '60', pub_id = '' WHERE ... ^
Cannot save entity
```

spficklin:
Hi @patricksis I can confirm that this does occur as you have indicated.  I've marked this as a bug and added it to our milestones to fix before our v3.1 release which we hope to put out in the next few weeks. 

In the meantime I'll look into this and see what's going on and try to make a fix.

spficklin:
@patricksis I believe I have fixed the problem and I've committed a pull request for review by others on the Tripal core development team.  If you feel comfortable with pulling branches would you mind testing to see if the fix works for you?  If not, we'll perform the test and let you know if successful when it's available in the main branch.

bradfordcondon:
merged and confirmed.  thanks for bringing this to our attention!

PR #781, opened 2018-12-08
Can't remove publication

spficklin:
# Bug Fix

Issue #769 

## Description

If a publication is added to an Organism content type it cannot later be removed without an error:

## Testing?

1. Pull this branch
2. Test removing pub from a linker table:
     - Create or edit an "Organism" content type and link to any publication and save the organism.
    -  Attempt to edit and remove publication, then save.
    -  The publication should be gone and no error should be reported.
    - You can repeat this as often as you like with any number of publications.
3.  Test removing pub from base table.
    - Create or edit a "Study" content type and repeat the same test above.
    - You should be able to add and remove the publication.


almasaeed2010:
I did a functional test following the provided steps and this has worked for me without any errors. Thanks @spficklin !

almasaeed2010 (approved):

spficklin:
Thanks for the test @almasaeed2010

bradfordcondon:
this PR looks good to me.  thanks all.

PR #782, opened 2018-12-10
Remove php 5.6 support from travis

almasaeed2010:
Issue #

## Description

This PR adds support for php7.2 and remove 5.6 on Travis.

PHP 5.6 is nearing end-of-life. Our PRs have already dropped support for 5.6. For example, PR #678 uses type declarations that are only supported on php 7.0+. Therefore, many of our PRs will fail on travis if we don't remove 5.6 support and instead add 7.2.

screen shot 2018-12-10 at 10 12 21 am

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ sudo: required
 php:
   - 7.0
   - 7.1
-  - 7.2
 
 env:
   - BASE_URL="http://127.0.0.1:8080"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ services:
 sudo: required
 
 php:
-  - 5.6
   - 7.0
   - 7.1
+  - 7.2
 
 env:
   - BASE_URL="http://127.0.0.1:8080"


bradfordcondon:
just to provide a link:

https://dri.es/the-end-of-php-5

December 31st 2018, Drupal 7 will no longer support php 5.6 officially.  Unofficially they will continue to support it as long as security fixes are being provided by a third party.

spficklin:
PHP 7.2 still doesn't work with Drupal even though there has been a patch.  I still get errors. Perhaps we leave it off for now?

.travis.yml
```diff
@@ -7,9 +7,9 @@ services:
 sudo: required
 
 php:
-  - 5.6
   - 7.0
   - 7.1
+  - 7.2
```
almasaeed2010:
```suggestion
```

spficklin:
This is such a simple change. it only needs 1 review. and given the Slack discussion in which @laceysanderson agreed it was okay to drop off PHP 5 testing I'm going to merge.

Issue #784, opened 2018-12-10
GOslim ontology runs out of memory in test


bradfordcondon:
frustrating loading the goslim ontology exhuasts my memory when running local unit tests.  I see this happen in hte travis environment all the time as well, causing tests to fail.  rerunning fixes the issue.

we should probably change this test to not load the full goslim plant obo.

```
.Downloading URL http://www.geneontology.org/ontology/subsets/goslim_plant.obo, saving to /private/var/folders/ms/jc3nwzvd57z7c2hrf7n264yc0000gn/T/obo_FMTgQ3
.......Building cvterm paths for vocabulary: tripal_obo_test
Clearing the cvtermpath table for this vocabulary...
Clearing completed.
Retrieving relationships...
Processing terms...
.PHP Fatal error:  Allowed memory size of 2684354560 bytes exhausted (tried to allocate 12288 bytes) in /Users/bc/tripal/includes/database/database.inc on line 2227
PHP Stack trace:
PHP   1. {main}() /usr/local/bin/phpunit:0
PHP   2. PHPUnit\TextUI\Command::main() /usr/local/bin/phpunit:570
PHP   3. PHPUnit\TextUI\Command->run() phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:148
PHP   4. PHPUnit\TextUI\TestRunner->doRun() phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:195
PHP   5. PHPUnit\Framework\TestSuite->run() phar:///usr/local/bin/phpunit/phpunit/TextUI/TestRunner.php:545
PHP   6. PHPUnit\Framework\TestSuite->run() phar:///usr/local/bin/phpunit/phpunit/Framework/TestSuite.php:755
PHP   7. PHPUnit\Framework\TestSuite->run() phar:///usr/local/bin/phpunit/phpunit/Framework/TestSuite.php:755
PHP   8. Tests\tripal_chado\TaxonomyImporterTest->run() phar:///usr/local/bin/phpunit/phpunit/Framework/TestSuite.php:755
PHP   9. PHPUnit\Framework\TestResult->run() phar:///usr/local/bin/phpunit/phpunit/Framework/TestCase.php:894
PHP  10. Tests\tripal_chado\TaxonomyImporterTest->runBare() phar:///usr/local/bin/phpunit/phpunit/Framework/TestResult.php:698
[...]

PR #788, opened 2018-12-11
remove remote obo test

bradfordcondon:
# Bug fix

Issue #784

The test in question loads the entire Goslim OBO hosted remotely.  Loading OBO files is memory intensive, even GOslim, so automated checks keep failing because the test environment runs out of memory.

I propose that we simply remove the test.  We have testing for a small, custom OBO file hosted locally, this is much better and more specific than loading goslim.


spficklin (approved):

spficklin:
This is easy enough that I think a code review is only needed.

Issue #724, opened 2018-10-19
minor bug: NCBI Taxon db URL linkout is incorrect


bradfordcondon:
### System information

* Tripal Version: 3

### Issue description

please see #722 
the default linkout is to http://purl.obolibrary.org/obo/ncbitaxon#%7Baccession%7D

this is no good.  it links to, for the tutorial organism, purl.obolibrary.org/obo/ncbitaxon#NCBITaxon:2711

i propose https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?&id={accession}

#### Steps to reproduce
use the ncbi taxon importer.  Check the dbxref linkout.

bradfordcondon:
pr merged

Issue #722, opened 2018-10-19
CrossReference


kyrenya:
Hi
Using UBUNTU 18.04 and last Tripal v3.0 Release installation
I'm trying to follow the tutorial of the citrus example: 

When creating the organism, my link to CrossReference never works.. 

![image](https://user-images.githubusercontent.com/43243429/47213363-aadb8800-d392-11e8-94ba-a5d2a3e42fd8.png)

should link to: 
purl.obolibrary.org/obo/ncbitaxon#NCBITaxon:2711
never loads. 
is that an issue of the OBO or my tripal crossreference config? 

Thanks!

bradfordcondon:
is `purl.obolibrary.org/obo/ncbitaxon#NCBITaxon:2711` the built URL or the desired URL?

If its teh desired URL and the built URL is wrong, what is the built URL?

you can check the configuration of your database crossref by going to `admin/tripal/loaders/chado_db/ ` can you share the relevant DB information?

kyrenya:
This is the address that tries to achive when click in the CrossRef..  so.. then I guess is the built URL isnt it?
I can't find anything related to NCBITaxon in admin/tripal/loaders/chado_db/. 
I just followed the steps in here https://tripal.info/tutorials/v3.x/example-genomic-site/organism for import taxonomy for existing species (Chado NCBI Taxonomy Loader).

bradfordcondon:
Please follow the tripal 3 guide at https://tripal.readthedocs.io/en/latest/user_guide/example_genomics/organisms.html?highlight=organisms.  It's more up to date.  @spficklin is there a reason we still have the tripal 3 user guide up on tripal.info?
OK thanks, in the future if you link what part of the guide you're following that will be helpful for me to answer your questions.

In this case, the bug is that the NCBI taxon db entry is wrong. here's mine: (found by going to the page i linked and searchign for NCBI.)

<img width="1335" alt="screen shot 2018-10-19 at 10 18 42 am" src="https://user-images.githubusercontent.com/7063154/47223989-7a8eec00-d388-11e8-9a6d-291aa2a19bc8.png">


you can see the linkout tripal will build for a given accession is http://purl.obolibrary.org/obo/ncbitaxon#{accession}. This isnt correct.  it should instead be, perhaps, https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?&id={accession}

[...]

PR #793, opened 2018-12-13
update the NCBI taxon db to not go to purl but to NCBI

bradfordcondon:
# Bug fix

Issue #724 
see also #722 for user encountering this problem.

## Description

the NCBI taxon db linked out to PURL.  This isn't ideal, because if you have a dbxref using NCBI taxon you probably want to link out to NCBI taxon, not the "ontology" itself.

ie visit https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=7004 not http://purl.obolibrary.org/obo/ncbitaxon#%7B720%7D

### DIFF
--- a/tripal_chado/includes/tripal_chado.semweb.inc
+++ b/tripal_chado/includes/tripal_chado.semweb.inc
@@ -2135,7 +2135,7 @@ function tripal_chado_populate_vocab_NCBITAXON() {
     'name' => 'NCBITaxon',
     'description' => 'NCBI organismal classification.',
     'url' => 'http://www.berkeleybop.org/ontologies/ncbitaxon/',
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
Functionality Tested 👍 
Also, code-wise, looks good 👍

laceysanderson:
Since it looks like the test failed due to an out of memory error, should we merge anyway?

bradfordcondon:
what i suggest is check the builds.  Generally what is happening is one build is passing and one is failing.  The out of memory error is definitely environment related.  Since one of the builds did succeed, and no tests fail, IMO its clear to merge.

We do really need to get this problem fixed though :|

laceysanderson:
That was my thoughts as well 👍

Issue #749, opened 2018-11-14
request: support field permissions module in web services


bradfordcondon:
## Feature Request/Discussion

Background #742 

I wrote a guide demonstrating field-level permissions implemented ia the drupal field_permissions module, but web services dont support them.  If you make a field admin only, two things happen:
* admin only fields show up in web service requests
* You get this warning:

```
Notice: Undefined offset: 0 in TripalContentService_v0_1->addEntityField() (line 383 of /Users/bc/tripal/sites/all/modules/tripal/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc).
```

## resolution
At first glance, looks easy to resolve the warning by adding this to the web services class just prior to here: 

https://github.com/tripal/tripal/blob/bfca9765a5ef7aece7b4355c886d91513a89cadb/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc#L383

```php
  if (empty($values)){
        return;
      }

      // If the value is an array and this is the field page then all of those
      // key/value pairs should be added directly to the response.
      if (is_array($values[0])) {
```

adding this also causes the ws to work how i'd expect: it isnt returned since we didnt have the permission enabled for an anonymous user.

probably also need to add a case for where cardinality != 1....  what do you think?  This would make the ws behave how i expect (field isnt shown on web services), and it resolves the php warning.

bradfordcondon:
@bradfordcondon will make the pr and explicitly ask @spficklin

bradfordcondon:
doing as i suggest above resolves the warning, but the content still appears...

https://github.com/tripal/tripal/tree/749_webservices_field_permission

ah ok, but there is a field_permission key.  it has a tpe value, but i think just retruning if the key is set at all is good enough?

spficklin:
I think this is now resolved with the PR.  Closing but if not please reopen.

bradfordcondon:
I thought of an edge case i want to test before i close this issue.

if field permissions are set for editing, but anonymous users are allowed to view the field, will it show up in web services?

bradfordcondon:
unfortunately i've confirmed my fix was too simplistic.
if we set custom permissions allowing admins to set the value, but anyone to view it....
![screen shot 2018-12-18 at 5 42 29 pm](https://user-images.githubusercontent.com/7063154/50187540-5437ef00-02ec-11e9-8050-3003d358977b.png)


The field will appear on the content page, but not on web services: we exclude it because the field permission is 2.

[...]

PR #794, opened 2018-12-14
749 webservices field permission

bradfordcondon:
# Bug fix

Issue #749 

## Description
I added a guide for using field-specific permissions, to hide some fields from certain user types on input and on display.
https://tripal.readthedocs.io/en/latest/user_guide/content_types/field_permissions.html

However web services still displayed content that was hidden.  This PR should resolve that.

## Testing?

Follow the guide to hide a particular field (uncheck all boxes except for admin). You can visit the page as anon/admin to verify the field is hidden to anonymous users.  Then visit that content with web services: the field should display.

If you then checkout this branch and view via web services, th field should no longer display.

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -380,7 +380,7 @@ class TripalContentService_v0_1 extends TripalWebService {
         // TODO: handle this case.
       }
     }
-    if (isset($field['field_permissions']['type'])) {
+    if (isset($field['field_permissions']['type']) && $field['field_permissions']['type'] != 0) {
       return;
     }
 
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -44,9 +44,9 @@ class TripalContentService_v0_1 extends TripalWebService {
   public function handleRequest() {
 
     // Get the content type.
-    $ctype = (count($this->path) > 0) ? $this->path[0] : '';
+    $ctype     = (count($this->path) > 0) ? $this->path[0] : '';
     $entity_id = (count($this->path) > 1) ? $this->path[1] : '';
-    $expfield = (count($this->path) > 2) ? $this->path[2] : '';
+    $expfield  = (count($this->path) > 2) ? $this->path[2] : '';
 
     // is this a valid content type?
     if ($ctype) {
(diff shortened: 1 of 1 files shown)


bradfordcondon:
dangit, i ran this with the linter apparently.  

the only actually code change is:

```php

 if (isset($field['field_permissions']['type']) && $field['field_permissions']['type'] != 0) {
      return;
    }

    if ($hide_fields == TRUE and empty($values[0])) {
      return;
    }



    // If the field cardinality is 1.
    if ($field['cardinality'] == 1) {

```

want me to redo the pr @spficklin ?

tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
```diff
@@ -370,12 +380,17 @@ class TripalContentService_v0_1 extends TripalWebService {
         // TODO: handle this case.
       }
     }
+    if (isset($field['field_permissions']['type']) && $field['field_permissions']['type'] != 0) {
```
bradfordcondon, line 374: the only code change is right here!

tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
```diff
@@ -90,15 +90,19 @@ class TripalContentService_v0_1 extends TripalWebService {
...
-    else if ($ctype and $entity_id and !$expfield) {
-      $this->doEntity($ctype, $entity_id);
-    }
-    else if ($ctype and $entity_id and $expfield) {
-      $this->doExpandedField($ctype, $entity_id, $expfield);
-    }
-    // Otherwise just list all of the available content types.
     else {
-      $this->doContentTypesList();
+      if ($ctype and $entity_id and !$expfield) {
```
spficklin: Just curious why this if statement was changed?

bradfordcondon: its an automated style change.

the original:

```
else if
else if
else
```

the change:

```
else 
if 
else 
if 
else

```

If youre concerned about the style changes I can remake the PR with just the inserted line.  Its hard to remember NOT to run my formatter when changing files...

bradfordcondon: As you can see its a recommendation of the drupal coder code sniffer

screen shot 2018-12-17 at 11 45 48 am


although `elsif` seems to be fine, so seems to me like the change here was more extensive than it had to be.

spficklin: If I'm reading this right, it looks like the else if's from the two lines above got merged into a single else.... I'm not sure that's the same logic....

bradfordcondon: look at lines 95-105

```php
    else {
      if ($ctype and $entity_id and !$expfield) {
     $this->doEntity($ctype, $entity_id);
      }
      else {
        if ($ctype and $entity_id and $expfield) {
          $this->doExpandedField($ctype, $entity_id, $expfield);
        }
        // Otherwise just list all of the available content types.
        else {
          $this->doContentTypesList();
        }
      }
```

youre right the elsifs got merged but they are branched in that else within if statements.  It looks like the same logic to me.

spficklin: Just to satisfy my paranoia can we put it back the way it was?  I don't have time to test that this reworking of those if statements is the same.  I can test your permissions fix easy enough.

bradfordcondon: no problem.  but if its alright with you im going to revert the whole file: that way ill be more confident im not changing hte logic :)

spficklin: Yep sounds good.

spficklin (approved):

spficklin:
Merged!  Thanks @bradfordcondon

Issue #763, opened 2018-11-22
Issue with tripal_ds_update_ds_layout function 


mboudet:
### System information

Tripal: 7.x-3.0
Drupal : 7.59
Postgres: 9.5.8
Php: 7.1.23

### Issue description
I'm trying to setup fields and fields group programmatically (As in, enable fields, set them up in the field group, and place them).
The tripal_ds_update_ds_layout function should take some field groups and fields, enable them and set them up. However, nothing happen when the function is used.

The tripal_ds_create_field function use this function, so the same issue should occurs (meaning, the panel is created but not set up)

### Issue details

The issue is the [drupal_write_record](https://github.com/tripal/tripal/blob/3606ca5cf8891a068c0dc5ebe6ac48f245252136/tripal_ds/tripal_ds.module#L435)

We feed layout_info to the drupal_write_record, but the layout_info is just an array, it does not refer to any element of the table, nor a field.

Either layout_info should be a full-on object (like [here](https://github.com/tripal/tripal/blob/7f35a59c5e48bd173d06da51e28d8abbfebc8594/tripal_ds/includes/tripal_ds.field_group.inc#L10)), or we should use db_update to update the select field..

Something like 
```
db_update('ds_layout_settings')
        ->fields(array(
            'settings' => serialize($layout_info),
        ))
        ->condition('ds_layout_settings.id', $ds_identifier, '=')
        ->execute();
```

Also, this function does not check if the field/field group is already in the layout.. it might be good to add a check.

#### Steps to reproduce
Use the tripal_ds_update_ds_layout function. Nothing happen (no change in DB, no error).

#### Others

For now, my workflow to create a field group, add a field to it, and publicize it is as follows : 

1. Create a field group with tripal_ds_create_field
2. Edit the field group in DB to add the field as children
3. Use tripal_ds_update_ds_layout to add both to the layout.

Is there any simpler way? I feel like the tripal_ds_create_field function should have an optional argument to customize the created panel (at least, to add children or parents to it)

bradfordcondon:
>I'm trying to setup fields and fields group programmatically

Have you checked out the Drupal features module?  It lets you do exactly this.  Granted, its for exporting and sharing your field configuration, so if you must define the groupings programmatically to begin with it won't help you.
  I wrote a full tutorial:

https://tripal.readthedocs.io/en/latest/dev_guide/exporting_field_settings.html

spficklin:
I believe this is now resolved given the merged pull request.

PR #795, opened 2018-12-17
Fix tripal_ds_update_ds_layout in tripal_ds.module

mboudet:
# Bux Fix

Issue #763 

## Description
Current implementation was non-functional (Update in DB was not working).

## Additional Notes (if any):
It might be good to implement a check to avoid duplicate fields. (There might be a check beforehand, but it wouldn't hurt to do it again?)

### DIFF
--- a/tripal_ds/tripal_ds.module
+++ b/tripal_ds/tripal_ds.module
@@ -432,7 +432,12 @@ function tripal_ds_update_ds_layout($bundle_name, $field_name, $tripal_pane_name
       }
     }
     //Update the ds_layout_settings table with the new layout info.
-    drupal_write_record('ds_layout_settings', $layout_info);
+    db_update('ds_layout_settings')
+        ->fields(array(
+            'settings' => serialize($layout_info),
+        ))
+        ->condition('ds_layout_settings.id', $ds_identifier, '=')
+        ->execute();
   }
 }
 /*


shawnawsu:
This works as expected now. Thanks for the fix. 

When you say check for duplicate fields, do you mean adding the same field more than once or trying to create two fields with the same chado references but different names?

mboudet:
Adding the same field (or panel) twice.
This part 
```
    if(!empty($field_name)){
      $layout_info['regions']['right'][$index] = $field_name;
      $incremented_index = $index++;
      $layout_info['fields'][$field_name] = 'right';
    }
    if(!empty($tripal_pane_name)){
      if(!empty($incremented_index)){
        $layout_info['regions']['right'][$incremented_index] = $tripal_pane_name;
        $layout_info['fields'][$tripal_pane_name] = 'right';
      }
      else {
        $layout_info['regions']['right'][$index] = $tripal_pane_name;
        $layout_info['fields'][$tripal_pane_name] = 'right';
}
```
will add $tripal_pane_name and $field_name to $layout_info['regions']['right'] even if they are already in it. I don't know how Tripal would react to that though. There might be a check when doing it with the UI.

spficklin (approved):

spficklin:
Given @shawnawsu  positive review and my code review looks good!  Thanks @mboudet for the fix.

Issue #790, opened 2018-12-13
[Tripal3]Aliases token not filled (empty)


mboudet:
### System information

System info:
Tripal: 7.x-3.0
Drupal : 7.59
Postgres: 9.5.8
Php: 7.1.23

### Issue description
Hi,
I noticed that some of my automatically created aliases are malformed (missing some tokens).
This occurs for the mRNA entity type.
My alias pattern is as follows : features/[obi__organism,TAXRANK:0000005]/[obi__organism,TAXRANK:0000006]/[rdfs__type]/[data__identifier]

An expected alias would be : features/Citrus/sinensis/mRNA/PAC:18136223

But the created alias looks like this : features///mRNA/
(So there are duplicate errors)

#### Steps to reproduce
This one is complicated, because the problem only occurs if the token fields are not in cache. Depending on your setup, you will not see it occurs.

#### Issues

Looking at the code, there are actually two separate issues:
1) **In tripal.entities.api.inc**
[Here](https://github.com/tripal/tripal/blob/3606ca5cf8891a068c0dc5ebe6ac48f245252136/tripal/api/tripal.entities.api.inc#L1331)

We use the $field_ids variable to monitor the fields missing from cache, however the variable is reset in the first loop:

```
foreach ($attach_fields as $storage_type => $details) {
      $storage = $details['storage'];
      $fields = $details['fields'];
      $field_ids = array();
            foreach ($fields as $field) {
        $field_ids[$field['id']] = array($entity->id);
      }
```
So only the last loop actually keep the values. Meaning, if there are more than one storage type for the fields, information is lost. (In my case, the [rdfs__type] token storage is tripal_no_storage, and the other tokens are field_chado_storage.
This is easily fixed by moving the $field_ids = array(); before the first loop.

2) **In tripal_field_storage_load**

This one is a bit more complicated, but basically, we ignore any fields which are not TripalField class. (like text..)

If we compare the hook_field_storage_load implementation in tripal and tripal_chado, we can see that the tripal implementation completely ignore non-TripalField tokens.

```
tripal_load_include_field_class($field_type);
      if (class_exists($field_type)) {
        $tfield = new $field_type($field, $instance);
        $tfield->load($entity);
      }

```
While tripal_chado implements something to take care of it
```
tripal_load_include_field_class($field_type);
        if (class_exists($field_type) and is_subclass_of($field_type, 'TripalField')) {
          $tfield = new $field_type($field, $instance);
[...]

PR #796, opened 2018-12-17
Fix issue with aliases token not filled (#790)

mboudet:
# Bux Fix

Issue #790 

## Description
1) Put the $field_ids definition outside the loop to avoid a reset.

2) Fixed token management for text fields (based on implementation in tripal_chado_field_storage_load)

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -1324,10 +1324,10 @@ function tripal_replace_entity_tokens($string, &$entity, $bundle_entity = NULL) 
 
   // If we have any fields that need attaching, then do so now.
   if (count(array_keys($attach_fields)) > 0) {
+    $field_ids = array();
     foreach ($attach_fields as $storage_type => $details) {
       $storage = $details['storage'];
       $fields = $details['fields'];
-      $field_ids = array();
       foreach ($fields as $field) {
         $field_ids[$field['id']] = array($entity->id);
       }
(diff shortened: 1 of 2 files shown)


mboudet:
The travis fails with an out of memory error, but I'm not exactly sure why the changes would affect that..

bradfordcondon:
the out of memory error is a problem with our travis setup/the OBO loader: its a consistent bug right now where 7.0 passes and 7.1 doesn't: please ignore it for now.

bradfordcondon:
hi @mboudet thank you very much for the contribution.  I'd like @spficklin to weigh in before i pass judgement, but my hunch is that using the Chado implementation isn't correct, as base Tripal entities wont have the keys you added (chado_record, chado_table ,etc).

mboudet:
You might be right, I only checked for the mRNA entity type, so it might bug out for others.
There might be an easier way to access the field value. 

Some tokens also seems to crash out (like the [data__image].. though I do not know why someone would want to use that, nor what the url would look like) when calling field_info_field

mboudet:
While i'm doing some tests, we should probably add a check for the existence of $element in the $value array in _tripal_replace_entity_tokens_for_elements(). We just assume there is a value associated with the element, but it might not be the case.

(Example, the default page title format is [taxrank__genus] [taxrank__species] [taxrank__infraspecific_taxon,TAXRANK:0000045], but the infraspecific_name might not be set..)

spficklin:
Hi @mboudet,

Yes @bradfordcondon is correct there should be no code in the tripal module that pulls data from Chado.  Anything that pulls data from Chado to populate fields should be in the tripal_chado module.  On the issue you posted #790 I approve of the fix for the first problem. If you want to cancel out this PR and submit that fix by itself we can get that merged in.  For the second issue I think we need to better understand why you're not getting values and that conversation. I'll post a follow-up quesion there.

mboudet:
That's fine, I will submit the PR again. Thanks.

PR #798, opened 2018-12-17
Pr template typo

bradfordcondon:
Fix the PR template to replace Bux fix with Bug fix.  

Probably not worth its own PR but it was getting on my nerves!


bradfordcondon:
this requires no review as it's simply a typo fix on the PR . template.

PR #799, opened 2018-12-18
Fix issue with aliases token not filled

mboudet:
# Bug Fix

Issue #790 

## Description
Fix part 1 of the issue.
Moved the variable definition to avoid reset when setting tokens


spficklin (approved):

spficklin:
Thanks @mboudet for the fix!

PR #802, opened 2018-12-21
Type Hinting problems with PHP 5.6: WSOD

laceysanderson:
# Bug Fix

## Description
PHP 5.6 only supports type hinting for objects and arrays. However there were a couple of places where it was used for int and bool, which caused WSOD on PHP 5.6 sites. Although we are moving on from supporting 5.6 since PHP will soon cease support, we don't want site maintainers struggling to upgrade to be met with a WSOD.

Since this PR only changes type hinting in function parameters, it should only require a single review.

## Testing?

Pull this branch and refresh your site. You should not see a WSOD ;-P


bradfordcondon:
I approve (non functional review).  Removing type hints wont have any negative impact though, and if it keeps us 5.6 compatible while our users need it im all for it.

bradfordcondon (approved):

spficklin (approved):

Issue #805, opened 2018-12-28
Fields disappearing even when not empty


spficklin:
### System information

* Tripal Version: 3.x
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: N/A

### Issue description

When hiding of empty fields for a content type is turned on, then fields that are marked to not auto attach (meaning they should be loaded via AJAX) are automatically removed from the Tripal Pane table.  This is because they are technically "empty" as data has not yet been added to them (it gets added via AJAX).  This is incorrect behavior.  They should not be removed from the table until after their data is loaded.  The `tripal_field_group_table_rows_alter` function is the culprit.

bradfordcondon:
#806 was merged so closing this as resolved.

PR #806, opened 2018-12-28
Fix for disappearing fields in tables that are ajax loaded.

spficklin:
# Bug Fix

Issue #805 

## Description

When hiding of empty fields for a content type is turned on, then fields that are marked to not auto attach (meaning they should be loaded via AJAX) are automatically removed from the Tripal Pane table. This is because they are technically "empty" as data has not yet been added to them (it gets added via AJAX). This is incorrect behavior. They should not be removed from the table until after their data is loaded. The tripal_field_group_table_rows_alter function is the culprit.

## Testing?

To test that there is a problem
1.  Do not pull this branch just yet
2.  Make sure you have checked from the mRNA content type the hide empty fields and the load via ajax settings.
3. Go to any mRNA page that has a sequence.  I use the Citrus sinensis data from the Tripal tutorial.  
4. Click the 'Reload' tab and the page should reload.  The mRNA sequence is missing!
5. Turn off the 'hide empty fields' setting for the mRNA content type, return to the mRNA page and then click the 'Reload' tab. You should now see the sequence.

To test the fix:
1.  Pull the branch
2.  Perform steps 1-4 above.  This time however, the mRNA sequence is present.

### DIFF
--- a/tripal/includes/TripalEntityController.inc
+++ b/tripal/includes/TripalEntityController.inc
@@ -101,7 +101,7 @@ class TripalEntityController extends EntityAPIController {
 
     return TRUE;
   }
-  
+
   /**
    * Sets the title for an entity.
    *
@@ -649,7 +649,7 @@ class TripalEntityController extends EntityAPIController {
 
           // Add this field to the entity with default value.
           if (!isset($queried_entities[$id]->{$field_name})) {
-            $queried_entities[$id]->{$field_name} = array();
+            $queried_entities[$id]->{$field_name} = [];
           }
 
           // Options used for the field_attach_load function.
@@ -688,14 +688,19 @@ class TripalEntityController extends EntityAPIController {
           // attach then we will ignore it. It can only be set by providing
           // the id in the $field_id array handled previously.
           else {
-            
+
+            // Does the field instance have an 'auto_attach' setting?
+            $auto_attach = FALSE;
+            if (array_key_exists('settings', $instance) and
+                array_key_exists('auto_attach', $instance['settings']) and
+                $instance['settings']['auto_attach'] == TRUE) {
+              $auto_attach = TRUE;
+            }
+
             // Do not load fields that are not auto attached.  Instead set
             // their value to an empty string and set the #processed key to
             // FALSE.
-            if (array_key_exists('settings', $instance) and
-                array_key_exists('auto_attach', $instance['settings']) and
-                $instance['settings']['auto_attach'] == FALSE) {
-
+            if (!$auto_attach) {
                // Add an empty value. This will allow the tripal_entity_view()
                // hook to add the necessary prefixes to the field for ajax
                // loading.
--- a/tripal/includes/tripal.user.inc
+++ b/tripal/includes/tripal.user.inc
@@ -3,7 +3,7 @@
 
 function tripal_user_get_files($uid){
   $directory = [];
-  
+
   // Get all of the files that have been uploaded by the user.
   $sql = "
      SELECT FM.fid, FM.filename, FM.uri, FM.uid, FM.filesize, TGEF.expiration_date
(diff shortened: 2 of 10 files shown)


tripal/includes/TripalEntityController.inc
```diff
@@ -688,14 +688,19 @@ class TripalEntityController extends EntityAPIController {
           // attach then we will ignore it. It can only be set by providing
           // the id in the $field_id array handled previously.
           else {
-            
-            // Do not load fields that are not auto attached.  Instead set
-            // their value to an empty string and set the #processed key to
-            // FALSE.
+
+            // Does the field instance have an 'auto_attach' setting?
+            $auto_attach = FALSE;
```
spficklin, line 693: Just a note... I changed these lines of code to hopefully prevent problems in the future if a developer forgets to set the 'auto_attach'  we need it to have a default.

tripal/includes/tripal.user.inc
```diff
@@ -3,47 +3,47 @@
...
+
   // Get all of the files that have been uploaded by the user.
   $sql = "
      SELECT FM.fid, FM.filename, FM.uri, FM.uid, FM.filesize, TGEF.expiration_date
      FROM {file_managed} FM
        INNER JOIN {file_usage} FU on FM.fid = FU.fid and FM.uid = :user_id
        LEFT JOIN {tripal_expiration_files} TGEF on TGEF.fid = FU.fid
      WHERE FU.module = 'tripal'
-     ORDER BY FM.filename
+     ORDER BY FM.uri
```
spficklin, line 14: Whoops. I didn't mean for this fix to make it in to this PR.  But it's so simple I'll leave it.  It just fixes a sorting issue with the user's uploaded file list.

tripal/tripal.module
```diff
@@ -1565,18 +1551,18 @@ function tripal_field_group_table_rows_alter(&$element, &$children) {
       $bundle = $element[$child]['#bundle'];
       $bundle_info = tripal_load_bundle_entity(array('name' => $bundle));
 
+      $processed = FALSE;
+      if (array_key_exists('#processed', $element[$child]['#object']->$child)) {
+        $processed = $element[$child]['#object']->$child['#processed'];
+      }
+
```
spficklin, line 1558: This set of code is the beginning of the true fix for the problem.

spficklin:
Sorry for all of the formatting changes in this PR.  I turned on automatic removal of white space in my IDE.  Also, I fixed a few other field formatters (e.g. publication) that weren't properly dealing with empty values.  They don't affect this bug, but it took me a while to track down the problem and I wound up fixing these things as well.

laceysanderson:
I'm having difficulties reproducing this problem...

## On 7.x-3.x (f543712afd3b21860d029fbf3d9dc69c112ede76)
 - [x] Make sure you have checked from the mRNA content type the hide empty fields and the load via ajax settings.
 - [x] Go to any mRNA page that has a sequence. **I created a new mRNA through the UI with dummy nucleotide sequence**
- [ ] Click the 'Reload' tab and the page should reload. **I see the sequence, even after reloading and clearing the full cache multiple times**
- [x] Turn off the 'hide empty fields' setting for the mRNA content type, return to the mRNA page and then click the 'Reload' tab. You should now see the sequence.
screen shot 2018-12-31 at 10 37 02 am

spficklin:
@laceysanderson it may be a timing issue and the order that the `tripal_field_group_table_rows_alter` function runs and when the load function gets called for the field.    If so you might not be able to reproduce the problem.

bradfordcondon:
I also cant reproduce the problem.

bradfordcondon:
on master with hide empty fields

screen shot 2019-01-03 at 8 46 38 am

on your branch with hide empty fields:

screen shot 2019-01-03 at 8 47 42 am


So im able to reproduce the problem and confirm the fix resolves it.  approved.

bradfordcondon (approved):

laceysanderson (approved):
Did a code review and it all looks reasonable and makes sense 👍

laceysanderson:
Since we have two reviews including one functional review, I'm merging this! Thanks for the fix @spficklin

spficklin:
Thanks for the reviews!

PR #808, opened 2018-12-30
Bug in update function 7336

spficklin:
# Bug Fix

Issue # N/A

## Description

The update function `tripal_chado_update_7336` fails if I don't have any contact linker field instances for my bundles.  This is just a simple fix to make sure the instance exists before trying to update it.  
## Testing?

This changes just wraps an if statement around the update.  I don't think it needs a functional test. Just one code review.

### DIFF
--- a/tripal_chado/tripal_chado.install
+++ b/tripal_chado/tripal_chado.install
@@ -1846,11 +1846,12 @@ function tripal_chado_update_7336() {
       $field_name = $base . '_contact';
       $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);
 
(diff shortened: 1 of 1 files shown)


bradfordcondon:
oops thanks for the catch.  Out of curiosity when does this case arise?  If you delete the default bundles that use the contact linker table?

bradfordcondon (approved):
looks good, thank you

spficklin:
@bradfordcondon I believe I manually removed the contact fields because they were causing problems.  I suppose it's a fringe use case and probably wouldn't happen to most people.

spficklin:
Thanks for the review @bradfordcondon

Issue #762, opened 2018-11-21
Relationship Widget Minor Fixes


spficklin:
### System information

* Tripal Version:  3.x
* Drupal Version:  N/A
* PostgreSQL Version  N/A:
* PHP Version: N/A

### Issue description

After @laceysanderson did a wonderful job of fixing the relationship widget, there are a few minor outstanding items that need fixing on the widget.  I've moved them to this new issue from the discussion on PR #726.

1.  When adding a new entity for a content type that has relationships it's not clear how to refer to the record being added such that it can be used in a relationship. It's only possible to set relationships for entities that already exist and editing them.
> Its own name wont fill out the autocomplete box. I can infer that it needs to be genus species based on the autocomplete... Maybe the field instance setting description should be dynamic and describe what the name field is going to consist of? Note that despite the warning, the form does submit and correctly create the relationship. way to go.

> Maybe a more intuitive way to do this is to use form states to hide the subject/object boxes. Then you start with a radio button that says "this record is the subject" and "this record is the object", and only display the other box for looking up the "other" record.

~~2.  If a relationship type (e.g. 'part_of') has two words rather than one then the widget gives a validation error.~~  (see new issue here #801 BC)[original screenshot](https://user-images.githubusercontent.com/1719352/48812484-6ae82780-ece7-11e8-94c7-80bafec84140.png)
3.  When creating a relationship for the project content type the word "project" seems to have an odd pattern for bolded words. [original screenshot](https://user-images.githubusercontent.com/1719352/48812563-c0bccf80-ece7-11e8-9a2f-42063333b675.png)
4. Unable to remove a relationship.
> Its not possible to remove a relationship with this widget. I can manually unset everything, however, the field then fails validation.

5. Formatter for shows "feature" rather then the content type (e.g. "Genetic Marker"). [original screenshot](https://user-images.githubusercontent.com/7063154/48785041-e929d700-ecb1-11e8-8b52-594e839cbf2e.png)
6. On some theme's, the widget expands out of the box. [original screenshot](https://user-images.githubusercontent.com/7063154/48778007-20908780-eca2-11e8-9878-45dd643cbe14.png)

bradfordcondon:
>If a relationship type (e.g. 'part_of') has two words rather than one then the widget gives a validation error. original screenshot

I can use the term `part_of` no problem.

[...]

PR #809, opened 2018-12-30
Relationship widget fixes

laceysanderson:
# Bug Fix

Issue #762 

## Description
As described in Issue #762 there were a couple of issues with the relationship widget. This PR fixes some of them (listed in testing).

### Unable to remove existing relationship
1. Create Tripal Content with a relationship to an existing record.
2. Edit the page and empty the subject, object and type textfields.
3. Click Save
On this branch: the relationship should be removed with no errors.
On 7.x-3.x: you would be told the object/subject cannot be empty. 

### Weird Bolding of the word "project" on project view.
1. Create Tripal Content of type "project" with a name including the word "project". Also relate this content to an existing project and click save.
On this branch: there should be no highlighting of the word "project"
On the 7.x-3.x branch: the word project in the name would also be bolded.

NOTE: I simply removed the bolding. This bug is caused by the fact the field creates the statement and then the formatter added the bolding. This meant it had to use find/replace to do so which had the unintended consequence of bolding that word if it appeared in the content name as well.

### Relationship formatter has the word "feature" rather then "gene" in statement
1. Create Tripal content with relationships for any content type based on the feature table.
On this branch: you should see the cvterm.name of the type in the statement (e.g. gene)
On 7.x-3.x: you would see the name of the table in the statement (e.g. feature)

NOTE: If you re-label your content type, you won't see that label here but rather the cvterm.name... This is still not ideal but I didn't want to add the additional loading time to retrieve the entity label.

### DIFF
--- a/tripal_chado/tripal_chado.install
+++ b/tripal_chado/tripal_chado.install
@@ -1846,11 +1846,12 @@ function tripal_chado_update_7336() {
       $field_name = $base . '_contact';
       $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);
 
-      $instance_info['type'] = 'chado_linker__contact';
-      $instance_info['widget']['type'] = 'chado_linker__contact_widget';
-      $instance_info['formatter']['type'] = 'chado_linker__contact_widget';
-      field_update_instance($instance_info);
-      
+      if ($instance_info) {
+        $instance_info['type'] = 'chado_linker__contact';
+        $instance_info['widget']['type'] = 'chado_linker__contact_widget';
+        $instance_info['formatter']['type'] = 'chado_linker__contact_widget';
+        field_update_instance($instance_info);
+      }
     }
   }
 }
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
@@ -438,7 +438,7 @@ class sbo__relationship extends ChadoField {
     // The above relationship will be shown both on marker and variant pages
     // and as such both subject and object names need to be shown.
     $clause = 'The ' . $subject_type . ', ' .
-        $subject_name . ', ' . $verb . ' '  . $rel_type_clean . ' the '  .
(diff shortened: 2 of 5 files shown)


spficklin (approved):

spficklin:
This all worked for me!  Thanks @laceysanderson. I did a full functional review.  Although I encountered two issues not related that I fixed.

1.  I could not add a new relationship.  I'm not sure how this happened because we had all this working previously.  But it seems in the tripal storage API was including the pkey when checking for duplicates. I don't know what was up with that. But I added a quick band-aid fix to get around it.  
2.  I adjusted the written "clause" part a bit to add support 'is_a' as a type.
3.  My IDE automatically removed blank spaces at the end of lines, so unfortunatly there are quire a few updated lines where those were removed.

I approve of the PR.

laceysanderson:
Thanks for the review @spficklin and for your fixes! I functionally tested your fixes and all works for me :-)

tripal_chado/tripal_chado.install
```diff
@@ -1846,11 +1846,12 @@ function tripal_chado_update_7336() {
       $field_name = $base . '_contact';
       $instance_info = field_info_instance('TripalEntity', $field_name, $bundle_name);
 
-      $instance_info['type'] = 'chado_linker__contact';
-      $instance_info['widget']['type'] = 'chado_linker__contact_widget';
-      $instance_info['formatter']['type'] = 'chado_linker__contact_widget';
-      field_update_instance($instance_info);
-      
+      if ($instance_info) {
```
laceysanderson, line 1849: Was this accidentally added to this PR @spficklin?

bradfordcondon: looks like a commit that has already been merged.  you can probably ignore

https://github.com/tripal/tripal/blob/f543712afd3b21860d029fbf3d9dc69c112ede76/tripal_chado/tripal_chado.install#L1849-L1854

bradfordcondon:
looks sane to me, 👍 w/ no functional testing

laceysanderson:
Merging since there are two approvals :-) Thank you both!

Issue #813, opened 2019-01-03
Unable to create cvterm-based content type


laceysanderson:
### System information

* Tripal Version: 7.x-3.x (668b07ce27c26f0481fa8887005608a9b2fcc67b)
* Drupal Version: 7.61

### Issue description
I get a WSOD when I trying to create a content type where the base table is cvterm. This seems to be caused by a missing function to generate the options for a cvterm content type.

#### Steps to reproduce
1. Go to Admin > Structure > Tripal Content Types > Add Tripal Content Type
2. Choose a term, storage method is chado
3. Select 'cvterm' as the base table and click continue.

#### Error messages and screenshots

```
PHP Fatal error:  Call to undefined function tripal_chado_field_storage_bundle_mapping_form_add_cvterm() in /var/www/sites/all/modules/tripal/tripal_chado/includes/tripal_chado.field_storage.inc on line 1218
```

bradfordcondon:
https://github.com/tripal/tripal/blob/668b07ce27c26f0481fa8887005608a9b2fcc67b/tripal_chado/includes/tripal_chado.field_storage.inc#L1211-L1228

i searched for similar function names, im guessing it was stubbed and never defined?

laceysanderson:
It was defined a long time ago. I had to go back to before 7.x-3.0 to find it.
I started work on branch 813-cvterm_entitytype

laceysanderson:
@spficklin I seem to be missing something... 

I can now fully submit the create content type form for a cvterm-based type. However, when running the job, I get:
```
2019-01-03 06:38:21: Calling: tripal_create_bundle(Array)
WD tripal_entities: Failed to create content type: Cannot create content type. Missing the "data_table" for this bundle..                                                                        [error]
ERROR (TRIPAL_ENTITIES): Failed to create content type: Cannot create content type. Missing the "data_table" for this bundle..
[site http://default] [TRIPAL ERROR] [TRIPAL_ENTITIES] Failed to create content type: Cannot create content type. Missing the "data_table" for this bundle..
Failed to create content type: <em class="placeholder">Cannot create content type. Missing the &quot;data_table&quot; for this bundle.</em>.
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] Failed to create content type: Cannot create content type. Missing the "data_table" for this bundle..
Failed to create content type: Cannot create content type. Missing the "data_table" for this bundle..  
```

[...]

PR #814, opened 2019-01-03
 Unable to create cvterm-based content type

laceysanderson:
# Bug Fix

Issue #813 

## Description
Originally I got a WSOD when I trying to create a content type where the base table is cvterm. This was caused by a missing function (`tripal_chado_field_storage_bundle_mapping_form_add_cvterm`) which generates the options for a cvterm content type. Additionally, cvterm logic had been removed from the whole process. This PR brings back this functionality.

## Testing?

Attempt to create as content type where the base table is cvterm and...
### All records in a cv become pages of the selected type.
1. Go to Admin > Structure > Tripal Content Types > Add Tripal Content Type
2. **Step 1:** choose any term (e.g. "Trait")
3. **Step 2:** storage method is chado
4. **Step 3:** 
        - chado table: cvterm; 
        - Select "All records that belong to a single controlled vocabulary "
        - Select any cv with terms.
5. Click "Create content type"

On 7.x-3.x you would see a blank screen after clicking continue with "cvterm" chosen as the chado table. On this branch you will be able to create the content type.

## Known Limitations
Choosing "All child records of the specified term" is still not supported as I do not know the storage arguments needed for this. @spficklin do you know what they should be?

### DIFF
--- a/tripal_chado/includes/tripal_chado.field_storage.inc
+++ b/tripal_chado/includes/tripal_chado.field_storage.inc
@@ -1148,6 +1148,11 @@ function tripal_chado_field_storage_bundle_mapping_form_summary(&$form, &$form_s
     ];
   }
 
+  //
+  // Add the row for cvterm options
+  //
+  // @todo @lacey
+
   $table = array(
     'header' => $headers,
     'rows' => $rows,
@@ -1808,8 +1813,13 @@ function tripal_chado_field_storage_bundle_mapping_form_validate($form, &$form_s
   if ($clicked_button == 'chado-cvterm-settings-continue') {
     // If the user chose to differentiate type based on cv_id and supplied a cv_id
     // then we're done!
-    if (($form_state['values']['chado_type_use_cv'] == 'cv') AND ($form_state['values']['chado_type_cv_id'] > 0)) {
-      $form_state['chado-stage'] = 'complete';
+    if ($form_state['values']['chado_type_use_cv'] == 'cv') {
+      if ($form_state['values']['chado_type_cv_id'] > 0) {
+        $form_state['chado-stage'] = 'complete';
+      }
+      else {
+        form_set_error('chado_type_cv_id', 'Please select a controlled vocabulary which contains records for this content type.');
+      }
     }
 
     // If the user chose to diffentiate type based on the current term,
@@ -1883,7 +1893,8 @@ function tripal_chado_field_storage_bundle_mapping_form_submit($form,
   // If the user indicated they want to use a cv_id to control which cvterm
   // records become pages then we'll set the stroage args and return.
   elseif ($default['use_cvterm'] == 'cv' AND $default['cv_id']) {
-    $storage_args['data_table'] = $default['table'];
+    // @debug dpm($form_state, 'form state');
+    $storage_args['data_table'] = 'cvterm';
     $storage_args['type_id'] = $form_state['values']['selected_cvterm_id'];
     $storage_args['type_value'] = $default['cv_id'];
   }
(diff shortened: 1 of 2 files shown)


bradfordcondon (commented):
functional test: I was able to create a bundle for a CV as specified, and create a new term using that bundle with no errors.

tripal_chado/includes/tripal_chado.field_storage.inc
```diff
@@ -1139,6 +1156,38 @@ function tripal_chado_field_storage_bundle_mapping_form_summary(&$form, &$form_s
     ];
   }
 
+  //
+  // Add the row for cvterm options
+  //
+  if ($default['use_cvterm']) {
+    $message = 'No. All records belong to a single controlled vocabulary.';
+    if ($default['use_cvterm'] == 'parent') {
+      $message = 'Yes. Records should include all child records of the specified term.';
```
bradfordcondon, line 1165: I think I see what direction this is headed in, but does it really make sense to include?  Is the use case something like having a "branch" of a CV dedicated to user-submitted terms?  Is that really the right solution as opposed to a user-specific CV (which would be the other case)?

If theres not a good use for this alternate schema i propose dropping.

bradfordcondon (approved):

tripal_chado/includes/tripal_chado.field_storage.inc
laceysanderson: I'm good with dropping this option as I don't actually need it... @spficklin since you originally added it, do you think it still needs to stay?

spficklin: Sorry, I don't understand the question....

spficklin: The highlighted code is just printing summary info, so I'm not making the connection with the question.

bradfordcondon: We are discussing the "All child records of the specified term" option.

screen shot 2019-01-04 at 6 42 07 pm

see also the "known limitations" in lacey's PR message

spficklin: Oh, I see. Yes, we can get rid of it. It doesn't work so no point in offering the option when it doesn't work.

spficklin: I'll apply a simple fix to remove that option.

spficklin:
Creation of the trait content type worked but now my site won't load and Apache consumes all memory when trying to load the page after creating the content type.  I can't get my site to load. I'm investigating what's wrong.

spficklin:
@laceysanderson as I mentioned above after I created the 'Trait' content type my site completely overran the memory.  It took some time but I traced it down to  when fields of properties are added to the content type.  We didn't have the correct use case for finding properties so what it ended up doing was finding all distinct property types in the `cvtermprop` table.  I have lots of ontologies loaded on my site after all of that OBO work I did and that created the problem for me.    I've fixed this by adding code that specifically handles adding properties when the `cvtermprop` table is the property table.   

So, it works just fine now for me.  I approve of the change, although for safety sake you may want to just check the code one more time.

spficklin:
Oh, I also removed that second option of using a parent term.  There's now only one option for using the cvterm table for content types.

laceysanderson:
I confirmed the test failure was due to memory exhausted on the obo test:

> ....Building cvterm paths for vocabulary: tripal_obo_test
> Clearing the cvtermpath table for this vocabulary...
> Clearing completed.
> Retrieving relationships...
> Processing terms...
> .PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 12288 bytes) in /home/travis/build/tripal/drupal/includes/database/database.inc on line 2227

I also confirmed @spficklin (thanks btw!) added fixes and retested the functionality 👍 

Since there are two approvals and I've confirmed the test failure is not the fault of this PR, I'm merging!!!

Thanks for the reviews/suggestions/fixes @bradfordcondon and @spficklin!!!

Issue #815, opened 2019-01-03
OBO importer tests run out of memory at > 2GB in Travis


bradfordcondon:
The travis build is inconsistently failing, resulting in half passes/half fails. its really annoying and it makes it easy to ignore actual failures in the build/test.

I previously modified it to not test on the GOslim OBO, but thats not enough.  https://travis-ci.org/tripal/tripal/jobs/474957282

As the below excerpt from the log shows, the `tripal_obo_test` vocabulary exhausts memory (sometimes).

While I'd prefer we not remove the OBO loader test, it might be unavoidable.  Maybe instead we need to cover the individual methods, and get by without a functional test of the whole loader.  Alternatively we need to somehow account for limited memory environments.

```
Building cvterm paths for vocabulary: tripal_obo_test
Clearing the cvtermpath table for this vocabulary...
Clearing completed.
Retrieving relationships...
Processing terms...
.PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 12288 bytes) in /home/travis/build/tripal/drupal/includes/database/database.inc on line 2227
PHP Stack trace:
PHP   1. {main}() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/phpunit:0
PHP   2. PHPUnit\TextUI\Command::main() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/phpunit:53
PHP   3. PHPUnit\TextUI\Command->run() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/src/TextUI/Command.php:151
PHP   4. PHPUnit\TextUI\TestRunner->doRun() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/src/TextUI/Command.php:198
PHP   5. PHPUnit\Framework\TestSuite->run() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:529
PHP   6. PHPUnit\Framework\TestSuite->run() /home/travis/build/tripal/drupal/sites/all/modules/tripal/vendor/phpunit/phpunit/src/Framework/TestSuite.php:776
[...]

PR #816, opened 2019-01-03
up mem limit

bradfordcondon:
# Bug Fix

Issue #815 

## Description

This PR modifies the PHP memory limit setting.  This prevents the test suite from running out of memory when running the OBO loader tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,10 @@ before_script:
   - drush en -y field_group, field_group_table, field_formatter_class, field_formatter_settings, ctools, date, devel,
               ds, link, entity, libraries, redirect, token uuid, jquery_update, views, webform
 
+  # up memory limit of PHP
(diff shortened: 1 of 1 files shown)


bradfordcondon:
this works but we **STILL** run out of memory for the OBO test some times:

```
Building cvterm paths for vocabulary: tripal_obo_test
Clearing the cvtermpath table for this vocabulary...
Clearing completed.
Retrieving relationships...
Processing terms...
.PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 12288 bytes) in /home/travis/build/tripal/drupal/includes/database/database.inc on line 2227
PHP Stack trace:
```

we should probably merge this either way but let's discuss in #815

spficklin:
Approved I see no objections.

bradfordcondon:
OK, abdullah and i discussed this so im merging, thanks.

PR #817, opened 2019-01-03
check field permissions explicitly if configured

bradfordcondon:
finish resolving #749 , after #794

this PR deals with custom field permission cases properly, by using `global $user` and checking permissions for that user.
This PR assumes that it makes sense for web services to have a user associated with them (rather than needing to assume that the user is anonymous).

to check if the user accessing web services has permission for the field, in the event of custom field permissions.

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -377,7 +377,9 @@ class TripalContentService_v0_1 extends TripalWebService {
 
       if ($field['field_permissions']['type'] === '1') {
         // Field is private.
-        return;
+           if (!user_access('view ' . $field_name, $user)) {
+          return;
+        }
       }
       if ($field['field_permissions']['type'] === '2') {
         // Field is custom permissions: Check
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -371,9 +371,21 @@ class TripalContentService_v0_1 extends TripalWebService {
       }
     }
 
-    if (isset($field['field_permissions']['type']) && $field['field_permissions']['type'] != 0) {
-      return;
-    }
+    if (isset($field['field_permissions']['type'])) {
+
+      global $user;
+
+      if ($field['field_permissions']['type'] === '1') {
+        // Field is private.
+        return;
+      }
+      if ($field['field_permissions']['type'] === '2') {
+        // Field is custom permissions: Check
+        // if user has access to view this field in any entity.
+        if (!user_access('view ' . $field_name, $user)) {
+          return;
+        }
+      }
 
       if ($hide_fields == TRUE and empty($values[0])) {
       return;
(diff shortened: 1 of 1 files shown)


bradfordcondon:
the failed test in `tripal_ws` fails for me on current master as well as on this branch.  However that test isnt failing on the travis.... not sure what the problem is.

if someone can get the test to pass locally on master/this branch would help to clarify: `phpunit tests/tripal_ws/http/TripalWebServicesContentTest.php`

I see that the test doesnt get an explicit bundle: `    $label = db_query('SELECT label FROM tripal_bundle LIMIT 1')->fetchField();` so therefore it might not be checking the same bundle in each environment.  Presumably it will be organism? 


tests are passing on travis, even though they are failing for me locally still.  not sure what ot make of that, but i think it clears this PR.

tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
```diff
@@ -371,8 +371,21 @@ class TripalContentService_v0_1 extends TripalWebService {
       }
     }
 
-    if (isset($field['field_permissions']['type']) && $field['field_permissions']['type'] != 0) {
-      return;
+    if (isset($field['field_permissions']['type'])) {
+
+      global $user;
+
+      if ($field['field_permissions']['type'] === '1') {
```
spficklin, line 378: Just curious, where do these 'types' (e.g. 1 and 2) come from? Is this a field_permission module thing?

bradfordcondon: yep!  we could link out to the field permissions module in a comment if that would help although i only found it in the code, not in the online documentation.
0:  Public (author and administrators can edit, everyone can view)
1: Private (only author and administrators can edit and view)
2: Custom
although that makes me realize, maybe the action for 1 & 2 should be the same- to check the user's permission.

spficklin (approved):

spficklin:
The code changes look good. I approve via a code review.  I will get to a functional review shortly.

spficklin:
code changes are so few that I think only 1 review is needed.

bradfordcondon:
thanks, let me know if you have questions and/or check out the docs for using field specific permissions: https://tripal.readthedocs.io/en/latest/user_guide/content_types/field_permissions.html?highlight=permissions

laceysanderson (approved):
I did a thorough functional review and feel this is ready to be merged.

tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
laceysanderson: Are you content @bradfordcondon or did you want to change this before merging? I feel your original approach accomplishes the same goal.

bradfordcondon: thanks for the review, and the reminder about this @laceysanderson 

Basically the difference is that if the admin user is using web services, or if the user is the author, they wont be able to access the field via web services.

It can be easily resolved by taking the same approach for permissions 1 & 2 :

```php

   if (!user_access('view ' . $field_name, $user)) {
          return;
        }

```

tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
```diff
@@ -371,8 +371,21 @@ class TripalContentService_v0_1 extends TripalWebService {
...
+        // Field is private.
+        return;
```
bradfordcondon:
```suggestion
           if (!user_access('view ' . $field_name, $user)) {
          return;
        }
```

PR #820, opened 2019-01-07
comment out all tests for obo loader

bradfordcondon:
Issue #815 

## Description

the OBO tests are running out of memory inconsistently and causing the test environment to fail.  I see that in each case the OBO is in a data provider, i wonder if this is part of the problem...

We comment out the tests in this PR as discussed in #815


bradfordcondon:
this PR has been approved in the thread, and its just commenting out tests, so im going to merge.

Issue #824, opened 2019-01-17
Tripal DS should trigger a window resize event after showing a pane


almasaeed2010:
### System information

* Tripal Version: 3
* Drupal Version: 7.6
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

Tripal DS uses jquery animations to show/hide Tripal panes. Many modules, including tripal_analysis_interpro and tripal_analysis_blast, draw charts that show up after the page has loaded. Since they are in a hidden container, the width/height of the container is 0 by default. This is messing up how modules compute the dimensions of the chart and causing them to display blank results. See screenshot below.

#### Steps to reproduce

see issue https://github.com/tripal/tripal_analysis_interpro/issues/24 for more details on reproducing.

#### Error messages and screenshots

![issue](https://user-images.githubusercontent.com/17642511/51313489-f8c56200-1a4d-11e9-82a2-c4021a7fe1cb.png)

### Solution 

By triggering a resize event on the `window` object after the animation has completed, we can simulate the default behavior of the browser. See below for a working example:

```js
           if (typeof Event !== 'undefined') {
               window.dispatchEvent(new Event('resize'));
           }
           else {
               // Support IE
               var event = window.document.createEvent('UIEvents');
               event.initUIEvent('resize', true, false, window, 0);
               window.dispatchEvent(event);
           }
```

By adding this code right after this line, the issue would be fixed.
https://github.com/tripal/tripal/blob/3244f2dc184513d1c4576fb29b2d72e6f61d4413/tripal_ds/theme/js/tripal_ds.js#L65

Thanks!

bradfordcondon:
merged. thanks for the fix.

bradfordcondon:
I'm reopening this issue: it appears this fix isn't effective for safari or firefox

almasaeed2010:
My opinion is that since manually resizing the window on Safari /FF doesn't fix, then this is a library issue not a Tripal issue. The library may have implemented the event listeners to support only Chrome.

almasaeed2010:
If you want to verify it is not a Tripal issue, simply add a `console.log()` and make sure it fires when you click on the pane link.

```js
$(window).on('resize', function(e) { console.log('it fired!') })
```

bradfordcondon:
updating our featureviewer library to 1.0.4 resolved the problem.  abdullah is correct its a library issue not a Tripal issue.

PR #825, opened 2019-01-17
Trigger a window resize event when the tripal_ds animation is complete

almasaeed2010:
# Bug Fix

Issue #824

## Description

Fixes an issue with charting modules as described in #824.

## Testing?

The easiest way to test is to use tripal_analysis_interpro or tripal_analysis_blast modules to try to display a chart.

Thanks!

### DIFF
--- a/tripal_ds/theme/js/tripal_ds.js
+++ b/tripal_ds/theme/js/tripal_ds.js
@@ -34,7 +34,7 @@
           }
         });
       });
-      
+
       // Move the tripal pane to the first position when its TOC item is clicked.
       $('.tripal_pane-toc-list-item-link').each(function (i) {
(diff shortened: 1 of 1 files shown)


bradfordcondon:
Functional review: I've confirmed this fixes the problem.  (Thank you!)

@almasaeed2010 is there any potential problem if a module has its own resize event triggering to deal with this problem?  Sorry for the ignorant question.

almasaeed2010:
Technically it would trigger the event twice causing double the processing/memory usage. But it's extremely rare that that would cause any noticeable issues.

laceysanderson (approved):
Code-wise this looks good 👍 I also tested it on my site and it didn't break any of my existing charts.

laceysanderson:
Thank you for the fix @almasaeed2010!

PR #827, opened 2019-01-22
Allow complex filtering when only one level is set

guignonv:
#

Issue #

## Description
When using complex filtering, one could use as filter argument ($values):
```
array(
  'fieldname' => array(
    'op' => 'someop',
    'data' => 'somevalue',
  ),
);
```
or
```
array(
  'fieldname' => array(
    array(
      'op' => 'someop',
      'data' => 'somevalue',
    ),
    array(
      'op' => 'someop2',
      'data' => 'somevalue2',
    ),
  ),
);
```

but the following syntax was not supported:
```
array(
  'fieldname' => array(
    array(
      'op' => 'someop',
      'data' => 'somevalue',
    ),
  ),
);
```
This last syntax should be supported as complex filters might be auto-generated by some process that should not have to worry, in the end, if they provide just one or several complex filters for a field.
This patch does the trick.

## Testing?
Tested with the 3 case of complex filter above and also with a simple filter with multiple values like:
```
array(
  'fieldname' => array(
    array(
      'test' => '1',
    ),
  ),
);
```

### DIFF
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -1368,7 +1368,7 @@ function chado_select_record($table, $columns, $values, $options = NULL) {
       // CASE 1a: If there is only one element in the array, treat it the same
       // as a non-array value.
       if (count($value) == 1 AND is_int(key($value))
-          AND !(isset($value[0]['op']) && isset($value[0]['op']))) {
+          AND !(isset($value[0]['op']) && isset($value[0]['data']))) {
 
         $value = array_pop($value);
         $op = '=';
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -1367,7 +1367,8 @@ function chado_select_record($table, $columns, $values, $options = NULL) {
 
       // CASE 1a: If there is only one element in the array, treat it the same
       // as a non-array value.
-      if (count($value) == 1 AND is_int(key($value))) {
+      if (count($value) == 1 AND is_int(key($value))
+          AND !(isset($value[0]['op']) && isset($value[0]['op']))) {
 
         $value = array_pop($value);
         $op = '=';
--- a/tests/tripal_chado/api/ChadoQueryTest.php
+++ b/tests/tripal_chado/api/ChadoQueryTest.php
@@ -37,7 +37,7 @@ class ChadoQueryTest extends TripalTestCase {
       'uniquename' => array(
         array(
           'op' => 'LIKE',
-          'data' => '01%',
+          'data' => 'octopus%',
         ),
       ),
     );
@@ -49,7 +49,7 @@ class ChadoQueryTest extends TripalTestCase {
 
     // Test 3 Pass an array of filters with multiple items.
     $selector = array(
-      'uniquename' => array(
+      'type_id' => array(
         array(
           'op' => '>',
           'data' => ($stock->type_id - 1),
@@ -60,7 +60,8 @@ class ChadoQueryTest extends TripalTestCase {
         ),
       ),
     );
-    $object = chado_generate_var('stock',[], $selector);
+
+    $object = chado_generate_var('stock', $selector);
     $this->assertNotNull($object->stock_id);
     $this->assertEquals($stock->stock_id, $object->stock_id);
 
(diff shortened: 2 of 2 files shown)


bradfordcondon:
>This last syntax should be supported as complex filters might be auto-generated by some process that should not have to worry, in the end, if they provide just one or several complex filters for a field.

I'd agree with this.

My gut says this is a fine change but i'd like to do a bit of confirmation first.  I see its the `chado_select_record` method so i can figure it out from there.

tripal_chado/api/tripal_chado.query.api.inc
```diff
@@ -1367,7 +1367,8 @@ function chado_select_record($table, $columns, $values, $options = NULL) {
 
       // CASE 1a: If there is only one element in the array, treat it the same
       // as a non-array value.
-      if (count($value) == 1 AND is_int(key($value))) {
+      if (count($value) == 1 AND is_int(key($value))
+          AND !(isset($value[0]['op']) && isset($value[0]['op']))) {
```
almasaeed2010: I think this line has a duplicate?

```suggestion
          AND !isset($value[0]['op'])) {
```

almasaeed2010: Maybe you meant for it to be something else?

bradfordcondon:
>Tested with the 3 case of complex filter above and also with a simple filter with multiple values like:

also @guignonv  would you please give more concrete examples for testing purposes?  For example an organism query that demonstrates the problem/each of the three cases.

  That would make it much easier for me to review this PR.

guignonv:
These kind of tests:
```
// Test 1 (no fix needed).
$selector = array(
  'stock_id' => 1085,
  'uniquename' => array(
    'op' => 'LIKE',
    'data' => '01%',
  ),
);
$object = chado_generate_var('stock', $selector);

// Test 2 (fix needed).
$selector = array(
  'stock_id' => 1085,
  'uniquename' => array(
    array(
      'op' => 'LIKE',
      'data' => '01%',
    ),
  ),
);
$object = chado_generate_var('stock', $selector);

// Test 3 (no fix needed).
$selector = array(
  'uniquename' => array(
    array(
      'op' => '>',
      'data' => '1084',
    ),
    array(
      'op' => '<',
      'data' => '1086',
    ),
  ),
);
$object = chado_generate_var('stock', $selector);
```

bradfordcondon:
thanks @guignonv .  That was incredibly helpful. I created three tests for the three cases you described.  I'll PR them to your branch on your fork.  (actually i created one test with the three cases you described, possibly not the best practice haha)

The PR is here: https://github.com/guignonv/tripal/pull/1

Please confirm I pr'd to the correct branch (patch-2)


Also, @almasaeed2010 has pointed out that the logic you added to case 1a is probably not what you intended, can you address?

bradfordcondon:
i confirmed the original issue; running the tests i pr'd to this branch on master gives us this error:

```

Runtime:       PHP 7.1.19 with Xdebug 2.5.5
Configuration: /Users/bc/tripal/sites/all/modules/tripal/phpunit.xml

E                                                                   1 / 1 (100%)

Time: 3.1 seconds, Memory: 58.00MB

There was 1 error:

1) Tests\ChadoQueryTest::test_filter_level
PDOException: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near ","
LINE 1: ...ock WHERE stock_id = '480' AND uniquename = 'LIKE', 'octopus...
                                                             ^
```

tests pass with @guignonv 's PR so thank you.

Can you please just  address https://github.com/tripal/tripal/pull/827#discussion_r250619969 ?  Then we should be good to merge.

tripal_chado/api/tripal_chado.query.api.inc
guignonv: Well, I wanted to also test 'data' and copy&paste but forgot to turn the second 'op' into 'data'. However, it (your correction) should work because we would expect to have a 'data' key when using a 'op' key.
I'll change 'op' into 'data' and it'll be fine. Thanks @almasaeed2010  for pointing out my mistake! :-)

guignonv:
[#827 (comment)](https://github.com/tripal/tripal/pull/827#discussion_r250619969) has been addressed. Code has been adjusted. The duplicate was on purpose but had to be modified: 'op' into 'data'. I guess I was distracted when I started to do it. :p

bradfordcondon:
this PR just needs one more review.  Doesn't need to be a full functional review as we have tests, and can be from any member of the community.

laceysanderson (approved):
Looks good to me! Thank you @guignonv for the contribution and @bradfordcondon for the tests!

PR #829, opened 2019-01-23
Fix #790 (+ revert #799) : Alias issues

### LINKED ISSUE #790: [Tripal3]Aliases token not filled (empty)

mboudet:
# Bug Fix

Issue #790
## Description
When writing #799, I misread the logic behind the loops.
The issue was not that the array was reset every loop, it was that 

`module_invoke($storage['module'], 'field_storage_load', 'TripalEntity',
          $entities, FIELD_LOAD_CURRENT, $field_ids, array());`

was **not** called every loop

Before #799, it meant that only one storage type was properly loaded (the last one in the loop).
After #799, one storage type is properly loaded (the last one), but we try to load fields not belonging in that storage type (and fail silently).

I think the intended logic was to have one loop per storage type, select all the fields belonging in that storage type, and load them. This PR fixes that.

## Additional Notes (if any):

It fixes #790 because the two missing fields (text) are properly loaded by tripal_chado_field_storage_load (with the else clause)

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -1324,17 +1324,17 @@ function tripal_replace_entity_tokens($string, &$entity, $bundle_entity = NULL) 
 
   // If we have any fields that need attaching, then do so now.
   if (count(array_keys($attach_fields)) > 0) {
-    $field_ids = array();
     foreach ($attach_fields as $storage_type => $details) {
+      $field_ids = array();
       $storage = $details['storage'];
       $fields = $details['fields'];
       foreach ($fields as $field) {
         $field_ids[$field['id']] = array($entity->id);
       }
       $entities = array($entity->id => $entity);
+      module_invoke($storage['module'], 'field_storage_load', 'TripalEntity',
+          $entities, FIELD_LOAD_CURRENT, $field_ids, array());
     }
-    module_invoke($storage['module'], 'field_storage_load', 'TripalEntity',
-        $entities, FIELD_LOAD_CURRENT, $field_ids, array());
   }
 
   // Now that all necessary fields are attached process the tokens.


bradfordcondon:
hi @mboudet is there any way you can provide some instructions for how to test this PR?  Kind of confusing with the fix/revert language, i'd like to functionally test it.

thank you!

mboudet:
Well, it is going to be a bit complicated since it's a cache issue (and based on the comment on the issue itself, it might come from a specific cache setup..). If your cache is properly populated, you will not see an issue.

Basically, I use docker-tripal (https://github.com/galaxy-genome-annotation/docker-tripal). When I try to set up an automatic alias for an entity type, the url created is malformed due to some token not being replaced.

The easiest way to test would be to setup the docker, setup the alias for the mRNA (as in #790), and upload some mRNA. You should be able to see the malformed aliases.

I'm aware the fix/revert is confusing. Basically, I thought I partially fixed the problem with #799, but it was just a coincidence that it worked.

The would-be logic for the tripal_replace_entity_tokens() function is this (if the cache is not set) : 
Get all missings fields ->get the storage back-end associated with each field -> iterate on each backend to load the fields related to it.

Due to looping issues, it was not properly loading fields for each backend, just the last one.
#799 made it try to load every missing field from the last backend, instead of just the related field.

Sorry if that's not clear @bradfordcondon

laceysanderson:
Hi @mboudet, I was trying to confirm you fix and ended up writing a unit test which I believe demonstrates your problem. I submitted a PR for it to your branch (https://github.com/mboudet/tripal/pull/1) if you're interested.

mboudet:
Thanks @laceysanderson , looking into it!

mboudet:
So sadly, the unit test fail is unrelated to the issue I think.
The easiest way to see if the code works would be to manually unset the field [here](https://github.com/tripal/tripal/blob/7.x-3.x/tripal/api/tripal.entities.api.inc#L1315), with an `unset $entity->{$field_name}`.
That would make sure the field cache is clear, and needs to be re-populated.

And then either manually call tripal_replace_entity_tokens(), or use the bulk update alias tool.
Without this PR, it should empty tokens in the generated urls.

(Not the easiest way to proceed I know, but I still do not know why my cache is not properly populated at this point..)

abretaud:
Any chance to get this merged? I tested the patch and it works great

bradfordcondon:
Thanks @abretaud I'll mark that as a review for this PR.  We just need one more, it looks like @laceysanderson is working on it.

spficklin:
With the upcomging 3.1 release imminent. Let's hold off on a merge of this until after the second review and the 3.1. release is out.,

bradfordcondon:
I'm not able to do a functional test this with that docker- too slow w/ lots of volume mappings, and i dont have a linux box to run it on.

That said the logic does make sense to me.  


I'm approving this based on @abretaud 's functional review.  We can merge once 7.x-3.1 is released.

Thank you very much for the contribution @mboudet and sorry for the slow review.  If we're looking to speed up reviews, I think having a clear PR message goes a long way, as do tests.  That said this is an exceptionally difficult PR to write good instructions for, and writing tests for fields is almost impossible right now.  Thanks both for your patience.

spficklin:
3.1 now released.  Merging this in.

PR #834, opened 2019-01-23
Resolve travis mem

bradfordcondon:
# Bug fix

Issue #815 

## Description
Please see #815 for full description.

We run out of memory because we have too many organisms when we run the taxonomy importer test.  It makes too many requests/second to NCBI, which spams errors until we run out of memory.

This fix

* checks to make sure we got a response from NCBI
* adds a timer to prevent overloading ncbi request limit (sleep)
* truncates the organism table prior to running hte test to ensure the test is fast.

Note there is one lingering issue:

- why do we have organisms?  we use transactions everywhere.  Are transactions not working as intended?

### testing

The unit test should fail if you have more than a few organisms (say, 10) on your test site on 7.x-3.x.
On this branch it should pass.

Reminder, you can unit test a single file:
`phpunit tests/tripal_chado/loaders/TaxonomyImporterTest.php`


laceysanderson:
I'm mildly worried about dumping all the organisms. While tests should be run on a development box, removing all the data in the site like this will make me run the tests less often. I feel that PHPUnit testing is designed with setUp() and tearDown() so that we don't alter the environment the tests are done on too much... Perhaps other tests should be doing a better job cleaning up after themselves?

bradfordcondon:
> Perhaps other tests should be doing a better job cleaning up after themselves?

i plan on going through and figuring out which transaction rollback is failing and why, but it will take some time.  This test is itself in a db transaction so your site will not be affected, something i've confirmed on my own site.   (I've run this test and i still have organisms afterwards).

As I added the sleep, we could remove the truncate.  the test would be slow but it also should pass.

tests/tripal_chado/loaders/TaxonomyImporterTest.php
```diff
@@ -26,8 +28,13 @@ public function testImportExistingTaxonomyLoader() {
...
     ];
 
+    // Speed up test an ensure no organisms
+    // in db since it will check all.
+    $prev_db = chado_set_active('chado');
+    chado_query('TRUNCATE TABLE {organism} CASCADE');
+    chado_set_active($prev_db);
     $organism = factory('chado.organism')->create($org);
-    //  $this->publish('organism');
+    // $this->publish('organism');.
```
bradfordcondon:
```suggestion
```

tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
```diff
@@ -483,6 +487,12 @@ class TaxonomyImporter extends TripalImporter {
       // Get the search response from NCBI.
       $rfh = fopen($search_url, "r");
       $xml_text = '';
+
+      if (!$rfh){
+
+        $this->logMessage("Could not look up " . $sci_name, [], TRIPAL_WARNING);
```
bradfordcondon: oops this is pretty lazy of me.  lets fix.

```suggestion
        $this->logMessage("Could not look up !sci_name" , ['!sci_name' =>  $sci_name], TRIPAL_WARNING);
```

bradfordcondon:
@laceysanderson i've removed that truncation from the test.

tests/tripal_chado/loaders/TaxonomyImporterTest.php
```diff
@@ -5,15 +5,17 @@
...
 class TaxonomyImporterTest extends TripalTestCase {
 
   use DBTransaction;
 
-
-  /*
+  /**
    * Adds an organism and checks that the importer runs and adds some properties to it.
    *
+   * @group waffle
```
bradfordcondon:
```suggestion
```
remove test code

laceysanderson (approved):
Confirmed that all tests pass locally. Also code-wise, everything looks great 👍

laceysanderson:
Since the code changes are minimal and make perfect sense, I'm marking this as needing only 1 review. Since I provided that review, I'm merging :-)

PR #835, opened 2019-01-24
Do not create DB records in sbo__relationship_widget test

almasaeed2010:
# Bug Fix

Issue #815

## Description

This PR makes sure DBTransaction is called and data does not get permanently stored in the DB by the end of the test.

It turns out that PHPUnit's dataProviders are executed before setUp and tearDown making it impossible for us to automatically start and rollback database transactions. Therefore, to keep all the test cases I suggest we keep the data provider and simply call it directly in the test function rather than use the PHPUnit decorator.

## Testing?

On a clean DB, prior to applying this fix, you'll have 28 organisms created when running `phpunit`. After switching to this branch, we no longer have records created.

## Additional Notes (if any):

Here is the related issue on TripalTestSuite: https://github.com/tripal/TripalTestSuite/issues/136

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,11 @@
     "doctrine/instantiator": "1.0.*",
     "statonlab/tripal-test-suite": "1.*"
   },
+  "autoload": {
+    "files": [
+      "tests/TripalFieldTestHelper.php"
+    ]
+  },
   "require": {
   }
 }
--- a/tests/tripal_chado/fields/sbo__relationship_widgetTest.php
+++ b/tests/tripal_chado/fields/sbo__relationship_widgetTest.php
@@ -1,12 +1,15 @@
 <?php
+
 namespace Tests\tripal_chado\fields;
 
 use StatonLab\TripalTestSuite\DBTransaction;
 use StatonLab\TripalTestSuite\TripalTestCase;
 
-module_load_include('php', 'tripal_chado', '../tests/TripalFieldTestHelper');
-
+/**
+ *
+ */
 class sbo__relationship_widgetTest extends TripalTestCase {
+
   // Uncomment to auto start and rollback db transactions per test method.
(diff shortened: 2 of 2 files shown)


almasaeed2010:
**NOTE** please run `composer update` or `composer dump-autoload` if you are getting errors. This is because I've moved the `TripalFieldTestHelper` to use composer's autoload functionality, which makes the class available to all of our tests rather than having to `require` or `module_load_include` it in every test.

Thanks!

almasaeed2010:
Sorry about all of the code style changes! I didn't realize it was this bad. Here is a summary of I did:

- Added this to all test functions
```
foreach ($this->provideEntities() as $dataEntry) {
      list(
        $bundle_name, $field_name, $widget_name, $entity_ids, $expect
        ) = $dataEntry;

...

}
```

Where `...` is the code that was already there.


- Removed all parameters from the test function decleration
- Removed module_load_include in favor of composer.json autoload as explained above
- removed all `@dataProvider` decorators

bradfordcondon:
i confirmed that we no longer leave extra records for project.  thanks!

I'm not marking this as a 1 review issue because of the number of line changes, even if they are largely style.

laceysanderson (approved):
Tests pass locally and a thorough code review looks good 👍  It makes me sad we can't use `@dataProvider` directly due to transactions but this is a great, clean work-around :-) Thank you @almasaeed2010!

bradfordcondon:
to be fair, you can still use data providers, you just cant have the data providers be inserting the records.

So instead of passing an inserted record, you can pass an array of values that would go into the factory.   So your test then starts out with `$my_record_specified_by_a_provider_but_inserted_in_the_test-so_transactions_work = factory($provider_table)->create($provider_array)` for example.

Issue #732, opened 2018-10-31
Gene page improvement: child feature field


bradfordcondon:
## Feature Request/Discussion
This issue is a splinter from #100 .  It's a feature request as @laceysanderson and I agree that having better gene pages is a **very desirable thing**. 

### Description

Gene pages aren't great out of the box, and fields just arent designed to work with a typical gene setup.  What do I mean by that?  In short, **Genes need to access child feature information** and they dont, or they do in limited ways.

Lets consider three fields.  I have a vanilla gene here at 
http://167.99.232.220/gene/185
I loaded FASTA sequences for scaffolds, mRNA, CDS, and proteins.  NONE of this information is available on the gene page!  Neither would annotations associated these feature types (unless we loaded them associating via the parent gene via regexp.).

#### Field 1 - transcripts

[Transcripts](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/so__transcript/so__transcript.inc)
The transcript field on the gene page doesnt provide sequences, or any information other than location.

#### Field 2 - sequence 

[Sequences](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/data__sequence/data__sequence.inc) . Per tripal's general philosophy, the sequence field only look as the targeted feature, not related features.  Because I didnt load in the gene FASTA, theres nothing to display, even though i loaded in 4 other feature type fastas, plus the GFF. I thould be able to infer the gene sequence via the GFF, just for starters.

#### Field 3 - protein sequence
[Field code](https://github.com/tripal/tripal/tree/7.x-3.x/tripal_chado/includes/TripalFields/data__protein_sequence)
There are protein sequences associated with this gene, but **indirectly** via the mrna -> CDS.    The protein sequences field therefore does not find them.

# Solution

Some considerations...
* I'd like the solution to work for all feature types, if possible.  Meaning, if im on feature X, it should find and display all features linked to it, nomatter how indirectly.  The field config might allow excluding display of certain SO terms if thast too much.
* Solution should be fast.  We are iterating through each feature_relationship.  I dont want a suite of fields doing that independently and redundantly.  So, we're probably either working on one mega field that displays all child-feature info, or, developing a shared architecture so that it only gets called once per page and cached.
* Sites will want access to, at a minimum, the following info from all children:
  - annotations (feature_cvterm, also blast data)
  - properties  (featureprops)
  - sequences 
   -  coordinate info (featureloc)

[...]

PR #837, opened 2019-01-25
Gene specific fields for child features

bradfordcondon:
# New Feature

Issue #732 

## Description
Gene fields.  We want to display info on gene pages, but information is typically associated with child features of said genes.  interpro annotations are on the proteins.  Perhaps some custom properties are on the mRNA.  Users expect to get a summary of the gene function on the gene page.

I accomplish this here by adding three fields.  The master field, data__sequence_features, stores all the information about child features.  It does so where each item is a direct child (assumed to be an mRNA).  Each item has two keys: `info` and `children`.  `Children` is an array of child features (via feature_relationship), each with two keys: `info` and `children`.  The `info` array holds an object from chado_expand_var: it has `featureloc`, `residues`, `featureprop`, and `feature_cvterm`.  This approach preserves the underlying data structure.   This field comes with a formatter to display a gene map (the featureloc info, see screenshot below)

 Additional fields (local__child_annotations, local__child_properties) are for displaying just the annotations and properties, respectively.

I have a bit of a guide written up here, but im not sure where to place it in the tripal docs if at all:

https://github.com/statonlab/tripal_manage_analyses/blob/master/docs/gene_fields.rst

## Testing?
The field is thinking that you have child features of genes that are mRNA/transcripts, and child features of that which are proteins, UTRs, CDS, etc.  Load such data, go to the gene bundle, and check for new fields.  enable display of the three gene page fields:

data__sequence_features
local__child_annotations
local__child_properties

screen shot 2019-01-25 at 12 06 22 pm

Child properties

screen shot 2019-01-25 at 12 19 59 pm

## Remaining questions

- I'm finding weighting the fields to be an inconsistent solution to ensuring the master field loads first.
- Do we need the guide info? if so, where?
- I didnt do query/queryOrder etc.
- The first child feature (assumed transcript) is open, the others are in a collapsed fieldset.  Is there an existing GFF property that should be used to determine the primary transcript to list first/open first?


spficklin:
I had to rebuild my Tv3 site from scratch so I tested this on a brand-new site using only data described in the User's Guide tutorial (i.e. just the citrus gene).

All went well until I tried to view the citrus gene page.  Here's what I'm getting:


837-problems

The new fields aren't appearing.  Thoughts on how to fix this?

bradfordcondon:
i'll load the citrus dataset (note to self, i'd love to create a test suite importer so i can do this with a single command... [WIP is here](https://github.com/tripal/TripalTestSuite/blob/citrus_seeder/stubs/CitrusSeeder.php)) so I can try to recreate the problem.  One possible issue is the main field's formatter requires a gene sequence to draw the glyph, so if theres no sequence, you get no drawing.  The sequence field currently JUST checks the feature.residues column it doesnt infer from feature.relationship/featureloc.  However that wouldnt explain those warnings.  Its possible those warnings come from the field weighting not working (a child field trying ot read data from the parent field, before the parent field loads), which is one of the issues im asking for help with.

bradfordcondon:
- the main field should now at least appear.  The problem arose if the sequence field wasn't set.  I try to build a string of N's based on the gene length, but if that isnt set either, we just build an arbitrary sequence of 5,000 N's.
- I'm getting inconsistent results with the property field.  This is because sometimes, the property field loads before the parent field.  Setting the weight doesn't resolve this.  If you have this problem, rearrange the fields.  If you can help me fix it, awesome!  If you still cant get it, just have the child fields load the parent field in the formatter instead of the load.
- the fields are much slower with the citrus data, im guessing because there are so many transcripts associated with the gene.

Issue #838, opened 2019-01-28
Weird field behavior with Ajax calls in field formaters


mboudet:
Hello,

I'm currently working on updating tripal_analysis_go for Tripal 3.
Thanks to #607, I managed to include an Ajax call in the field, allowing me to replicate the module features.

However, when I make the Ajax call, I have some weird behavior : the 'close' button on all fields is duplicated.

Example: 

![image](https://user-images.githubusercontent.com/17642511/51840623-d284a980-230c-11e9-8bc0-fe304a3dc60b.png)

After Ajax call

![image](https://user-images.githubusercontent.com/17642511/51840655-ea5c2d80-230c-11e9-985d-6f7ad165efa5.png)

Did anyone have this issue? It might come from my implementation, but I do not know what would cause it.

almasaeed2010:
Hello @mboudet 

Could you please provide your system information?

tripal version:
drupal version:
php version:
browser name and its version:

It'd be very helpful if we can also take a look at the code that builds the field.

Thanks!

bradfordcondon:
hey @mboudet total aside but did you use the guide i wrote in that issue?  was it helpful?

I decided against adding it to the official documentation but if you had the same question and that helped maybe i should reconsider :)

mboudet:
Sure : 

System info:
Tripal: 7.x-3.0
Drupal : 7.59
Postgres: 9.5.8
Php: 7.1.23

The code for the field would be available here : https://github.com/mboudet/tripal_analysis_go/tree/7.x-3.x/includes/TripalFields/local__go_results

Most of it was auto-generated with the field generator. I added a drupal_get_form to generate the form.

The form itself is build like this : 

```
$form['tripal_analysis_go_select'] = array(
    '#title' => t('Select a GO report to view'),
    '#description' => t('Any analysis with GO results related to this organism are available for viewing. For further information, see the analysis information page.'),
    '#type'  => 'select',
    '#options' => $analyses,
    '#ajax' => array(
      'callback' => "tripal_analysis_go_org_report",
      'wrapper' => 'tripal-analyis-go-report',
      'effect'   => 'fade',
      'method'   => 'replace',
    ),
    '#suffix' => '<div id="tripal-analyis-go-report"></div>'
  );
```

mboudet:
@bradfordcondon Yes it was great, since I had no idea how to make it work.
I think having it part of the official documentation would be a great help for module developer, since it helps creating complex fields.

almasaeed2010:
[...]

PR #839, opened 2019-01-28
Fix multiple close icons with ajax calls

almasaeed2010:
# Bug Fix

Issue #838 

## Description

It seems that Drupal executes all JS code that's registeres as a Drupal.behavors module whenever an AJAX call is made in a form. Tripal DS adds the `x` icon to panes on page load. Checking to make sure the `x` doesn't already exist before adding it fixes the issue.

```
if($(this).find('.tripal_pane-fieldset-buttons').length > 0) {
  return;
}
```

## Testing?

See issue #838 for a full example.

Screenshots have been posted by @mboudet 

screenshot "screenshot"

Thanks!

### DIFF
--- a/tripal_ds/theme/js/tripal_ds.js
+++ b/tripal_ds/theme/js/tripal_ds.js
@@ -3,6 +3,9 @@
     attach: function (context, settings){
       // Add a close button for each pane except for the te_base
       $('div.tripal_pane').each(function (i) {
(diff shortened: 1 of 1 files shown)


almasaeed2010:
@mboudet has tested this code so we can add `approved by reviewer #1` tag 👍

bradfordcondon:
looks really straight forward to me.  thanks for the fix.

Issue #607, opened 2018-08-28
help: using AJAX in a field formatter?


bradfordcondon:
I'm trying to add a feature a simple feature to a field formatter: a select box which will render different stuff via AJAX.

Here's the relevant code:

```php

      $element[0]['select'] = [
        '#type' => 'select',
        '#options' => $select,
        '#title' => 'Please select a map to view',
        '#default_value' => 0,
        '#ajax' => [
          'callback' => 'tripalmap_organism_featuremap_callback',
          'wrapper' => 'tripalmap-featuremap-organism-selector-wrapper',
          'effect' => 'fade',
        ],
      ];

      $selected = 0;

      if (isset($element['#values']['select_featuremap'])) {
        $selected = $element['#values']['select_featuremap'];
      }
      
      $element[0]['rendered_maps'] = [
        '#type' => 'fieldset',
        '#collapsible' => FALSE,
        '#prefix' => '<div id="tripalmap-featuremap-organism-selector-wrapper">',
        '#suffix' => '</div>',
      ];

      if ($selected != 0){
        $form = drupal_get_form('tripal_map_genetic_map_overview_form', $selected);
        $content = drupal_render($form);
        $element[0]['rendered_maps']['map'] = [
          '#type' => 'markup',
          '#markup' => $content,
        ];

      }
      return $element;
    }
```

here's the callback, which i placed in the main `.module` file.

```


function tripalmap_organism_featuremap_callback($form, &$form_state){
  ddl("it never fires");
  return $form[0];
}

```

As you can guess from the message, I never get the "it never fires" message logged.

Question 1, answered thanks to @laceysanderson :

>One other thing to note, fields will only render the number of $elements as there are $items. So if you have only a single item, you can't have your select as `0` and your fieldset as `1`. This has caught me many times! Instead you should have `$element[0] = ['#type' => 'markup', '#tree' => TRUE];` then `$element[0]['select']` and `$element[0]['fieldset']`

So next question, why wont the callback ever fire?  I'm fairly certain this code would work if i were in a $form.

bradfordcondon:
im writing this in an issue because i think AJAX enabled fields should go in a guide topic probably...

spficklin:
AJAX can be used in in field formatters.  I use it in the tripal_network module to allow users to select a network module for viewing when they are viewing a Network page.  It uses a select box to allow the end-user to select the module and then draw it.  

To make AJAX work in a formatter you would have to 
[...]

PR #840, opened 2019-01-28
add ajax formatter guide

bradfordcondon:
# Documentation
Issue #607
Adding AJAX to formatters is kind of tricky, since you have to add a separate form and then include that form.  This guide walks you through it.

## Description
I wrote a guide detailing how to add AJAX to a formatter field, in #607 .  in #838 @mboudet brought up they had used the guide and found it helpful.  considering another user had the same problem i did and the guide helped them figure it out, I think its worth putting in the manual.

### DIFF
--- a/docs/dev_guide/custom_field/ajax_custom_formatter.rst
+++ b/docs/dev_guide/custom_field/ajax_custom_formatter.rst
@@ -1,0 +1,105 @@
+AJAX Responsive Formatters
+===========================
+
+
+Some fields need to be responsive.  For example, a user might select an analysis or organism to display data from. Drupal developers often use AJAX to rebuild the page based on user input.
+
+Drupal and AJAX
+---------------
+
+Drupal has its special way of doing AJAX!  You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_ To Drupal, AJAX only makes sense as on forms, and field formatters **are not forms**.  Instead, they are `renderable arrays <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_
+This may seem obvious in hindsight: rather than accepting ``$form`` and ``&$form_state``, they accept ``&$element``, ``$entity_type``, ``$entity``, ``$langcode``, ``$items``, and ```$display``, where ``$element`` is the renderable array.
+
+This means if you want to add an AJAX callback to a field formatter, you  need a **seperate form file** that gets added in using ``drupal_get_form()``.  If you do this, you can build the AJAX as Drupal expects it.
+
+
+Example form and field
+----------------------
+
+Here's an example form file below: as you can see it's a standard form following Drupal AJAX conventions.  We provide a ``rendered_maps`` fieldset with the prefix defining the wrapper (``examplemap-featuremap-organism-selector-wrapper``).  This is what we want to re-draw depending on what the user selects.
+
+The selector has specified that wrapper, and the AJAX callback function ``examplemap_organism_featuremap_callback``.  We then define that function to simply return  the piece of the form that should be rebuilt: the ``rendered_maps`` fieldset!
+
+
+
+.. code-block:: php
+
+  #./includes/form_for_the_field.inc
+
+  function tripal_example_map_organism_featuremap_selector($form, &$form_state, $select) {
+
+    $selected = 0;
+
+    if (isset($form_state['values']['featuremap_select'])) {
+      $selected = isset($form_state['values']['featuremap_select']);
+    }
+
+
+    $form['rendered_maps'] = [
+      '#type' => 'fieldset',
+      '#collapsible' => FALSE,
+      '#prefix' => '<div id="examplemap-featuremap-organism-selector-wrapper">',
+      '#suffix' => '</div>',
+    ];
+
+
+    $form['rendered_maps']['featuremap_select'] = [
+      '#type' => 'select',
+      '#options' => $select,
+      '#title' => 'Please select a map to view',
+      '#default_value' => $selected,
+      '#ajax' => [
+        'callback' => 'examplemap_organism_featuremap_callback',
+        'wrapper' => 'examplemap-featuremap-organism-selector-wrapper',
+        'effect' => 'fade',
+      ],
+    ];
+
+
+    $chosen = 0;
+
+    if (isset($form_state['values']['featuremap_select'])) {
+      $chosen = $form_state['input']['featuremap_select'];
+    }
+
+    if ($chosen != 0) {
+
+
+      $mini_form = tripal_example_map_genetic_map_overview_form([], $form_state, $chosen);
+
+      $form['rendered_maps']['map'] = $mini_form;
+
+      return $form;
+    }
+
+    return $form;
+  }
+
+  /**
+  * The callback will return the part o the form you want to re-draw.
+  *
+   */
+  function examplemap_organism_featuremap_callback($form, &$form_state) {
+
+    return $form['rendered_maps'];
+  }
+
+
+
+In the field formatter, we simply add this form and put the markup in the element:
+
+.. code-block:: php
+
+      //Our__field_formatter.inc
+
+      //multiple maps for this organism, let user select.  Create a special form for that so we can have an AJAX select box
+      $select= $select + $select_add;
+
+      $form = drupal_get_form('tripal_example_map_organism_featuremap_selector', $select);
+      $content = drupal_render($form);
+        $element[] = [
+          '#type' => 'markup',
+          '#markup' => $content,
+        ];
+        return $element;
+    }
--- a/docs/dev_guide/custom_field/ajax_custom_formatter.rst
+++ b/docs/dev_guide/custom_field/ajax_custom_formatter.rst
@@ -24,17 +24,20 @@ The selector has specified that wrapper, and the AJAX callback function ``exampl
 
 .. code-block:: php
 
-  #./includes/form_for_the_field.inc
-
-  function tripal_example_map_organism_featuremap_selector($form, &$form_state, $select) {
+  /**
+   * AJAX-enabled form for [field formatter name].
+   */
+  function tripal_example_map_organism_featuremap_selector_form($form, &$form_state, $select) {
 
     $selected = 0;
 
+    // $form_state['values'] will be set if the form has been submitted via AJAX
     if (isset($form_state['values']['featuremap_select'])) {
       $selected = isset($form_state['values']['featuremap_select']);
     }
 
-
+    // We need to provide a container for Dupal AJAX to replace.
+    // Here we use a fieldset with a set ID which we can refer to below.
     $form['rendered_maps'] = [
       '#type' => 'fieldset',
       '#collapsible' => FALSE,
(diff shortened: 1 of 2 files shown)


docs/dev_guide/custom_field/ajax_custom_formatter.rst
```diff
@@ -0,0 +1,105 @@
+AJAX Responsive Formatters
+===========================
+
+
+Some fields need to be responsive.  For example, a user might select an analysis or organism to display data from. Drupal developers often use AJAX to rebuild the page based on user input.
+
+Drupal and AJAX
+---------------
+
+Drupal has its special way of doing AJAX!  You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_ To Drupal, AJAX only makes sense as on forms, and field formatters **are not forms**.  Instead, they are `renderable arrays <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_
```
laceysanderson:
```suggestion
Drupal has its own special way of doing AJAX! This is important to ensure that changes are executed in the correct order. You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_  The Drupal AJAX API works best on forms, and field formatters **are not forms**.  Instead, they are `renderable arrays. <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_
```

docs/dev_guide/custom_field/ajax_custom_formatter.rst
```diff
@@ -0,0 +1,105 @@
...
+This may seem obvious in hindsight: rather than accepting ``$form`` and ``&$form_state``, they accept ``&$element``, ``$entity_type``, ``$entity``, ``$langcode``, ``$items``, and ```$display``, where ``$element`` is the renderable array.
```
laceysanderson:
```suggestion
As such, rather than accepting ``$form`` and ``&$form_state``, they accept ``&$element``, ``$entity_type``, ``$entity``, ``$langcode``, ``$items``, and ```$display``, where ``$element`` is the renderable array.
```

laceysanderson (commented):
Helpful :-) 👍 In addition to the suggestions I made inline,

- the code samples could use some explanatory in-line documentation and function headers following Drupal coding standards
- Convention is for the form function to end in `_form` so maybe we want to show that here as well.

laceysanderson:
Added my further suggestions in a PR here: https://github.com/tripal/tripal/pull/844

docs/dev_guide/custom_field/ajax_custom_formatter.rst
bradfordcondon: >This is important to ensure that changes are executed in the correct order. 

I dont actually understand what this sentence is saying...

bradfordcondon: >  The Drupal AJAX API works best on forms

You imply that the Drupal way of handling AJAX is do-able on renderable arrays, but that its better on forms.  So, can we do it on renderable arrays *without forms* as this sentence implies?  I didn't think so, thats why we have to use a form instead.

laceysanderson: Drupal AJAX does work outside of forms. See https://www.drupal.org/docs/7/api/javascript-api/simple-drupal-ajax-load-with-jquery-and-delivery-callback. Unfortunately I've never seen a way to make this other method work with renderable arrays and since form-based ajax is more common and better documented 🤷‍♀️

bradfordcondon:
thanks for the suggestions @laceysanderson

laceysanderson (approved):
Did one last check rendering it and it looks great 👍

bradfordcondon:
@mboudet any chance you'd be willing to leave some feedback on the new guide, given you've already looked at a previous version of it?   We like all pull requests to have 2 reviews before we merge.

docs/dev_guide/custom_field/ajax_custom_formatter.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Some fields need to be responsive.  For example, a user might select an analysis or organism to display data from. Drupal developers often use AJAX to rebuild the page based on user input.
+
+Drupal and AJAX
+---------------
+
+Drupal has its own special way of doing AJAX! This is important to ensure that changes are executed in the correct order. You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_  The Drupal AJAX API works best on forms, and field formatters **are not forms**.  Instead, they are `renderable arrays. <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_
+As such, rather than accepting ``$form`` and ``&$form_state``, they accept ``&$element``, ``$entity_type``, ``$entity``, ``$langcode``, ``$items``, and ```$display``, where ``$element`` is the renderable array.
+
+This means if you want to add an AJAX callback to a field formatter, you  need a **seperate form file** that gets added in using ``drupal_get_form()``.  If you do this, you can build the AJAX as Drupal expects it.
```
mboudet: 'Separate form file' is a bit misleading I think. You need a separate form function, which should be defined in another file (usually the theme file)

docs/dev_guide/custom_field/ajax_custom_formatter.rst
```diff
@@ -0,0 +1,114 @@
...
+      '#options' => $select,
+      '#title' => 'Please select a map to view',
+      '#default_value' => $selected,
+      '#ajax' => [
+        // Your Drupal AJAX callback
+        // which simply returns the form element to be re-rendered.
+        'callback' => 'examplemap_organism_featuremap_callback',
+        // This should be the ID you set above on your container to be replaced.
+        'wrapper' => 'examplemap-featuremap-organism-selector-wrapper',
+        'effect' => 'fade',
```
mboudet, line 60: Doesn't this need a 'method'   => 'replace' too?

docs/dev_guide/custom_field/ajax_custom_formatter.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. code-block:: php
+
+    /**
+     * In our Our__field_formatter.inc
+     */
+    public function view(&$element, $entity_type, $entity, $langcode, $items, $display) {
+
+      // Multiple maps for this organism, let user select.  Create a special form for that so we can have an AJAX select box
+      $select = $select + $select_add;
```
mboudet: These are not defined, are they? $select should be an array key->values

docs/dev_guide/custom_field/ajax_custom_formatter.rst
bradfordcondon: No, thats optional (actually "replace" isn't a valid choice).  see: https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x

>#ajax['method']: The jQuery method to use to place the new HTML. Defaults to 'replaceWith'. May be: 'replaceWith', 'append', 'prepend', 'before', 'after', or 'html'. See the jQuery manipulators documentation for more information on these methods.

docs/dev_guide/custom_field/ajax_custom_formatter.rst
bradfordcondon: yeah, looks like that part of the code got cut out generalizing the example.  Thanks!

docs/dev_guide/custom_field/ajax_custom_formatter.rst
mboudet: I see. How about setting ''replaceWith' then? It's optional, but it does indicate that there are other options available

bradfordcondon: there are a lot of optional keys to pass there (effect, speed, event, prevent, method, progress).  I don't think we need to stub them all, especially in cases like this where the replaceWith effect is generally the desired effect.  I THINK we link to the ajax documentation in this guide, where a user can see them all.  If we don't, we definitely need to.

mboudet: Ok, If the user has a linkto the ajax doc it should be fine I think.

bradfordcondon: yep we do a few lines above, so i'm going to keep as-is (makes me wonder if i should remove the "effect" key actually)
Thanks!
```Drupal has its own special way of doing AJAX! This is important to ensure that changes are executed in the correct order. You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_  The Drupal AJAX API works best on forms, and field formatters **are not forms**```

File: docs/dev_guide/custom_field/ajax_custom_formatter.rst (hunk 3 above)
Comment (bradfordcondon) [reply docs/dev_guide/custom_field/ajax_custom_formatter.rst]: changing to `seperate form function`.
I'm not going to lie, i dont put my forms in `theme` i put them in `includes/thing_myform_does.form.inc`.  Is that poor practice?

**Comment by mboudet** [comment]
To be honest, I'm a bit confused by the mix of ajax callback, and the ajax call from submiting the form.
It might be good to include screenshots of the results.

Do we need to have this mix? Couldn't we just use the tripal_example_map_genetic_map_overview_form() in the callback function itself? No need to actually submit anything.

(I might just be confused about the logic in the form :p )

File: docs/dev_guide/custom_field/ajax_custom_formatter.rst (hunk 5 above)
Comment (bradfordcondon) [reply docs/dev_guide/custom_field/ajax_custom_formatter.rst]: the selectors get passed from the `load` of the base field, with some extra code to make it clean.  im going to replace it with this:

```php

      // Select choices would be loaded in the base field's load method.
      
      $select = $items[0]['select_choices'];

      $form = drupal_get_form('tripal_example_map_organism_featuremap_selector_form', $select);
```

bradfordcondon:
What are you suggesting we have a screenshot of?  The built select box that does something when selected? 


>Do we need to have this mix? Couldn't we just use the tripal_example_map_genetic_map_overview_form() in the callback function itself? No need to actually submit anything.

My understanding is the AJAX callback function shouldn't ever have logic in it.    It should just return the piece of form that needs to be rebuilt. 



>No need to actually submit anything.

You're right that we never use the `submit` function for the form.  But, we need the `form` because thats what drupal's AJAX wants to work on, a `form`.  it wont work on a renderable array.

So, i think the answer to your question is this section from the guide:

```
Drupal has its own special way of doing AJAX! This is important to ensure that changes are executed in the correct order. You should `read the documentation carefully! <https://api.drupal.org/api/drupal/includes%21ajax.inc/group/ajax/7.x>`_  The Drupal AJAX API works best on forms, and field formatters **are not forms**.  Instead, they are `renderable arrays. <https://www.drupal.org/docs/7/api/render-arrays/render-arrays-overview>`_
As such, rather than accepting ``$form`` and ``&$form_state``, they accept ``&$element``, ``$entity_type``, ``$entity``, ``$langcode``, ``$items``, and ```$display``, where ``$element`` is the renderable array.
```

We're saying that the AJAX has to go in a form because Drupal AJAX callbacks only work on forms, and thats why they pass `form` and `form_state` and not, for example, `element`.

mboudet:
Ok, I was missing that the ajax callback actually reload the form (thus setting $form_state['values']..)
My bad! No need for screenshot.

Otherwise, it seems fine to me.

bradfordcondon:
ok, great.  Basically the challenge here is to not write a Drupal AJAX guide, because that exists on Drupal's site.... BUT its still kind of confusing so i foresee lots of questions etc.

I think that the comment on the callback addresses this actually-
```
 /**
   * The callback will return the part of the form you want to re-draw.
   */
```
Sounds like it has your approval?


Thank you so much for spending the time to review @mboudet really appreciate it.

PR #847, opened 2019-02-05
Update job_management.rst

laceysanderson:
# Documentation 

Issue #387 

## Description
There was a very small typo in the Tripal Daemon installation docs. This PR fixes it.

## Testing?
This change is so small you should be able to test by simply looking at the change. It turns out that Drush commands are case-sensistive 😝


bradfordcondon (approved):

bradfordcondon:
thanks for the fix

PR #855, opened 2019-02-18
Fix #854 (empty field behaviour)

### LINKED ISSUE #854: Incorrect behaviour for empty fields
### System information
System info:
Tripal: 7.x-3.0
Drupal : 7.59
Postgres: 9.5.8
Php: 7.1.23

### Issue description

Hello,

Currently, there is an issue for several fields (at least two), where they do not display properly when empty.

image

Based on the code, both fields should display a message such as 'No *** is available', however they are currently empty or malformed.

The issue comes from the default value set in these fields : 
`$entity->{$field_name}['und'][0]['value'] = ''` or `$entity->{$field_name}['und'][0]['value'] = array()`, so it goes into the for each loop even though it is technically empty.

mboudet:
Fix #854 by adding a check for an empty value in $item['value].
Without it, Tripal behaves as if there was an actual value in it, and does not display properly.

image

There might be others fields with this behaviour.

### DIFF
--- a/tripal_chado/includes/TripalFields/schema__alternate_name/schema__alternate_name_formatter.inc
+++ b/tripal_chado/includes/TripalFields/schema__alternate_name/schema__alternate_name_formatter.inc
@@ -15,7 +15,9 @@ class schema__alternate_name_formatter extends ChadoFieldFormatter {
     $list_items = array();
 
     foreach ($items as $delta => $item) {
-      $list_items[] = $item['value'];
+      if(! empty($item['value'])){
+        $list_items[] = $item['value'];
+      }
     }
     $list = 'There are no synonyms.';
     if (count($list_items) > 0) {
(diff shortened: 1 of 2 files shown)


almasaeed2010:
This makes sense to me. It's obviously a mistake to let the field an empty line when there is no data. This solution also makes sure that "hide empty fields" works by keeping the value array empty and prints the message in the formatter if the the "hide empty fields" is not checked. so win win.

Thanks @mboudet!

almasaeed2010 (approved):

almasaeed2010 (approved):
Did a functional test and the results are as expected.

bradfordcondon:
thanks @mboudet !  looks good to me as well.  

This will get merged in after the 3.1 release.

bradfordcondon:
sorry for the delay, we were waiting for the 3.1 release.  Thank you for the contribution!

Issue #863, opened 2019-02-27
chado_db_query must be improved


bradfordcondon:
## Feature Request/Discussion

* This feature [does ] attempt to solve an existing problem with Tripal
* This feature is [somewhat URGENT]

edit: i wrote this issue thinking we need to fix the lack of join support but really i think the base class is just broken.  see this [basic test](https://github.com/tripal/tripal/commit/c400380bcf71) which is failing right now.

### Description
the `chado_db_query()` function returns a `ChadoDatabaseConnection` [see here](https://github.com/tripal/tripal/blob/7f35a59c5e48bd173d06da51e28d8abbfebc8594/tripal_chado/includes/ChadoDatabaseConnection.inc).    I'm using it as per our discussion in the #774 instead of `db_select('chado.table_name')`. 

```
    $connection = chado_db_select('pub', 'p');
    $pub = $connection->fields('p')->condition('p.uniquename', $uname)->execute()->fetchObject();
```

However, what if i need to join to another chado table?
```
$connection = chado_db_select('pub', 'p');
$connection->join('analysis_pub', 'ap');
    $pub = $connection->fields('p')->condition('ap', $analysis_id)->execute()->fetchObject();
```
this wont work. It should function similarly to chado_query where the chado table names go in a `{}` and are replaced by the API.

consider this failing test:

```php

  /**
   * @group sad_panda
   */
  public function testChadoInsertRecordCanJoin(){

    $pub = factory('chado.pub')->create();
    $analysis = factory('chado.analysis')->create();

    chado_insert_record('analysis_pub', ['pub_id' => $pub->pub_id, 'analysis_id' => $analysis->analysis_id]);
    $connection = chado_db_select('pub', 'p');
    $connection->join('analysis_pub', 'ap', 'ap.pub_id = p.pub_id');
//  $connection->join('{analysis_pub}', 'ap', 'ap.pub_id = p.pub_id');
    $connection->fields('p');
    $connection->condition('ap.pub_id', $analysis->analysis);
    $result = $connection->execute()->fetchObject();

  // this assertion fails
    $this->assertNotFalsel($result);

// so does this one obviously
    $this->assertEquals($result->pub_id, $pub->pub_id);

  }
```

I can think of other problems. How do i join across public and chado schema with this method?  What if i want to start from a public table and join into a chado table?

## additional problems?
switching to chado_db_select breaks my functionality but perhaps i'm using it wrong.
[...]

Issue #431, opened 2018-06-13
General table prefixing issue in chado_query


guignonv:
Tripal Versions: 2.x and 3.x (13/06/2018)
Drupal Version: not relevant
PostgreSQL Version: not relevant
PHP Version: not relevant

In tripal_core.chado_query.api.inc, the chado_query function prefixes tables using the curly brackets {} for Chado tables and square brackets [] for Drupal tables. The following piece of code is currently used:
```
 $chado_schema_name = chado_get_schema_name('chado');
    $drupal_schema_name = chado_get_schema_name('drupal');
    $sql = preg_replace('/\{(.*?)\}/', $chado_schema_name.'.$1', $sql);
    $sql = preg_replace('/\[(\w+)\]/', $drupal_schema_name.'.$1', $sql);
```
This raises 2 issues:
1) when the user wants to use curly or square brackets for other things that table names, it will be a problem. For instance, if I'd like to select any feature with a uniquename that begins with capital A or B, I would write a query like that (use "drush php" for testing):
```
$sql_query = "SELECT * FROM {feature} WHERE uniquename ~ '^[AB]';"
$results = chado_query($sql_query);
drupal_set_message('Query: ' . $results->getQueryString());

```
Such code will in fact return nothing while there should be some results as the query will be turned into:
`SELECT * FROM chado.feature WHERE uniquename ~ '^public.AB';`

Then

2) replacing the square brackets with just "public." will break Drupal behavior: Drupal uses curly brakets arround table names in order to prefix them with the database prefix used by the Drupal installation (@see https://api.drupal.org/api/drupal/includes%21database%21database.inc/7.x ). For some reasons, people may want or need to prefix all Drupal tables with something and removing curly brackets around Drupal table names prevents Drupal from doing its job.
Some people may wonder why using a table prefix: there are many cases in which it can actually be useful. The first one I think of is when you want to share Drupal users between several Drupal instances. It is possible to specify in your settings.php files that the user table are not prefixed while all the other Drupal tables are. Then when a user is created on a Drupal instance or another, it will be also created on other (well it's a bit more tricky than this but it works).

I'm sorry I don't have a pull request to fix that (yet) and I didn't came up with a good solution yet but if people have ideas to discuss for a fix, feel free. I just wanted to raise the issue so it would be known and people having trouble using queries like mines would understand why sometimes, they might not have expected results... (and also, using Tripal prevents them from using Drupal table prefixing feature).

guignonv:
[...]

PR #867, opened 2019-03-01
863 fix chado_db_select function

bradfordcondon:
# Bug Fix

Issue #863 

## Description
The `chado_db_select`  API function was incompatible with db transactions.  This was a problem for both tests and importers.  Furthermore, it only resolved `select` statements, and not joins.  It also didnt allow joining across public/chado schemas.

  To resolve, the transaction issue, we no longer create a new connection.  In addition, we utilize a [Drupal Query Extender](https://www.drupal.org/docs/7/api/database-api/dynamic-queries/extenders) to modify the query returned by `chado_db_select`.

## Testing?
I've added a unit test that demonstrates that the problem (`test_chado_db_select`)- the query with `chado_db_select` fails because the test takes place in a transaction, but the identical query succeeds.  On this branch, the test will pass.    

## Additional Notes (if any):
This PR was co-authored with @almasaeed2010


bradfordcondon:
the breaking test made us realize that `chado_query` and `chado_db_select` behave differently.  We should [discuss what we want to do before merging](https://github.com/tripal/tripal/issues/863#issuecomment-468747801).

almasaeed2010:
I think this is ready. [The new extender class](https://github.com/tripal/tripal/pull/867/files#diff-2afcc316886a934e8cf98c911333ddbb) no longer establishes a new connection and instead overrides  the select  and the join (and its variants) functions of the SelectQuery class to prefix chado tables with the correct schema. It also looks for `public.` and replaces it with the proper schema name if different.

bradfordcondon:
pointing this out because its hard to distinguish failing tests from the current NCBI-related build error:

Looks like `test_chado_db_select_recognizes_non_chado_tables` is failing on Travis
```

Time: 4.14 minutes, Memory: 60.00 MB
There was 1 error:
1) Tests\ChadoQueryTest::test_chado_db_select_recognizes_non_chado_tables
PDOException: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "FROM"
LINE 2: FROM 
        ^
/home/travis/build/tripal/drupal/includes/database/database.inc:2227
/home/travis/build/tripal/drupal/includes/database/pgsql/database.inc:111
/home/travis/build/tripal/drupal/includes/database/select.inc:1280
/home/travis/build/tripal/drupal/includes/database/select.inc:724
/home/travis/build/tripal/drupal/sites/all/modules/tripal/tests/tripal_chado/api/ChadoQueryTest.php:108

```

almasaeed2010:
Looks like the tests are passing now. 
Thanks @bradfordcondon for the help!

I'll mark this as ready for review.

spficklin:
@almasaeed2010 in the process of trying to fix merge conflicts with the big reformatting that was just merged into the master branch there were some new functions added that showed up as new and when asking @bradfordcondon about them on Slack he wasn't 100% sure about them.  Can you confirm that these functions: `chado_replace_schema_prefix` and `chado_replace_table_prefix` should be included in the PR?

almasaeed2010:
Hello @spficklin. Yes both should be included although we might want to rename them.

|Function|Description|
|--------|------------|
|`chado_replace_schema_prefix`|Takes in a table name with or without schema such as `chado.analysis`, `public.users`, `users` and finds the correct name for the schema. For example, chado.analysis will be replaces with `chado_get_schema_name('chado').'analysis'`.|
|`chado_replace_table_prefix`| Takes in SQL and replaces braces with table names.|

Since the `chado_replace_schema_prefix` function is only used in the extender class, I'll move it there as a static function.

almasaeed2010:
Moved function to the extender class as `static::getRealSchema()` and added unit tests for it.

spficklin:
Thanks @almasaeed2010. Do you have a suggestion for renaming of those?  Once they are in the API it's really hard to change names.  So, we should do that now rather than later.

almasaeed2010:
The `chado_replace_schema_prefix` is not part of the extender so we don't need to worry about that. And I think `chado_replace_table_prefix` can be left as is since it's no longer confusing.

bradfordcondon:
@almasaeed2010 if these were in a class both of those functions would be `private` right? 

Can we can move them out of the `.api` file and/or prefix them with `_` to get around this issue?


`chado_replace_schema_prefix` I no longer see this in the code.  however its still referenced in `getTable` so we should remove that
`chado_replace_table_prefix` is only used 1x so it could be put directly in the code (which was how it started!) or prefixed with `_` to make clear its not part of the API...

tripal_chado/includes/ChadoPrefixExtender.inc
```diff
@@ -0,0 +1,262 @@
...
+   * If the table name has a schema name as a prefix, replace it with the
+   * correct schema name.
+   *
+   * @param string $table
+   *   The table name.
+   *
+   * @return string
+   *   The table with the correct prefix.
+   *
+   * @see chado_replace_schema_prefix()
```
bradfordcondon: This function was removed
```suggestion
```

almasaeed2010:
To clarify, `chado_replace_schema_prefix` no longer exists. And I just removed `chado_replace_table_prefix` and used the code directly in chado_query as it was before so it's no longer an API function.

Thanks!

spficklin (approved):

spficklin:
After having looked at the code changes and testing I see no problems with this fix.  I approve.

bradfordcondon:
grateful to Stephen for the review.  Is anyone else able to review this? It's blocking me from submitting Tripal HQ as a gold standard module. Apologies for the pings, know we're busy.  @laceysanderson @shawnawsu @guignonv @Ferrisx4 @chunhuaicheng

guignonv:
@bradfordcondon: Too busy this week unfortunately but if nothing is done by then, I'll check/validate the changes next week.

guignonv:
I've created a new dev website to test this. I think the changes may have a deep impact on how the multi-chado module works so I'll have to dig a bit and test a couple of things. I think we should also reference issue #431 as it is related.
I'll do some tests and come back to you.

I'll just add a point about the Multi-Chado module: to trick Drupal/Tripal and handle different linking tables between Chado instances, it dynamically replace the linking tables (chado_XXX for Tripal 2.x and chado_bio_data_XXX for Tripal 3.x) with a temporary view which will "redirect" to the appropriate tables corresponding to the right chado schema. I know it's a bit short to explain how it works and to understand... but basically, it means that if a "public." prefix is hardcoded in queries (and there's no hook to remove it), the multi-chado won't work.

**Edit**: just to add more information about how postgres handles schema, when querying a table, postgres not only uses the "search_path" but first look at the temporary objects. So if you do `SELECT * FROM foo;` and uses a search_path like `"$user", public, chado`, postgresql will first query the temporary table/view `foo`, then the one from the user schema, then the one from the public schema and finally the one from the chado schema. However, if your query is `SELECT * FROM public.foo;`, there's no ambiguity and postgres will use the public schema without taking into account any temporary table.

spficklin:
Hi @guignonv good catch on the hard-coded `public` and `chado` schemas. I fixed those and just pushed to this same branch. They were hardcoded in the `getTable` function.  But all was good in the `getRealSchema` function.

Can you review again please?  If you don't have time to do a full functional test, as long as you are happy with a visual code review I think we can merge, as I've already tested it.

To keep our PRs more simple I think we should fix issue #431 in a different PR rather than try to fit it into this one. I'm not sure if that's what you meant when you referred to it.

guignonv:
I have quite a by stack of tasks this week. I'll try to find a moment next week.

laceysanderson:
A note regarding **"Some checks were not successful"**. As you can see [here](https://travis-ci.org/tripal/tripal/jobs/524709325#L991) all tests against the 7.x-3.x branch pass ✅ . The failure is happening during the [tripal 2 upgrade tests](https://travis-ci.org/tripal/tripal/jobs/524709325#L1117) which is a problem already solved in previous PRs.

laceysanderson (approved):
I did both a code and functional test and it all looks good 👍 

Specifically, I tested 
- [x] a number of different fetch methods
- [x] queried a non-standard chado table (phenotype_ to ensure additional columns showed up and that I could filter by them
- [x] queried a chado custom table (genotype_call) and joined on it to a standard chado table (project)
- [x] Union between feature and stock selecting the name, uniquename

laceysanderson:
Since this has two approvals as required by our guidelines, I'm merging it. If you run into any problems with it @guignonv please comment on the associated issue!

Issue #868, opened 2019-03-03
Description field assumes text filter machine name


bradfordcondon:
## BUG/ERROR report
After migrating a tripal 2 site to tripal 3 and migrating content, all content with description fields wont display descriptions.
The description widget has the full description, with the text format set to: `-select-`.    Picking any option and submitting results in still no description, and, if i go back to edit, the text format is set back to `-select-`

<img width="1055" alt="screen shot 2019-03-03 at 5 39 01 pm" src="https://user-images.githubusercontent.com/7063154/53703039-57b32080-3ddb-11e9-929b-592b9a7273d9.png">

<img width="942" alt="screen shot 2019-03-03 at 5 40 59 pm" src="https://user-images.githubusercontent.com/7063154/53703054-87622880-3ddb-11e9-93dd-5803ca6cd180.png">

some similarity to #721 , i had those WYSIG editors enabled prior to migration, I deleted them afterwords but no help.

If I change the bundle-level field setting to use hte plain text processing, then the field shows up (but without the html tags being processed, so thats not a fix.  but it suggests to me that something is going on at the widget level)

<img width="527" alt="screen shot 2019-03-03 at 5 45 16 pm" src="https://user-images.githubusercontent.com/7063154/53703119-31da4b80-3ddc-11e9-9648-d76e3c7700f2.png">

The problem also occurs if i create a new organism- i can select the filter type but the field doesnt display and editing shows no filter tpe is selected.

bradfordcondon:
This looks to be the problem: I dont have a `full_html` filter availalbe on this site (despite it showing up in the select box?  and despite it working if i manually create a multiline text field and select the `full_html` filter?)


![screen shot 2019-03-04 at 9 40 08 am](https://user-images.githubusercontent.com/7063154/53740200-ac00e380-3e61-11e9-8d22-5a581e654c09.png)

almasaeed2010:
In a previous version of drupal the text formats were represented as numbers. The new drupal way is to give them a machine name (using alpha numerical characters instead of only numbers). Unfortunately, I don't remember finding a way to fix this issue. I ended up having to create new fields and marking old ones with a do not use label.

almasaeed2010:
Go to `/admin/config/content/formats` to see the formats. Click configure to see the machine name.

bradfordcondon:
Yep that looks like whats going on:

```
select * from filter_format;
format	name	cache	status	weight
4	File Tree	1	0	-7
[...]

PR #871, opened 2019-03-06
run phpstorm base style formatter

bradfordcondon:
# Style Fix

Issue #310

## Description

This PR runs the PHP storm drupal style standard. This is less intense than the PHPCS drupal standard.

what gets changed:

* `array()` -> []
* white space 
* line length
* weird nested if elseifs
* other stuff...
  
screen shot 2019-03-05 at 8 09 47 pm


tripal_chado/includes/tripal_chado.fields.inc
```diff
@@ -208,6 +212,9 @@ function tripal_chado_bundle_fields_info_base(&$info, $details, $entity_type, $b
       $base_info['type'] = 'text';
       $base_info['settings']['text_processing'] = 0;
     }
+
+    dpm($base_info);
```
laceysanderson: How did this get added? Was testing and saw dpms when finding new fields 🤣 

```suggestion
```

laceysanderson: Confirmed this is the only one that was introduced.

laceysanderson (commented):
Other then the introduced dpm, everything else looks good. I checked a number of pages, cleared the cache 3X, looked for new fields, created content types, etc. (overkill really 😝 ) and everything worked the same way it did before

tripal_chado/includes/tripal_chado.fields.inc
bradfordcondon: ack!  I was troubleshooting issue #868 , thats where this DPM came from.
 Thank you for confirming this was the only one.

laceysanderson (approved):
I checked locally and all seems good 👍 Ignoring the failed test since the log implies it's the pubmed test being unable to query Entrez

spficklin:
I did not do as thorough a review as Lacey. Just checked for a few major admin pages and it seems good. Nice to have the reformatting done!

spficklin (approved):

Issue #870, opened 2019-03-05
Warning with the new version Tripal v3.1


Yosra3:
Hello 
After updating tripal v3.1 i have this warning on the home page : 

Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in require_once() (line 64 of  /tripal/sites/all/modules/tripal/tripal_chado/tripal_chado.module).

what it means? it is a problem of updating ? 

Thank you

almasaeed2010:
Hello @Yosra3, 

Could you please provide information on your system?

```
Drupal version:
PHP version:
```

Thanks!

almasaeed2010:
You can get this information by visiting the status report page at `/admin/reports/status`

Yosra3:
@almasaeed2010  thanks for your reply 

drupal-7.64
PHP 7.3.2

almasaeed2010:
Ok I managed to get my hands on a vm that has php7.3 installed. This error only happens in php7.3:

<img width="1440" alt="screen shot 2019-03-06 at 9 08 51 am" src="https://user-images.githubusercontent.com/1512664/53887095-89043a00-3fef-11e9-8696-41e09fa41cdf.png">

There is a simple fix for this. I'll push the changes today but it probably won't be available until tripal version 3.2. The easiest way to get rid of this for now is to add the following lines to the end of your `sites/default/settings.php` file:

```php
error_reporting(0); 
$conf['error_level'] = 0;
ini_set('display_errors', false);
ini_set('display_startup_errors', false);
```

Thanks!

Yosra3:
thank you i have no more this warning !!

bradfordcondon:
Thank you @Yosra3 @almasaeed2010  for reporting and fixing.

PR #873, opened 2019-03-06
Fix loop error

almasaeed2010:
# Bug Fix

Issue #870

## Description

This issue appears on php version 7.3+. As indicated in the [PHP documentation](http://php.net/manual/en/control-structures.continue.php), the switch statement treats `continue` the same as `break`. The code is obviously intended to skip this iteration of the parent foreach loop. Therefore, switching to `continue 2` allows us to tell php we don't mean the switch statement but the foreach loop.

## Testing?

Run the code in a php 7.3 environment before and after the fix.

## Additional Notes (if any):

Drupal has encountered this error before and [they fixed it](https://www.drupal.org/project/drupal/issues/2989734) using the same method.

Thanks!

### DIFF
--- a/tripal_chado/includes/tripal_chado.fields.inc
+++ b/tripal_chado/includes/tripal_chado.fields.inc
@@ -151,7 +151,7 @@ function tripal_chado_bundle_fields_info_base(&$info, $details, $entity_type, $b
         break;
       case 'blob':
         // not sure how to support a blob field.
-        continue;
+        continue 2;
         break;
       case 'int':
         $base_info['type'] = 'number_integer';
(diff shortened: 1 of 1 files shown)


bradfordcondon:
hi @almasaeed2010 can you confirm that `continue 2` is supported by PHP 7.1/7.2? Sorry if that should be obvious to me from the documentation link, im tired 🛌

laceysanderson (approved):
Confirmed that this has the desired effect in 7.1 through `execute php code`:
```
for ($i=0;$i<=5;$i++) {

  switch ($i) {
    case 0:
      dpm('aaaaaaa');
      break;
    case 1:
      dpm('bbbbb');
      break;
    case 2:
        dpm('ccccccc');
    default:
       continue 2;
  }
  dpm('MEMEME');
}
```
which only showed `MEMEME` 3 times using `continue 2` where it shows 5 times using `continue`.

I also ensured this branch didn't cause any unexpected errors when clearing the cache, visiting the home page, loading entities or looking for new fields.

almasaeed2010:
@bradfordcondon, confirmed that `continue N` has been supported since PHP 4 👍 

Thanks for the quick and thorough review @laceysanderson 😃

bradfordcondon:
great thanks all

PR #883, opened 2019-03-08
Add error checking in chado_get_minimal_pub_info

almasaeed2010:
# Bug Fix

## Description

This PR adds simple error checking. When data is missing, this function issues the following errors:
```
[site http://127.0.0.1] [TRIPAL ERROR] [TRIPAL_CHADO] Missing $object argument to the chado_expand_var function from chado_get_minimal_pub_info (line: 1211)
[site http://127.0.0.1] [TRIPAL ERROR] [TRIPAL_CHADO] Missing $object argument to the chado_expand_var function from chado_get_minimal_pub_info (line: 1229)
```
This PR mitigates that by making sure data exists before passing it to the next function. This same method is used everywhere in the function except for the `$abstract` and `$authors` variables.

## Testing?

To test, you'll need to provide the `chado_get_minimal_pub_info` with a publication that doesn't have an abstract or authors associated with it.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.pub.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.pub.api.inc
@@ -70,7 +70,7 @@ function chado_get_publication($identifiers, $options = []) {
     );
   }
 
-  // If one of the identifiers is property then use 
+  // If one of the identifiers is property then use
   // chado_get_record_with_property().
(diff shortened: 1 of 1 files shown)


bradfordcondon:
Looks good!  I think this qualifies as a single review.  Thank you!

Issue #884, opened 2019-03-08
Cross Reference Field Adds Empty Records


almasaeed2010:
### System information

* Tripal Version: master branch
* Drupal Version: 7.6
* PostgreSQL Version: 9.6
* PHP Version: 7.1

### Issue description

The `sbo__database_cross_reference` field creates an empty placeholder that does not get removed in the load function. This is causing the field to always to display the message "There are no cross references," even though the `hide empty fields` option is enabled on the bundle. See the lines below:

https://github.com/tripal/tripal/blob/202275b3c3252fd5aab68f53a282b0e3ea79513d/tripal_chado/includes/TripalFields/sbo__database_cross_reference/sbo__database_cross_reference.inc#L139-L147

#### Steps to reproduce

- Attach the field to any bundle.
- visit an entity in that bundle that doesn't have a dbxref record.

#### Error messages and screenshots

<img width="367" alt="screen shot 2019-03-08 at 12 28 58 pm" src="https://user-images.githubusercontent.com/1512664/54044768-c5759880-419d-11e9-8561-a0e759752745.png">

bradfordcondon:
merged, thank you @almasaeed2010

PR #885, opened 2019-03-08
Remove default empty record in cross reference field

almasaeed2010:
# Bug Fix

Issue #884 

## Description

Simply removed extra code.

## Testing?

Using this branch:
- Visit an entity that does not have cross reference data
- When the `hide empty fields` option is enabled, the field should disappear. Otherwise, the message `There are no cross references` should appear.


laceysanderson (approved):
I tested this on a fresh Tripal site. The dbxref field acted as expected with a no results message when "hide fields" was unchecked and disappearing when it was checked. I also double checked that it doesn't affect addition of cross references and all is good 👍

laceysanderson:
Since this is a small code change and I feel this can be a single review PR. Since I did a thorough functional test, I'm going to merge it :-) Thanks for the fix @almasaeed2010!

Issue #852, opened 2019-02-13
unable to publish many content types due to no module widget callback


bradfordcondon:
## BUG/ERROR report
I'm currently unable to publish most content types on my site.

undefined function `_field_widget_form()` in `tripal_chado.publish.inc` line 96.

<img width="1274" alt="screen shot 2019-02-13 at 10 54 36 am" src="https://user-images.githubusercontent.com/7063154/52724700-d86dc200-2f7d-11e9-8f2f-d3ab471b5550.png">

It seems the only content i don't have this error for is project.

https://github.com/tripal/tripal/blob/52bec75150e4518b3f617c179abc34105d1cdaba/tripal_chado/includes/tripal_chado.publish.inc#L91-L97

here are some `$instance`s that will trigger this error:

<img width="1010" alt="screen shot 2019-02-13 at 11 01 07 am" src="https://user-images.githubusercontent.com/7063154/52725212-be80af00-2f7e-11e9-90e3-d811dd34212d.png">

<img width="870" alt="screen shot 2019-02-13 at 11 01 21 am" src="https://user-images.githubusercontent.com/7063154/52725234-ca6c7100-2f7e-11e9-9e7a-a6b7b422fef8.png">

in both cases, $instance['widget']['module'] is null.
The obvious workaround is to wrap the callback function in something like ` if (!$instance['widget']['module']) {}` so we dont get this error.  But considering this error just popped up for me, im wondering if something has gone wrong...

laceysanderson:
Hmmm.... I just tested this on a new Tripal site and am not able to reproduce. Specifically, I published analysis, genes and cultivars. In all three cases `$instance['widget']['module']` is correctly set to `tripal`. Are you still experiencing this problem @bradfordcondon ?

bradfordcondon:
no, but i never figured out what caused the issue :|

spficklin:
This fix got merged, so I'm closing.

PR #900, opened 2019-03-21
Add more error checking to the publish form.

laceysanderson:
# Bug Fix
Issue #852 

## Description
Adds a little more error checking to the lines causing the error reported in #852. The original issue is not reproducible but the error checking and default I added are pretty generally applicable.

## Testing?
Try to publish content and ensure I haven't broken the form ;-p Other then that, a code review should be sufficient.

### DIFF
--- a/tripal_chado/includes/tripal_chado.publish.inc
+++ b/tripal_chado/includes/tripal_chado.publish.inc
@@ -90,7 +90,7 @@ function tripal_chado_publish_form($form, &$form_state) {
 
         // Now call the widget callback function to let the module adjust the
         // element as needed.
-        if (!isset($instance['widget']['module'](( {
+        if (!isset($instance['widget']['module'])) {
           tripal_report_error(
             'tripal_chado',
             TRIPAL_ERROR,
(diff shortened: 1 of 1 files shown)


bradfordcondon:
publish page now crashes

```
ParseError: syntax error, unexpected '{' in menu_execute_active_handler() (line 93 of /Users/bc/tripal/sites/all/modules/tripal/tripal_chado/includes/tripal_chado.publish.inc).
```

looks like a missed brace

tripal_chado/includes/tripal_chado.publish.inc
```diff
@@ -90,10 +90,22 @@ function tripal_chado_publish_form($form, &$form_state) {
 
         // Now call the widget callback function to let the module adjust the
         // element as needed.
+        if (!isset($instance['widget']['module'](( {
```
almasaeed2010:
```suggestion
        if (!isset($instance['widget']['module'])) {
```

almasaeed2010: @bradfordcondon you should try again with this fix

bradfordcondon: Parenthesis in wrong direction 😈 
```suggestion
        if (!isset($instance['widget']['module'])) {
```

bradfordcondon: yep agreed

spficklin:
This worked fine for me when testing. I understand the code changes, and I see the fix @bradfordcondon made.  I think this is good and can be merged.   Travis CI is failing but on the custom quota and not related to this fix, so I'm merging.

Issue #561, opened 2018-08-20
Error message: array_key_exists() expects parameter 2 to be array


risharde:
Hi guys, any idea what causes this error message? I keep seeing this at the top of the pages. It doesn't seem to affect functionality per say but would like to get to the bottom of this one. Please let me know, thanks!
> 
Warning: array_key_exists() expects parameter 2 to be array, boolean given in tripal_get_term_details() (line 396 of /var/www/Drupal/sites/all/modules/tripal/tripal/api/tripal.terms.api.inc).

spficklin:
Hi @risharde we'll look into it.  I think you're right though, it's just an annoyance.   Can you give some examples of pages where you see this appear so we can confirm?

laceysanderson:
@risharde are you still experiencing this? Can you provide us with some examples of pages where you see this appear so we can confirm?

bradfordcondon:
i haven't seen this error in a long while. 

Could be the `array_key_exists` call at 358 (different from 396 in the log, but the file has changed  since this issue was posted )

https://github.com/tripal/tripal/blob/13679f3c55c555b33e87ced5c9aab2204bf0cdaf/tripal/api/tripal.terms.api.inc#L339-L361

if im right, its cases where `$term` is null.  We know `$vocabulary` and `$accession` arent null because we check for that first... 

we could add a `!$term` check before `array_key_exists`?  what do y'all think...

laceysanderson:
I like the idea of adding a `!$term` check as it's own "if stanza". That way, we can log an error and return FALSE since if there's no term, we can't find details for it.

```php
if (!$term) {
  tripal_report_error(...);
  return FALSE;
}
```

spficklin:
This looks like it was fixed but never closed.

PR #901, opened 2019-03-21
Add term check when retrieving term details.

laceysanderson:
# Bug Fix
Issue #561 

## Description
This PR adds a simple check to function `tripal_get_term_details()` to ensure we don't try to access array keys without first checking to see if we have a term.

## Testing?
I'm not able to duplicate the original problem. However you can prove the function still works as follows:
```php
// Works.
$details = tripal_get_term_details('EFO','0000269');
dpm($details, 'details');

// Should not work but no errors.
$details = tripal_get_term_details('UFO','0000269');
dpm($details, 'details');
```

### DIFF
--- a/tripal/api/tripal.terms.api.inc
+++ b/tripal/api/tripal.terms.api.inc
@@ -361,7 +361,7 @@ function tripal_get_term_details($vocabulary, $accession) {
         tripal_report_error(
           'tripal',
           TRIPAL_ERROR,
-          "Unable to find term for :vocab, :accession.",
+          "Unable to find term for :vocab, :accession using :function.",
           [':vocab' => $vocabulary, ':accession' => $accession, ':function' => $function]
         );
         return FALSE;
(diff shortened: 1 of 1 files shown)


bradfordcondon (approved):
Thank you for providing test code, made reviewing this even easier!

What do you think about also logging the function name?

tripal/api/tripal.terms.api.inc
```diff
@@ -355,7 +355,17 @@ function tripal_get_term_details($vocabulary, $accession) {
     $module = $stores[$keys[0]]['module'];
     $function = $module . '_vocab_get_term';
     if (function_exists($function)) {
+
       $term = $function($vocabulary, $accession);
+      if (!$term) {
+        tripal_report_error(
+          'tripal',
+          TRIPAL_ERROR,
+          "Unable to find term for :vocab, :accession.",
```
bradfordcondon:
```suggestion
          "Unable to find term for :vocab, :accession using :function.",
```

tripal/api/tripal.terms.api.inc
```diff
@@ -355,7 +355,17 @@ function tripal_get_term_details($vocabulary, $accession) {
...
+          [':vocab' => $vocabulary, ':accession' => $accession]
```
bradfordcondon:
```suggestion
          [':vocab' => $vocabulary, ':accession' => $accession, ':function' => $function]
```

spficklin:
Code changes look fine, I tested and no issues.  This is such a simple fix I think it only needs 1 review.

Issue #904, opened 2019-03-21
Tripal 2 upgrade process fails


almasaeed2010:
### System information

* Tripal Version: 7.x-2.x
* Drupal Version: 7.6
* PostgreSQL Version: 9.6
* PHP Version: 7.3

### Issue description

Starting from scratch with a new drupal site, installing tripal 2, installing chado and enabling all sub modules such as tripal_feature and tripal_stock then disabling tripal_core causes an error.

#### Steps to reproduce

- Create a new Drupal site
- Install tripal 7.x-2.x (from github directly)
- Install chado using the GUI
- Enable all sub modules: `drush pm-enable tripal_core, tripal_views, tripal_db, tripal_cv, tripal_analysis, tripal_organism, tripal_feature, tripal_pub, tripal_stock`
- Disable tripal_core: `drush dis -y tripal_core`

#### Error messages and screenshots

```
 drush dis -y tripal_core
The following extensions will be disabled: tripal_core, tripal_views, tripal_db, tripal_cv, tripal_analysis, tripal_contact, tripal_organism, tripal_feature, tripal_pub, tripal_stock
Do you really want to continue? (y/n): y
tripal_analysis was disabled successfully.                                                                     [ok]
tripal_contact was disabled successfully.                                                                      [ok]
tripal_core was disabled successfully.                                                                         [ok]
tripal_cv was disabled successfully.                                                                           [ok]
tripal_db was disabled successfully.                                                                           [ok]
tripal_feature was disabled successfully.                                                                      [ok]
tripal_organism was disabled successfully.                                                                     [ok]
tripal_pub was disabled successfully.                                                                          [ok]
tripal_stock was disabled successfully.                                                                        [ok]
tripal_views was disabled successfully.                                                                        [ok]
Unable to find a view by the name of 'tripal_stock_user_stocks'. Unable to disable this view.                  [error]
Unable to find a view by the name of 'tripal_stock_admin_stocks'. Unable to disable this view.                 [error]
Unable to find a view by the name of 'tripal_pub_admin_publications'. Unable to disable this view.             [error]
Unable to find a view by the name of 'tripal_feature_user_feature'. Unable to disable this view.               [error]
Unable to find a view by the name of 'tripal_feature_admin_features'. Unable to disable this view.             [error]
Unable to find a view by the name of 'tripal_organism_user_organisms'. Unable to disable this view.            [error]
Unable to find a view by the name of 'tripal_organism_admin_organisms'. Unable to disable this view.           [error]
[...]

PR #906, opened 2019-03-25
URGENT: Fix Tripal 2 => 3 upgrade

laceysanderson:
# Bug Fix

Issue #904 

## Description
Due to backport of the user file quota system to Tripal 2, the upgrade process now fails since the `tripal_custom_quota` and `tripal_expiration_files` tables already exist. This PR uses the typical Tripal3 method of moving the original tables out of the way, allowing Drupal to create the new one and then replacing it with the original table.

**NOTE: In order for tests to pass the Tripal2 docker needs to be updated. This is due to another commit on 7.x-2.x that fixes a change to views (i.e. views automagically disables views when module is disabled and doesn't allow suppressing of error if tripal tries to re-disable them).**

## Testing?
In this case, Travis will do the testing for you. **As long as the tests now pass, 👍**. 

Excerpt from Travis:
(note the following failed without this PR)
```
$ docker exec -it tripal2 bash -c "rm -rf /modules/tripal && ln -s /tripal /modules/tripal"
The command "docker exec -it tripal2 bash -c "rm -rf /modules/tripal && ln -s /tripal /modules/tripal"" exited with 0.
2.31s

$ docker exec -it tripal2 drush en -y tripal
The following extensions will be enabled: tripal
Do you really want to continue? (y/n): y
tripal was enabled successfully.                                     [ok]
tripal defines the following permissions: administer tripal, access tripal content overview, manage tripal content types, publish tripal content, upload files, view dev helps
The command "docker exec -it tripal2 drush en -y tripal" exited with 0.
```

If you would like to be **extra diligent**:
1. test Tripal2: tripal core can be disabled without error on most recent 7.x-2.x
2. Test upgrade (disable tripal2 and enable tripal3) works without error.
3. Check the tripal file quota system still works after upgrade.

### DIFF
--- a/tripal/tripal.install
+++ b/tripal/tripal.install
@@ -118,6 +118,18 @@ function tripal_enable() {
     db_rename_table('tripal_variables2', 'tripal_variables');
   }
 
+  if (db_table_exists('tripal_custom_quota2')) {
+    $sql = "DROP TABLE tripal_custom_quota";
+    db_query($sql);
+    db_rename_table('tripal_custom_quota2', 'tripal_custom_quota');
+  }
+
+  if (db_table_exists('tripal_expiration_files2')) {
+    $sql = "DROP TABLE tripal_expiration_files";
+    db_query($sql);
+    db_rename_table('tripal_expiration_files2', 'tripal_expiration_files');
+  }
+
   // schema change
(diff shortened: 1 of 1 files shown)


bradfordcondon (approved):
thanks for resolving this!

almasaeed2010 (approved):

Issue #909, opened 2019-03-26
Important caveat for field specific permissions


bradfordcondon:
## BUG/ERROR report

https://tripal.readthedocs.io/en/latest/user_guide/content_types/field_permissions.html?highlight=field%20permissions

i recommend the field permissions module to users looking to customize input options for different users.  @almasaeed2010 has pointed out that configuring the permissions for a given field will also configure those permissions for **all other instances** of that fields.  IE if we change the publication field on organism, its also going to change it for analysis.  Not surprising i guess because they are all the same entity type (TripalEntity)

We should at least update the documentation to warn about this.

spficklin:
This looks like the PR got merged but this issue wasn't closed.

PR #910, opened 2019-03-26
add warning about cross-bundle field permissions

bradfordcondon:
# Documentation

Issue #909 

## Description

Add the warning described in #909.  Altering field permissions for one bundle will alter it on **all** bundles.


almasaeed2010 (approved):
We've experienced this issue on HWG so I think the warning is worth it here. Thanks @bradfordcondon

laceysanderson (approved):

Issue #911, opened 2019-03-27
relationship formatter does not find subject or object analysis entity name


bradfordcondon:
### System information

* Tripal Version: 7.x-3.1
* Drupal Version: 7.59
* PostgreSQL Version:  9.2.23
* PHP Version: 7.1.15

### Issue description

Working on a side upgrading from t2 to t3, i came across the relationship reporting a relationship as:  "The analysis, , derives from analysis, "  Both analyses have `analysis.names`, and the name field is populated.  Yet, `schema:name` is empty for both the subject and object (see dpm).  This problem occurs on bot the subject and object.

![Screen Shot 2019-03-27 at 1 15 13 PM](https://user-images.githubusercontent.com/7063154/55097534-67e4b580-5092-11e9-969d-7d974ea7c0c4.png)

bradfordcondon:
Easy fix:

The constructor doesnt set`  $this->base_name_columns` for analysis, biomaterial, cell_line, quantification, and acquisition.

https://github.com/tripal/tripal/blob/d0d52df57057600aa987d2372c6ab36254828784/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc#L237-L278



setting it to `['name']` populates the field (although i still dont have entity links).

![Screen Shot 2019-03-27 at 1 24 20 PM](https://user-images.githubusercontent.com/7063154/55098162-ae86df80-5093-11e9-8f0e-c63a593e3867.png)


resolved as well, using `substr_replace` instead of regexp replace:


![Screen Shot 2019-03-27 at 1 49 01 PM](https://user-images.githubusercontent.com/7063154/55099992-2d314c00-5097-11e9-9313-be34b72e1ddc.png)

laceysanderson:
PR Merged 🎉

PR #912, opened 2019-03-27
fix bug for names of tables without type_id.  Fix bug where names wit…

bradfordcondon:
…h regexp characters wouldnt replace with the link.

# Bug Fix

Issue #911

### Bug

Analysis relationships did not have the name:

Screen Shot 2019-03-27 at 1 06 32 PM

Relationships between entities whose names have regexp characters (`*` `.` for example) didnt have their title replaced with the link properly.

Screen Shot 2019-03-27 at 1 24 20 PM

### Resolution

* The relationship base accidentally didnt set the "type" for some type-less columns. I compile all said columns into the same case in this PR (and add a note that chado 1.4 will introduce a type for sure...)

* we used `preg_replace` to replace the entity title with the linked entity title.   I dont think we have a need for `preg_replace` so `str_replace` is the better choice and wont break when titles have odd characters in them.

## Testing?

Create an `analysis._A` and `analysis._B` (note regexp characters).  Make sure your relationship formatter is set to display for analysis. 
 Create a relationship between the two with the relationship widget.  You should see the relationship listed in the relationship widget, but without the names.

With this PR

Screen Shot 2019-03-27 at 1 49 01 PM

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
@@ -239,34 +239,29 @@ class sbo__relationship extends ChadoField {
     $this->base_type_column = 'table_name';
     switch ($instance['settings']['chado_table']) {
 
+      // TODO: note that Chado 1.4 will add types to, at least,
+      // project and analysis, at which point you should use the default instead.
       case 'acquisition_relationship':
       case 'analysis_relationship':
       case 'biomaterial_relationship':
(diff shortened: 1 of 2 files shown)


spficklin (approved):

spficklin:
I did not do a functional test. Just a code review.  It seems simple enough that this may only need 1 review.

laceysanderson (approved):
I did both a functional and code review and this looks good 👍 Thanks for the fix @bradfordcondon!

Issue #928, opened 2019-04-26
BUG: Expanding field names in Web services with Spaces


spficklin:
### System information

* Tripal Version: 3.x
* Drupal Version: N/A
* PostgreSQL Version: N/A
* PHP Version:

### Issue description

If you are using web services and you want to expand a field that has a link (expandable field) and if that field name has a space in it, then web services breaks.  

#### Error messages and screenshots

For example if you had the following JSON for mRNA

![ws1](https://user-images.githubusercontent.com/1719352/56775873-ff1a5700-677d-11e9-95f3-157abe19e8d8.png)

Then you could expand `cds`, `annotation`, `relationship` and `sequence_coordinates` but the link for `sequence_coordinates` has a space and that breaks.
```
error: "Could not find a matching field for the name: Sequence coordinates"
```

PR #929, opened 2019-04-26
Web Services field causing problems

spficklin:
# Bug

Issue #928

## Description

If you are using web services and you want to expand a field that has a link (expandable field) and if that field name has a space in it, then web services breaks.

## Testing?

Use the default Citrus data from the Tripal User's Guide tutorial.  Be sure that the `mRNA` content type is published.   Prior to pulling this PR, open web services, find an `mRNA` record follow the `sequence_coordinate` link.  It should give you the following error:

```
error: "Could not find a matching field for the name: Sequence coordinates"
```
Pull the code for this PR and try following the link again.

Alternatively, you can test with any web services field that has a space in it.

Here are before and after screenshots:
ws_before
ws_after

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -161,18 +161,27 @@ class TripalContentService_v0_1 extends TripalWebService {
       $field_name = $instance['field_name'];
       $field = field_info_field($field_name);
       $field_type = $field['type'];
+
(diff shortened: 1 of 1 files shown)


spficklin:
This fix is so simple, I think only one review is needed.

almasaeed2010:
Confirmed this works on my local site using `Sequence coordinates` as an example.

Thanks!

almasaeed2010 (approved):

spficklin:
Thanks @almasaeed2010 for the review!

PR #930, opened 2019-04-26
chado_query table prefixing

spficklin:
# Bug Fix

Issue #431

## Description

The `chado_query` function adds an appropriate schema prefix to database tables prior to execution of a query.  For example the curly bracesare used around tables in the Chado schema  (e.g. `{feature}`) and square braces are used around Drupal tables (e.g. `[tripal_entity]`).  The function then uses a regular expression to replace the braces with the table prefixes with the appropriate Chado and Drupal schemas.  However, if someone needed to use those braces in strings such as a regular expression in the query then those strings would get a table prefix added! That's not right.  

This fix corrects that by only adding a prefix to known Chado or Drupal tables. If the text between the braces is not a Chado or Drupal table then no prefix is added.

## Testing?

Try the following two SQL statements (using Drush) before and after pulling the branch for the PR. The first uses the square braces in a regular expression to pull out any features that start with the letter `o` or `L` and prints the first matching record.  This should work with the orange data used in the Tripal tutorial.  Change the `[oL]` to match any feature names you might have in your database.  

```sql
drush php-eval "\$results = chado_query(\"select * from {feature} where name similar to '[oL]%'\"); print_r(\$results->fetchObject())"
```
After running the query above you should see a result printed in the terminal.  This next SQL statement performs the same query as above, but rather than print out the results, it prints out the SQL statement.
```sql
drush php-eval "\$results = chado_query(\"select * from {feature} where name similar to '[oL]%'\"); print_r(\$results->getQueryString())"
```
Before the fix is applied the SQL statement above was converted to this:

```sql
select * from chado.feature where name similar to 'public.oL%'
```
After it should be converted to this:
```sql
select * from chado.feature where name similar to '[oL]%'
```

### DIFF
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -1747,7 +1747,7 @@ function chado_query($sql, $args = []) {
       }
     }
 
-    // Now set the Drupal prefix if the table is surrounded by square brakets.
+    // Now set the Drupal prefix if the table is surrounded by square brackets.
     if (preg_match_all('/\[(.*?)\]/', $sql, $matches)) {
       $matches = $matches[1];
       $drupal_tables = array_unique(array_keys(drupal_get_schema()));
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -349,9 +349,9 @@ function chado_set_active($dbname = 'default') {
   $search_path = chado_get_schema_name('drupal');
 
   // Change only if 'chado' has been specified.
-  // Notice that we leave the active_db set as chado but use the possibly 
-  // user-altered  schema name for the actual search path. This is to keep 
-  // outward facing mentions of chado as "chado" while still allowing the user 
+  // Notice that we leave the active_db set as chado but use the possibly
+  // user-altered  schema name for the actual search path. This is to keep
+  // outward facing mentions of chado as "chado" while still allowing the user
   // to alter the schema name used.
   if ($dbname == 'chado') {
     $active_db = 'chado';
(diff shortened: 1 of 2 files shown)


tripal_chado/api/tripal_chado.query.api.inc
```diff
@@ -1728,30 +1728,51 @@ function chado_query($sql, $args = []) {
...
+      $matches = $matches[1];
+      $chado_tables = array_unique(array_keys(chado_get_table_names(TRUE)));
+      foreach ($matches as $match) {
+        if (in_array($match, $chado_tables)) {
+          $sql = preg_replace("/\{$match\}/", $chado_schema_name . '.' . $match, $sql);
+        }
+      }
+    }
+
+    // Now set the Drupal prefix if the table is surrounded by square brakets.
```
bradfordcondon:
```suggestion
    // Now set the Drupal prefix if the table is surrounded by square brackets.
```

tripal_chado/api/tripal_chado.schema.api.inc
```diff
@@ -392,11 +392,16 @@ function chado_get_version($exact = FALSE, $warn_if_unsupported = FALSE) {
     $version = "1.11 or older";
   }
   else {
+    // We need to hardcode the schema here rather than rely on the db_query
+    // function to settle it cecause the db_query function needs to know the
```
bradfordcondon:
```suggestion
    // function to settle it because the db_query function needs to know the
```

bradfordcondon (approved):
Thanks for the great testing instructions.  Looks good, works as advertised, thank you!

spficklin:
Thanks @bradfordcondon .  I'm adding the label for your approval

spficklin:
@guignonv I know you're busy, but if you have a little bit of time this should be a quick test for you, and resolves an issue you reported.  We're hoping to get a v3.2 release out in May so every bit helps!

guignonv (approved):
So far, I just reviewed the code and it looks good to me. I approve the changes.

spficklin:
Thanks @guignonv !

Issue #177, opened 2017-10-26
Custom Tripal tables unavailable content type


btski:

spficklin:
Hi @btski .  Apologies for the slow reply on this.  I noticed you closed it without a response and removed the message.  Were you able to figure something out, or would you still like some help?

btski:
Response time wasn't an issue. It was just a poorly conceived question. I've been having trouble depositing information into Chado with the bulk loader, likely due to constraints on input data. I should be able to sort that out, but I've shifted focus to creating a custom Chado table to house all of my metadata (some of which fails to fit well within the existing Chado schema). I'm planning to populate the base Chado tables as well as possible, but I'll also have all metadata stored within a custom table for easier access. 

Currently, I'm trying to figure out how to create a 'Content Type'. By default, Tripal allows new content to link to 20 Chado tables. Most of the available tables are not represented. I've successfully loaded sample data onto my custom table (using the bulk loader), but I'm unable to access it. For reference: I'm storing everything on an AWS instance and am using Tripal v3. To create a new page, I'm going to the 'Structure' tab --> 'Tripal Content Types' --> 'Add Tripal Content Type' --> reference an ontology (I'm using 'Database') --> select Chado table (20 available options, new Chado table unavailable). Also, I'm not entirely sure how the pages work (in terms of displaying data linked to Chado tables). I've been able to deposit and sync/publish organism data (using the bulk loader and following instructions on the Tripal site), but when I successfully load authors into the 'Contact' table I'm unable to publish the data (zero records published).

Thanks for help in advance @spficklin or whoever else happens to see this. I also appreciate that you followed up with my original post despite my attempts to remove it.

Edited for clarification.

spficklin:
Okay, it looks like we have a bug here with custom tables now showing up in that list for creating new content types.  I've marked this as a bug and task to look at.   

Can you describe your custom tables and the data you're putting into them?

btski:
[...]

PR #931, opened 2019-04-27
Custom tables as base tables

spficklin:
# New Feature

Issue #177 

## Description

Users who create custom tables cannot use those tables as base tables.  This is not good because custom tables may be needed for new data types not currently supported by Chado.  Someone may create a custom module that creates custom tables and they may want to publish content from those tables.  

## Testing?

Test with the following:
1) Pull the branch and clear the cache.
2) Run the `drush updatedb` to perform the `tripal_custom_tables` table update.
3) Navigate to Admin > Tripal > Data Storage > Chado > Custom Tables and either edit an existing table or add a new one.  Be sure to check the 'Is this a base table?.  For my testing I just used the `blast_organism` table that comes with the Tripal Analysis Blast module.
4) Go to the Admin > Tripal > Data Storage > Chado > Semantic Web page, find the table name and add vocabulary terms to the fields.
5) Create a new content type that uses the custom table.
6) Publish records for the new content type.  You should have new pages!

The only weirdness with this fix is that I had to force new fields added to custom tables to have a `c_` prefix before all field names. This is because we cannot guarantee that custom tables are using the fields in the Chado style and our built-in Tripal fields may not work. For example.  If someone does use the `blast_organisms` table as a base table, and then use the `OBI:organism` term as the term for the `organism_name` field of that table, it will create a field named `c_obi__organism`.  This is because our `obi__organism` field expects an `organism_id` field and the custom table does not.  Therefore, any custom module that wants to use a custom table, will have to specifically override this.


laceysanderson (changes requested):
I used my genotype_call custom table as a test and was unable to create a content type for it. I set all the semantic web terms:
Screen Shot 2019-05-10 at 2 07 27 PM
and then created the content type but when I run the job it fails:
```
2019-05-10 15:09:24: Calling: tripal_create_bundle(Array)
WD tripal_entities: Failed to create content type: Attempt to create field c_so__sequence_variant with no type..                             [error]
ERROR (TRIPAL_ENTITIES): Failed to create content type: Attempt to create field c_so__sequence_variant with no type..
[site http://default] [TRIPAL ERROR] [TRIPAL_ENTITIES] Failed to create content type: Attempt to create field c_so__sequence_variant with no type..
Failed to create content type: <em class="placeholder">Attempt to create field c_so__sequence_variant with no type.</em>.
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] Failed to create content type: Attempt to create field c_so__sequence_variant with no type..
Failed to create content type: Attempt to create field c_so__sequence_variant with no type..                                                 [error]
```

As you can see if it unable to create my variant_id field even though there is a term attached via the semantic web form.

Additionally, we need an update to the documentation for this PR. Specifically, documentation for how to override the fields to use the default Tripal Fields. For example, I may have an OBI:organism referencing an organism_id and want to use the Tripal Field... how would I do this?

bradfordcondon:
hey y'all, closing this due to inactivity and the failed tests.  Let's resubmit when we're able to revisit!

Issue #783, opened 2018-12-10
Field Pagers


spficklin:
### System information

* Tripal Version: 3.x
* Drupal Version:  7.x
* PostgreSQL Version: N/A
* PHP Version: N/A

### Issue description

When creating a pager for a field you are expected to give it a pager index.  Drupal expects that all pagers have an index this is because you can have more than one on a page and the index is used to distinguish which one the user is expecting to use.  Unfortunately, with fields, we never know how many pagers will be on the page and therefore the field developer cannot properly pick a proper index for the pager.  Instead, we need a way for the entity object to maintain a count of the number of pagers so that it can then be used to set the index number for each pager.

spficklin:
Also, the documentation in the developer's User's Guide needs to be updated.

laceysanderson:
I see the documentation made it in https://tripal.readthedocs.io/en/latest/dev_guide/custom_field/custom_formatter.html?highlight=pager#creating-pagers 🎉 

Is this resolved @spficklin?

spficklin:
I'm not sure!  I forgot all the changes made.  I need to check it.

spficklin:
There is no problem it works as advertised!  I added a PR to fix some bug in the spinner but that's all.

PR #932, opened 2019-04-29
Pager AJAX fix.

spficklin:
# Bug Fix

Issue #783 

## Description

In the process of trying to discover if we had a problem with multiple pagers on the same page (see issue #783) I found out we do not! They work great.  But, there was a bug in that all of the spinners for all pagers on a page would spin anytime one of the pagers was clicked on. While this isn't a functional problem, it's a bit annoying.  This PR removes some older code that is no longer needed and fixes the ajax spinner problem.  

## Testing?

The only way to test this is to run it on a page that has multiple fields with pagers.  A good example would be an gene with blast results and lots of relationships.  

This is such an easy fix I think if no one has time to hunt for a field, that a simple code review is all that is needed.


laceysanderson (approved):
Code review-wise this looks good. I didn't get the chance to do a functional review though...

spficklin:
Thanks @laceysanderson I don't think a functional review is necessary.  This is just a cleanup Thanks for checking.

Issue #938, opened 2019-05-14
some SO- based bundles defined with wrong term name


bradfordcondon:
## BUG/ERROR report

See:  https://github.com/UofS-Pulse-Binfo/nd_genotypes/issues/34

Take the Sequence Variant bundle.  

```php
  //
  // Create the 'Sequence Variant' entity type.
  //
  $args = [
    'vocabulary' => 'SO',
    'accession' => '0001060',
    'term_name' => 'Sequence Variant',
    'storage_args' => [
      'data_table' => 'feature',
      'type_column' => 'type_id',
    ],
    'category' => 'Genetic',
  ];
```
the term is `sequence_variant` not `Sequence Variant`.  unfortunately its important to fix because when fields attach and if they chose to attach via the term name instead of the accession, it will cause problems.

Confirmed that the fix on install is this simple:

```php
//
  // Create the 'Sequence Variant' entity type.
  //
  $args = [
    'vocabulary' => 'SO',
    'accession' => '0001060',
    'label' => "Sequence Variant",
    'term_name' => 'sequence_variant',
    'storage_args' => [
      'data_table' => 'feature',
      'type_column' => 'type_id',
    ],
    'category' => 'Genetic',
  ];
```

bradfordcondon:
PR submitted with an updatedb.

This raises the question... why can we create a  tripalterm without a === matching cvterm?  Answer: because we dont want to be Chado dependent, right?

Issue #681, opened 2018-09-25
Duplicate field EFO:0000269 for Assay


abretaud:
## BUG/ERROR report

During startup, at the "Creating common Tripal Content Types..." step, I get this warning:

`[TRIPAL WARNING] [TRIPAL_FIELDS] The field, efo__array_design, uses a term, EFO:0000269, that is already in use on this content type. The field instance cannot be added.`

Indeed, the same term is used for [array design](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/tripal_chado.semweb.inc#L748) and [assay design](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/tripal_chado.semweb.inc#L778) fields.

### System information

* Tripal Version: 3.x
* Drupal Version: 7.59
* PostgreSQL Version: 9.5
* PHP Version: 7.x

#### Error messages and screenshots

```
error message
```

spficklin:
Yes, I've noticed that too. We need to track it down but it won't cause a problem for you.  It's just trying to create the same field twice.

bradfordcondon:
I'll point out that its the array_design field in both cases: attaching to array, and to assay.  With that in mind, correcting the typo at https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/tripal_chado.semweb.inc#L778 to be the exact same term (IE array design) should fix the error.


edit: actually this whole block is kind of wonky.


https://github.com/tripal/tripal/blob/dc48c7867f546b793a57b5217b8eea427ff3709b/tripal_chado/includes/tripal_chado.semweb.inc#L761-L783



the preceeding term is defined and assigned 2x.  The "assay design" term is defined but not assigned.

laceysanderson:
How about these changes to fix it?
```diff
diff --git a/tripal_chado/includes/tripal_chado.semweb.inc b/tripal_chado/includes/tripal_chado.semweb.inc
index 4d63e9cc..3f412620 100755
--- a/tripal_chado/includes/tripal_chado.semweb.inc
+++ b/tripal_chado/includes/tripal_chado.semweb.inc
@@ -759,7 +759,7 @@ function tripal_chado_populate_vocab_EFO() {
     'cv_name' => 'efo',
     'definition' => 'An instrument design which describes the design of the array.',
   ]);
-  chado_associate_semweb_term(NULL, 'arraydesign_id', $term);
[...]

PR #940, opened 2019-05-15
resolve SO based bundles using wrong term name

bradfordcondon:
# Bug Fix

Issue #938

## Description

Some SO bundles were added using the human readable names instead of the machine readable names. 
These were Sequence Variant, Genetic Marker, and Heritable Phenotypic Marker ie instead of `sequence_variant`.  This is problematic not just because using bad terms is bad, but other modules looking for these bundles wont find them if they look by term name.

This PR changes the term name to match the SO term while setting the label to remain the human readable label. 

## Testing?

Note your bundle names and terms at`/admin/structure/bio_data` for the above terms.  The term column will be the human readable name instead of the machine name.  Run `updatedb` and confirm the term names have been fixed.
Additionally/alternatively build a new site with this branch and confirm bundles have the correct terms.

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -433,6 +433,9 @@ function tripal_add_notification($title, $details, $type, $actions, $submitter_i
  *     - term_name: A human-readable name for this term.  This will became
  *       the name that appears for the content type.  In practice, this
  *       should be the name of the term. (E.g. the name for SO:0000704 is gene).
+ *     - label: An alternative bundle label.  This will be used instead of the
+ *       term name, for cases where the vocabulary term is not human readable.
+ *
  * @param $job
  *  The job ID if this is launched via a job.
  *
@@ -446,6 +449,7 @@ function tripal_create_bundle($args, $job = NULL) {
   $accession = $args['accession'];
   $term_name = $args['term_name'];
   $storage_args = $args['storage_args'];
+  $label = $args['label'];
 
   $message_args = [
     'job' => $job,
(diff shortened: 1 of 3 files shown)


almasaeed2010:
The changes work for me. 
However, I have multiple bundles that use human readable names as terms, see below. Do we need all of those changed? Or is it only SO terms?

Screen Shot 2019-05-15 at 8 40 47 AM

almasaeed2010:
Also, there is a typo in the testing instructions. It should be `updatedb` not `dbupdate` 👍

bradfordcondon:
> Do we need all of those changed? Or is it only SO terms?

it depends entirely on the term vocabulary.  For example, Genome annotation in operation- the term name is [in fact Genome Annotation](https://www.ebi.ac.uk/ols/ontologies/swo/terms?iri=http%3A%2F%2Fedamontology.org%2Foperation_0362) instead of genome_annotation.  If I have time tonight I can double check each term.  However, to my knowledge, its just the SO names.

e also please feel free to edit my posts to correct typos at your discretion :)

bradfordcondon:
i checked the other bundle terms.  Unfortunately you are correct, many other bundles have case issues for the term.
 Assay Design should be assay design but the accession is mixed up with array design, [so i commented in that existing issue](https://github.com/tripal/tripal/issues/681#issuecomment-492875147) so im excluding it from the below list.

- Genetic Map -  should be Genetic map
- Genome Assemby- should be Genome assembly
- Genome Annotation- should be Genome annotation
- Physical Map -> Physical map
- Study -> study

The Crop Ontology terms are crazy.  For example, Recombinant Inbred Line should be   414 inbred line .


@almasaeed2010 I've updated all terms except Assay Design

laceysanderson (commented):
The update works perfectly 👍 
When creating a new site from scratch the labels do not get used but rather the term name does. This is most notably a problem with the CO_010 terms (see green circles below) which should be `Recombinant Inbred Line`, `Germplasm Accession` and `Cultivar (Germplasm Variety)` respectively.
Screen Shot 2019-05-16 at 8 37 30 PM

bradfordcondon:
thanks for the catch.

So weird that it ignores label key and just converts machine readable names to human automatically on install, but, when using the controller directly, it works as i'd expect.  I guess I need ot factor out the updatedb and what, submit a job to run it after install?  Perhaps it will be easier to look at the Tripal API and make sure its handling the label key....

bradfordcondon:
@laceysanderson I modified `tripal_create_bundle` to accept an additional `label` argument which will override the default behavior of humanizing the term name if supplied.

the resulting bundles on install:

Screen Shot 2019-05-17 at 10 09 03 PM

laceysanderson (approved):
Changes look perfect from a code perspective and work as expected 👍 Thanks for the fix @bradfordcondon

almasaeed2010:
Works great. Thanks @bradfordcondon

almasaeed2010 (approved):

bradfordcondon:
2 approvals, merging.  Thank you both for the reviews.

PR #944, opened 2019-05-26
Tripal Blast: Request for Gold Badge Status

laceysanderson:
# Documentation

## Description

We are applying for the Gold-level badge for Tripal Blast as we meet the following requirements. I have provided examples below to make review easier.

Repository: https://github.com/tripal/tripal_blast/

## Bronze

| Requirement                                                                              | Status        |
|-------------------------------------------------------------|------------|
| Has a public release.  | [Yes](https://github.com/tripal/tripal_blast/releases)   |
| Should install on a Tripal site appropriate for the versions it supports. |  [Yes](https://www.hardwoodgenomics.org/blast) |
| Defines any custom tables or materialized views in the install file (if applicable). | [Yes](https://github.com/tripal/tripal_blast/blob/7.x-1.x/blast_ui.install#L55)  |
| Adds any needed controlled vocabulary terms in the install file (Tripal3). |  **Not Applicable** (No cvterms needed.) |
| Provides Installation and admin instructions README.md (or RTD). |  [Yes](https://tripal-blast-ui.readthedocs.io/en/latest/user_guide/install.html) |
| Has a license (distributed with module). | [Yes](https://github.com/tripal/tripal_blast/blob/7.x-1.x/LICENSE.txt)  |

## Silver
| Requirement           | Status |
|------------------- |---------------------|
| Follows basic Drupal Coding standards; specifically, code format and API documentation. | Yes |
| Uses Tripal API functions. Specifically, it should use the Chado Query API for querying chado (if using chado as the storage system). | Yes (e.g. [chado.db for linkouts](https://github.com/tripal/tripal_blast/blob/7.x-1.x/includes/blast_ui.node.inc#L230))   |
| Tripal Jobs API for long running processes. |  [Yes](https://github.com/tripal/tripal_blast/blob/7.x-1.x/includes/blast_ui.form_per_program.inc#L556-L562) |
| TripalField class to add data to pages (Tripal3). |  **Not Applicable** |
| Provides ways to customize the module (e.g. drush options, field/formatter settings, admin UI). | Yes (e.g. [admin UI](https://tripal-blast-ui.readthedocs.io/en/latest/user_guide/install.html#configure-tripal-blast), [custom linkouts hook](https://tripal-blast-ui.readthedocs.io/en/latest/dev_guide/custom_linkouts.html)) |
[...]

PR #947, opened 2019-06-03
Update TPPS Extension Module Rating

par12005:
# Documentation

## Description

Added Bronze badge to the TPPS Extension Module description in docs.

## Testing?

Test that docs still build correctly

## Additional Notes:
This is my first time contributing to Tripal, so please let me know if I'm doing anything wrong!

### DIFF
--- a/docs/extensions/data_input.rst
+++ b/docs/extensions/data_input.rst
@@ -38,6 +38,10 @@ A BibTEX importer for Tripal Publications. Currently this module only provides a
 Tripal Plant PopGen Submission
 -------------------------------
 
+.. image:: https://tripal.readthedocs.io/en/7.x-3.x/_images/Tripal-Bronze.png
+  :target: https://tripal.readthedocs.io/en/7.x-3.x/extensions/module_rating.html#Bronze
+  :alt: Tripal Rating: Bronze
+
 The Tripal Plant PopGen Submit (TPPS) Module supports a flexible submission interface for genotype, phenotype, environmental, and metadata for population, association, or landscape genetics studies. The portal walks the user through specific questions and collects georeferenced coordinates on plant accessions and also supports ontology standards, including the Minimal Information About a Plant Phenotyping Experiment (MIAPPE) (http://www.miappe.org/) and standard genotyping file formats, such as VCF.
 
 `Documentation <https://tpps.readthedocs.io/en/latest/>`__


spficklin:
Hi @par12005. Thanks for submitting your module for a badge!   This is great.  Could you help us out a bit and just confirm with a "Yes" that the module meets the Bronze requirements?  An example is here:  https://github.com/tripal/tripal/pull/944.   You don't need to format as a nice table, just confirm that you acknowledge that the module meets each of the requirements, or if a requirement doesn't apply.

par12005:
Thanks @spficklin, I'm not familiar with releasing software, so please let me know if there is any problem with the way I did that. I have provided links to all of the Bronze requirements:

Has a public release: [Yes](https://github.com/par12005/TPPS/releases)
Should install on a Tripal site appropriate for the versions it supports: [Yes](https://treegenesdb.org/tpps)
Defines any custom tables or materialized views in the install file (if applicable): [Yes](https://github.com/par12005/TPPS/blob/master/tpps.install#L59)
Adds any needed controlled vocabulary terms in the install file (Tripal3): [Yes](https://github.com/par12005/TPPS/blob/8fea747e1c24358b59396beb33ca29715d4624a6/tpps.install#L231)
Provides Installation and admin instructions README.md (or RTD): [Yes](https://tpps.readthedocs.io/en/latest/install.html)
Has a license (distributed with module): [Yes](https://github.com/par12005/TPPS/blob/master/LICENSE)

bradfordcondon:
after enabling the module and visiting /admin/config/content/tpps, i get the following database error

```
PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "cartogratree_groups" does not exist LINE 3: cartogratree_groups g ^: SELECT g.group_name AS group_name, g.group_id AS group_id FROM {cartogratree_groups} g; Array ( ) in tpps_admin_settings() (line 39 of /Users/bc/tripal/sites/all/modules/custom/TGDR/admin/config.php).

```

I dont see a `cartogratree_groups` table definition anywhere in the module, is it perhaps defined in a dependency module?   Modules should define, or declare as dependencies modules that do define, all tables and fields they query for a bronze rating.

par12005:
Thanks @bradfordcondon! TPPS is supposed to be able to use the cartogratree_groups table from the CartograTree module if that module is enabled. I forgot to put a check for that in the administrative configuration form. TPPS does not depend on CartograTree, but it does have more features if the CartograTree module is enabled. Should TPPS still define that table in the install file?

bradfordcondon:
I think your solution of checking if the table exists is the correct one if it's not a "true" dependency of the module (users might be interested in TPPS without CartograTree, right?).
What's essential is that your module doesn't accidentally crash without CartograTree installed, as that's really what we want to achieve by "Defines any custom tables or materialized views in the install file".

So, I'd add the check to your admin area (really all places where you touch a Cartogratree table or field) and I'll look again.  If your module is looking for Cartogratree in many places and thats a pain, then I'd instead declare it as a dependency in the `.info` file.

par12005:
I've updated TPPS with the CartograTree checks, if you would like to take another look.

bradfordcondon:
Better, good job.  I'd say you're all clear for bronze.

 You might want to put a guard on the completed submissions table, as it crashes if there are no submissions (presumably)

```
Notice: Undefined index: id in tpps_completed_display() (line 21 of /Users/bc/tripal/sites/all/modules/custom/TPPS/includes/completed_display.inc).
TypeError: Argument 1 passed to tpps_table_display() must be of the type array, null given, called in /Users/bc/tripal/sites/all/modules/custom/TPPS/includes/completed_display.inc on line 39 in tpps_table_display() (line 17 of /Users/bc/tripal/sites/all/modules/custom/TPPS/includes/table_display.inc).
```

par12005:
Thanks! I added a check to the completed submission page which redirects the user to the main TPPS form if there is not an accession provided.

spficklin:
Thanks @par12005. I'm merging into the docs!

PR #953, opened 2019-06-19
Update governance.rst

spficklin:
Replace occurances of TPMC with PMC, Steering with "Advisory" and TSC with TAC, to match terminology we officially settled on.

# Documenation

Issue # N/A (fixed inline on GitHub)

## Description

In the governance docs we used the phrase "Tripal Steering Committee" and called it TSC.  But we have settled on "Tripal Advisory Committee".  So I fixed the documentation to match.  I also changed occurrences of TPMC to just PMC as that's what we tend to call it.  

I"m requesting @mestato review as she's current chair of the TAC.


mestato (approved):
Works for me. Thanks, Stephen!

laceysanderson:
Nice, thanks for the changes @spficklin and the quick review @mestato. I'm merging since it has the required approval :-)

PR #968, opened 2019-07-02
PHP 7.2 migration displays a "Countable" error

guignonv:
Solves "Warning: count(): Parameter must be an array or an object that implements Countable in include() (line 9 of .../modules/tripal/tripal_stock/theme/templates/tripal_stock_collections.tpl.php)." (and issues at line 25 after when $collections is not a non-empty array). This issue appears in PHP 7 because "count()" behavior changed and does not return "1" when a scalar is provided as argument.

#

Issue #


guignonv:
Forgot to highlight it's for the old "Tripal 2.x".

spficklin (approved):

PR #973, opened 2019-07-05
Update bulk_loader.rst

spficklin:
# Documentation

Issue # https://github.com/tripal/tripal_analysis_blast/issues/40

## Description

I removed language about the location on tripal.info for sharing bulk loader templates as we no longer have that resource.


bradfordcondon:
any interest in distributing them with core in an examples folder or in the documentation site?  Or even just as a github issue?  I like the idea of at least having a couple of examples.

spficklin:
I'm okay, with having an 'examples' directory and putting them in there.  Then the PMC can make sure that the way data is stored meets expected norms.

laceysanderson:
What about putting them in the Tripal examples module? We could link to it from the docs and then we don’t clutter up Core

spficklin:
That would make more sense.  I often forget we have an example module, because it's separate from Tripal and I don't think to update it. Would it make sense to put the tripal_example module back into tripal?

spficklin:
@bradfordcondon and @laceysanderson any time to look this over?  Should be quick.

bradfordcondon (approved):
if we want to deal with example module, i think that merits a separate issue.  As far as fixing broken links, this is fine for now and only requires one review.  thanks @spficklin

Issue #75, opened 2017-04-01
Improper regex for loading FASTA files causes errors


spficklin:
Copied from the Drupal issue queue: https://www.drupal.org/node/2691787

It was reported on the mailing list (https://sourceforge.net/p/gmod/mailman/message/34924927/) that if the regular expression used for importing a FASTA file is incorrect that an error message occurs. This should be handled gracefully with a more intuitive notice given to the user. Here is the message recieved:
```
WD trp-fasta: [error]
array_keys() expects parameter 1 to be array, string given tripal_core.tripal.api.inc:117 [warning]
[site http://default] [TRIPAL ] [TRP-FASTA] Array
strtr() expects parameter 1 to be string, array given syslog.module:115 [warning]
syslog() expects parameter 1 to be long, string given syslog.module:118 [warning]
html_entity_decode() expects parameter 1 to be string, array given unicode.inc:463 [warning]
WD trp-fasta:
```

spficklin:
This is low hanging fruit and has been in the queue for 2 years. We should fix this.

bradfordcondon:
I couldnt recreate this exact error, but i did add validation that I think is pretty straight forward in the form.  In addition i improved the error reporting.

spficklin:
Thanks @bradfordcondon for working on this older issue :+1:

PR #976, opened 2019-07-30
add regexp validation for FASTA importer and fix bugs

bradfordcondon:
# Bug Fix

Issue #75 

## Description
The original issue cited cryptic errors when providing an invalid regexp. When investigating i saw we had a bunch of typos prevent proper error reporting in the importer.  In addition, I added validation on the regexp fields.

## Testing?

visit the fasta loader (`/admin/tripal/loaders/chado_fasta_loader).
provide bad regexps for your FASTA importer (for example `test//////`.  Remember that this importer APPENDS the leading trailing `/` in your regexp, so leave those out).  The fields are in the advanced options of the form.  You'll see the importer passes validation, but will give unhelpful errors on run time.  Repeat on this branch- you wont get past validation with an invalid regexp.

### DIFF
--- a/tripal_chado/includes/TripalImporter/FASTAImporter.inc
+++ b/tripal_chado/includes/TripalImporter/FASTAImporter.inc
@@ -319,6 +319,21 @@ class FASTAImporter extends TripalImporter {
       form_set_error('db_id', t("Please select a database"));
     }
 
+    //check to make sure the regexps are valid
+
+    if ($re_accession && @preg_match("/$re_accession/", null) === false) {
+
+      form_set_error('re_accession', t("please provide a valid regular expression."));
+    }
+    if ($re_name && @preg_match("/$re_name/", null) === false) {
+
+      form_set_error('re_name', t("please provide a valid regular expression."));
+    }
+    if ($re_name && @preg_match("/$re_uname/", null) === false) {
+
+      form_set_error('re_uname', t("please provide a valid regular expression."));
+    }
+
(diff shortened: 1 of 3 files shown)


tripal_chado/includes/TripalImporter/FASTAImporter.inc
```diff
@@ -651,7 +669,7 @@ class FASTAImporter extends TripalImporter {
       ], $values);
       if (count($results) > 1) {
         $this->logMessage("Multiple features exist with the name '!name' of type '!type' for the organism.  skipping",
-          ['!name' => $name, '!type' => $type], TRIPAL_ERROR);
```
bradfordcondon, line 654: `$type` is not an initialized variable.  I think the cvterm name is the right thing here (and below).

spficklin: Yes, I agree.

spficklin (approved):

spficklin:
@bradfordcondon I added another if statement to check the regular expression for the relationship parent RE and I added some additional text to each error statement so that if more than one are messed up you don't get the exact same message multiple times.   I tested and it works I approve.  I don't think we need another functional review as this is really a simple fix.  But I'll let you decide and if you agree I think this can be merged.

bradfordcondon:
@spficklin thanks for the review, i like the additions. If you think 1 review is OK considering you seem to have given it a thorough look.

PR #999, opened 2019-09-26
Add entity permissions tests

laceysanderson:
# Automated Testing
**This PR adds automated testing ONLY**

## Description
This PR adds automated testing for Tripal Content CRUD permissions. Specifically, `view [bundle name]`, `create [bundle name]`, `edit [bundle name]`, `delete [bundle name]`. I'm working on a Tripal Access Extension module and wanted to ensure I was not breaking core permissions. After developing the tests, I though I might as well contribute back :-) 

## Testing?
This will be automatically tested by Travis :-) 

There are two tests:
1) Simply tests that the permissions are available.
2) Tests that given a test entity and test user/role, users with the permission can access the content.

### DIFF
--- a/tests/tripal/entities/PermissionsTest.php
+++ b/tests/tripal/entities/PermissionsTest.php
@@ -47,7 +47,8 @@ class PermissionsTest extends TripalTestCase {
     $faker = Factory::create();
 
     // Create organism entity for testing.
-    $bundle_name = 'bio_data_2';
+    $bundle_id = db_query("SELECT bundle_id from {chado_bundle} where data_table='organism'")->fetchField();
+    $bundle_name = 'bio_data_' . $bundle_id;
     $bundle = tripal_load_bundle_entity(['name' => $bundle_name]);
 
     $genus = $faker->word(1, TRUE);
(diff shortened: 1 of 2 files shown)


tests/tripal/entities/PermissionsTest.php
```diff
@@ -0,0 +1,167 @@
...
+   * NOTE: We only test one bundle since it should be the same for all
+   * of them (done in a loop).
+   *
+   * @group permissions
+   */
+  public function testPermissionsForUser() {
+    $faker = Factory::create();
+
+    // Create organism entity for testing.
+    $bundle_name = 'bio_data_2';
```
bradfordcondon: I'm always scared assuming bundle number.  Can we instead look it up, o reven just double check the term_id make sense and if it doesnt exit with a warning?

bradfordcondon (approved):
Love to see new tests.  I read through these, great documentation.  I had one suggestion.  Thanks @laceysanderson

tests/tripal/entities/PermissionsTest.php
laceysanderson: Good point! I looked it up now :-)

laceysanderson:
Thank you for the review @bradfordcondon! Based on our current guidelines which say we only need "1 review by a contributor" I'm merging in these tests :-)

PR #1000, opened 2019-09-27
Cannot Sync cvterm-based entities

laceysanderson:
# Bug Fix

Issue n/a

## Description
Currently, if you create a content type and then immediately try to publish records, it will fail. This is due to the change to `chado_query()` which requires the drupal table to be retrievable via `drupal_get_schema()`. Since the table was just created as part of the create tripal bundle job, it is not in the list returned by `drupal_get_schema()`. To fix this, I cleared the drupal schema cache after creating a new bundle.

## Testing?
1. Create a new Tripal Content Type with base chado table `cvterm`.
2. Run the Tripal job to create it.
3. Publish records.

Not on this branch it will fail with a SQL syntax error due to un-replaced `[chado_bio_data_##]` in the query. On this branch it will work perfectly :-)

NOTE: This PR also fixes a few other minor bugs related to publishing cvterm-based content types. Specifically, adds defaults to validating widget bundle/delta and ensure the bundle can be retrieved for cvterm-based content types in SIO:vocabulary field.


laceysanderson:
This is quite serious on KnowPulse... I would really appreciate a review on the PR! <3

bradfordcondon (approved):
I couldn't reproduce your problem-  I defined a custom cvterm bundle for SIO cvterms, and published with no errors on 7.x-3.x.

That said i did see the removed warnings on publishing so thats nice, and clearing the cache after creating the bundle wont create problems, so I approve.

bradfordcondon:
sorry to make you wait. Im not sure what we should do as i couldnt reproduce your problem... but looking at the code i have no issues so im happy to go ahead with this.

laceysanderson:
Just reproduced the problem on a clean Tripal site 😓 Since you're happy code-review wise, I think I'm just going to merge.

Issue #1010, opened 2019-10-17
Web Services - Invalid Content Type


reynoldtan:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7
* PostgreSQL Version: 11.3
* PHP Version: 7.1.30

### Issue description

When content types comprise of multiple words separated by spaces, Web Services format the url with a + (plus sign) instead of _ (underscore) as seen in other section. This causes invalid content type error.

### **AN UPDATE**
I was able to load an expanded field (beyond the invalid content type error) and found out that the + plus sign issue was also present at this point. 

![invalid-content-type](https://user-images.githubusercontent.com/15472253/67040360-91be0200-f0e0-11e9-9a9e-ce6de439a395.gif)

#### Steps to reproduce

Access /web-services/content/v0.1 in your site and click on @id link of a content type that contains multiple words (ie. Genetic Marker, F1 Breeding Crosses etc), keep accessing @id until you reach content resource where @id now contains plus symbol as part of the url.

#### Error messages and screenshots

@see image above
```
error message
`Invalid Content Type`

bradfordcondon:
Thank you for contributing this fix

PR #1011, opened 2019-10-17
Proposed change for issue #1010

reynoldtan:
Apply the same treatment to service path and or content type name as seen in doEntityList method.

#Issue #1010 Web Services: Invalid Content Type.

## Description

The use of urlencode() replaces any white spaces into plus symbol (+) causing field label to mismatch and trigger an invalid content type error. This proposal makes it consistent to apply the same treatment (as seen in other sections) to content type labels when used in constructing the field url.

## Testing?

After implementing this proposal, access /web-services/content/v0.1 in your site and click on @id link of a content type that contains multiple words (ie. Genetic Marker, F1 Breeding Crosses etc), keep accessing @id while observing that the string url should now not contain any plus symbol to concatenate words in content type name/label. Do this until you reach content resource.

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -110,7 +110,7 @@ class TripalContentService_v0_1 extends TripalWebService {
    * Creates a resource for an expanded field of an entity.
    */
   private function doExpandedField($ctype, $entity_id, $expfield) {
-    $service_path = $this->getServicePath() . '/' . urlencode($ctype) . '/' . $entity_id;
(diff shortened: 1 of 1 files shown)


laceysanderson:
First off, thank you for your PR! Your tests fail due to a change in the Travis environment. If you could please update your patch from 7.x-3.x then we could confirm tests pass on your PR.

Steps:
```
git clone https://github.com/reynoldtan/tripal.git
git remote add upstream https://github.com/tripal/tripal.git
git fetch upstream
git checkout 7.x-3.x
git merge upstream/7.x-3.x
git checkout patch-1
git merge 7.x-3.x
git push
```
This will update your fork with the most recent changes from github.com/tripal/tripal and update your patch with them.

bradfordcondon (approved):
Reproduced the issue and confirmed the fix.  The code contribution is styled correctly.  Thank you for the fix!

Issue #1012, opened 2019-10-21
travis builds failing w/ postgres


bradfordcondon:
## BUG/ERROR report
bug is in CI builds.

```
 connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"

```

Probably a change in how it supports postgres:  https://docs.travis-ci.com/user/database-setup/#postgresql

laceysanderson:
PR Merged -Thank you for the fix @bradfordcondon!

PR #1013, opened 2019-10-26
fix travis build

bradfordcondon:
fix postgres usage in travis

# Bug fix

Issue #1012 

## Description

our travis builds are failing.  Turns out we need to change our config from using `postgres` to `postgresql`.  Crazy I Know,  [but here it is working on  a fork](https://travis-ci.org/bradfordcondon/tripal/jobs/603105056?utm_medium=notification&utm_source=github_status)

## Testing?
If this build passes, that should be all we need to see.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 services:
   - docker
-  - postgres
+  - postgresql
 
 sudo: required
 


laceysanderson (approved):
Thanks for tracking this down!!!

Issue #1005, opened 2019-10-08
Deleting a Database results in the following error message


risharde:
Tripal: 3.0
Postgres: 9.x
PHP: 7.2

Deleting a created 'TEST' database results in the following error, just a heads up? I have added a test vocabulary and also a test cv term. I suspect it may be failing since this data exists? Maybe a better way would be to have a checkbox that forces the removal of the database and all tied in data?

```
PDOException: SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "dbxref_id" violates not-null constraint DETAIL: Failing row contains (2989, 57, test_content_type, , null, 0, 0). CONTEXT: SQL statement "UPDATE ONLY "chado"."cvterm" SET "dbxref_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "dbxref_id"": DELETE FROM chado.db WHERE db_id = :db_id ; Array ( [:db_id] => 44 ) in chado_query() (line 1715 of /var/www/html/sites/all/modules/tripal/tripal_chado/api/tripal_chado.query.api.inc).

```

bradfordcondon:
its true.  I dont think we get a ton of use out of the delete db interface, but, if someone did want it to cascade, the would probably want to know how many attached records would be deleted.

I could also see us just adding instructions that you need to remove all dbxrefs from the db before this would work.

bradfordcondon:
i built it out to just block the delete and provide a notice as its easiest/safest and definitely laziest: https://github.com/tripal/tripal/commit/cd5a9c55a7c8a7645cd418e979276bba594018ce#diff-ad8ca83a8fb3d11bfdc15152c7a9a956

what do we think?  worth adding to prevent the SQL error?

laceysanderson:
PR Merged

PR #1014, opened 2019-10-26
prevent SQL error screen deleting databases in use

bradfordcondon:
# Bug fix

Issue #1005 

## Description
This is a pretty weak solution to a minor problem, i was just looking for something small to do.

The chado.database interface lets you delete a database record... but it crashes if there are linked records, since it doesnt cascade.
Cascading on delete here seems like a pretty big deal, so instead, i just block deletion (since deleting would crash anyway) and let the user know they have dbxrefs they should delete first.

## Testing?

- create a db, a cv, and a cvterm linked to the two.  You will now see a warning when trying to delete the db.

## Additional Notes (if any):
this is a silly PR, and definitely not a great solution, but it does blocks off a SQL crash so i figured why not submit.

### DIFF
--- a/tripal_chado/includes/tripal_chado.db.inc
+++ b/tripal_chado/includes/tripal_chado.db.inc
@@ -84,7 +84,8 @@ function tripal_chado_db_edit_form($form, &$form_state) {
   ];
 
 
-  // if we don't have a db_id then we can  return the form, otherwise
+  // if 
+  we don't have a db_id then we can  return the form, otherwise
   // add in the other fields
   if ($dbid) {
     tripal_chado_add_db_form_fields($form, $form_state, $dbid);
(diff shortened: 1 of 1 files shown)


tripal_chado/includes/tripal_chado.db.inc
```diff
@@ -84,20 +84,29 @@ function tripal_chado_db_edit_form($form, &$form_state) {
...
       '#type' => 'submit',
       '#value' => t('Update'),
     ];
-    $form['delete'] = [
-      '#type' => 'submit',
-      '#value' => t('Delete'),
-      '#attributes' => ['onclick' => 'if(!confirm("Really Delete?")){return false;}'],
-    ];
+
+    // Assert we have some configuration in our database.
```
bradfordcondon:
```suggestion
```

laceysanderson (approved):
Tested and it works as expected. I think this is a good solution for now.

PR #1016, opened 2019-10-26
Attempt to fix Issue #681.

laceysanderson:
# Bug Fix

Issue #681 

## Description
See the issue for a full description. Error to be removed:
```
[TRIPAL WARNING] [TRIPAL_FIELDS] The field, efo__array_design, uses a term, EFO:0000269, that is already in use on this content type. The field instance cannot be added.
```

## Testing?
Switch to this branch and create a new tripal site. The error normally appeared at the end of the prepare step but will not in this branch.


bradfordcondon (approved):
well done!

Issue #723, opened 2018-10-19
publishing mRNA Date Limit Format Warning


kyrenya:
Hi there! 
Ubuntu 18.04, tripal v3.0 (last release). php7.2.10 

when publishing gene (following the citrus example) 
I got warning like here: https://www.drupal.org/project/date/issues/2843367. I applied this patch: 

2843367-php71-string-offset-26.patch
it seemed that removed the warnings despite the patch being for 7.1 for what I understood.

When publishing mRNA I got this one
![image](https://user-images.githubusercontent.com/43243429/47218967-5f7ea500-d3a5-11e8-830d-2064df02f0df.png)

is there a fix for that one? (I couldnt find it).

Thanks in advance!

bradfordcondon:
here are the warnings on my site
(tripal 3, php 7.1, OSX)


```
Warning: Illegal string offset 'value' in date_combo_element_process() (line 295 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Cannot assign an empty string to a string offset in date_combo_element_process() (line 295 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Illegal string offset 'value' in date_local_date() (line 145 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Notice: Uninitialized string offset: 0 in date_local_date() (line 145 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Illegal string offset 'value' in date_combo_element_process() (line 298 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Cannot assign an empty string to a string offset in date_combo_element_process() (line 298 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: implode(): Invalid arguments passed in date_limit_format() (line 2276 of /Users/bc/tripal/sites/all/modules/date/date_api/date_api.module).
Warning: Illegal string offset 'value' in date_combo_element_process() (line 295 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Cannot assign an empty string to a string offset in date_combo_element_process() (line 295 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
Warning: Illegal string offset 'value' in date_local_date() (line 145 of /Users/bc/tripal/sites/all/modules/date/date_elements.inc).
[...]

PR #1018, opened 2019-10-27
Fix for Issue #723: publishing mRNA Date Limit Format Warning

laceysanderson:
# Bug Fix

Issue #723 

## Description
This PR removes datetime fields from the publish form in order to avoid errors.

## Testing?
1. Go to Admin > Content > Tripal Content and click on "Publish Tripal Content"
2. Choose a content type with datetime fields (e.g. genes, mRNA, QTL)
On 7.x-3.x you will see a whole string of errors. On this branch you will not.


bradfordcondon (approved):
I'd definitely prefer if we could allow filtering on date time but its better to remove so we dont have errors than wait for a fix that won't come.  Thanks lacey!

Issue #1019, opened 2019-11-01
Web Services - hydra:member not rendered


reynoldtan:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7
* PostgreSQL Version: 11.3
* PHP Version: 7.1.30

### Issue description

An empty array or list in members.

![Screen Shot 2019-11-01 at 10 48 32 AM](https://user-images.githubusercontent.com/15472253/68041391-15a5eb80-fc96-11e9-8dda-42de0d9e1b8e.png)

When implementing an instance of hydra:member in the field's load() function, such as the code below;

```
foreach ($items as $i => $item) {
  $entity->{$field_name}['und'][$delta]['value']['hydra:member'][$i] = array( 
    'rdfs:label' => $item['label'],
    'local:count' => $item['count'], 
   );
}
```
expected result should look like
![Screen Shot 2019-11-01 at 11 09 03 AM](https://user-images.githubusercontent.com/15472253/68042266-16d81800-fc98-11e9-8069-6ccd121dcac6.png)

but instead, the output is an empty array.
![Screen Shot 2019-11-01 at 11 11 26 AM](https://user-images.githubusercontent.com/15472253/68042404-69b1cf80-fc98-11e9-9019-5c412a614e3d.png)

#### Steps to reproduce

In the load() method of your field, create additional information for Tripal Web service and use hydra:member as the key (see code sample above). Clear your cache and reload page.

#### Error messages and screenshots

```
error message
```

laceysanderson:
PR Merged

PR #1020, opened 2019-11-05
Proposed Fix to Issue #1019 - Hydra:member empty

reynoldtan:
Added additional condition in sanitize function to handle case when key is a non-vocabulary term such as the key/index (0,1,2...) of a hydra:member array.

# Bug Fix

Issue https://github.com/tripal/tripal/issues/1019 - hydra:member array not rendered.

## Testing?

Implement a hydra:member instance to any field. Below is a sample code:

```
foreach ($items as $i => $item) {
  $entity->{$field_name}['und'][$delta]['value']['hydra:member'][$i] = array( 
    'rdfs:label' => $item['label'],
    'local:count' => $item['count'], 
   );
}
```

Save your code and clear the cache. Reload web service page containing your field.

A sample output of hydra:member implementation.
image

### DIFF
--- a/tests/tripal_ws/http/TripalWebServicesContentTest.php
+++ b/tests/tripal_ws/http/TripalWebServicesContentTest.php
@@ -58,4 +58,105 @@ class TripalWebServicesContentTest extends TripalTestCase {
     $json = $response->json();
     $this->assertEquals($json['label'], "$label Collection");
   }
+
+  /**
+   * Tests the sanitizeFieldKeys() method.
+   * @group tripal_ws
+   * @group tripalWS-ServiceResource
+   */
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
I've added automated testing for the `sanitizeFieldKeys()` method in web-services which is what was causing the issue. The automated tests fail on `7.x-3.x` and pass on this branch. The test checks a number of conditions beyond just those effected by this PR.

Additionally, I've done a full functional review on a generic Tripal site using the organism germplasm summary field from the [tripal example module](https://github.com/tripal/tripal_example) and it works as expected!
Screen Shot 2019-11-14 at 10 39 58 AM

Thank you for the fix @reynoldtan !

Issue #1022, opened 2019-11-06
Web Services - Limited access works only in viewing an entity (for anonymous user).


reynoldtan:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7
* PostgreSQL Version: 11.3
* PHP Version: 7.1.30

### Issue description

Limited access to content only works in viewing an entity and does not work in listing and viewing content types. For example:

I've limited access to Genetic Maps to anonymous user in the permissions panel.
![Screen Shot 2019-11-05 at 1 10 17 PM](https://user-images.githubusercontent.com/15472253/68245968-86723e00-ffdd-11e9-9217-644c73e95aa0.png)

It works on **web-services/v0.1/content type/entity id**
![Screen Shot 2019-11-05 at 1 19 45 PM](https://user-images.githubusercontent.com/15472253/68246068-b3beec00-ffdd-11e9-8ea8-c57cb5323720.png)

but not in the following sections:

A. EntryPoint
**web-services/v0.1**
![Screen Shot 2019-11-05 at 1 20 24 PM](https://user-images.githubusercontent.com/15472253/68246161-e1a43080-ffdd-11e9-8829-d485fbf3cf7a.png)

B. Content Type
**web-services/v0.1/content type**
![Screen Shot 2019-11-05 at 1 21 30 PM](https://user-images.githubusercontent.com/15472253/68246205-f4b70080-ffdd-11e9-8a24-4fc33514c70f.png)

#### Error messages and screenshots

```
error message
```

laceysanderson:
PR Merged

PR #1023, opened 2019-11-06
Proposed Fix to Issue #1022 - A and B

reynoldtan:
# Bug Fix

Issue https://github.com/tripal/tripal/issues/1022 - Limited access works only in viewing an entity (for anonymous user)

## Description

A. EntryPoint
When resources are listed in **/web-services/content/v0.1**, all content types become available to user regardless of permission settings. The proposed fix in doContentTypeList() function excludes any content types based on permission configuration.

B. Content Type
While users will not be able to see content types they don't have permission through EntryPoint, they are not restricted to access a specific content type directly through the url. The proposed fix in doEntityList() ensures that content type is first checked for permission before rendering any information.

Configure a permission to an anonymous user by clicking on/off the checkbox for each content type you want to permit/restrict.

Screen Shot 2019-11-05 at 1 10 17 PM

To test

A. EntryPoint 
head to **web-services/content/v0.1**, content types where you have set to have permission to to view should become part of the list, otherwise excluded.

B. Content Type
for content type you have no view permission, access **web-services/content/v0.1/your restricted content type** (since this link is not available in A). A message "Permission Denied" will be displayed for this content.

Screen Shot 2019-11-06 at 9 34 56 AM

### DIFF
--- a/tests/tripal_ws/http/TripalWebServicesContentTest.php
+++ b/tests/tripal_ws/http/TripalWebServicesContentTest.php
@@ -38,10 +38,18 @@ class TripalWebServicesContentTest extends TripalTestCase {
    */
   public function testGettingListOfEntitiesInABundle() {
     // Get bundle label
-    $label = db_query('SELECT label FROM tripal_bundle LIMIT 1')->fetchField();
+    $label = db_query('SELECT label, name FROM tripal_bundle LIMIT 1')->fetchObject();
+
+    // Grant user permission to view bundle.
+    $bundle_name = 'view ' . $label->name;
+    if (!user_access($bundle_name)) {
+      // Either the user is anonymous or registered.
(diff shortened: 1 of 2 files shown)


bradfordcondon:
i see that your change fails two tests in `TripalWebServicesContentTest`.  Do you think you could address this?

laceysanderson (approved):
I've tested these changes on a generic Tripal site as well as a KnowPulse clone and it works in both cases. Specifically, I've confirmed that
- when logged in I can see all the content my user has permission to see in the content listing. I can also click through to records
- after logging out, I can only see the content available to the anonymous user. URLs in web-services which I could access before now show permission denied if not available to the anonymous user
- pages available to the anonymous user were always available whether logged in or not.

Also regarding the changes to tests: The tests were previously failing because they assumed no access control. The changes made to make them pass simply include granting permissions. This is good 👍 There was one other small change to the tests which related to a previous PR ($ctype change).

Overall, I approve this PR -Thank you for the contribution @reynoldtan!

Issue #272, opened 2018-01-29
Tripal 3: time executed parameters?


bradfordcondon:
Tripal 3 question:

My time executed column only lets me pick a date since 2015.  This is problematic for loading in entries to archive older datasets.

![screen shot 2018-01-29 at 1 44 12 pm](https://user-images.githubusercontent.com/7063154/35528033-8ec731e4-04fa-11e8-87d9-caac98f9da02.png)

I can't find this field  in tripal core, and the field settings dont give me a way to modify the start date.  

Any ideas where i can address this?

spficklin:
Line 919 in the tripal_chado/includes/tripal_chado.fields.Inc file is where the settings for date time fields are set.  I'm not able to dig into it at the moment but I suspect he date range is some field default as it is not set in Tripal.

Are you able to edit the field and change the date range in the field GUI?

bradfordcondon:
Hi @spficklin -- I looked before and did not find them (honest).  I found them this time around hidden in a fieldset!

So on the bundle settings page, edit the field, and under the bundle specific settings  (the first box) open the fieldset and you can configure the start/end there.

<img width="686" alt="screen shot 2018-01-30 at 3 38 51 pm" src="https://user-images.githubusercontent.com/7063154/35590074-bdcef04c-05d3-11e8-8d1b-27a4afd26b73.png">


Thanks!

spficklin:
Should we set some larger defaults in Tripal for these fields?

bradfordcondon:
I set mine to 20 years in the past, to the present.  The UX is pretty easy for picking a date, we're confronted with setting our birthday from a dropdown where the option is 1900-2017 pretty regularly.  I think it's ok to start at, say, 1990, and go till present day.

 don't know what the use case is for the future date (ie 2019).

bradfordcondon:
i got bored and looked into this tonight. (well i was looking for easy issues to close)

`['years_back']['#maxsize']` defaults to 10, as does `years_forward` in the date module.  So i actually dont see how this gets translated into +/- 3 :|

spficklin:
This is low hanging fruit, I think.  I'm marking for v3.2 release.

laceysanderson:
I've submitted a PR to fix this: #1024. If this issue is bugging you, please review the PR to get it fixed 😃

laceysanderson:
PR Merged

PR #1024, opened 2019-11-11
Fix the range for datetime fields to +/-10

laceysanderson:
Issue #272 

## Description
The default range for datetime fields is +/- 3 which does not make sense for analysis records. As such, I've changed the default to +/-10 for Tripal-added datetime fields to make date entry easier. This setting can be configured on each Tripal site through the field settings.

## Testing?
1. Go to Structure > Tripal Content Types > Analysis (or any other content type with a datetime field)
2. Delete the datetime field (e.g. Date Performed for Analysis')
3. Check for new fields which will re-add the datetime field with the new settings.
4. Add/Edit an analysis to see the new extended range for the year.
Screen Shot 2019-11-10 at 9 48 10 PM


bradfordcondon (approved):
Simple and nice change, thank you!

laceysanderson:
Thanks for the review!

Issue #1028, opened 2020-01-09
Registration page opt-out confusion


spficklin:
### System information

* Tripal Version: 3.2
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: N/A

### Issue description

On a brand new install of Tripal the user is asked to register the site. They can opt-out if they want. When they check the opt-out checkbox, nothing happens, and the message at the top of the page asking to opt-out doesn't go away.

laceysanderson:
Good catch, I'll try to get to the associated PR today!

laceysanderson:
PR Merged

PR #1029, opened 2020-01-09
Registration form fixes

spficklin:
# Bug Fix

Issue #1028

## Description

If the site admin decides to opt out of the registration there is a check-box for them to click on.  However, clicking the box doesn't properly turn off the notice at the top of the page.  The user would have to click the button at the bottom that reads 'Register Site'.  That makes no sense.   This PR fixes that as well as a few other bugs related to the number of funding sources.

## Testing?

1. You will need to install a brand new instance of Tripal.  Go to the registration page and try to opt-out. (Do not click the "Register Site" button or you'll have to start all over). You get the behavior described above.  
2.   Pull this branch, and retry the test. It should be clear now that you have opted out.

### DIFF
--- a/tripal/includes/tripal.registration.inc
+++ b/tripal/includes/tripal.registration.inc
@@ -337,6 +337,8 @@ function tripal_registration_form_submit($form, &$form_state) {
   // Now send the updated info to the Tripal Site.
   // Only register with tripal.info if the user has not opt'd out.
   if ($form_state['values']['disable_tripal_reporting'] == FALSE) {
+
+    $registration = variable_get('tripal_site_registration');
     tripal_registration_remote_submit($registration);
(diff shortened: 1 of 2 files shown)


laceysanderson (commented):
Testing:
1. Opt-out of registering -the message disappears like it should 🎉 
2. Uncheck the checkbox and fill it in minimally (site status, description, PI name, Admin name + email, funding agency name only) and submit... I see the following errors:
Screen Shot 2020-01-09 at 9 25 04 PM

Also my registration did not show up in tripal.info

bradfordcondon:
My submission status with the committed change:

Screen Shot 2020-03-17 at 5 51 41 PM

The site shows up on tripal.info as well.

laceysanderson (approved):
Just re-tested with @bradfordcondon change on a fresh install and it works perfectly 👍

PR #1032, opened 2020-02-12
This typographical error causes 404 errors

risharde:
Particularly with link checkers, this typo causes nearly every page to show up with an unnecessary 404 error due to this link

#

Issue #


laceysanderson (approved):
Good catch! Thank you for the fix

Issue #1055, opened 2020-05-01
National Ag Library publication loader UTF-8 mixed with HTML entities


dsenalik:
### System information

* Tripal Version: 7.x-3.x
* Drupal Version: 7.69
* PostgreSQL Version: 12.2
* PHP Version: 7.2

### Issue description
Some publications imported from the National Ag Library will have UTF-8 characters mixed with characters encoded as HTML entities. When the information, for example the abstract, is passed through
mb_convert_encoding($pub['Abstract'], 'UTF-8', 'HTML-ENTITIES')
the entites are converted to UTF-8, but existing UTF-8 characters are corrupted. To make the matter worse, some HTML entities are double encoded e.g. `&amp;#x2018;` and these are only half-converted.

A second issue is that rarely, a publication will have two fields in the MARC record 35a "System Control Number", which represents the publication accession number at AGL. The second number will have a "ns" prefix. The simple solution is to ignore values with a "ns" prefix.

I am providing a pull request #1056 to correct these problems.

#### Steps to reproduce
Example of a publication with second "ns" accession number, create an importer for a publication with the title "Integrated Molecular and Morphological Studies of the Daucus guttatus Complex" and click Test Importer.
```
<datafield tag="035" ind1=" " ind2=" ">
    <subfield code="a">ns3947</subfield>
  </datafield>
  <datafield tag="035" ind1=" " ind2=" ">
    <subfield code="a">4782120</subfield>
  </datafield>
```

Example of a publication with UTF-8 + HTML entity corruption problems, load publication with the title "First Report of Ceratocystis lukuohia on Metrosideros polymorpha"

spficklin:
Fixed with PR. Thanks @dsenalik

PR #1056, opened 2020-05-01
National Ag Library publication loader HTML entity bugfix

dsenalik:
#

Issue #1055 

## Description

This pull request fixes issues described in issue #1055, corruption of UTF-8 in publications loaded by the AGL importer and accession number filtering of "ns" prefix values.

### DIFF
--- a/tripal_chado/includes/loaders/tripal_chado.pub_importer_AGL.inc
+++ b/tripal_chado/includes/loaders/tripal_chado.pub_importer_AGL.inc
@@ -574,13 +574,8 @@ function tripal_pub_AGL_count($yazc, $search_str) {
 function tripal_pub_AGL_decode($text) {
   // first handle double encoding situations by replacing &amp;
   $text = preg_replace("/&amp;/", "&", $text);
-
-  // then only replace things that look like an HTML entity, i.e.
-  // ampersand followed by semicolon, in order to leave UTF-8 intact
-  $text = preg_replace_callback("/(&[^;\p{C}\p{M}\p{Z}]{1,31};)/",
-    function($m){return(mb_convert_encoding($m[1], 'UTF-8', 'HTML-ENTITIES'));},
-    $text);
-  return($text);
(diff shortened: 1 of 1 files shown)


tripal_chado/includes/loaders/tripal_chado.pub_importer_AGL.inc
```diff
@@ -553,6 +553,36 @@ function tripal_pub_AGL_count($yazc, $search_str) {
...
+ *
+ * @ingroup tripal_pub
+ */
+function tripal_pub_AGL_decode($text) {
+  // first handle double encoding situations by replacing &amp;
+  $text = preg_replace("/&amp;/", "&", $text);
+
+  // then only replace things that look like an HTML entity, i.e.
+  // ampersand followed by semicolon, in order to leave UTF-8 intact
+  $text = preg_replace_callback("/(&[^;\p{C}\p{M}\p{Z}]{1,31};)/",
```
bradfordcondon: if the goal is to decode html entities, [can we use `html_entity_decode` instead?](https://www.php.net/manual/en/function.html-entity-decode.php)

dsenalik: Thank you Bradford, I didn't know about that function, it is a much cleaner way to do it. I have updated and tested with this new method.

bradfordcondon (approved):
approving as a non-functional review.  Thank you.

spficklin:
Hi @dsenalik sorry this sat in the queue so long. I can do a functional review, but you can provide an example publication I can use to test it?

dsenalik:
I put the examples in issue https://github.com/tripal/tripal/issues/1055, it would probably have been better to put them in the pull request!

Doug

spficklin:
Sorry for the delay on this. The functional test passed. Thanks @dsenalik for the submission!  I'm merging.

spficklin (approved):

Issue #1040, opened 2020-04-08
Long load times for gff3 loads


mpoelchau:
### System information
* Tripal Version: 2
* Drupal Version: 7.69
* PostgreSQL Version: 9.6.17
* PHP Version: 7.3.16

### Issue description
Loading gff3 files into the feature table via the GFF3 loader has been taking a really long time. The last GFF3 file that we loaded had 16,017 genes  and 24,093 mRNAs (we load subfeatures such as exons as well). It took about 4 days to load on our production site. We’re wondering if the problem is the feature table size - it currently has 3,303,837 entries. The same file took 14 hours to load on our dev site, where the feature table has 2,091,629 entries (although there are some other differences between the sites, as well - but the php, postgres, and drupal versions are the same). Other larger files similarly take much longer to load on our production site than our dev site. 

@srobb1 suggested to dump/backup of the db, then in the webform in the advanced options deselect the 'Use a transaction'. Also vacuuming the database before loading can sometimes help as well. 

@spficklin wonders if the GFF3 loader needs to be looked at for performance issues: "For every single feature property, dbxref, etc that gets added it must check for integrity constraints and that adds a lot of time.  The suggestions @Sofia Robb gives can help, but I suspect the underlying problem is a very large feature table and the time it takes to check integrity before inserts." 

If there are any suggestions others might have about how to reduce the load time - in particular with larger feature tables - I would love to hear them!

#### Steps to reproduce

- Loaded genome fasta file with Tripal fasta loader
- Loaded gff3 file with Tripal fasta loader
-  I can provide the files that I loaded if it's helpful.

<img width="654" alt="Screen Shot 2020-04-03 at 14 20 41" src="https://user-images.githubusercontent.com/5487172/78401625-665ba080-75b6-11ea-9f32-e0c05259e148.png">

par12005:
@spficklin, I have submitted PR #1057 which tries to resolve some performance issues and implements some unit tests. Please take a look when you get the chance!

spficklin:
Hi all, Just an update on this issue.  

[...]

PR #1057, opened 2020-05-11
Gff3 performance and unit tests

par12005:
# Bug Fix

Issue #1040 

## Description

 - Use cvterm lookup table in helper methods to increase performance. Previously, we were using a lookup table for cvterms within the main `loadGFF3` method, but we were not using the table in the helper methods, which do most of the actual inserting to the database. I have added a private method called `getCvterm` which first checks the lookup table before starting a db query. This has improved load time on a 1000 line gff file from about 23 seconds to about 11 seconds. More optimizations may be necessary in the future, but this seems like a good improvement for now.
 - Remove "use transaction" option from GFF Importer. All Tripal Importers are currently wrapped in a transaction, and the GFF Importer itself never begins a transaction. This means that the "use transaction" option does not actually have any functional purpose, and is therefore a bit confusing and misleading, so I have removed it.
 - Added unit tests for GFF Importer Attributes. The tests I have added do not test every single possible case, but they test for most of the main GFF attributes that are coded for in the GFF Importer, including the feature, alias, dbxref, ontology term, parent, target, properties, and derives from relationship.
 - Added small .gff and .fasta files to be used for attribute unit tests.

## Testing?

- It may be necessary to clear your cache after pulling this code, which can be done via `drush cc all`.
- Load your .fasta file with the Tripal FASTA importer
- Navigate to `admin/tripal/loaders/chado_gff3_loader`
- Fill out the form and submit your GFF Loader job
- Verify that the loader works and completes faster than before.
- If you would like to run the unit tests locally:
  - Navigate to the tripal module directory
  - Install dependencies with `composer install`
  - Run all unit tests (`./vendor/bin/phpunit`) or run only the new unit tests added in this PR with
```
./vendor/bin/phpunit --filter testGFFImporterAttribute tests/tripal_chado/loaders/GFF3ImporterTest.php
```


par12005:
@Ferrisx4 pointed out to me that the performance improvements made previously in this PR didn't seem to affect large files as much as I had hoped. It looks like the loader still has a time complexity larger than O(n), and as such the cvterm lookup table does not make a very big difference for very large gff files.

So, I've added another performance improvement - I noticed that almost all of the calls to `chado_insert_record` on standard chado tables can safely use the `skip_validation` option, since they are preceded by code which checks that there will be no unique key constraint conflicts and provide values for all of the `not null` fields of each table. This change has further increased the performance of the loader in my tests, including for larger files.

spficklin:
@par12005 I'm going to close this out as I've started a complete rewrite of the GFF3 loader to take advantage of multi inserts as described in issue #1068 .  Your contributions won't be lost as I've built off of this branch.  See comments on issue #1040 .

par12005:
Thanks Stephen! That sounds good to me, hopefully I will have some time to contribute to the new branch this week!

Issue #660, opened 2018-09-18
Offer an Unpublish AND delete button for tripal content


bradfordcondon:
## Feature Request/Discussion

- [x] does this feature attempt to solve an existing problem with Tripal?

### Description

When a user deletes an entity via the GUI, the corresponding chado record is not deleted.

from @chunhuaicheng 

>However, I think it [deleting an entity] should delete the corresponding chado record as well. Here is our use case:
>Our curator uses the web interface to delete unwanted pubs that were imported by the Tripal pub importer. After she deletes the pubs, they still show up in the Tripal Pub Search (/find/publications) results, even after applying you fix.

I'm inclined to agree with him.  here was my response:

>Deleting a tripal entity, intentionally, does not delete the chado record (thats my recollection of this discussion i've had with stephen).

>I think this is not always the desired behavior, for example in the case described. I would go so far as to say that deleting the chado record should be the default behavior. That said, maybe we need to support both? Deleting the record brings up a dialogue that lets you specify you want to also delete the associated chado record ?

### note

Please do not confuse this issue with the bug reported in #652 and its corresponding PR #653

chunhuaicheng:
I second that supporting both is a good idea. I would suggest changing the name of deleting entity to 'Unpublish' record though. Simply because creating the entity is called 'Publish' in T3.

laceysanderson:
I like changing the label of the delete link/button to unpublish but worry this may be more difficult then we think since it's provided by "Drupal Magic".

My Current Vote: Do not delete from chado --too Dangerous ;-p

I'm very open to re-visiting this but feel currently there is not enough information for me vote "Support both". Primarily we should discuss how this would be configured? What would you prefer @chunhuaicheng since you brought this issue to light?

chunhuaicheng:
It's fine for me not to change the current setup. I'll still need a mechanism to delete data though. Any suggestion how to 'really' delete data from chado without having to issue SQL?

bradfordcondon:
>Do not delete from chado --too Dangerous ;-p

I just don't understand this: would you elaborate?  I think of Tripal as a way to serve and administer Chado content via Drupal.  Why leave the D out of CRUD?  Is it really only meant to be a one-way look into Chado?

How about instead of supporting both, we add an admin setting where you can check "Delete both Chado & Drupal records on delete" and "Delete only Drupal records on delete"?  That would be relatively painless wouldnt it?  Tripal checks a global variable before deleting the chado record?

[...]

PR #1061, opened 2020-05-25
Fix for issue #660

spficklin:
# Bug Fix

Issue #660 

## Description

This PR adds a new 'Unpublish' button to each entity form.  When clicked, the entity is unpublished from the site. This is the functionality that the 'Delete' button used to provide.  Also, this PR, redefines the 'Delete' button so that when clicked it both unpublishes and deletes from the underlying database (i.e. Chado).  Both the unpublish and the delete have a confirmation form.

Also, If a base record is deleted and there are other dependent records (e.g. feature records from an organism) that get removed via a cascade delete but those records have published content, then they get removed by a job to remove orphans that gets added after the delete.

## Testing?

Just try out each button.  One thing that happens when deleting a base record is that all ancillary records are cascade deleted via Chado table constraints.  So, if you delete an organism, you will also end up deleting all features, stocks, etc....  So, test that out and be careful.  

## Additional Notes (if any):

There are two possible issues here.   
1. If the delete will be large and cascade it may cause the site to timeout (i.e. trying to delete an organism with thousands of records).  
2. While there is a new 'Unpublish' button, the 'Delete'  button still looks the same. I have some concern that someone may not notice the change and think they are unpublishing when they click the 'Delete' button. If the person is paying attention the warning is there.  What do you think?


bradfordcondon (approved):
I tested unpublishing and deleting a simple record and this worked as expected.  The submitted code looks good to me.

I was thinking if there would be a good place to add this info to the readthedocs site, but I dont see an obvious place and I think the modal warnings are informative enough (for delete, anyway)

tripal/includes/TripalEntityUIController.inc
```diff
@@ -876,11 +901,56 @@ function theme_tripal_add_list($variables) {
...
+ *   The tripal_entity to delete
+ *
+ * @see confirm_form()
+ */
+function tripal_entity_unpublish_form($form, &$form_state, $entity) {
+  $form_state['entity'] = $entity;
+  $form['#submit'][] = 'tripal_entity_unpublish_form_submit';
+
+  $form = confirm_form($form,
+    t('Unpublish the record "%title"?',
```
bradfordcondon: Maybe a sentence to match the delete content message?

```suggestion
    t('Unpublish the record "%title"?  This will delete the Drupal record, but not the underlying (eg Chado) record.',
```

spficklin:
Getting back to this one.  Thanks @bradfordcondon for the review.  I committed the warning message text change you suggested. Do we need another review prior to merging this?

laceysanderson (changes requested):
I'm looking at this right now. Still working on a second functional review but the first thing that jumped out at me is that the new delete button is where the old unpublish button (labelled delete for max confusion) was. I'm worried that my users won't see the difference and will just click the delete labelled button in the same spot and accidentally delete from the database... Could we switch the location of the delete and unpublish buttons so that unpublish is on the far right. This way, if you click without reading, you're still achieving the same thing you were before. It just feels like it would need to be more intentional to delete from the database this way.
Screen Shot 2020-08-10 at 2 57 23 PM

UPDATE: I completed my functional review and this is my only comment. I just found it too easy for muscle memory to delete the record more fully then I intended ;-p

spficklin:
The delete button is on the far right to prevent accidental clicking of it.  We do have that confirmation page to catch accidents but if someone goes through too quickly they could just ignore it.   So the thinking is to keep that button far away.  

I agree, for folks used to the other layout with the 'delete' button meaning 'unpublish' this might cause problems.  I guess we could put the 'unpublish' and 'delete' buttons side-by-side on the right side.  What do you think of that?

laceysanderson:
>  I guess we could put the 'unpublish' and 'delete' buttons side-by-side on the right side. What do you think of that?

I like this idea. That should at least give people pause when they go to click the delete button that something obvious has changed. 🤞

bradfordcondon:
My preference in a vacuum is both on the right side, so if @laceysanderson approves too, perfect.

spficklin:
Okay, the button has been  moved to the right. I also retested everything to make sure it all still works. Which it does.  @laceysanderson it's ready for review again.

laceysanderson (approved):
Functional test looks perfect 👍 The new position of the button is great and everything still works as expected :-)

Issue #1064, opened 2020-05-27
Chado 1.2-1.3 upgrade - nonexistent sequences are not created


Ferrisx4:
### System information

* Tripal Version: 7.x-3.1 (540e0795e4bd63d97254e2d8ee058f361a8ed667)
* Drupal Version: 7.68
* PostgreSQL: 10.12
* PHP Version: 7.2.24

### Issue description

Upgrading from Chado 1.2 to 1.3 via the GUI process causes numerous errors about relations not existing. These relations are the sequences listed in the tripal_chado.install.inc file (linked below in **Steps to reproduce**) in the `tripal_chado_upgrade_chado_1_2_to_1_3_pre_alter()` function.

This code [was removed](https://github.com/tripal/tripal/commit/6cd8b469b8ead4304970eac7b650e8ee8431d78c#diff-1d2ba65756645ba7ee3de90f8f1066da) from the raw SQL file (default_schema-1.2-1.3-diff.sql) and executed with PHP due to 
```sql
CREATE SEQUENCE IF NOT EXISTS ...
``` 
functionality not existing prior to Postgres 9.5.

#### Steps to reproduce
Upgrade Chado from version 1.2 to 1.3 on a site/database where the sequences from https://github.com/tripal/tripal/blob/d40b939ed9330350b6b66896e8cf7e42c2047693/tripal_chado/includes/tripal_chado.install.inc#L504 do not exist for whatever reason (likely legacy site issues).

#### Errors
```
[site http://default] [TRIPAL ERROR] [TRIPAL_CHADO] FAILED. Line  3025, 0SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "analysis_cvterm_analysis_cvterm_id_seq" does not exist:ALTER SEQUENCE analysis_cvterm_analysis_cvterm_id_seq OWNED BY analysis_cvterm.analysis_cvterm_id;
```
#### Full error screenshot: 
![Screen Shot 2020-05-27 at 3 09 48 PM](https://user-images.githubusercontent.com/9593623/83062215-47184700-a02c-11ea-8a74-ba3715dc2b75.png)

Ferrisx4:
Note: for the purposes of i5k and Tripal 2 to Tripal 3 migration, including Chado upgrade from 1.2 to 1.3, uncommenting the CREATE SEQUENCE statements from the upgrade SQL file and adding IF NOT EXISTS allows the Chado upgrade to succeed on a test version of the site.

[...]

PR #1067, opened 2020-06-16
Fix for #1064 - Chado Upgrade 1.2-1.3 SQL

Ferrisx4:
# Bug Fix

Issue https://github.com/tripal/tripal/issues/1064

## Description

Reverts (and slightly modifies) changes made to the Chado install process in [2016](https://github.com/tripal/tripal/commit/6cd8b469b8ead4304970eac7b650e8ee8431d78c#diff-1d2ba65756645ba7ee3de90f8f1066da). Sequence creation is now (again) handled in the SQL file rather than in PHP.

As discussed in several meetings and in the related issue, this change will require databases to be using Postgres 9.5 or later. This is required due to the use of "CREATE SEQUENCE **IF NOT EXISTS**" in the upgrade process which did not exist prior to Postgres 9.5.

## Testing?

These changes can be tested by upgrading a Tripal 7 site from Chado 1.2 to Chado 1.3. 

## Additional Notes (if any):

Two commits in this PR are unrelated, (0f07995, 21d5f93), but they cancel each other out.


bradfordcondon (commented):
>As discussed in several meetings and in the related issue, this change will require databases to be using Postgres 9.5 or later. 

So this was OK'd by the community?    Are we incrementing the required Postgres version to 9.5?  According to our docs we support it at 9.3: should that be updated? (here) https://github.com/tripal/tripal/blob/7.x-3.x/docs/user_guide/install_tripal/server_setup.rst#server-setup



edit: after re-reading the issue, I see that @laceysanderson  was ok with `9.5`.  So please if we could include a note about that in the documentation i linked i'll be happy.   ie something like

> PostgreSQL 9.3 or higher (9.5 required for Chado 1.2 to 1.3 upgrade)

Lacey also requested a note in the README and a warning in the upgrade Chado form which would also be nice.

Thanks!

Ferrisx4:
Updated the documentation, readme, and the upgrade form to reflect the new requirement for PostgreSQL 9.5

bradfordcondon (approved):

bradfordcondon:
Thanks for adding the documentation!  if @spficklin  or @laceysanderson can confirm we're happy with this change we'll get it merged.

laceysanderson (approved):
Looks good to me! Thanks @Ferrisx4

Issue #1073, opened 2020-07-27
TripalImporter errors related to NCBI importing


risharde:
The following error is seen in /var/log/php-fpm/php-errors.log on a massive scale. In our case, the log files quickly filled our VM disk allocation which also causes other issues. This is on our live server.

As seen, it seems to be related to the TripalImporter where TaxonomyImporter.inc file contains the problematic code. The job being executed is importing NCBI data. 

[27-Jul-2020 16:39:47 America/New_York] PHP Warning: feof() expects parameter 1 to be resource, boolean given in /var/www/Drupal/sites/all/modules/tripal/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc on line 685 pid 16026
[27-Jul-2020 16:39:47 America/New_York] PHP Warning: fread() expects parameter 1 to be resource, boolean given in /var/www/Drupal/sites/all/modules/tripal/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc on line 686 pid 16026
[27-Jul-2020 16:39:47 America/New_York] PHP Warning: feof() expects parameter 1 to be resource, boolean given in /var/www/Drupal/sites/all/modules/tripal/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc on line 685 pid 16026

* Tripal Version: 3.1

bradfordcondon:
Thank you @risharde 

Do you have an example taxonomy ID or import job that is failing for you?

If we look at the line in question,  we can see that `fopen()` is failing with your built URL and returning `false` instead of a file resource.  Also possible that your `fpm` isnt set up to allow `fopen` on a remote URL?

https://github.com/tripal/tripal/blob/dd88e9fa9b497fe046913e97110b4f4721431edc/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc#L676-L688


We can make an easy PR to to check that our request was successful from NCBI.  However this wont fix your actual NCBI request, ~i wonder if its a valid taxon ID or if the API has changed~.  I get back XML at https://www.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=taxonomy&id=1 for example so i'd imagine the problem is your php configuration.


edit: see this issue

https://stackoverflow.com/questions/9038694/fopen-is-not-working-on-my-server

check that `allow_url_fopen` is set in your `php.ini`.  since you're using `fpm` you might have to find the `php-fm.conf` or equivalent config file instead

laceysanderson:
[...]

PR #1074, opened 2020-07-28
NCBI Taxonomy importer

par12005:
# New Feature/Bug Fix

Issue #1073 

## Description

Without these changes, if more than 3 requests to NCBI are made in one second, then the call to `fopen()` on [line 683 of `importRecord()`](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc#L683) will return a boolean (due to request frequency limit [set in place by NCBI](https://www.ncbi.nlm.nih.gov/books/NBK25497/#_chapter2_Usage_Guidelines_and_Requiremen_)), which causes the following while loop to continue infinitely and log warnings each time it calls `feof()` and `fread()`. On our server, this quickly filled the php error log and used up all of our disk space. I have implemented a few checks to prevent the infinite loop and the warnings, as well as an optional NCBI API key field so that the importer can operate at 10 EUtils requests per second, rather than 3.

- Added Optional NCBI API key field to `admin/tripal/loaders/chado_taxonomy`
- Sleep in between calls to `TaxonomyImporter::importRecord()` so as not to pass the limit [set in place by NCBI](https://www.ncbi.nlm.nih.gov/books/NBK25497/#_chapter2_Usage_Guidelines_and_Requiremen_)
- If for some reason an EUtils call fails, do not attempt to read the contents of the resulting boolean, and do not call `addItemsHandled()`.

## Testing?

- Navigate to `admin/tripal/loaders/chado_taxonomy`
- (optional) input your NCBI API key
- Run the Taxonomy importer
- See that it does not produce the warnings mentioned in #1073 

## Additional Notes:
These changes were based off of previous work we did with the [NCBI Publications importer](https://github.com/tripal/tripal/issues/1033)

### DIFF
--- a/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
+++ b/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
@@ -715,7 +715,7 @@ class TaxonomyImporter extends TripalImporter {
     ]);
 
     // Get the details for this taxonomy.
-    $fetch_url = "http://www.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?" .
+    $fetch_url = "https://www.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?" .
       "db=taxonomy" .
       "&id=$taxid";
 
--- a/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
+++ b/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
@@ -130,6 +130,30 @@ class TaxonomyImporter extends TripalImporter {
         already exist on this site.  This loader will also construct
         the taxonomic tree for the species loaded.'),
     ];
+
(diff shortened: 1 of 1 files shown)


bradfordcondon (approved):
- Importing single taxon looks good
- importing 4+ taxa without key looks good
- setting/unsetting API key works

Everything looks good here to me, thank you!

risharde:
Applied PR branch submitted by @par12005 
No tripal warnings / errors seen on dev site with new code
Performed a test of NCBI import via /admin/tripal/loaders/chado_taxonomy - which created a job for submission
Job was manually ran in drush successfully

laceysanderson:
Awesome! Thanks for the fix @par12005 and thanks for the reviews @bradfordcondon and @risharde! Merging with 2 reviews

PR #1075, opened 2020-08-07
Filtering sub elements in Tripal Entity Views.

spficklin:
# Bug Fix

Issue: https://github.com/tripal/tripal_analysis_interpro/issues/31

## Description

An issue was reported on the Tripal Analysis InterPro project by @martacds.  The problem is that when trying to add a filter to a view it's not working. The example is when trying to add a filter for annotations (CV terms)_ on a feature page (i.e. for InterPro terms).  Adding the filter works just fine, but it doesn't actually do any filtering.  

The problem is actually much bigger than just filtering on annotations (cv terms). It occurs for any field that has sub elements. For example, adding a 'Genus' or 'Species' filter also has no effect as those are sub elements of the `obi__organism` field.

## Testing?

To test this before pulling the branch, try testing adding a 'Genus' or 'Species' filter to any View that searches a feature.   It will not work.   then pull the branch and rerun the View. The 'Genus' or 'Species' filter should now work.  

You can also test with any other field that has sub elements (e.g. relationships, cv terms).

### DIFF
--- a/tripal/tripal_views_query.inc
+++ b/tripal/tripal_views_query.inc
@@ -112,7 +112,7 @@ class tripal_views_query extends views_plugin_query {
       // For fields compatible with the Tripal storage API, the
       // incoming $field_name is a combination of the entity term ID,
       // followed by the field name and the the sub element string, with
-      // sub elements children separated by a comma.  For non Tripal 
+      // sub elements children separated by a period.  For non Tripal
       // storage API the $field_name is a combination of the table name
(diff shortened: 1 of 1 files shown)


bradfordcondon (approved):
Thank you!

bradfordcondon:
https://github.com/tripal/tripal_analysis_interpro/issues/31#issuecomment-670372748 is a functional approval, so with that i will merge.  Thank you!

Issue #1077, opened 2020-08-12
Handling legacy content from T2 GFF Loader in Tripal 3


Ferrisx4:
In the Tripal 2 GFF loader, mRNA features such as polypeptides were loaded as `part_of` the parent whereas in the Tripal 3 GFF loader, these are now loaded as `derives_from`.
As such, in Tripal 3, the Protein Sequence field expects the `derives_from` relationship as well, which will only show features loaded with the Tripal 3 GFF loader, and **not** ones loaded from the Tripal 2 GFF loader.

There seems to be two possible paths for sites upgrading from Tripal 2 to Tripal 3 to allow these fields to be properly populated:
 1. Update the existing relationships of this type in the database from `part_of` to `derives_from` _during the migration process_.
 2. Alter the Protein Sequence field to show sequences of both relationship types.

laceysanderson:
I personally think option 2 is best as it will support sites which have already completed the upgrade. Furthermore, I think it's important to minimize altering legacy chado data as much as possible.

There are two paths to fix this issue:
1. Simply add the `part_of` cvterm [here](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc#L96)
2. Add configuration to the field to allow the relationship cvterm be a configurable list of terms including these two by default. This is more involved and would involve creating an [instance_settings_form() as done in the relationship field](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc#L693).

Both are acceptable though the second future proofs the field in case others have data with different terms.

bradfordcondon:
I vote for @laceysanderson 's first proposal (adding `part_of` to the field query).  If others have different relationship terms we could vote to support but i don't think its likely, as this is arising specifically because of how Tripal 2 loaded the data.  So building a UI etc not worth the work.

I have to say that I would also support converting the relationships from `part_of` to `derives_from` given that [I don't believe part-of is the proper term for the mRNA/polypeptide relationship](http://wiki.geneontology.org/index.php/Annotation_Extension_Relation:part_of) but then we're writing a migration script that is messing with people's data which is probably an "overbearing" solution.

Ferrisx4:
Thanks @laceysanderson and @bradfordcondon 
After discussing it, we agree, Lacey, that adding `part_of` to the clause will be the simplest solution that should work for all sites affected.
[...]

PR #1082, opened 2020-09-08
1077 legacy content

Ferrisx4:
# Bug fix

Issue #1077

## Description
For protein sequences in Tripal 3, "part_of" relationships are also shown to make the field backwards compatible with features loaded with the Tripal 2 version of the GFF loader.

As discussed in the issue, this change _does not_ migrate legacy content to use the 'derives_from' relationship. This is still being left up to the individual site to enact that change. 

## Additional Notes (if any):
Also fixes a tiny typo in so__transcript_formatter.inc

### DIFF
--- a/tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc
+++ b/tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc
@@ -93,7 +93,7 @@ class data__protein_sequence extends ChadoField {
       WHERE
         FR.object_id = :feature_id and
         CVT.name = 'polypeptide' and
-        RCVT.name = 'derives_from'
+        (RCVT.name = 'derives_from' or RCVT.name = 'part_of')
       ORDER BY FR.rank ASC
     ";
     $proteins = chado_query($sql, [':feature_id' => $feature->feature_id]);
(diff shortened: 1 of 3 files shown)


tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc
```diff
@@ -93,7 +93,7 @@ class data__protein_sequence extends ChadoField {
       WHERE
         FR.object_id = :feature_id and
         CVT.name = 'polypeptide' and
-        RCVT.name = 'derives_from'
+        (RCVT.name = 'derives_from' or RCVT.name = 'part_of')
```
bradfordcondon: `IN('derives_from', 'part_of')` is actually faster IF the the column is indexed (see: https://stackoverflow.com/questions/3074713/in-vs-or-in-the-sql-where-clause).   I don't know off the top of my head if it is or not.

(I also think it looks cleaner this way but i wouldnt be annoying about that :P)

bradfordcondon: It is: https://chado.readthedocs.io/en/rtd/_static/schemaspy_integration/tables/cvterm.html#Indexes

I therefore recommend using an IN clause instead.

bradfordcondon:
```suggestion
        RCVT.name  IN ('derives_from', 'part_of')
```

bradfordcondon (changes requested):

bradfordcondon (approved):
great thanks!

bradfordcondon:
merging this as a minor change.  Thanks for the contribution @Ferrisx4

Issue #1086, opened 2020-10-01
OBO Importer fails loading obi.obo


dsenalik:
### System information

* Tripal Version: 3.4
* Drupal Version: 7.70
* PostgreSQL Version: 12.4
* PHP Version: 7.2.33

### Issue description

A pull request #1087 is provided to fix several errors, and a few assorted typographical errors.

Loading the OBI ontology as found at http://www.obofoundry.org/ontology/obi.html
failed due to several problems.

Line 310 the example URL is no longer correct, this was updated here and at line 358

This obo file does not include a default namespace, and the loader web form indicated "Additionally, if a default namespace is not provided in the OBO header this name will be used as the default_namespace." but was not the case. New code was added at line 890 to make this true.

Some warnings due to more strict behavoir with a more recent php version have received additional array_key_exists() checks, e.g. this error:
array_key_exists() expects parameter 2 to be array, null given OBOImporter.inc:925

Line 986 "decrease" changed to "increase" in printed message.

Error message "TypeError: Argument 2 passed to t() must be of the type array, integer given, called in .../sites/all/modules/tripal/tripal/includes/TripalImporter.inc on line 580" was due to an extra t() call wrapping a logMessage() call line 1019.

Line 2082 a variable was assigned to itself instead of using the variable created in the preceding line.

#### Steps to reproduce
Tripal → Data Loaders → Chado Vocabularies → OBO Vocabulary Loader → ADD A NEW ONTOLOGY OBO REFERENCE
New Vocabulary Name: obi
Local File: path/obi.obo
Import OBO File

laceysanderson:
PR Merged

PR #1087, opened 2020-10-01
Oboimporter20201001

dsenalik:
# Bug Fix
#

Issue #1086 

## Description
A pull request is provided to fix several errors, and a few assorted typographical errors.

Loading the OBI ontology as found at http://www.obofoundry.org/ontology/obi.html
failed due to several problems.

Line 310 the example URL is no longer correct, this was updated here and at line 358

This obo file does not include a default namespace, and the loader web form indicated "Additionally, if a default namespace is not provided in the OBO header this name will be used as the default_namespace." but was not the case. New code was added at line 890 to make this true.

Some warnings due to more strict behavoir with a more recent php version have received additional array_key_exists() checks, e.g. this error:
array_key_exists() expects parameter 2 to be array, null given OBOImporter.inc:925

Line 986 "decrease" changed to "increase" in printed message.

Error message "TypeError: Argument 2 passed to t() must be of the type array, integer given, called in .../sites/all/modules/tripal/tripal/includes/TripalImporter.inc on line 580" was due to an extra t() call wrapping a logMessage() call line 1019.

Line 2082 a variable was assigned to itself instead of using the variable created in the preceding line.

## Testing?

Tripal → Data Loaders → Chado Vocabularies → OBO Vocabulary Loader → ADD A NEW ONTOLOGY OBO REFERENCE
New Vocabulary Name: obi
Local File: path/obi.obo
Import OBO File

### DIFF
--- a/tripal_chado/includes/TripalImporter/OBOImporter.inc
+++ b/tripal_chado/includes/TripalImporter/OBOImporter.inc
@@ -894,6 +894,10 @@ class OBOImporter extends TripalImporter {
         and array_key_exists('obo_name', $this->arguments['run_args'])) {
       $namespace = $this->arguments['run_args']['obo_name'];
     }
+    if (!$namespace and array_key_exists('run_args', $this->arguments)
+        and array_key_exists('uobo_name', $this->arguments['run_args'])) {
+      $namespace = $this->arguments['run_args']['uobo_name'];
+    }
 
     // If we can't find the namespace or the short_name then bust.
     if (!$namespace and !$short_name) {
--- a/tripal_chado/includes/TripalImporter/OBOImporter.inc
+++ b/tripal_chado/includes/TripalImporter/OBOImporter.inc
@@ -169,7 +169,7 @@ class OBOImporter extends TripalImporter {
 
 
   /**
-   * An array of used cvterm objects so that we don't have to lookup them
+   * An array of used cvterm objects so that we don't have to look them
    * up repeatedly.
    */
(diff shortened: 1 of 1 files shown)


spficklin:
Hi @dsenalik . Thanks much for these fixes!  The OBO loader is a bit complex so great job working through it to make fixes.  

I tested your fixes and it seems to work just fine.  it's currently running but I do have one fix that I'd like to ask if you could add into your code.  

Just after the block of code you added on line 894-897 can you add this additional block:
```
    if (!$namespace and array_key_exists('run_args', $this->arguments)
        and array_key_exists('uobo_name', $this->arguments['run_args'])) {
      $namespace = $this->arguments['run_args']['uobo_name'];
    }
```
There are actually two variables where the OBO name is stored in the form:  `obo_name` (for newly added OBOs) and `uobo_name` (when using a saved OBO).   

I hit on the error because I first tried to produce the originally error by loading the OBO as  "new" OBO.  The second time, I downloaded your code and tried it again. But because I already had it saved I didn't add it again.  And I hit on that same error again.  Adding that block of code fixed it for me.  If you could just add that into your code I can test it one more time and then I think all is good!

dsenalik:
I squashed my mess of initial commits into one commit, and now I have added your suggestion in a second commit. I'm glad it is (nearly) working!

laceysanderson:
Did a second functional review and the OBI OBO loads perfectly now. I also did a code review and 👍 Thank you for the contribution @dsenalik!

Issue #1089, opened 2020-10-04
Alter hooks for the field info and instances info arrays.


spficklin:
### Description
When creating a new extension module that creates custom tables we want to be able to create fields for the columns of those tables, but the `tripal_chado` module automatically generates fields for all new content types that are mapped to Chado tables.   This isn't a problem if the extension module comes before the `tripal_chado` module when the `hook_file_bundle_fields_info` is called. But if not, then the the extension module has no way to adjust any of the default settings set by the `tripal_chado` module.  We need an alter hook.

I propose two new hooks:
`hook_bundle_fields_info_alter(&$info, $entity_type, $bundle)`: for altering the info array of the fields.
`hook_bundle_instances_info_alter(&$info, $entity_type, $bundle) `:   for altering the info array of the field instances.

Unfortunately, these hooks can allow a module to change field settings for fields that they do not manage, but this isn't an uncommon problem with any Drupal hook.

spficklin:
I should also add. The reason I need this is because I'm working on the Tripal v3 compatible Tripal File module and it adds several new custom tables to Chado for associating files with Chado data.  the `tripal_chado` module automatically creates the fields (yeah!) but I need to adjust some of instance defaults.

spficklin:
So apparently, the alter functions already were supported but not documented in the API and also were not called everywhere they should have been.  I updated the PR to fix this.  So this is actually a bug too because the code in the cron job that updates the dashboard notifications isn't calling the alter hooks like it should.

laceysanderson:
PR Merged

PR #1090, opened 2020-10-04
Alter hooks for field/field instance info arrays

spficklin:
#

Issue #1089

## Description

See the issue for a description of the problem that this PR tries to address.

<s>This PR adds two new hooks:

hook_bundle_fields_info_alter(&$info, $entity_type, $bundle): for altering the info array of the fields.
hook_bundle_instances_info_alter(&$info, $entity_type, $bundle) : for altering the info array of the field instances.</s>

Update:  this is no longer an enhancement, because the hooks already exists but were not documented.  This PR now does the following:
1.  Documents the hooks in the API by adding "hook" functions
2.  Calls the alter functions in places where they were missing. This is anywhere the `hook_bundle_field_info` and `hook_bundle_instances_info` functions were called.
3.  Fixes the bug in the notification dashboard where fields could be created but were not using the alter hooks.

## Testing?

<s>The easiest way to test is to install the [tripal_file](https://github.com/tripal/tripal_file) module.  Be sure to checkout the `for_PR` branch as that will not have active development code. It contains just the basic installation that includes two content types: File and License.  
To test:
1. Do not yet pull this PR code.
2. Download the  the Tripal file module and checkout the `for_PR` branch.
3. Install the tripal_file module.  This will create two new content types: `File` and `License`.  
4. Create a new License content entity.  It will appear under 'Other' category when on the 'Add Tripal Content' page. 
5. Notice that the URI field is a textarea... we don't want this.  
6. Remove the 'File' and 'License' content types.
7. Disable and Unistall the `tripal_file` module
8. Switch Tripal to use this PR's code
9. Re-enable the `tripal_file` module
10.  Create a new license content and you should now see that the field is a textfield and not a textarea.  This is because the alter hook was called and the settings were adjusted.
</s>

Updated Testing instructions:

1. Do not yet pull this PR code.
2. Download the the Tripal file module and checkout the `for_PR` branch.
3. Install the `tripal_file` module.  This will create two new content types: `File` and `License`.  
4. Go to the Admin >> Structure >> Tripal Content Types.  Click  `manage fields` next to the `License` content type.  Notice that the URI field is a "text field".  This is correct. It means the alter hook was called properly.  
5. Remove the URI field from the License content type.  (leave this page open for easy access later).
6. In another tab, run the Drupal Cron at: Admin >> Reports >> Status Report and click the link 'run cron manually.' This will submit a Tripal job to update the Tripal notifications.  Run that job.  It will look at all bundles and check if any have new fields that are available and add a notification to the Dashboard to alert the admin user.
7.  Now go to the Dashboard, and find the section "Tripal Administrative Notifications":
image You should see the notification that a new `data__uri` field is available.
8.  Click the linked named `Import`.  This should add the new field.  Go back to the mange field page for the `License` content type (should have been left open from step 5).  Refresh.  You should see the URI field as a 'textarea' instead of a 'text field'. This is wrong.  This is the bug.  It means the alter hook was not called. 
9.  Now pull the code for the PR and repeat steps 6-8.  This should fix the bug and the URI field should be a 'text field' like it is supposed to be.

### DIFF
--- a/tripal/api/tripal.entities.api.inc
+++ b/tripal/api/tripal.entities.api.inc
@@ -636,6 +636,7 @@ function tripal_tripal_cron_notification() {
   foreach ($all_bundles as $bundle_name) {
     // Get the bundle object.
     $bundle = tripal_load_bundle_entity(['name' => $bundle_name->name]);
+    $term = tripal_load_term_entity(['term_id' => $bundle->term_id]);
     if (!$bundle) {
       tripal_report_error('tripal', TRIPAL_ERROR, "Unrecognized bundle name '%bundle'.",
         ['%bundle' => $bundle_name]);
@@ -647,7 +648,7 @@ function tripal_tripal_cron_notification() {
       $function = $module . '_bundle_fields_info';
       $entity_type = 'TripalEntity';
       $info = $function($entity_type, $bundle);
-      drupal_alter('bundle_fields_info', $info, $entity_type, $bundle);
+      drupal_alter('bundle_fields_info', $info, $bundle, $term);
       foreach ($info as $field_name => $details) {
 
         // If the field already exists then skip it.
@@ -674,7 +675,7 @@ function tripal_tripal_cron_notification() {
       $function = $module . '_bundle_instances_info';
       $entity_type = 'TripalEntity';
       $info = $function($entity_type, $bundle);
-      drupal_alter('bundle_instances_info', $info, $entity_type, $bundle);
+      drupal_alter('bundle_instances_info', $info, $bundle, $term);
       foreach ($info as $field_name => $details) {
 
         // If the field is already attached to this bundle then skip it.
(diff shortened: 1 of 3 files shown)


spficklin:
Well... apparently the hooks are already supported! But it wasn't documented (i.e. no hook_ api function)  Here's the lines of code the are called:

https://github.com/tripal/tripal/blob/04dee243fb329eff7dfaff0e26cb43c7e8f166e7/tripal/api/tripal.entities.api.inc#L749

and here:

https://github.com/tripal/tripal/blob/04dee243fb329eff7dfaff0e26cb43c7e8f166e7/tripal/api/tripal.entities.api.inc#L784

There are two problems
1.  These implements are slightly different from what I did
2.  They are not called in all locations where they should be. 

For backwards compatibility I think I should change what I have done to match these.  But I'm curious if this was something we added and never fully implemented?  Does anyone know if these hooks are used by any other module?

laceysanderson (approved):
Looks great!

Issue #1091, opened 2020-10-04
License can be added in readme file


sam-2200:
one can get through license by readme

PR #1092, opened 2020-10-04
Update README.md

sam-2200:
License is added in readme

Readme is updated

Issue #1091 

## Description
License is added in readme file 
One can get through the license wile reading readme which can get easy

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ PHPUnit tests will also be run in the Travis CI build.
 
 Read our [testing guidelines](https://tripal.readthedocs.io/en/latest/contributing/tests.html).
 
+# License
+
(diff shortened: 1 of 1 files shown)


spficklin:
Hi @sam-2200 Thanks for your contribution.  This isn't a significant contribution that would count under the Hacktoberfest expectations, and isn't a change we were looking for on the README as the license is clearly noted in the Github description of the project.   I'm going to close this out, but please feel free to look over our issue queue if you'd like to try to tackle any of those.  If you are looking for an issue that might be more appropriate for a first time contributor let us know and we can try to point you towards one, although we do try to tag those with a 'good first issue' label.

Issue #1088, opened 2020-10-04
PHP 7.4 + 8.0 Issues


spficklin:
Just pointing out that for PHP 7.4 the `array_key_exists` [function is deprecated.](https://www.php.net/manual/en/migration74.deprecated.php#:~:text=Using%20array_key_exists()%20on%20objects%20is%20deprecated.,property_exists()%20should%20be%20used.)  Ubuntu 20.04 uses php 7.4 so as folks upgrade they will see this warning on their Tripal sites.  

![image](https://user-images.githubusercontent.com/1719352/95007709-9e748800-05c7-11eb-9a7b-04b97a906cec.png)

spficklin:
I'm renaming this as a place to post problems that occur with PHP 7.4

bradfordcondon:
> array_key_exists

I think it’s only deprecated for use on objects, not arrays

For others finding and fixing this usage, note that `property_exists` reverses the arguments for the property/key vs the object/array.  See my PR #1096

spficklin:
More issue with 7.4 that don't show up on earlier versions of PHP.

![image](https://user-images.githubusercontent.com/1719352/96526114-8d678000-1231-11eb-8bc7-89abaa33aa62.png)

laceysanderson:
![Screen Shot 2020-11-13 at 6 13 43 PM](https://user-images.githubusercontent.com/1566301/99132467-f1385b00-25db-11eb-8c2b-2d97645dc2ba.png)

lilicamd:
Tripal Crop Docker
![Screenshot from 2021-02-11 16-41-34](https://user-images.githubusercontent.com/8963698/107713332-3ece6d80-6c88-11eb-9ef0-c50a1d0461a7.png)

lilicamd:
Tripal Crop Docker
![Screenshot from 2021-02-11 17-13-16](https://user-images.githubusercontent.com/8963698/107715453-8e169d00-6c8c-11eb-8eb6-300b91a5641f.png)

laceysanderson:
Heads up: I need to upgrade KnowPulse + Tripal Crop Docker to PHP8 thanks to urging from our IT department. Since there are still a number of incompatibilities, I've started a branch `1088-update-to-php8` to resolve them :-) Feel free to post more that you've noticed here so I can do a more thorough job!

dsenalik:
I have a tripal field issue that was a warning before, but with php8 becomes an error rendering a page not editable. This is while editing an "assay" page:

[...]

PR #1096, opened 2020-10-05
Use property exists instead of array_key_exists

bradfordcondon:
# Bug Fix
Issue #1088
Issue does not resolve all #1088 just warning pointed out by @spficklin  explicitly.

### description

`array_key_exists()` is deprecated when used on an object.  Switch usage to `property_exists()` for this call.


spficklin (approved):

spficklin:
Just a simple code fix so I think this only needs 1 review.  Thanks @bradfordcondon

PR #1104, opened 2020-10-14
Bug fix for duplicate properties generated by parseFeature() function when running test case testGFFImporterAttributeProperty()

risharde:
The parseFeature() function parses a single line in the GFF3loader and processes the GFF Features. In the test case provided by Peter, the parseFeature() within the GFF3loader duplicated the Gap property as it kept appending to the nested array. This caused the Gap property nested array to contain 3 elements (2 of 3 were duplicate values), instead of 2 elements (which contain unique values). This affected the rank values saved within the database. 

This error focused particularly on small_gene.gff 3rd line (test_gene_001) when it original reported a failed assertion.


risharde:
@spficklin found a way to resolve without this commit since the loader should only accept a single gap and throw an error otherwise.

Issue #1100, opened 2020-10-09
sio__references tripal field hits php execution limit, time out if too many features present


dsenalik:
I have a genome publication, and if I try to view the publication it times out due to php execution time limit (currently set at 30 seconds). The reason is that this publication is linked to a large number of genetic markers in several linkage maps used for that publication.
If I increase the php time limit, eventually the page is rendered in a couple of minutes, with an enormously long list of markers.

The time seems to be occupied mostly in the call to `chado_generate_var()`

It would be helpful to have a mechanism to limit the number of records of a given type that are returned. Does anyone have any suggestions?

spficklin:
HI @dsenalik I'm actually working on creating new fields for an extension module I'm working on.  So since I'm in the "fields" zone I can take a look at this field for you.   Increasing the PHP limit could potentially get you around the problem but it still means your users are waiting a very long time for the page to load, which isn't good.

spficklin:
Just curious @dsenalik how many records do you have linked to this publication?

dsenalik:
I have 2505 total references. 2473 of those are genetic markers, the major culprit. The page took just over 2 minutes to render on my test site with php execution limit increased.

spficklin:
@dsenalik can you test this branch and see if it fixes the problem: `1100-tv3-references_field`

dsenalik:
After about 40 seconds I received these errors:

- Notice: Undefined index: foreign keys in sio__references->load() (line 106 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sio__references/sio__references.inc).
- Warning: Invalid argument supplied for foreach() in sio__references->load() (line 107 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sio__references/sio__references.inc).
- Warning: array_key_exists() expects parameter 2 to be array, null given in sio__references->load() (line 119 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sio__references/sio__references.inc).
[...]

PR #1105, opened 2020-10-14
Fix for long running sio__references field

spficklin:
# Bug Fix

Issue #1100 

## Description

This fixes the `sio__references` field in the tripal_chado module.  The field was taking an extreme amount of time to load when thousands of records were associated with the publication. This is because it was using the `chado_generate_var` function.  To fix it, I replaced it with some more efficient SQL code.  

## Testing?

@dsenalik has thoroughly tested this as noted on the issue #1100 but I've asked him to approve it here. I think this now just needs a code review.  

But if you want to test, you have to have the following:
1.  A published publication entity
2. Thousands of records associated with the publication.   

To do this testing on a dev site I manually added entries to the `feature_pub` table just to have thousands of records

### DIFF
--- a/tripal_chado/api/tripal_chado.variables.api.inc
+++ b/tripal_chado/api/tripal_chado.variables.api.inc
@@ -119,13 +119,13 @@
  *     property table (e.g. featureprop) and you want the CV and accession
  *     but do not want the DB the following array would work:
  *
- *        $table_options =  array(
- *          'include_fk' => array(
- *            'type_id' => array(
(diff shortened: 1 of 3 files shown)


dsenalik (approved):
Functionally tested and a massive improvement in page rendering time

spficklin:
Since this has been tested by @dsenalik I'm marking as approver 1 reviewed but it needs a code review still.

laceysanderson (approved):
Code review done 👍 Looks good!

PR #1107, opened 2020-10-15
October corrections for typos

dsenalik:
As I encounter spelling errors or other typos, I keep a list. It's time to fix them!
25 various corrections, mostly in comments but a few in documentation or are visible to users.


spficklin:
Thanks @dsenalik . Much appreciated as always!

Issue #1108, opened 2020-10-16
Tripal field URLs should be clickable


dsenalik:
## Feature Request/Discussion

* This feature does attempt to solve an existing problem with Tripal
* This feature is Not Urgent

### Description
When the tripal field ```tpub__url``` displays a URL for a publication, it is not clickable

I propose a simple change to ```chado_linker__prop_formatter``` that will make any string matching a regex unto a clickable link ```/^https?:/i```

Pull request #1109 

### Additional information/screenshots
Example

![url-not-clickable](https://user-images.githubusercontent.com/8419404/96269147-99ccae00-0f8f-11eb-96cf-41bbca6f2af7.png)

becomes

![url-is-clickable](https://user-images.githubusercontent.com/8419404/96269169-a0f3bc00-0f8f-11eb-8b5c-0cca1c463012.png)

spficklin:
The PR that fixes this got merged so closing this out. Thanks @dsenalik

PR #1109, opened 2020-10-16
Tripal field URLs made clickable

dsenalik:
# New Feature

Issue #1108

## Description
This is a change to ```chado_linker__prop_formatter``` that will make a URL in a tripal field, specifically ```tpub__url``` into a clickable link, as described in issue #1108

### DIFF
--- a/tripal_chado/includes/TripalFields/chado_linker__prop/chado_linker__prop_formatter.inc
+++ b/tripal_chado/includes/TripalFields/chado_linker__prop/chado_linker__prop_formatter.inc
@@ -27,7 +27,12 @@ class chado_linker__prop_formatter extends ChadoFieldFormatter {
 
     $list = [];
     foreach ($items as $index => $item) {
-      $list[$index] = $item['value'];
+      $value = $item['value'];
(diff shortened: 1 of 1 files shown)


spficklin:
Hi @dsenalik .  Have you tired the Drupal [url function](https://api.drupal.org/api/drupal/includes%21common.inc/function/url/7.x) for creating this link?  It should be able to handle both the internal Drupal URIs and those that are external with an `http://` prefix.

dsenalik:
I don't think the ```url()``` function will actually change anything here, the value can be either a full external URI, or other text, and if it's not actually a URI then it gets a bit mangled, thus the need for the HTTP prefix check.  
However, I should have used the ```l()``` function instead of ```t()```.  
I have now provided my updated version, adding a ```target=_blank``` so that these external URIs will open in a new window|tab.

Maybe I'm wrong though and there would be a desire to have internal URIs handled similarly? The problem would be distinguishing those from ordinary text.

spficklin:
Looks great!  Thanks.

spficklin (approved):

Issue #1112, opened 2020-10-20
Bug in web services


spficklin:
### System information

* Tripal Version: 3.4
* Drupal Version:  7.x
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

Fields that link to entities are missing the link to the entity. For example, for any entity from the feature table (e.g. Gene) if you look at the record, the organism field will not have the URI for the organism record.  For example:

![image](https://user-images.githubusercontent.com/1719352/96528736-4c269e80-1238-11eb-9808-790f3bc9feba.png)

There should be a corresponding `@id` element in the organism stanza linking to the organism if it is published on the site.

spficklin:
Closing as PR with fix is merged.

PR #1113, opened 2020-10-20
Added fix for issue #1112

spficklin:
# Bug Fix

Issue #1112 

## Description

While I was working on the Tripal File module (which we need full compatibility with web services) I noticed that fields that link to other entities seem to be missing the `@id` element that links to the corresponding entity. For example, the organism field for any "feature" entity (e.g. Gene, mRNA, SNP, etc.) is missing the `@id` element for the `organism` field.  This very small PR fixes that.

## Testing?

1. Before pulling the branch for this PR, use web services to find an entity that derives from the feature table of Chado. 
2. Load the JSON from web services for the entity (use the JSON fomratter plugin in Chrome to view it nicely).
3.  Note that the `@id` is missing from the organism field.
4.  Pull the branch, and then reload the JSON.  The `@id` should be there.

Before:

image

After: 

image

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -520,6 +520,9 @@ class TripalContentService_v0_1 extends TripalWebService {
           // Index number (0, 1, 2...) of a hydra:member array.
           $temp[$k] = $this->sanitizeFieldKeys($resource, $v, $bundle, $service_path);
         }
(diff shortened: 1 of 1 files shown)


Ferrisx4:
Tested on a development site, the organism @id link appears as advertised.

spficklin:
Thanks @Ferrisx4 for the review!

laceysanderson:
Awesome, thanks for the fix @spficklin and thanks for the quick review @Ferrisx4!

Issue #1116, opened 2020-10-20
OBO Loader can't import EDAM vocaublary


spficklin:
### System information

* Tripal Version:  3.4
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

The OBO loader cannot load the EDAM vocabulary this is because of several problems:

1. The EDAM vocabulary OBO file does not have a default-namespace (cv.name)
2. In the OBO file, the IDs are listed as `EDAM_format:0005`, for example, but the actual term ID is  `format:0005`
3.  The OBO has multiple `idspace` entries in the header. These are the prefixes used for the terms, so not all of the terms belong to the same db record in Chado. The OBO loader currently doesn't support multiple `idspace`s, so we need to update the OBO to support this.
4. The `idspace` names include an `EDAM_` prefix which is not correct as the term IDs don't include `EDAM_`.
5.  The OBO file contains some `alt_id` entries that are improperly formated. This causes the OBO loader to fail, but it would be better to report the problem and continue on.

#### Error messages and screenshots

If you try to load the EDAM vocabulary http://edamontology.org/EDAM.obo you will get an error message that it cannot cache the term because there is no default-namespace.  This is just the first issue the loader encounters.

PR #1117, opened 2020-10-20
Fixes to load EDAM ontology.

spficklin:
# Bug Fix
Issue #1116

## Description

It is not possible to load the EDAM vocabulary into Chado using the Tripal OBO loader for the reasons mentioned in the issue linked above.   Most of them are issues with the formatting of the EDAM OBO and the organization of the EDAM vocabulary. One is a limitation in the OBO loader.

Unfortunately, because the EDAM OBO does not follow the standard properly some of the fix involved hardcoding some cleaning steps that are unique to EDAM.  I didn't like doing this, but I think it's the easiest way to deal with it.

## Testing?

1. Before pulling the code for this PR, add the EDAM vocabulary at Admin >> Tripal >> Data Loaders >> Chado Vocabularies >> OBO Vocabulary Loader.  Use the following 
   - Vocabulary Name:  EDAM
   - Remote URL:  http://edamontology.org/EDAM.obo
2.  Submit a job to import the EDAM vocabulary. You should get an error.
3. Pull this code.
4.  Re-submit a job to import the EDAM vocabulary. It should go in.
5.  Inspect your database tables.   
     - The EDAM record in the `cv` table should already exist because it was loaded by the Chado Prepare step.  the EDAM cv record will not the the highest cv_id value.
     - The `format`, `data`, `operation` and `topic` db records  should already exist because they are loaded with a Chado prepare.
     - Browse your Tripal site using these relative paths and check that the terms were uploaded for the vocabulary:  
        - `/cv/lookup/data`
        - `/cv/lookup/format`
        - `/cv/lookup/operation`
        - `/cv/lookup/topic`


Ferrisx4:
As per https://github.com/tripal/tripal_file/issues/8#issuecomment-716552950, this fix was seen to work on a dev site.

spficklin:
Thanks @Ferrisx4 for the functional review. Now it just needs  a code review.

laceysanderson (approved):
Code Review looks good 👍

Issue #1118, opened 2020-10-21
Bugs in Web Services


spficklin:
### System information

* Tripal Version: 3.4
* Drupal Version: 7.x
* PostgreSQL Version:  N/A
* PHP Version:  7.4

### Issue description

In the process of creating the File module which makes use of web services to share files, I have found the following bugs with web services

1.  If an entity has never been cached, all of the non "auto attach" fields have no link This makes it seem like those entities don't have that data, but they just might.   This happens because the code checks if there are no values and if none then it hides the field (similar to the behavior on the site). The problem is that in web services we don't load all of the fields so for an entity that has never been cached the non 'auto_attach' fields won't have values.  The solution is just to provide all of the fields regardless if they have a value or not.  The problem code is shown below:
![image](https://user-images.githubusercontent.com/1719352/96677303-2b3b7780-1324-11eb-85af-b0af6da8a109.png)

2.  If an "auto attach" field has a cardinality > 1 but has no elements the web services gives one record in the "member" list with no values. It should just give one.
![image](https://user-images.githubusercontent.com/1719352/96677419-68076e80-1324-11eb-9863-34b01f692966.png)

3.  Some of the fields with multiple cardinality result in error. The following is  the 'transcript' field for mRNA content types:
![image](https://user-images.githubusercontent.com/1719352/96677635-e401b680-1324-11eb-99f9-cfff54686617.png)

4.  When a field has cardinality >1 it will contain any number of elements that are either "resources" (i.e. published content) or non resources (i.e. non publisehd content). An example of a non-resource element is sequence coordinates for a gene.  There is no page for just sequence coordinates. Therefore, items without a published page should be considered properties and should not have an `@id` or `@type` keys.   An example is shown below:

![image](https://user-images.githubusercontent.com/1719352/96684779-013c8200-1331-11eb-9ff1-95d41d56038a.png)

PR #1119, opened 2020-10-21
Fixes for web services bugs

spficklin:
# Bug Fix

Issue #1118 

## Description

In the process of trying to make the new Tripal File extension module work with web services I noticed several bugs. They are described with screenshots on the issue listed above.  

## Testing?

1. Before loading of the code in this PR do the following.
2. Make sure your browser has a JSON viewer that allows you to view the web services responses and click on the links. It will make testing this a whole lot easier. 
3.  Also, if you have the demo citrus data loaded that makes testing easy too. But if you have genes with transcripts (mRNA) relationships that will work just fine as well. 
2. To test fix for bug 1:  
    - clear the entire site cache (this will remove cached fields): `drush cc all`. This is important because the problem only occurs with fields that are not cached.
   - using web services, go to any gene page. You will see that many of the fields that are 'null' even if you know they should have values. The relationship field is a good one because there are usually relationships. 
   - switch to this branch (`git checkout 1118-tv3-ws_bugs`) and reload the page. You should now see links to the fields that used to be null, and if you click on `relationships` you will see the data.  
3. To test fix for bug 2:
   - do not clear the cache. For this test we need a cache field.
   - switch back to the master branch:  `git checkout 7.x-3.x`
   - reload the same gene page, and find a field that has a cardinality greater than one but has no value.  A good one that is usually empty is the `contact` field.  Click that field to view it.  You should see that in the `members` array there is an `@id` and a `@type` but nothing else.  This is wrong.
   - switch back to this PR branch: `git checkout 1118-tv3-ws_bugs`
   - reload the page with the contact, the `mebmers` list should not be empty as it should.
4. To test fix for bug 3:  
   - do not clear the cache. For this test we need a cache field.
   - switch back to the master branch:  `git checkout 7.x-3.x`
   - on the gene page click on the link for the `transcript` field.  You should get an error stating:  `The key, @id, has not yet been added to the context. Use the addContextItem() function to add this key prior to adding a value for it.`
   -  switch back to this PR branch: `git checkout 1118-tv3-ws_bugs`. 
   -  reload the transcript page and you should see the list of transcripts for the gene.
5. To test fix for bug 4:
   - do not clear the cache. For this test we need a cache field.
   - switch back to the master branch:  `git checkout 7.x-3.x`
   - click on the link for the field named `map_position`.  You will that the members elements have `@id` and `@type` tags. This is wrong because those are not published resources. They are just properties about the position.  
   -  switch back to this PR branch: `git checkout 1118-tv3-ws_bugs`. 
   -  reload the map_position page and you should see the those elements are now gone.

### DIFF
--- a/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
+++ b/tripal_ws/includes/TripalWebService/TripalContentService_v0_1.inc
@@ -147,7 +147,8 @@ class TripalContentService_v0_1 extends TripalWebService {
         FIELD_LOAD_CURRENT, ['field_id' => $field['id']]);
     }
 
-    $this->addEntityField($this->resource, $term, $entity, $bundle, $field, $instance, $service_path, $expfield);
+    $field_service_path = $service_path . '/' . urlencode($term['name']);
+    $this->addEntityField($this->resource, $term, $entity, $bundle, $field, $instance, $field_service_path, $expfield);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
Function test ✅ 
Code Review ✅ 

Thanks for the fixes @spficklin

spficklin:
Thanks @laceysanderson for the review!

PR #1121, opened 2020-10-21
Update in_development.rst

katherynbuble:
Remove TripalMap, move to Extension Modules / Visualization display

### DIFF
--- a/docs/extensions/in_development.rst
+++ b/docs/extensions/in_development.rst
@@ -5,14 +5,6 @@ In Development
 The following modules are not yet ready for production or not fully Tripal 3 compatible.
 
 
-TripalMap
------------
-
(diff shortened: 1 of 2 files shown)


spficklin:
Hi @katherynbuble .  Looks great! Thanks  for making the change to update the module. Before we merge it you might also want to fix the listing on this page too:  https://tripal.readthedocs.io/en/latest/extensions.html

It still lists it as "In Development"

laceysanderson:
@spficklin that should be updated automatically when the docs are rendered since it's just a toc-tree directive which links to the category rather then each extension module.

spficklin:
Ahh... Thanks @laceysanderson for the reminder.   Sorry @katherynbuble sounds like all is fine then. I'll merge it!

Issue #214, opened 2017-12-08
Tripal version as global variable


spficklin:
Consider adding a global variable that indicates the current Tripal version.  Suggestion by @almasaeed2010

bradfordcondon:
this can be accessed via [`system_get_info`](https://api.drupal.org/api/drupal/modules!system!system.module/function/system_get_info/7.x).  so, might not be necessary.  

`dpm(system_get_info('module', 'tripal'));` yield this:

<img width="948" alt="Screen Shot 2019-07-29 at 10 09 52 PM" src="https://user-images.githubusercontent.com/7063154/62095206-a0b5b680-b24d-11e9-9e77-a42e8b795f98.png">

so we would have to *ahem* remember to update the `.info` file for this to work.  We could wrap this in an API call if you thought it would be helpful.  @almasaeed2010 do you still think it would be helpful or is the system_get_info call easy enough?

spficklin:
Oh, I agree @bradfordcondon.  I think this should be sufficient.  We could easily wrap this up in a nice little Tripal API function to make it easy on module developers.

laceysanderson:
Moving this over to Tripal 4 as it will not be done in Tripal 3.

PR #1127, opened 2020-10-30
214 tripal version api

Ferrisx4:
# New Feature

Issue #214

## Description

As suggested in #214, this adds two methods for developers to retrieve the current Tripal version:
 - via the tripal_version() command
 - via the trp-version drush command

## Testing

From a module, or with `drush php-eval`, call tripal_version(). It should return (not print) the version string.
From the command line, call `drush trp-version`. It should print the version string.

## Additional Notes

In the usual place for API functions, i.e. tripal/api/, there did not seem to be an appropriate place to include this function, so I stuck it at the bottom of the tripal.module file among a number of other Tripal functions. This may not end up being the best place for it.


spficklin:
Hey @Ferrisx4 I tested it and it works fine.  The only problem with it is related to what @bradfordcondon brought up in the issue, is that we keep forgetting to update the version of Tripal! Currently in the `.info` file it is set at `7.x-3.1` but we are already at `7.x-3.4`.  Since you posted the fix can you go through all of the .info files in the tripal core modules (e.g. tripal, tripal_chado, etc) and update them to say `7.x-3.5` since that is the next version we will release?

Ferrisx4:
hi @spficklin, sure no problem. For legacy modules such as `/legacy/tripal_analysis/` are we also keeping those up to date with the latest version number? I am guessing _yes_.
While I'm at it, I can also update the docs unless it is too early for that.

spficklin:
HI @Ferrisx4 yep, those legacy modules should have the same version as all of the core modules too.

spficklin:
Great! Thanks @Ferrisx4

PR #1132, opened 2020-11-07
Score php unit test + GFF fix for strands specified by question marks + Strand php unit test

risharde:
3 updates

GFF fix so that the loader interprets strands specified by question marks. Without this fix, GFF loader fails to create a valid db record since the question mark (char) is never replaced with an integer value which is needed for the strand column in chado.featureloc to do a valid insert. 

Score php unit test - 3 assertions/tests (positive integer, double, negative double)
Strand php unit test - 5 assertions/tests (+,-,?,. strands) + row count validation


spficklin:
Hi @risharde .  Thanks for the fix and updated tests. It seems, though, there are some extra files committed (0, 1, `__DIR__`, etc.).  Can you remove those?

risharde:
Will try to recommit

Issue #1135, opened 2020-11-21
mRNA pages with chromosomal FASTA loaded exceed memory


spficklin:
### System information

* Tripal Version: 3.4
* Drupal Version: 7
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

After loading a genome with large supercontigs, and after loading the corresponding GFF3 file, the mRNA pages would not load because they would exceed the default memory limits. Granted I only have it set to 2GB but it still shouldn't do that.   The problem was a result of a call to `chado_expand_var` on the `featureloc` table that caused `residues` column to be included.  We can control which foreign keys get included in the object but we can't exclude columns.  The two fields that caused this problem are the `so__cds` and teh `data__sequence_coordinates`. 

Also, related to sequences, in Tripal v2, mRNA pages would load the gene sequence, full length mRNA, CDS and protein sequences (if available).  We had code in the `data_)sequence` field to do this but it was commented out and this functionality is missing from Tripal v3.  The reason it was commented out was because the `data__sequence` is meant to handle the `feature.residues` column alone.

spficklin:
This has been fixed by the PR.

PR #1136, opened 2020-11-21
Fix for bugs with displaying sequences

spficklin:
# Bug Fix

Issue #1135 

## Description

This PR fixes two problems
1.  The `so__cds` and `data__sequence_coordinates` fields will cause an mRNA page to not load if the reference genome sequences are loaded and memory is set too low (but not unreasonable).  This is caused by a call to `chado_expand_var` that includes the `feature.residues` column of the reference sequence.
2.  When we ported the fields from Tripal v2 we never finished the sequence fields.  For example, Tripal sites will only show sequences if they are present in the `feature.residues` field and will not show derived sequences from the reference. Most notably, the mRNA pages were missing the full length mRNA (with introns spliced out), and parent gene sequences.  

## Testing?

Note:  The only way to test the first bug is to load a large genome (~3GB) with large supercontigs.  If you are not able to do this or don't have the time, then I think just checking existing gene and mRNA pages to make sure this code doesn't break them is sufficient.  If you don't have mRNA/gene pages then loading of the Citrus demo data from the Tripal tutorial should work.  

1. Prior to testing the code from this branch:
    1.  First go to an mRNA/gene page.
    2.  Make sure you have the sequence fields enabled:  sequence, sequence coordinates, sequence length, sequence checksum and if an mRNA page there will also be Coding Sequence and Protein Sequence.  If the fields are not enabled then you'll want to do that just to confirm they are working. If you have a large chromosome loaded and the PHP memory set at about 2GB the page will not load.
     3.  Also, make sure you have chromosomal, contig or some sort of reference sequence set that has mRNA associated with it loaded into your Tripal site.  
2.  Pull this code
     1.  Run a `drush updatedb` to push a new database update.  It will tell you a new sequence field is added.
     2.  Go to the Admin > Structure > Tripal Content Types and for both mRNA and genes click the "manage fields" link and then click the link at the top of the page to find new fields.  It should add a new 'Sequences' field.
     3. Click on the "Manage Display" tab. You'll see the new 'Sequences' field in the Disabled section, Move it up into the 'Sequences' table and save the changes.
     4.  Go to an mRNA page, and look for the new field in the 'Sequences' fieldset.  You should now see sequences for  the gene, CDS, full length mRNA, and protein if you have that loaded.  It will show these sequences regardless if you have values in the `feature. residues` column as it pulls them from the reference sequence.
     5. You should now be able to go back to Admin > Structure > Tripal Content Types, click "Manage Display" and disable all of the sequence fields except this new field without any loss of information.
     6.  Also, if you are on the "Manage Dispaly" page for the content type then you can click the link at the top to revert to the default layout.  This will automatically reset any customizations but will also make all the other sequence fields disabled except the new one. This is important because disabling them prevents queries from running twice for the same information.
3.  You can also verify that the field is added to web services:   `/web-services/content/v0.1/mRNA`

### DIFF
--- a/tripal/includes/tripal.fields.inc
+++ b/tripal/includes/tripal.fields.inc
@@ -911,11 +911,13 @@ function tripal_form_field_ui_display_overview_form_alter(&$form, &$form_state, 
   $fields_names = element_children($form['fields']);
   foreach ($fields_names as $field_name) {
     $field_info = field_info_field($field_name);
-    if ($field_info['type'] == 'kvproperty_adder') {
-      unset($form['fields'][$field_name]);
-    }
-    if ($field_info['type'] == 'cvterm_class_adder') {
-      unset($form['fields'][$field_name]);
+    if ($field_info) {
+      if ($field_info['type'] == 'kvproperty_adder') {
+        unset($form['fields'][$field_name]);
+      }
+      if ($field_info['type'] == 'cvterm_class_adder') {
+        unset($form['fields'][$field_name]);
+      }
     }
   }
 }
--- a/tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc
+++ b/tripal_chado/includes/TripalFields/data__protein_sequence/data__protein_sequence.inc
@@ -93,18 +93,12 @@ class data__protein_sequence extends ChadoField {
       WHERE
         FR.object_id = :feature_id and
         CVT.name = 'polypeptide' and
-        RCVT.name  IN ('derives_from', 'part_of')
+        RCVT.name IN ('derives_from', 'part_of')
       ORDER BY FR.rank ASC
     ";
     $proteins = chado_query($sql, [':feature_id' => $feature->feature_id]);
     while ($protein = $proteins->fetchObject()) {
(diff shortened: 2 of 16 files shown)


risharde:
@spficklin I've tested #1 i,ii (iii - I'm not sure what this means)
I've also tested #2 i to vi (All acted correctly here)
For #3, if I visit an mRNA web-services page, I can confirm seeing sequence_record

Overall, this commit looks good and works as planned with the minor test I was not able to confirm in #1 iii.

If iii is not required, I can go ahead and confirm the review. Please let me know.

risharde (approved):
I've tested #1 i,ii
I've also tested #2 i to vi (All acted correctly here)
For #3, if I visit an mRNA web-services page, I can confirm seeing sequence_record

Overall, this commit looks good and works as planned!

risharde:
Just to confirm, part iii is in order after @spficklin clarified that the sample data for Orange which was used in the test meets this requirement.

laceysanderson (approved):
Looks good 👍

spficklin:
Thanks @laceysanderson and @risharde for the reviews.

Issue #1143, opened 2020-12-27
$use_button variable not working on TripalImporter


spficklin:
### System information

* Tripal Version: 3.4
* Drupal Version:  7.x
* PostgreSQL Version: N/A
* PHP Version:  7.4

### Issue description

In PR #955  @laceysanderson added some code to the `includes/tripal.importer.inc` file to support a $use_button variable to hide the submit button. I believe this was meant for extension modules that want to completely override the form such as with the Analyzed Phenotypes module.  This code seems to have accidentally gone missing when I was trying to fix some merging problems. I'm adding it back in.  

I noticed this because of a custom module we're working on that needs this functionality too.

spficklin:
Fixed with merged PR.  Closing

PR #1144, opened 2020-12-27
Tripal Importer button

spficklin:
# Bug Fix

Issue #1143

## Description

This is a very simple code fix that restores code that was added in PR #955 but that accidentally went missing.


laceysanderson (approved):
Good catch -looks good 👍

spficklin:
Thanks @laceysanderson .  As this is a very simple fix I'm merging with only one review.

PR #1145, opened 2021-01-06
Fix for previous PR on use_button for importers.

spficklin:
# But Fix

Issue # 1143

## Description

In PR #1143 that was just merged the `$use_button` variable was left out of the parent TripalImporter class.  So this means any importer that doesn't specifically include it will break.  The PR #1143 re-added the missing code but broke all importers. This fixes that.

## Testing?

Just try to use any importer (I tried using the Taxonomy importer) before and after the fix.

### DIFF
--- a/tripal/includes/TripalImporter.inc
+++ b/tripal/includes/TripalImporter.inc
@@ -57,6 +57,7 @@ class TripalImporter {
    */
   public static $use_analysis = TRUE;
 
+
   /**
    * If the $use_analysis value is set above then this value indicates if the
    * analysis should be required.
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code review is 👍 
✅ Checked the FASTA importer on 7.x-3.x and it is indeed broken 👀 
✅ Switched to this branch and the FASTA importer shows up as it should 👍 

## On 7.x-3.x right now
Screen Shot 2021-01-06 at 1 32 54 PM

## Same page on this branch
**Not only does the page load without error but the button referred to is present :-)**
Screen Shot 2021-01-06 at 1 33 43 PM

laceysanderson:
One code + functional review is enough for this small fix. Merging -Thank you for both fixes @spficklin!

Issue #1147, opened 2021-01-14
Date Time fields in Tripal are showing the current time.


spficklin:
### System information

* Tripal Version: 7.x-3.4
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

After importing new features into Chado, and then publishing them, the date accessioned and date last modified are always showing the current date/time.  If you edit the entity the date value appears just fine in the form.  If you then save the entity then all of the dates  show the correct time afterwards.

The problem is that Tripal borrows the Drupal datetime field for displaying dates stored in Chado.  When the values get set in Chado for the first time, Tripal is setting milliseconds, but the Drupal datetime field doesn't know what to do with milliseconds in the formatter so it just defaults to showing the current date and time.  However, if you edit and save the entity the the Drupal datetime field saves the date without milliseconds and it can the display is fine thereafter.  

One fix for this is to adjust the load function of all datetime values from Chado to exclude the milliseconds.  

#### Steps to reproduce

Just go look at any feature page that has the "Time Acceesioned" field turned on and which has never been edited through the web interface.

#### Error messages and screenshots

Here's an example from the Tripal 3 demo site

![image](https://user-images.githubusercontent.com/1719352/104638304-dc6b5880-565a-11eb-88bf-ae54862d48ad.png)

spficklin:
oops this is a duplicate of #1048

spficklin:
Fixed with PR #1148

Issue #1078, opened 2020-08-13
Content Page Time Values Incorrect


Ferrisx4:
### System information

* Tripal Version: 3 (multiple sites)
* Drupal Version: 7

### Issue description
Certain time-based fields, such as `Time Last Modified` and `Time Accessioned` on content pages such as Gene or mRNA do not read the values from the database but instead report the current time, or when the page was loaded. This is seen on multiple Tripal 3 sites.

#### Steps to reproduce

Load up a content page that has `Time Last Modified` and `Time Accessioned` fields and observe that these fields have the current time. Upon refresh, it will always reflect the current time (see caching note below).

##### Page load:
![1_49](https://user-images.githubusercontent.com/9593623/90169599-cfc19c00-dd6c-11ea-85c5-b4e0ec9b47a3.png)

##### After refreshing a few minutes later:
![1_56](https://user-images.githubusercontent.com/9593623/90169616-d51ee680-dd6c-11ea-9475-cf888c5f6bdf.png)

#### Notes
 - On the Hardwoods site, and possibly others, the time is reported on the page as +1 hour to whatever time it is (server is set to EDT). This may be an issue of time zone support on that field which is configurable.
 - Caching: This bug is affected by caching, refreshing on some sites is not enough to get the value to change, but the originally reported value is still incorrect.

Ferrisx4:
Due to some [comments](https://github.com/tripal/tripal/blob/c9940778f7c29f4640e9d5d8a313804a23cc0ed9/tripal/includes/tripal.fields.inc#L868) and [descriptions](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/tripal_chado.fields.inc#L1066) in the code, it appears that this might be partly intentional. Perhaps the Discussion tag might be appropriate here

Ferrisx4:
After discussing it briefly during a Tripal Dev meeting, it seems that having these two fields display the current time is **not** intentional. The code was only favoring the current time over the value stored in the database for the purposes of create/edit content forms.

Setting the field to display as "Plain" instead of "Date and Time" makes the field display the value from the database, but not in a pretty format: `2013-10-18 13:58:24.761174`.

spficklin:
Oops I just added a similar issue #1147 and a fix in PR #1148

spficklin:
I believe this is fixed with PR #1148.

PR #1148, opened 2021-01-14
Datetime Fields Showing Wrong Date

spficklin:
# Bug Fix

Issue #1147 & #1078

## Description

Tripal borrows the datetime fields from Drupal for showing data in Chado that are dates, but that field cannot handle milliseconds in the date.    It's weird because the datetime field form widget can deal with milliseconds but not the formatter. It just sets the time displayed to the current time.  

Since milliseconds probably aren't the most critical when setting dates when records are added or updated, this PR gets around the problem by removing the millisconds from a time string in Chado if it is there.  

## Testing?

1.  Find a gene or transcript or other entity that has the timeaccessioned or timelastmodified fields.   You may need to enable the field on the content type if you have it hidden.   The entity should not have ever been edited manually through the web interface.
2. Load the page... notice the dates are always the current date/time.
3. Pull this branch
4. Go back to the entity page, and click the 'reload' tab to ensure the cache is not being used.  The date time should now show the actual date in the database.

### DIFF
--- a/tripal_chado/includes/tripal_chado.field_storage.inc
+++ b/tripal_chado/includes/tripal_chado.field_storage.inc
@@ -398,6 +398,16 @@ function tripal_chado_field_storage_load($entity_type, $entities, $age,
             // as the column value.
             $entity->{$field_name}['und'][0]['value'] = $record->$field_column;
             $entity->{$field_name}['und'][0]['chado-' . $field_table . '__' . $field_column] = $record->$field_column;
+
+            // For datetime columns we need to strip out the milliseconds if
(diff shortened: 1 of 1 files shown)


Ferrisx4 (approved):

Ferrisx4:
Tested this out on a site (same site from #1078) and it seems to work as advertised. The time fields read from the database correctly (with timezone offset) and no longer report the current time at page load.

spficklin:
Thanks @Ferrisx4 for the functional review!  This is such a small change that I don't think it needs a code review.  I'll merge.

Issue #1151, opened 2021-01-14
Sequences are missing the FASTA definition lines


spficklin:
### System information

* Tripal Version: 3.4
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

The sequences in the new data___sequence_record field are missing the FASTA definition line. This was pointed out by Marta in the Codefest in the Gene Pages team.  

#### Error messages and screenshots

```
error message
```

spficklin:
This issue is fixed!

PR #1152, opened 2021-01-14
Missing sequence definition line

spficklin:
# Bug Fix

Issue #1151 

## Description

In a recent PR #1136, a new `data__sequence_record` field was added as much of the feature functionality from Tripal v2 was missing from Tripal v3.  But that field is missing the FASTA definition lines. This PR adds a new local term `local:fasta_definition` and uses it when returning the sequence record values.  

## Testing?

1.  Make sure you have the most recent 7.x-3.x version of Tripal so that you have the new `data__sequence_record` field.
2. Make sure that on your gene and mRNA content types you have the `Sequences` field enabled (you can turn off all other sequence related fields and just use this one).
3. Go to a gene and an mRNA page. You will see that the sequences being shown do not have the FASTA style definition line.
4. Pull this PR branch, and run the `drush updatedb` code to get the new vocabulary term.
5. Return to the gene and mRNA pages and click the `reload` button so that it doesn't pull the data from cache.
6. You should now see the sequences with the FASTA style definition line.  

Here's a screenshot of the sequences using the Citrus data from the Tripal tutorial showing the sequences with FASTA deflines.
image

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.feature.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.feature.api.inc
@@ -793,7 +793,7 @@ function chado_get_fasta_defline($feature, $notes = '', $featureloc = NULL, $typ
     $defline .= "length=" . $length . "bp|";
   }
   if ($featureloc) {
-    $defline .= "location=Sequence derived from alignment at " . chado_get_location_string($featureloc);
+    $defline .= "location=Sequence derived from: " . chado_get_location_string($featureloc);
     $defline .= " (" . $featureloc->srcfeature_id->organism_id->genus . " " . $featureloc->srcfeature_id->organism_id->species . ")|";
   }
   if ($notes) {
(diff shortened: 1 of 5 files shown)


laceysanderson (approved):
Code Review looks good ✅ 
I did a functional test on a fresh Drupal 7 docker with only the Citrus example. The field worked on a gene, mRNA and genetic marker. All three had the defline this module adds which was correct and informative.

My only comment is the lack of spaces in the defline cause weird wrapping which initially caused me to miss seeing part of the defline...

## Screenshots.
It worked on the Gene page:
Screen Shot 2021-01-26 at 5 54 54 PM
And on the mRNA page:
Firefox_Screenshot_2021-01-27T00-02-14 240Z

And on a genetic marker page (not ideal but as I would expect)
Screen Shot 2021-01-26 at 6 05 48 PM

spficklin:
Thanks for the review @laceysanderson . I've updated the code to add a semicolon and a space to separate elements in the defline. That should fix the wrapping problem. Below is a screenshot of the fix.  If this looks good to you then we can merge.

Also, in the process of doing this I found a bug in the the `data__sequence_record` field that was the result of not initializing the items array in the `load()` function which was causing the field to not display even though it had values. 

image

laceysanderson (approved):
Updates look Perfect 👍

Ferrisx4 (approved):
The changes also look good to me. I tested them on an existing site and the deflines are shown as advertised. 
Screen Shot 2021-01-27 at 9 20 11 AM

laceysanderson:
Awesome, thanks for the fix @spficklin  and the review @Ferrisx4! This is ready to merge!

PR #1153, opened 2021-01-14
Fix for PHP compatibility

spficklin:
# Bug Fix

Issue #1040

## Description

This is a very simple fix. I don't think it even needs a functional review, just a code review.   In the newly updated GFF3 importer a bit of PHP code was used that is not compatible with PHP 7.0.  This fixes it so that it is backwards compatible. It was noticed because I got an error from it on the Tripal v3 demo site when trying to make the new `sequence__data_record` field to work.   After changing the code the error went away.


laceysanderson (approved):
✅ Code Review Approved.

Also, I agree this doesn't need a functional review 👍

spficklin:
Thanks @laceysanderson !

PR #1158, opened 2021-01-28
Bug Fixes pre v3.5 release.

spficklin:
# Bug Fix 
This PR adds a documentation fix and a bug fix to the GFF3 loader that I noticed when I was testing Tripal prior to the v3.5 release.

The problem is that the GFF3 loader was not adding all of the records to the `feature_relationship` table.   The issue is caused by the fact that the GFF3 loader organizes children of each parent by their start location, but if two children have the same start location it ends up only keeping track of one of them.  This PR fixes the problem by storing the children in an array.  I'm amazed we missed this with all the testing we did!

# Testing
1 Start with a brand new Tripal site or a site that does not have the demo citrus data loaded.
2. Import the Citrus demo data that comes with the User's Guide
3. Publish the genes
4. Go to the gene page and click on the transcript and the relationships links. You'll see only one mRNA.
image
5. Pull this PR, rerun the GFF3 import
6. Go back to the gene page, click the reload tab, and now all the transcripts are present.
image


risharde:
Hmmm I'm guessing there was something I didn't think about testing via the unit tests? Or was this an issue more related with a particular GFF file?

Ferrisx4:
I'm able to confirm that all of the transcripts and relationships appear with the latest code, however I got the following warning 10 times during the GFF load on step 11 of 26 (Associate parents and children):
```
array_key_exists() expects parameter 2 to be array, null given GFF3Importer.inc:2157
```
[GFF3Importer.inc](https://github.com/tripal/tripal/pull/1158/files#diff-c0647b59c110acc2f33edc479da389f046a630914107f9fe6c2d3d783794b0dbR2157)

spficklin:
Hey @Ferrisx4 . I've put in a check in the code that should prevent that warning message you got.

Ferrisx4:
Thanks @spficklin. I pulled it and tested it, loader no longer issues that warning

spficklin:
Thanks @Ferrisx4 for the review.  I don't think it needs a code review as the change is fairly simple.   I'm going to merge.

Issue #1159, opened 2021-01-28
sbo__relationship tripal field gets confused if one term is a substring of another


dsenalik:
### System information
* Tripal Version: 7.x-3.3 (more or less)
* Drupal Version: 7.78
* PostgreSQL Version: 12.5
* PHP Version: 7.3.26

### Issue description
The sbo__relationship tripal field uses a regex to insert links to entities. It gets very confused if one of the subject or object terms is a substring of the other, only part of the text becomes a link, and it is to the wrong entity

#### Error messages and screenshots
Example, "USDA ARS" gets a link to the "USDA" entity:
![20200128example](https://user-images.githubusercontent.com/8419404/106204558-c4c7c000-6182-11eb-89b9-8cf31ca14bed.png)

I fixed this by converting the stored SIO:000493 clause to an array, and working only on the appropriate portions in the formatter, and imploding when done.

I was also able to then easily fix the TO-DO and get bold to work safely and correctly.

My only question is whether returning an array in place of the "clause" string can break anything else.

PR #1160, opened 2021-01-28
Fix for sbo__relationship field Issue #1159

dsenalik:
# Bug Fix
Issue # 1159

## Description
Fixes problem described in Issue #1159 
modifications to 
sbo__relationship.inc
sbo__relationship_formatter.inc

phrase is now stored as an array, and not imploded to a string until links and bold have been added, regex replacement is completely eliminated. This allowed me to easily fix this part
```
      // @todo add back in bolding...
      // @todo Fix Current Bug: if type name is in the object name, weird bolding happens.
```
although the code there appeared to want to replace the type, not the name, in conflict with the comment.

I have another tiny bug fix thrown in for the get_rel_verb() function, missing "break" and a missing term there.

## Testing?
You just need a relationship where one term is a substring of another

20200128example
becomes
20200128example2
(we are on the "USDA" page, so it is proper for "USDA" to not have a link since it would point to itself)

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
@@ -433,7 +433,6 @@ class sbo__relationship extends ChadoField {
     $object_type = $entity->{$field_name}['und'][$delta]['value']['local:relationship_object']['rdfs:type'];
     $object_name = $entity->{$field_name}['und'][$delta]['value']['local:relationship_object']['schema:name'];
 
-
     // Remember the current entity could be either the subject or object!
     // Example: The genetic_marker, MARKER1 , derives from the sequence_variant, VARIANT1.
     // The above relationship will be shown both on marker and variant pages
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
@@ -59,6 +59,9 @@ class sbo__relationship_formatter extends ChadoFieldFormatter {
    */
   public function view(&$element, $entity_type, $entity, $langcode, $items, $display) {
 
+    // Load the Field class into scope for use below.
+    tripal_load_include_field_class('sbo__relationship');
+
     // Get the settings
     $settings = $display['settings'];
     $rows = [];
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
@@ -78,10 +78,6 @@ class sbo__relationship_formatter extends ChadoFieldFormatter {
       $verb = sbo__relationship::get_rel_verb($rel_type);
       $rel_type_clean = lcfirst(preg_replace('/_/', ' ', $rel_type));
 
-      $clause_parts = [ 'The ', $subject_type, ', ', $subject_name, ', ',
-        $verb, ' ', $rel_type_clean, ' ', $object_type,
-        ', ', $object_name, '.' ];
-
       // Handle some special cases.
       // For mRNA objects we don't want to show the CDS, exons, 5' UTR, etc.
       // we want to show the parent gene and the protein.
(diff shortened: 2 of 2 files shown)


laceysanderson (changes requested):
Thank you for the contribution! <3 

I have tested the changes and they perform as advertised :-) This approach definitely fixes the bug and makes formatting much more reliable. Unfortunately it is not compatible with web-services which is why I didn't take this approach originally :-( This is also not a backwards compatible change due to web-services.

That said, this definitely needs to be fixed and I like this approach. **How about we move the exploded version of the clause outside the value (i.e. `$entity->{$field_name}['und'][$delta]['clause_parts']`. That way it does not show up in web-services and we can keep the `entity->{$field_name}['und'][$delta]['value']['SIO:000493']` matching the `entity->{$field_name}['und'][$delta]['value']['label']`.**

## Screenshots from this branch
Screen Shot 2021-01-30 at 5 31 07 PM
Screen Shot 2021-01-30 at 5 30 30 PM
Screen Shot 2021-01-30 at 5 26 51 PM

tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
```diff
@@ -72,7 +72,7 @@ class sbo__relationship_formatter extends ChadoFieldFormatter {
       $subject_type = $item['value']['local:relationship_subject']['rdfs:type'];
       $object_name = $item['value']['local:relationship_object']['schema:name'];
       $object_type = $item['value']['local:relationship_object']['rdfs:type'];
-      $phrase = $item['value']['SIO:000493'];
+      $clause_parts = $item['value']['SIO:000493'];
```
laceysanderson:
```suggestion
      $clause_parts = $item['clause_parts'];
```

tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc
```diff
@@ -438,15 +438,17 @@ class sbo__relationship extends ChadoField {
...
     // and as such both subject and object names need to be shown.
-    $clause = 'The ' . $subject_type . ', ' .
-      $subject_name . ', ' . $verb . ' ' . $rel_type_clean . ' ' .
-      $object_type . ', ' . $object_name . '.';
+    // Keep the clause as an array for now so we can apply formatting
+    // and links in the formatter
+    $clause = [ 'The ', $subject_type, ', ', $subject_name, ', ',
+                $verb, ' ', $rel_type_clean, ' ', $object_type,
+                ', ', $object_name, '.' ];
     $entity->{$field_name}['und'][$delta]['value']['SIO:000493'] = $clause;
```
laceysanderson:
```suggestion
    $entity->{$field_name}['und'][$delta]['clause_parts'] = $clause;
    $entity->{$field_name}['und'][$delta]['value']['SIO:000493'] = implode('', $clause);
```

laceysanderson (approved):
I've applied by suggested changes and retested. The following screenshots confirm that the bug is still fixed and that the web-services are now correct. Also this approach is backwards compatible if anyone wrote an alternative formatted for this field as all the original values remain the same format. 

Screen Shot 2021-01-30 at 10 15 09 PM
Screen Shot 2021-01-30 at 10 15 23 PM
Screen Shot 2021-01-30 at 10 15 47 PM

NOTE: Generally, it is advised not to use keys outside the value array for formatters to ensure that that the content shown through web-services is the same as that displayed in the page. However in this case, clause parts is simply an exploded version of the values|clause making this acceptable in my opinion.

dsenalik:
Very nice. I suspected I might be breaking something, this looks much better.
Thank you
I tested here and it looks good.

spficklin:
Hi @dsenalik and @laceysanderson .  Thanks for taking time to work on this.  I have this inner voice telling me this isn't quite the correct fix.  by adding in another "clause parts" element outside of the 'value' of the array we are building into the load function a fix that is really meant for formatting.  I think we should just rebuild the clause in the formatter so we can do the regex properly.  I'll make the fix and see what you think.

spficklin:
So, I have taken this PR and made changes into a new branch  `dsenalik-20210128sbo__relationship`.  If you both agree to this then I think  @dsenalik you can just pull this new branch, merge it into your PR branch and we can then merge.  If you don't agree we can just delete the new branch I created.  Here's what I did:

1. I restored the load() function to what it originally was.
2. I moved the creation of the clause parts to the formatter instead. This required a few additional lines of code for getting a human-readable verb and relationship type.
3. I changed the `get_rel_verb` function in the `sbo__relationship` class so that it is static and can be called in the formatter.

That's it. the formatter changes that expect an array is still as it was.

dsenalik:
Thanks for the changes. I incorporated those, but now since we are no longer passing the parts to the formatter as an array, I don't see the need for the array any more, so I changed this further to format the individual variables before concatenating into the clause, mirroring the code from load() exactly at that step.

Problem. After clearing the cache, it works perfectly the first time for any particular contact, but the second time the page is visited or if I refresh, I get an error:

Error: Class 'sbo__relationship' not found in sbo__relationship_formatter->view() (line 78 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc).

That comes from this line:
```$verb = sbo__relationship::get_rel_verb($rel_type);```

Any idea what is wrong?

**Comment by laceysanderson** [comment]
Hi @dsenalik, it's just that the first time the entity is loaded, both field and formatter classes are loaded. On all subsequent loads, the field::load() was cached and only the formatter class is imported. To fix this you just need to import the field class at the top of the view:

```patch
diff --git a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
index 1928f309..225291a4 100755
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_formatter.inc
@@ -59,6 +59,9 @@ class sbo__relationship_formatter extends ChadoFieldFormatter {
    */
   public function view(&$element, $entity_type, $entity, $langcode, $items, $display) {
 
+    // Load the Field class into scope for use below.
+    tripal_load_include_field_class('sbo__relationship');
+
     // Get the settings
     $settings = $display['settings'];
     $rows = [];

```

spficklin:
Whoops. Sorry I didn't catch that when I wrote it.

spficklin:
Okay, to get this merged in quickly so we can make the 3.5 release. I'm going to go ahead and merge in this PR so that you get credit for it @dsenalik and then work in the fix from @laceysanderson.

dsenalik:
Thanks @laceysanderson I incorporated your correction.

spficklin:
Well, apparently, I can't merge this into the `dsenalik-20210128sbo__relationship` branch via the online tool. So, I'll have to do this via the command-line.  I'll re-open a new PR.

spficklin:
Oh, I just saw your message @dsenalik .  I think we can merge  directly to 7.x-3.x then.

dsenalik:
I just finished testing and it now works perfectly!

laceysanderson:
Hooray! 🎉  Thanks for the contribution @dsenalik and for getting us going on this 😉

PR #1161, opened 2021-01-28
Display Issue problem

spficklin:
# Bug Fix

Issue # N/A

## Description

Bug # 1:
While testing the current dev branch I found a display bug when a site is first installed.  Several of the feature specific fields, related to the sequence (checksum, length, residues, CDS and protein) were not inside of a Tripal Panel.  They were just floating on the right side.  See screenshot.
image

The problem is that when the layouts are first generated when the site is prepared on the first install, there is no field group caching and there was an if statement that tried to fake a cached info array for the field groups but it had a mistake in it.  I was able to fix this and simplify the code by simply rebuilding the cache with the `field_group_info_groups` function call. that was already being used.  I also took the time to simplify the code logic as well to make it a bit more readable and easier to find the bug. 

Bug # 2:
Also, in the process of fixing the display issue related to the sequence fields I noticed that the "Sequences Pane" was just empty with nothing in it if there was actually no sequence data loaded into Chado.  There should be at least a message.  See screenshot below.

image

Both bugs are now fixed with this PR and the results look correct:

image

## Testing?

To test this fix you will unfortunately need to reinstall Tripal, from scratch... twice..

1. First installation use the current 7.x-3.x code
    - Install Chado
    - Prepare Tripal/Chado
    - Create the Citrus organism
    - Create the Citrus analysis
    - Load the Citrus demo data
    - Publish the gene and mRNA features.
    - Go to the gen page and you'll see the messed up display items described above
2. Repeat the same steps above with a new installation but this time use this branch. You'll notice the problems are fixed.

### DIFF
--- a/tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record.inc
+++ b/tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record.inc
@@ -98,6 +98,12 @@ class data__sequence_record extends ChadoField {
     else {
       $this->addGenericReference($entity, $feature, $field_name);
     }
+
+    // The field should always return a 'value' key so even if we don't
+    // have a sequence we still need to add it
+    if (count(array_keys($entity->{$field_name}['und'])) == 0) {
+      $entity->{$field_name}['und'][0]['value'] = '';
+    }
   }
 
   /**
--- a/tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record_formatter.inc
+++ b/tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record_formatter.inc
@@ -28,10 +28,8 @@ class data__sequence_record_formatter extends ChadoFieldFormatter {
     $phase_term = chado_get_semweb_term('featureloc', 'phase');
     $fasta_defline = 'local:fasta_definition';
 
-    $content = [];
-
     foreach ($items as $delta => $item) {
-      if (empty($item['value'])) {
+      if (!$item['value']) {
         continue;
       }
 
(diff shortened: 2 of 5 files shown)


risharde:
@spficklin I've tested a clean Tripal 3.4 version and I do not experience the bug specified, however I believe it is because during the process of importing Citrus in, the link to create an Analysis content type does not load at first. I resolved this by doing a drush cc all and reclicking the link.

After completing the steps to import Citrus, the Gene page on my side shows the fields as intended within the Sequence pane. I also checked the mRNA pages and the fields are also within the Sequence pane. 

So on install, what seems to work is a force cache clear after install (Tripal 3.4)

Please advise.

spficklin:
Hey @risharde . Yes, I noticed that the link to create an analysis page didn't load either without a cache  clear.  For me me that was a separate problem from the display.  It's curious that you didn't have the same problem with the sequence fields not showing up in the proper place.  Did you start from a completely new database with no prior Tripal installation and are you using the current 7.x-3.x code?

edit: I should add... if you're using the Docker image you probably won't get this behavior because it already comes pre-installed with Tripal and already has the content types created.   For this test to work you have to manually do the installation of Tripal.

risharde:
@spficklin Indeed I started with a clean install (var/www/html was totally deleted) and also (var/lib/pgsql/data was totally deleted) and not the docker image either (Virtualbox, CENTOS 7, PHP 7.2 using Tripal Rapid Installer for CENTOS7 (https://gitlab.com/TreeGenes/tripal_rapid_installer_for_centos_7)) running , so I'm not sure what other difference there could be between environments. Very strange indeed!

spficklin:
Hmm... interesting. I was able to reproduce the behavior repeatedly on my end.  Multiple times I wiped clean the database and consistently after reinstall of Tripal got the screenshots I show above.  After I made the code fixes the problem was resolved for me.  So, i guess if someone else wants to try and reproduce the problem and see.  If not, then it may just be a problem that some folks see in which case this should fox it for them.  The good news is that despite these changes you saw nothing wrong with the display so it would seem nothing was broken with the changes.  @laceysanderson and @Ferrisx4 what say ye?  Just do a code review and merge or another functional test?

laceysanderson (commented):
I just did a code review and I don't see any problems... that said, there are enough changes that I don't trust just a code review at this point. I will do one more functional review now.

laceysanderson (approved):
I just did 2 fresh installs. For the first one I was on the 7.x-3.x and I was able to reproduce the problem you experienced @spficklin. Then I completely dropped the database and re-installed on the tv3.5-pre-release-fixes branch and confirmed the bug was fixed.

I also checked a number of other content types, both feature-based and otherwise to confirm that the display configuration was correct with all fields in the expected panes. Looks good 👍

spficklin:
Thank you @laceysanderson !

PR #1163, opened 2021-02-01
Documentation and Minor Fix Pre v3.5 Release

spficklin:
# Documentation & Bug Fix

Issue # N/A

## Description

This PR is primarily updates to the User's Guide documentation to reflect recent changes. It updates text and images. 

I did find one more bug with manual creation of publications.  I don't know how this broke but the pubprop.rank column doesn't have a default value and the `chado_linker__prop` field wasn't setting it properly.  For other prop tables this wasn't a problem because there is a default rank on those tables.  This PR fixes  that bug when adding to the pubprop table.

## Testing?

1.  Try to manually add a publication following the instructions in the User's Guide.  You'll get an error.
2. Download this PR and try again. You should be able to add the pub.


laceysanderson (approved):
✅ Documentation changes look good.
✅ Code review looks good.
✅ Functional test: manually add publication

I was able to replicate the error you encountered on 7.x-3.x and this branch fixes it 👍

PR #1166, opened 2021-02-11
7.x 3.x add new extension category

leetaei:
#Documention

## Description

Added one line docs/extensions.rst
Added a new file 'breeding.rst' for new category.

I'd like to have a new category 'Breeding' for Tripal 'Extension Modules' plage.
https://tripal.readthedocs.io/en/latest/extensions.html
BIMS (Breeding Information Management System) can be added to the category.

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -8,6 +8,7 @@ The below modules are Tripal 3 compatible and grouped roughly by category:
 
    extensions/administrative
    extensions/analysis
+   extensions/breeding
    extensions/data_input
    extensions/developer
    extensions/3rd_party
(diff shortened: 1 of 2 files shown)


spficklin:
I think  this is a great addition!  Thanks @leetaei .   @laceysanderson before we merge any thoughts?  Perhaps we should cross list some other modules under this category as well?  IF so, I think it's easier to just merge this in and then do a new PR with any cross-listings.

laceysanderson:
I agree that this is a great addition! Thank you @leetaei 

@spficklin: we will also need to cross-list BrAPI from the 3rd party integration in this category but I agree that it should be a separate PR and that this can be merged!

spficklin:
Okay, Thanks @laceysanderson . I'm going to merge then.  @leetaei you should see the updates in the online docs in about 10 minutes or so.

Issue #1167, opened 2021-02-11
sbo__relationship tripal field does not work for publications


dsenalik:
### System information

* Tripal Version: 7.x-3.3 (more or less)
* Drupal Version: 7.78
* PostgreSQL Version: 12.5
* PHP Version: 7.3.26

### Issue description
If two publications are linked by the ```pub_relationship``` table, such as a chapter linked to the containing book, the ```sbo__relationship``` tripal field will produce an error and not display the relationship.

I have a pull request (#1168) to fix this, as well as a related issue with titles longer that the ```chado_expand_var()``` length limit not displaying.

#### Error messages and screenshots
```
     Notice: Undefined property: stdClass::$name in sbo__relationship->getRelationshipSubject() (line 292 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc).
    Notice: Undefined property: stdClass::$name in sbo__relationship->getRelationshipObject() (line 349 of /var/www/2020-02-18/drupal-7.69/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship.inc).
```

spficklin:
I believe this has been fixed with the pull request that was recently merged so I'm going to close this out.

PR #1168, opened 2021-02-11
20210211sbo  relationship

dsenalik:
#

Issue #1167 

## Description
The ```pub``` table does not have a ```name``` column, this is the source of the error. Changed to ```title``` and moved up a line to be consistent with the order in nearby code.
If one desired, ```uniquename``` could also be used, but generates rather long output.

That long output revealed another bug, where ```chado_expand_var()``` will by default not return long text strings, so titles longer that 256 characters were not displayed.
Additional code has been added to make sure that the title is expanded when this happens.

The topic for discussion would be whether this can affect other tripal fields, and could there be a way to generalize this variable expansion.

## Testing?
You need two publications linked using the ```pub_relationship``` table, by for example the tripal_pub cv term "part of". Make sure that one has a title longer than 256 characters. Then make sure the ```sbo__relationship``` is present on the publication page.
You would have to make the two changes separately, apply the ```name``` to ```title``` correction alone to see the error where the long title disappears.

Example of broken display of a pub_relationship:
pub-relationship-error

Examples of the field when working after these changes have been applied, chapter linking to a book:
ref-back-to-book

and the reverse relationship, book linking to its chapters:
book-to-chapters


dsenalik:
Okay, this breaks the field display on project pages! I'll fix that dumb mistake, and make the white space consistent there as well.

spficklin (changes requested):
Upon testing this PR I needed to add relationships between publications.  So I tried to manually add them via the GUI. I found that I was not able to and the error message was that it couldn't find the subject publication.  The problem is due to a missing space in the autocomplete added ID in the title (e.g. "[id: 8]"). Somehow the space got removed from that identifier string and it kept me from adding for testing. I  fix that and I have a patch for you. I'm trying to figure out how to get it applied.

spficklin:
Hey @dsenalik .  In the process of testing your fix, I found a different bug that prevented me from adding relationships between pubs using the web interface.  This is the problem:
- Edit a publication and scroll down to the relationship widget.
- Enter the publication titles for the subject and object parts of the relationships. It should autocomplete.
- After autocomplete, notice that if you scroll to the far right of the title you'll see an "[id:X]" text where X is an id number for the pub.  This helps the field identify what the publication is. Unfortunately, this should have a space in after the colon but it doesn't, and somewhere along the line we adjusted some parts to accept the lack of a space but not others.  So, it gives the error....

I fixed this problem. Now all works, but because this PR is from your forked repository I can't figure out how to get my fixes back into this PR without some rigamarole.   So, below I cut-and-pasted a patch.  Could you copy the patch text below into a patch file named `0001-relationship_bug2fix.patch` and then run this command in your forked tripal repository:

Here's the command to run:
```
git apply --check 0001-relationship_bug2fix.patch
```
Then if you commit it to this branch it should show up in the PR and I can approve the changes.

Here's the patch
```
From ee86ac5d95603ff7b228175fe8608518d418d250 Mon Sep 17 00:00:00 2001
From: Stephen Ficklin <spficklin@gmail.com>
Date: Fri, 5 Mar 2021 17:19:07 -0800
Subject: [PATCH] Added some additional fixes that correct for bugs when
 manually adding a relationship between pubs in the GUI

---
 tripal_chado/api/modules/tripal_chado.pub.api.inc             | 2 +-
 .../schema__publication/schema__publication_widget.inc        | 4 ++--
 2 files changed, 3 insertions(+), 3 deletions(-)

diff --git a/tripal_chado/api/modules/tripal_chado.pub.api.inc b/tripal_chado/api/modules/tripal_chado.pub.api.inc
index 4b62a9c3..695453c1 100755
--- a/tripal_chado/api/modules/tripal_chado.pub.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.pub.api.inc
@@ -315,7 +315,7 @@ function chado_autocomplete_pub($string = '') {
   ";
   $pubs = chado_query($sql, [':str' => $string . '%']);
   while ($pub = $pubs->fetchObject()) {
-    $val = $pub->title . " [id:" . $pub->pub_id . "]";
+    $val = $pub->title . " [id: " . $pub->pub_id . "]";
     $items[$val] = $pub->title;
   }
 
diff --git a/tripal_chado/includes/TripalFields/schema__publication/schema__publication_widget.inc b/tripal_chado/includes/TripalFields/schema__publication/schema__publication_widget.inc
index 8b6193f3..7860ad55 100755
--- a/tripal_chado/includes/TripalFields/schema__publication/schema__publication_widget.inc
+++ b/tripal_chado/includes/TripalFields/schema__publication/schema__publication_widget.inc
@@ -58,7 +58,7 @@ class schema__publication_widget extends ChadoFieldWidget {
       $pub_id = tripal_get_field_item_keyval($items, $delta, $pub_item_id, $pub_id);
       if ($pub_id) {
         $pub = chado_get_publication(['pub_id' => $pub_id]);
-        $title = $pub->title . ' [id:' . $pub->pub_id . ']';
+        $title = $pub->title . ' [id: ' . $pub->pub_id . ']';
       }
     }
 
@@ -122,7 +122,7 @@ class schema__publication_widget extends ChadoFieldWidget {
     $title = $form_state['values'][$field_name]['und'][$delta]['pub_title'];
     if ($title) {
       $matches = [];
-      if (preg_match('/^.*\[id:(\d+)]$/', $title, $matches)) {
+      if (preg_match('/^.*\[id: (\d+)]$/', $title, $matches)) {
         $pub_id = $matches[1];
         $pub = chado_generate_var('pub', ['pub_id' => $pub_id]);
         $form_state['values'][$field_name]['und'][$delta]['chado-' . $field_table . '__pub_id'] = $pub->pub_id;
-- 
2.25.1

```

dsenalik:
Thank you for debugging this probably-hard-to-find error. I didn't do things exactly right the first time by applying this to the wrong branch, so hopefully I have now incorporated your changes correctly!

spficklin:
Thanks @dsenalik . The fix works! I approve and merge.

Issue #1165, opened 2021-02-08
Bulk Loader null value error


btski:
### System information

* Tripal Version: 7.x-3.3
* Drupal Version: 7.78
* PostgreSQL Version: 9.5.24
* PHP Version: 7.0.33

### Issue description

Null value would trigger a false error when the null value is called as a foreign key in another table. My proposed solution can be found [here](https://github.com/btski/tripal/commit/be4642efe82578c77c241c6d969213586ab6b43e#diff-347452757b032417e3eef4c4019ba6bbecefc8a29cc595bd0ced49510d0a979c). If this change appears correct, I'll push the correction to the main branch.

#### Steps to reproduce

Attempting to link biomaterial to the contact table via "biosourceprovider_id" foreign key when contact "name" field contains null value.

#### Error messages and screenshots

Two errors below were encountered **while fixing**:

Error 1
```
[site http://default] [TRIPAL ERROR] [TRIPAL_CHADO] chado_insert_record: Too many records match the criteria supplied for biosourceprovider_id foreign key constraint (Array())
ERROR (TRIPAL_BULK): 
Line 1998 biomaterial (insert) Unable to insert record into biomaterial where values:Array
(
    [taxon_id] => Array
        (
            [abbreviation] => PRRSV
        )

    [biosourceprovider_id] => Array
        (
        )

    [dbxref_id] => 872192
    [name] => GB000001
)
```

Error 2
```
ERROR (TRIPAL_CHADO): chado_update_record: When trying to find update values, too many records match the criteria supplied for biosourceprovider_id foreign key constraint (Array
(
)
)
[site http://default] [TRIPAL ERROR] [TRIPAL_CHADO] chado_update_record: When trying to find update values, too many records match the criteria supplied for biosourceprovider_id foreign key constraint (Array())
ERROR (TRIPAL_BULK): 
Line 2 biomaterial (insert) Unable to insert record into biomaterial where values:Array
(
    [taxon_id] => Array
        (
            [abbreviation] => SVA
        )

    [biosourceprovider_id] => Array
        (
        )

    [dbxref_id] => 872194
    [name] => GB078389
    [description] => SVA_7_2
)
```

spficklin:
Closing out as I think this is fixed!

btski:
@spficklin problem has not been fixed. 

To reiterate the problem:
[...]

Issue #1164, opened 2021-02-06
Bulk Loader optional setting


btski:
### System information

* Tripal Version: 7.x-3.3
* Drupal Version: 7.78
* PostgreSQL Version: 9.5.24
* PHP Version: 7.0.33

### Issue description

Attempting to get the Bulk Loader optional setting working. Currently, the optional setting only works for required values (not null?) that are absent. I'm not sure what criteria is used to define "required" values. I believe the intended behavior for the optional setting is that if a null value is provided for a field that is required to be non-null, the field and all children are omitted from upload. To reflect this understanding, I made a simple [modification](https://github.com/btski/tripal/commit/63734c169e8c54cf6b8181baa09040b0576a8068#diff-6c01ec69165fffbaf8e13dc21f320027a78d8c66619bfac7a6d3858dc49001a1). If this logic appears correct, I'll push the correction to the main branch.

Edit - Corrected previous problem with my bulk loader template format to resolve an error.

spficklin:
I believe this is fixed as per the merged PR.

PR #1170, opened 2021-02-23
Fix for optional treatment for bulk loader

btski:
Corrected bulk loader behavior when processing optional fields. Current behavior fails to load optional null values when the field is required to be non-null. Expected behavior is that the field be omitted.

# Bug Fix

## Description

[Issue #1164](https://github.com/tripal/tripal/issues/1164): Fix for bulk loader optional setting. Currently, the optional setting only works for required values (not null?) that are absent. I'm not sure what criteria is used to define "required" values. I believe the intended behavior for the optional setting is that if a null value is provided for a field that is required to be non-null, the field and all children are omitted from upload. To reflect this understanding, I made a simple [modification](https://github.com/btski/tripal/commit/63734c169e8c54cf6b8181baa09040b0576a8068#diff-6c01ec69165fffbaf8e13dc21f320027a78d8c66619bfac7a6d3858dc49001a1). 

[Issue #1165](https://github.com/tripal/tripal/issues/1165): Null value would trigger a false error when the null value is called as a foreign key in another table. My proposed solution can be found [here](https://github.com/btski/tripal/commit/be4642efe82578c77c241c6d969213586ab6b43e#diff-347452757b032417e3eef4c4019ba6bbecefc8a29cc595bd0ced49510d0a979c). 

## Testing?

Successfully loaded all data hosted on my Tripal site. Previously, I used filler text on non-null fields and manually removed these fields after loading data.

### DIFF
--- a/tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
+++ b/tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
@@ -522,33 +522,30 @@ function process_data_array_for_line($priority, &$data, &$default_data, $addt) {
     $data[$priority]['error'] = TRUE;
     return;
   }
-
-  // Check that template required fields are present. if a required field is
-  // missing and this
-  // is an optional record then just return. otherwise raise an error
+  
+  // Check for optional records in template. If an optional field is not found
+  // and the record is required, raise an error
   $skip_optional = 0;
   foreach ($table_data['required'] as $field => $required) {
-    if ($required) {
-      // check if the field has no value (or array is empty)
-      if (!isset($values[$field]) or
-        (is_array($values[$field]) and count($values[$field]) == 0)) {
-        // check if the record is optional.  For backwards compatiblity we need to
-        // check if the 'mode' is set to 'optional'
-        if ($table_data['optional'] or preg_match('/optional/', $table_data['mode']) or
-          $table_data['select_optional']) {
-          $skip_optional = 1;
-          // set the values array to be empty since we all required fields are
-          // optional and we can't do a select/insert so we don't want to keep
-          // the values if this record is used in a later FK relationship.
-          $values = [];
-        }
-        else {
-          $msg = "Line " . $addt->line_num . ' "' . $table_data['record_id'] .
-            '" (' . $table_data['mode'] . ') Missing template required value: ' . $table . '.' . $field;
-          tripal_bulk_loader_throw_error($msg, [], TRIPAL_WARNING);
-          $data[$priority]['error'] = TRUE;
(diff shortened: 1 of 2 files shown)


spficklin:
Hi @btski can you provide an example data set and template that causes the problem that this PR is meant to fix?  I'm not 100% sure what the problem is.   

Thanks

dsenalik:
I realize that I have this exact problem too!
I am loading organisms with a chado bulk loader, and the data file includes NCBI taxid and GRIN taxid columns. Some organisms don't have those, so for those I have to have a fake id "NULL" and delete the dbxref after loading. Even though my bulk loader has those with "optional" checked.

I tested this branch, and it appears that I can now load with the "optional" columns being empty.
I am attaching my test file and my bulk loader template.

[pull_1170_taxaimport.template.txt](https://github.com/tripal/tripal/files/6103645/pull_1170_taxaimport.template.txt)
[pull_1170_test1.tsv.txt](https://github.com/tripal/tripal/files/6103646/pull_1170_test1.tsv.txt)

spficklin:
So, after looking into this problem I think this PR isn't ideal.  There are two issues
1)  The problem can somewhat be circumvented if the 'required' option is  turned on for 'optional inserts'.  I think the logic in the bulk loader needs correcting. When chatting with @btski in the dev meeting this morning he seemed to think this PR wasn't quite right either for that reason.
2)  This PR changes the `chado_insert_record`  and `chado_update_record` functions which shouldn't be needed for fixing the bulk loader and changing those functions makes me nervous.  I think the fix should be restricted to just the bulk loader code.

I suggested to @btski to leave this PR here to remind us to come up with a better solution and so this issue doesn't get lost in the other outstanding issues.

spficklin:
After chatting with @btski here's what we think the solution should be.

1.  When editing a field in the bulk loader, under the 'Additional Options' section, the 'Make this field required' checkbox doesn't hold it's check and doesn't pick up what the setting is in the template.  This needs fixing first.
2. Next, we need to makes sure the checkbox for 'Make this field required' is set by default to checked if the column in the Chado table has a 'NOT NULL' designation and the end-user cannot uncheck it. It needs to stay fixed.  A message should be shown to the end-user indicating that the box must stay checked because the underlying table requires it.  

I think this will fix the original problem, but I did notice there was another issue with the template provided by @dsenalik that might need fixing too.  Also, it would be nice if the field view, when editing the bulk loader would show the options so that the user can tell which fields are required.

spficklin:
I'm closing out this PR in favor of PR #1177 .  @dsenalik and @btski can you test the new PR to see if it solves the problem for you (see the testing instructions).

Issue #1071, opened 2020-07-06
ChadoRecordList Class


spficklin:
## Feature Request/Discussion

* This feature does  attempt to solve an existing problem with Tripal
* I am open to developing or collaborating on an extension module if this is not a good fit for Tripal Core

* This feature is URGENT

### Description

In order to manage multiple records at once it was suggested in the Tripal Dev Call that we add a new `ChadlRecordList` class.  

### Your Specific Use Case

Primarily, we're interested in a way to handle multiple inserts in one batch rather than individually.  The ChadoRecord class only handles one record at a time.   The idea is that someone could instantiate a ChadoRecordList class, add new ChadoRecords for insert and then those records could be batch "inserted" once a certain number of records are added (or manually).

### Generally Applicable

We believe this will help improve performance of loaders and can be used to test improvments to PR #1057.

### Additional information/screenshots

It was decided in the Core Dev Meeting that we would use this issue to discuss a design for this. Also, it requires approval by other members of the PMC (i.e. @laceysanderson and/or @bradfordcondon)

spficklin:
@risharde @par12005 had indicated they will work on this.

risharde:
Initial work, creation of the class file, added the insertMultiple code to the new class, included the ChadoRecordList.inc file to the module for boot

https://github.com/risharde/tripal/tree/7.x-3.x-chadoapi-chadorecordlist/

Thinking about if we need to change the current names used as the functions so any suggestions or confirmations about the current method names would be appreciated.

Also idea functions, something like insertRow($values) and maybe insert($rows). insertRow would append an additional row to the values variable while insert would set the values variable to what is in $rows.

Would an ordered insertRow be useful? something like insertRow($index_location, $values) which would insert the row into a particular index of the array?

laceysanderson:
I think the method names should try to follow the ChadoRecords class for consistency. For example, 
```
$multiInsert= new \ChadoRecordList('organism');
foreach($lines as $line) {
    // Add values to the object to be saved for later insert in bulk.
    $multiInsert->addValues( ['genus' => $line[0], 'species' => $line[1] );

    // Save values. This will insert/update if the number of records has been reached.
   $multiInsert->save();
}

// Alternatively set all values at once.
$multiInsert->setValues([
    ['genus' => 'Tripalus', 'species' => 'databasica'],
    ['genus' => 'Tripalus', 'species' => 'ferox'],
    ['genus' => 'Chado', 'species' => 'schemata'],
[...]

PR #1172, opened 2021-03-02
ChadoRecord class bug

spficklin:
#Bug Fix

Issue #1071 

## Description

Loading the EDAM ontology via the TripalFile module caused this error

```
ChadoRecord::setValue(). The column named, "is_obsolete", requires a value for the table: "cvterm".
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] ChadoRecord::setValue(). The column named, "is_obsolete", requires a value for the table: "cvterm".
```

The problem was due to comparison of values where a `===` should have been used instead of a  `==`.  I've fixed it. 

This is such a simple fix I don't think it needs a functional review. Just a code check.

### DIFF
--- a/tripal_chado/api/ChadoRecord.inc
+++ b/tripal_chado/api/ChadoRecord.inc
@@ -108,13 +108,6 @@ class ChadoRecord {
    */
   protected $required_cols = [];
 
-
-  /**
-   * @var array
-   *   An array of default values.
-   */
-  protected $default_values = [];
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
I have just functionally tested this and it does indeed correct the indicated problem

spficklin:
Wonderful. Thanks @dsenalik!

Issue #1122, opened 2020-10-21
Perpetual Files


spficklin:
## Feature Request/Discussion

* This feature [does ] attempt to solve an existing problem with Tripal
* I  [am ] open to developing or collaborating on an extension module if this is not a good fit for Tripal Core

* This feature is URGENT

### Description

Currently the user quote system does not allow a file to stay on the system perpetually.   

### Your Specific Use Case

I'm currently working on the File module that will allow users to attach files to content types in Chado.  It supports attaching files that were uploaded via Drupal such as the GFF/FASTA importer forms. This way those files that are uploaded can then be associated with the analysis.  However, once a file is associated it shouldn't be automatically removed or have an expiration date.  

### Generally Applicable

While I need this for the Tripal File module, I can see that it would be useful for anyone who needs to keep a file around for a while.

spficklin:
Here is  a possible solution:

1.  Provide a new permission that lets users mark files as permanent.  Users with this permission can mark their files to not expire. This would typically be given to an admin.
2.  Provide a hook that lets modules mark files as permanent  this way the user doesn't have to remember to do it.  As soon as a file is associated then the Tripal File module can mark the file as permanent.
3. We could implement this by simply setting the table column `tripal_expiration_files.expiration_date` value to zero for files that should never be deleted.

spficklin:
This has been marked as a release blocker so that the Tripal File module can tag the 3.6 release as compatible.

PR #1174, opened 2021-03-09
Permanent Files

spficklin:
# New Feature

Issue #1122 

## Description

This PR adds the ability for a site admin to mark an uploaded file as permanent.  This way, files  that need to stay on the site without getting automatically removed will be persistent.   

This fix does the following:
1.  Allows for a value of -1 to be a set in the `tripal_expiration_files.expiration_date` table which indicates that the file should not ever expire.
2. Updates file page on the User's account page.
    - Puts the usage and limitations in a collapsible fieldset so the page isn't so overwhelming.
    - Adds a 'Make Permanent' link as an Action for a file.
    - Removes the height of the file browser and sets a max-height instead.
3. Adds additional arguments to the `tripal_reset_file_expiration` API call to allow for setting a file as permanent and for setting a hard number of days. These are optional for backwards compatibility.
4. Updates the CSS class names so they have a 'tripal-user' namespace to avoid future conflicts.
5. Adjusts the CSS to remove styling that isn't generic (i.e. red colored selected file, blue colored file names)

## Testing?

To test this.
1.  Upload several files. The easiest way to test this is to use the FASTA or GFF uploader just to get the file uploaded (but don't actually submit an importer job).  For testing, you can use the Citrus [demo files](https://tripal.readthedocs.io/en/latest/user_guide/example_genomics/genomes_genes.html) in the Tripal User's Guide.
2. After uploading several files, go to your user account page and click on the 'Files' tab.  You should see the files you just uploaded.  Click on each one, then mouse over of the action links to see what file ID they have.  You'll see the File ID in the URL of the link.
3. Click the 'Make Permanent' link on them all, they should be updated to never expire.
4. If you look at the `tripal_expiration_files` table then you'll see the files that should not be expired should have a `-1` as the expiration date.
5. Using Drush run this command:  ```drush php-eval 'tripal_expire_files()'```. The files should remain.
6. Now run this Drush command to set their expiration date to right now:  ```drush php-eval 'tripal_reset_file_expiration(X,FALSE,0)'```. Be sure to change the `X` to the file ID of the file you want to remove.  The zero in the 3rd argument will force the file to have an expiration date of today.  Now if you run the ```drush php-eval 'tripal_expire_files()'``` again the file that was marked for expiration should disappear.
7. Return to the File page on the User account.  Click a file that is marked as permanent and then click the link 'Renew Expiration'. This should return the file to having an expiration date.

Here's a view of the files page with one file marked as permanent and the other with an expiration date.

image


spficklin:
Actually, hold off on reviewing this PR. I just realized that some users can have the permission to upload files but probably shouldn't have the ability to mark files permanent.  Give me a bit of time to update this PR a bit.

spficklin:
Okay, I've added one last update that adds a new permission that let's folks set files as permanent.  This PR can be tested.

dsenalik:
I am getting this output when I click on "Make Permanent"


The other actions work (Delete, Download, Renew Expiration)

Update - I was stupid. Clear cache! It works now

When testing step 6 I have a permission error 
```
$ drush php-eval 'tripal_expire_files()'
unlink(/var/www/2020-02-18/drupal-7.69/sites/default/files/tripal/users/1/test1.fna):        [warning]
Permission denied file.inc:2382
```
because files are owned by ```www-data```.
Instead I need to run
```
sudo -u www-data drush php-eval 'tripal_expire_files()'
```

Everything else works as advertised!

dsenalik (approved):
Functional testing successful

spficklin:
Thanks @dsenalik ! Much appreciated.

Ferrisx4 (approved):
Code looks good, no glaring issues.
I also did a functional test - after a cache clear, everything works as advertised.

spficklin:
Thanks @dsenalik and @Ferrisx4 for the review.

PR #1176, opened 2021-03-13
Tripal typos 20210312

dsenalik:
# Bug Fix
## Description
My accumulated corrections to typographical errors I have been collecting since January.

In one case, the errors are in code where ```==``` is used instead of ```=>```.
For example in ```tripal_chado/includes/loaders/tripal_chado.pub_importers.inc```
```
  $message_opts = [
    'watchdog' == FALSE,
```
is corrected to 
```
  $message_opts = [
    'watchdog' => FALSE,
```
Functionally, it worked as intended since the variable ```$message_opts['watchdog']``` was undefined and therefore was evaluated as false.
But note that in two cases it was 
```
 'watchdog' == TRUE,
```
so this correction might affect error or status logging by now logging as intended to the watchdog.

All other typos are in comments or documentation.


dsenalik:
Because the comment says "...We do not want to log messages to the watchdog..."
I decided to change one typo fix from ```== TRUE``` to ```=> FALSE``` so that the current actual behavior is maintained.

spficklin (approved):

spficklin:
Thanks for all these typo fixes!

PR #1177, opened 2021-03-15
Bulk Loader Bug in Required Fields

spficklin:
# Bug Fix

Issue #1164 and replacemenet for PR #1170

## Description

This PR fixes the bug described in issue #1164 and maybe #1165.   Here's a summary of the problem, but it's a bit complex...

The bulk loader allows the end-user to set a record's field as "required". This forces the input file to have a value for that field.  However, if the end-user can set a record as 'optional', meaning that if all of the required fields are not present then the record is skipped and nothing is updated/inserted/selected.  The problem occurs when database columns are set as "not null".  Consider an insert on the organism table. The `genus` and `species` are both required, but the user did not check both as "required" but suppose the end-user set the record as "optional".   Now suppose a species value was missing from a value in the data table. The bulk loader would fail because it would try to pass a NULL value to the species. It would do this because there are no required fields for the bulk loader record and it ignores the fact that the fields are required for the database table.

What should really happen is that the fields that map to database columns that are marked as 'not null' should always be marked as required.  There is no possible use case where a database column used in the bulk loader that is marked as "not null" should not have a value.  

This PR fixes that by
1.  Forcing all fields mapped to a database table column of "not null" to be required. The user cannot uncheck it. See the screenshot below for how this looks to the user.
image
2.  It also adds a new "Required" column  to the page when editing a template so the end-user can easily  tell which fields are required. 
image

## Testing?
To test this PR, do the following:

1. Download the template and test file provided by @dsenalik on PR #1170 
2. Using the 7.x-3.x version of Tripal, try to import the data file. It will fail.
3. Switch to this branch.
4. Edit each of the fields in the template just so you can ensure that the "mark this field required" checkbox is selected.  The box will be checked automatically for columns in Chado that are "not null".    You have to do this because the template has no required fields and it's hard-coded so we have to fix it.  But the web interface on this branch should automatically set it. Just edit and save each field. It's annoying to do all of them but is has to be done to fix the template.  
5. Now import the file and it should work.
6. You should have a new organism in the database.

### DIFF
--- a/tripal_bulk_loader/includes/tripal_bulk_loader.admin.templates.inc
+++ b/tripal_bulk_loader/includes/tripal_bulk_loader.admin.templates.inc
@@ -426,6 +426,10 @@ function tripal_bulk_loader_modify_template_base_form($form, &$form_state = NULL
             '#type' => 'hidden',
             '#value' => $priority,
           ],
+          'required' => [
+            '#type' => 'item',
+            '#markup' => $field['required'] ? '&#10004;' : '',
+          ],
           'field_name' => [
             '#type' => 'item',
             '#markup' => $field['title'],
@@ -1848,7 +1852,16 @@ function tripal_bulk_loader_template_field_form_default_values($mode, &$form_sta
   $v['foreign_record'] = (isset($form_state['values']['foreign_record'])) ? $form_state['values']['foreign_record'] : 'NULL';
   $v['foreign_field_options'] = $ref_chado_fields;
   $v['foreign_field'] = (isset($form_state['values']['foreign_field'])) ? $form_state['values']['foreign_field'] : current($ref_chado_fields);
-  $v['required'] = (isset($form_state['values']['required'])) ? $form_state['values']['required'] : FALSE;
+
+  // The required checkbox should be forced as checked if this database
+  // column is set to 'not null'. Otherwise, use the form state.
+  if (array_key_exists('not null', $table_description['fields'][$field]) and
+      $table_description['fields'][$field]['not null'] === TRUE) {
+    $v['required'] = 'FORCE_TRUE';
+  }
+  else {
+    $v['required'] = (isset($form_state['values']['required'])) ? $form_state['values']['required'] : $template_field['required'];
+  }
 
   if (isset($original_field['regex']) && empty($form_state['storage']['regex']['pattern'])) {
     $form_state['storage']['regex']['pattern'] = $original_field['regex']['pattern'];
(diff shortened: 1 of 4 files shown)


dsenalik:
**Edit: see below, all of this went away mysteriously**
I am getting the error
```
ERROR (TRIPAL_BULK): Line 1 NCBI Taxonomy ID (insert) Missing Database Required Value: dbxref.accession
[site http://default] [TRIPAL ERROR] [TRIPAL_BULK] Line 1 NCBI Taxonomy ID (insert) Missing Database Required Value: dbxref.accession
[site http://default] [TRIPAL NOTICE] [TRIPAL_BULK] Skipping processing of dbxref due to previous errors
Execution aborted due to errors.
```

When I look at the required status for that, which comes from the NCBI Accession line
ncbitaxid
it appears as not required, but editing the "NCBI Accession" field
ncbiAccession-required
it is grayed out checked.
This is because accession is not null in the dbxref table.

So the problem in this example is that the behavior wanted is for accession to be required _if the record is inserted_. If accession is absent the dbxref record should not be inserted.

This is complicated!

dsenalik:
I don't know if it will help but here is my template after editing and saving every field
[20210316template.txt](https://github.com/tripal/tripal/files/6150206/20210316template.txt)

dsenalik:
I thought that that template was corrupted (it wasn't). I re-did the entire testing procedure and **this time it worked**.

New template
[20210316template2.txt](https://github.com/tripal/tripal/files/6150138/20210316template2.txt)

And this now appears as expected.
ncbitaxid2

The only thing different in these two templates are the two required fields, but for some reason in the first case I couldn't change them.

I've tried editing in reverse order, but it is always setting the check for required.
If I load that first "bad" template I can now edit and change the field.
It's entirely possible I just never properly saved it since I can't reproduce it.
So, in conclusion, **it works**

spficklin:
Hey @dsenalik thanks for testing.  I think the problem that you found was that the original template did not have the "required" setting turned on, but the fix I made to the code forces it to be on in the GUI but doesn't actually change the template itself.   So, to get around the inconsistency you actually had to edit each field and just re-save it (making no changes). That would force the template (behind the scenes) to set the required fields.   It sounds like you did that and it is now working.  

@btski before I merge this PR, can you test it too to make sure the issue is fixed for you templates?   To reiterate, you'll have to do the same for you templates and edit/save all of the fields so your templates pick up that required setting behind the scenes.

btski:
@spficklin would it be possible to invoke tripal_bulk_loader_template_field_form_default_values() on tripal_bulk_loader_modify_template_base_form()? The current version works for creating new forms. However, each field in a old form has to be manually saved in order to correctly propagate the "required" label via tripal_bulk_loader_template_field_form().

btski:
@spficklin issue [#1165](https://github.com/tripal/tripal/issues/1165) appears to persist (Error 1).

spficklin:
Okay, well if this fixes issue #1170 for you then I think we can merge this in.

spficklin:
Given the positive response from both @btski and @dsenalik on the review. I merged this in. I'll mark issue #1170 as fixed but not #1165

spficklin:
Whoops I wrote #1170 but that was the closed PR. I meant issue #1164.

Issue #1178, opened 2021-03-16
Chado Bulk Loader miscounts number of lines in a file


dsenalik:
### System information

* Tripal Version: 7.x-3.5
* Drupal Version: 7.78
* PostgreSQL Version: 12.6 (Ubuntu 12.6-1.pgdg16.04+1)
* PHP Version: 7.3.27-9+ubuntu16.04.1+deb.sury.org+1

### Issue description
Chado Bulk Loader progress messages indicate one line too many for number of lines in data file.
This happens because after all bytes have been read, ```feof()``` still returns false until the next read occurs.
It works as intended only when the last line does not end in a return.
The reason is explained in detail here for c++: https://stackoverflow.com/questions/5431941/why-is-while-feof-file-always-wrong
I propose to copy the exact same logic from the loader section where blank lines are ignored. This will catch the case of counting a null string on the last loop, as well as better match the loader logic.

#### Steps to reproduce
Compare value from ```wc -l``` to value reported by loader

#### Error messages and screenshots
Drush output for a file with header line and one data line
```
File: /somepath/test.tsv (3 lines)
```
```
Progress:
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (3 of 3) Memory: 54444328
```

spficklin:
I think this is fixed with the PR so closing. Thanks @dsenalik

PR #1179, opened 2021-03-16
filter out blank lines in bulk loader line count

dsenalik:
# Bug Fix

#

Issue #1178

## Description
Resolve this issue by copying the logic for ignoring blank lines from the loader up to the routine that counts the number of lines

## Testing?
For any bulk loader job, compare the number of lines reported by drush to the number of lines in the file reported by ```sed -e 's/^\s*$/d' | wc -l```

For the same file used for the example in Issue #1178 the output is now correct
```
File: /somepath/test.tsv (2 lines)
```
```
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (2 of 2) Memory: 54448560
```

### DIFF
--- a/tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
+++ b/tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
@@ -152,11 +152,15 @@ function tripal_bulk_loader_load_data($nid, $job_id) {
   $node = node_load($nid);
   print "Template: " . $node->template->name . " (" . $node->template_id . ")\n";
 
-  // Determine the total number of lines in the file.
+  // Determine the total number of non-blank lines in the file.
   $total_lines = 0;
   $handle = fopen($node->file, "r");
   while (!feof($handle)) {
     $line = fgets($handle);
+    $line = trim($line);
+    if (empty($line)) {
+      continue;
+    } // skips blank lines
     $total_lines++;
   }
   fclose($handle);


spficklin:
Hi @dsenalik . I think this fix still miscounts lines in the middle of the file that are blank.  For example, if I insert just blank line (just a carriage return on the line) in the middle of the fill then the counts is too low and doesn't match what `wc -l` would return.

dsenalik:
**My mistake in describing how to test.**
With this change, the loader does not count blank lines, ever.
That was to match the exact logic in the loader at about line # 390.
So, ```wc -l``` is not expected to match now, ```sed -e '/^\s*$/d' | wc -l``` is expected to match, I have updated the pull request to say this now.
It could still be considered confusing since it counts the optional header line. That could be a whole other issue.

I tested with a file: header, data line, blank line, data line
and it counted as 3 lines
```
[|--------------------------------------------------]  0.00%. (0 of 3) Memory: 54329808
[|||||||||||||||||----------------------------------] 33.33%. (1 of 3) Memory: 54330416
[||||||||||||||||||||||||||||||||||-----------------] 66.67%. (2 of 3) Memory: 54348528
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (3 of 3) Memory: 54348240
```

I ran again with the blank line now containing data (no blank lines)
```
[|--------------------------------------------------]  0.00%. (0 of 4) Memory: 54329192
[|||||||||||||--------------------------------------] 25.00%. (1 of 4) Memory: 54329800
[||||||||||||||||||||||||||-------------------------] 50.00%. (2 of 4) Memory: 54347912
[||||||||||||||||||||||||||||||||||||||-------------] 75.00%. (3 of 4) Memory: 54348360
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (4 of 4) Memory: 54347248
```

And I put ten blank lines at the end, exact same line counts.

spficklin (approved):

spficklin:
I see, yes that makes sense. Thanks for the clarification.  This PR looks good!

Issue #1180, opened 2021-03-17
Chado Bulk Loader shouldn't count header line in line count


dsenalik:
### Description
I alluded to this in pull #1179 
Since pull #1179 was merged, the chado bulk loader reports the number of non-blank lines in a file. One of those lines is optionally a header line.
It would be conceptually more clear if the loader reports the exact number of lines of data loaded. The code already skips counting that line when loading, but the way the code was written, at the end it updates with the total number of lines, thus including the header line.

### Your Specific Use Case
My use case is loading any file with a header line lol.

spficklin:
This issue has been resolved with the recent PR. Thanks again @dsenalik

PR #1181, opened 2021-03-17
1180 tv3 bulk loader line count

dsenalik:
# New Feature

Issue #1180 

## Description
Alter the chado bulk loader so that the total line count of a data file does not included the header line.
In this way, it now reports actual number of lines of data that are loaded. The code as it existed actually did not count the header line while loading, but the earlier code to get the total number of lines in the file did.

## Testing?
Load any data file with a header line, and with "File has a header" set to "Yes".
After this pull request, the number of lines will be reduced by one.
Example of a file with two lines of data and a header line
**Before**
```
[|--------------------------------------------------]  0.00%. (0 of 3) Memory: 54331672
[|||||||||||||||||----------------------------------] 33.33%. (1 of 3) Memory: 54332752
[||||||||||||||||||||||||||||||||||-----------------] 66.67%. (2 of 3) Memory: 54351328
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (3 of 3) Memory: 54351048
```
**After**
```
[|--------------------------------------------------]  0.00%. (0 of 2) Memory: 54335776
[||||||||||||||||||||||||||-------------------------] 50.00%. (1 of 2) Memory: 54336856
[|||||||||||||||||||||||||||||||||||||||||||||||||||] 100.00%. (2 of 2) Memory: 54355152

```


tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
```diff
@@ -476,7 +483,7 @@ function tripal_bulk_loader_load_data($nid, $job_id) {
       $loaded_without_errors = FALSE;
       break;
     }
-    tripal_bulk_loader_progress_bar($total_lines, $total_lines);
```
dsenalik, line 479: Changed because this hid the fact that total_lines included a header while num_lines did not

tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc
```diff
@@ -164,10 +164,14 @@ function tripal_bulk_loader_load_data($nid, $job_id) {
     $total_lines++;
   }
   fclose($handle);
+  // do not count header line in the non-blank line count
+  if (($total_lines) and (preg_match('/(t|true|1)/', $node->file_has_header))) {
```
spficklin, line 168: Could you do this instead:
```
if (($total_lines) and ($node->file_has_header === TRUE)) {
```
The `preg_match` is meant for strings but I would think that `$node->file_has_header` is a boolean?

dsenalik: No, for some reason it's not. var_dump says for $node->file_has_header
```string(1) "1"```
I just copied that exact logic from pre-pull line # 335

Comment (dsenalik) [reply tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc:168]: It is super-ultra-not-boolean! If I set File has header to "No" it is now
```string(1) "2"```

From ```tripal_bulk_loader.chado_node.inc```:
```
  $form['loader']['has_header'] = [
    '#type' => 'radios',
    '#title' => t('File has a Header'),
    '#options' => [1 => 'Yes', 2 => 'No'],
    '#weight' => -7,
    '#default_value' => (isset($node->file_has_header)) ? $node->file_has_header : 1,
  ];
```
but right below in keep_track_inserted
```
    '#options' => [1 => 'Yes', 0 => 'No'],
```

dsenalik: I checked and the "No" value is never tested for.
Just because this bothers me a tiny bit I am going to change No from a 2 to a zero.

spficklin:
I tested and confirmed this fix works.  Thanks @dsenalik !

Issue #1183, opened 2021-03-22
schema__publication tripal field bug if multiple publications


dsenalik:
### System information

* Tripal Version: 7.x-3.5
* Drupal Version: 7.78
* PostgreSQL Version: 12.6 (Ubuntu 12.6-1.pgdg16.04+1)
* PHP Version: 7.3.27-9+ubuntu16.04.1+deb.sury.org+1

### Issue description
This appeared on an analysis page, specifically a Genome Annotation page but this bug would show up anywhere the field was used. For an analysis with two linked publications, when editing the analysis and trying to save, the following appeared:
Could not save the TripalEntity: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "analysis_pub_c1" DETAIL: Key (analysis_id, pub_id)=(9, 509) already exists.

The log message is a bit more informative:
PDOException: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "analysis_pub_c1" DETAIL: Key (analysis_id, pub_id)=(9, 509) already exists.: UPDATE chado.analysis_pub SET analysis_pub_id = :analysis_pub_id, analysis_id = :analysis_id, pub_id = :pub_id WHERE analysis_pub_id = :old_analysis_pub_id ; Array ( [:analysis_pub_id] => 14 [:analysis_id] => 9 [:pub_id] => 509 [:old_analysis_pub_id] => 14 )

To see why this fails, look at my current content
```
SELECT * FROM chado.analysis_pub WHERE analysis_id=9;

 analysis_pub_id | analysis_id | pub_id 
-----------------+-------------+--------
              12 |           9 |    509
              14 |           9 |    358
(2 rows)
```
So it is trying to save 9|509 under analysis_pub_id=14 on top of the 9|358 instead of properly on analysis_pub_id=12 thus leaving two 9|509 at the same time which triggers the duplicate key violation.

I found out why, too, it's this line in ```schema__publication.inc```:
```$entity->{$field_name}['und'][$i]['chado-' . $field_table . '__' . $pkey] = $linker->$pkey;```
where $linker is defined inside the loop above where we process each publication. Thus it only contains the last-encountered value.

I will submit a pull request for a possible fix #1184 

#### Steps to reproduce
I think that this will only trigger an error if the order of publications causes them to be updated in the (in-)appropriate order. I can get it to trigger if I:
1. Add a second publication to an Genome annotation analysis and save that
[...]

PR #1184, opened 2021-03-22
Bugfix for Issue #1183 schema__publications field

dsenalik:
# Bug Fix

Issue #1183

## Description

Fix bug described in issue #1183.
Save the pkey from the linker by creating another array level so we can pass on both $pub and this key

## Testing?
Create an uneditable analysis as described in issue #1183 and see if this fixes it


spficklin:
This PR works as described and the code changes look appropriate.  Thanks for the fix!  

At first it didn't work for me because I didn't clear my record cache. This means that even if we merge this PR fix and someone tries to edit the field that they still get the error like I did.  I think the fix is to force a cache reload when the `edit` button is clicked.  But for now I think this PR is good!

Issue #1196, opened 2021-04-23
Publish generating notices like "Notice: Undefined index: #field_name in theme_tripal_field_default()"


dsenalik:
### System information

* Tripal Version: 7.x-3.5
* Drupal Version: 7.78
* PostgreSQL Version: 12.6 (Ubuntu 12.6-1.pgdg16.04+1)
* PHP Version: 7.3.27-9+ubuntu16.04.1+deb.sury.org+1

### Issue description
Warnings showing up in the php log such as:
```
Notice: Undefined index: #field_name in theme_tripal_field_default()
Notice: Undefined index: #language in theme_tripal_field_default()
Notice: Undefined index: #delta in theme_tripal_field_default()
```

#### Steps to reproduce
Go the publish page at admin/content/bio_data/publish and select just about any content type so that the AJAX callback can be called. No more actions needed.
Now check the log at admin/reports/dblog

#### What it might be
I think the problem is in the file 
```tripal_chado/includes/tripal_chado.publish.inc```
The array ```$elements``` is set to some initial values on lines 83 to 95
```
        $element = [
          '#entity_type' => $instance['entity_type'],
          '#entity' => NULL,
          '#bundle' => $instance['bundle'],
          '#field_name' => $field_name,
          '#language' => $langcode,
          '#field_parents' => $parents,
          '#columns' => array_keys($field['columns']),
          '#title' => $title,
          '#description' => $description,
          '#required' => FALSE,
          '#delta' => 0,
        ];
```
and then on line 111 these values are cleared by
```
       $element = [];
```
prior to calling a callback.

I think this was an erroneous initialization in the wrong place added by commit 
https://github.com/tripal/tripal/commit/a665f1118c37d8a3d3744305e5ae7d01b7a3768e#diff-c73a026770ad781e34d2868f9b5d78ce38460c5dd9207d89871a981ea2af4389

If I just comment out line 111 (line 105 of that commit) I stop getting notices.

I bet this did not generate warnings in earlier php versions and so wasn't caught.

Am I interpreting this correctly? If so, I have a pull request to fix this.

spficklin:
Fix applied!  Thanks @dsenalik

PR #1197, opened 2021-04-23
1196 tv3 tripal field notices

dsenalik:
# Bug Fix

Issue #1196

## Description
Deletes one misplaced initialization of the ```$element``` array that is not needed and in fact is harmful, ```$element``` is initialized earlier in the code

## Testing?
Testing is described in the linked issue


spficklin:
Thanks @dsenalik . I think you're assessment is correct! That was an unfortunate mistake. Thanks for  finding and fixing it.

PR #1198, opened 2021-04-24
PHP 7.0 fix for explode

spficklin:
# Bug Fix
Issue # 1195

## Description

This fixes a very simple explode call that is incompatible with PHP 7.0 and prevented web services form appearing.


spficklin:
This is an extremely simple fix.  We discussed this in PMC meeting this morning so I'm going to commit without a code review.

Issue #1199, opened 2021-04-29
obi__organism field cannot display name for infraspecific type_id


dsenalik:
### System information
* Tripal Version: 7.x-3.6
* Drupal Version: 7.78
* PostgreSQL Version: 12.6 (Ubuntu 12.6-1.pgdg16.04+1)
* PHP Version: 7.3.27-9+ubuntu16.04.1+deb.sury.org+1

### Issue description
The ```obi__organism``` field has a "Rewrite Value" setting on the edit page to allow customization
of the way the organism name is presented. With chado 1.3 infraspecific nomenclature was
introduced, but an option missing from this Rewrite Value is a was to display the infraspecific type, e.g. subspecies or varietas.

#### Steps to reproduce
You will need an organism with an infraspecific nomenclature entry, then see how it is presented on the organism page.

#### Error messages and screenshots
Example
![Screenshot_2021-04-29_before](https://user-images.githubusercontent.com/8419404/116581711-3a564000-a8da-11eb-932d-00b9413d4625.png)

How it should appear
![Screenshot_2021-04-29_after](https://user-images.githubusercontent.com/8419404/116581750-45a96b80-a8da-11eb-9cd8-47d9ae5274fb.png)

PR #1200, opened 2021-04-29
1199 tv3 obi  organism

dsenalik:
# Bug Fix

Issue #1199 

## Description
This is a method of resolving Issue #1199 to allow displaying infraspecifc type (e.g. subspecies) in the obi__organism field.
I'm not very pleased with the method, a special handler has been created for a new token, because the table schema only allows incorporation of organism.type_id when ```chado_get_tokens()``` is called, not the name that that type_id refers to. However there does not appear to be any simple way to display this information.
This custom token is added to the list of available tokens only if an infraspecific_name token is encountered.

## Testing?
You need chado 1.3 and an organism with infraspecific nomenclature, then just observe how it is displayed by the obi__organism field

Before
Screenshot_2021-04-29_before

After
Screenshot_2021-04-29_after

The Edit page, a new ```[infraspecific_type]``` has been added
Screenshot_2021-04-29_field


spficklin:
Hey @dsenalik. I'm sorry I missed this PR!   I was looking it over this morning and I noticed there is a serious problem with these tokens.  They are incorrect which led you to create this solution.  The token should look something like this `[organism.type_id>cvterm.name]` and that would give you the infraspecific type.   I think you can withdraw this PR so that we can fix this problem with the tokens.

spficklin:
Hey @dsenalik I went ahead and closed this PR, and I'll be submitting another PR to fix the bug.

PR #1201, opened 2021-05-03
Fix for issue #1165: null value error

btski:
Error result when providing empty array to chado_insert_record(). Additional context available in issue #1165.


spficklin:
Hi @btski any chance you can give me a template for the bulk loader I can test with... and maybe a data file?

btski:
@spficklin you should be able to reproduce the error with the attached files. Let me know if you have any problems.
[chado_bulk_loader_data.txt](https://github.com/tripal/tripal/files/6496567/chado_bulk_loader_data.txt)
[chado_bulk_loader_template.txt](https://github.com/tripal/tripal/files/6496568/chado_bulk_loader_template.txt)

PR #1203, opened 2021-05-10
Tokens Bug

spficklin:
# Bug Fix

Issue #1199 

## Description

The issue reported by @dsenalik in issue #1199 was bigger than just the `obi__organism` field.  There was a bug in the chado token strings.  Tokens for Chado base tables always looked like `[base_table.field_name]`, and these worked just fine. However, when foreign keys are used the tokens are inccorrect. In previous version of Tripal a token, say for the infraspecific name would look like `[organism.type_id>cvterm.name]`. Instead, the token was showing up simply as `[cvterm.name]`. 

This PR fixes this problem for all tokens derived from foreign key relationships in Chado. 

## Testing?

To test this:
1. Find a page in your site that uses the Chado `organism_id` field (or `obi__organism` field). If you need to, add a dummy infraspecific type and name for testing.
2. Navigate to `Structure` > `Tripal Content Types` and select the content type for the page you just looked up.  Then edit the settings for the Organism field.   
3. Open the list of tokens for the organism display. You'll see the foreign key relationships are not in the tokens where they should be.  If you want to, experiment and use the `[cvterm.name]` to specify the infraspecific type. Save and view the data page... no infraspecific type.
4. Pull this PR and return to the content type page and edit the organism field.  Now you should see the tokens with their FKs embedded in the token.   Add the tokens `[organism.type_id>cvterm.name]` and `[organism.infraspecific_name]` to the text box and save.
5. Return to the content page, click the 'Reload' button and you should now see the infraspecific type and name appear on the page.

### DIFF
--- a/tripal_chado/api/tripal_chado.api.inc
+++ b/tripal_chado/api/tripal_chado.api.inc
@@ -359,16 +359,18 @@ function chado_get_tokens($base_table) {
       foreach ($details['columns'] as $left_field => $right_field) {
 
         $sub_token_prefix = $base_table . '.' . $left_field;
-        $sub_location_prefix = implode(' > ', [$base_table, $left_field]);
-
         $sub_tokens = chado_get_tokens($table);
-        if (is_array($sub_tokens)) {
-          $tokens = array_merge($tokens, $sub_tokens);
+        foreach ($sub_tokens as $key => $sub_token) {
+          $key = preg_replace('/[\[\]]/','', $key);
+          $new_key = '[' . $sub_token_prefix .'>' . $key . ']';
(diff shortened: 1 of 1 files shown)


spficklin:
@dsenalik would you mind reviewing this PR as well since it replaces the one you submitted at #1200

dsenalik:
I have tested and it works perfectly.
I remember tracing through the code and those values weren't loaded, I didn't realize that it was a bug!
Thanks for fixing this.

dsenalik (approved):

spficklin:
Thanks @dsenalik !

Issue #1202, opened 2021-05-10
sio__references using wrong cvterm as label when linking to a tripal map


dsenalik:
### System information

* Tripal Version: 7.x-3.5
* Drupal Version: 7.78
* PostgreSQL Version: 12.6 (Ubuntu 12.6-1.pgdg16.04+1)
* PHP Version: 7.3.27-9+ubuntu16.04.1+deb.sury.org+1

### Issue description
This issue occurs when the tripal field ```sio__references``` displays a link to a linkage map (Tripal Map).
The label (```rdfs:type```) is the unit type used for the linkage map. This is almost always CM (centiMorgans) for a genetic map, but may be bp (base pairs) for a physical map.
The issue is that the field will only display one or the other of these, regardless of the actual units. In my case it always displayed **Bp(s)** instead of **CM(s)**
This occurs because the code in ```tripal/tripal_chado/includes/TripalFields/sio__references/sio__references.inc```
looks for a ```type_id``` value and uses this to set the ```type_name``` value. A map uses ```unittype_id``` instead so it doesn't find a value. In this case it defaults to the value from ```$ref_type->name```, which in my case was always "bp", which came from
```$ref_type = chado_get_cvterm(['cvterm_id' => $ref_mapping->cvterm_id]);```
which just gets the one that by chance appears first in the ```chado_cvterm_mapping``` table.

### Discussion
I have a fix for this, but is this how we want to label linkage maps, though? **CM(s)**

#### Steps to reproduce
You would need a map of each of two different unit types and see how they are each displayed in the ```sio__references``` field.

#### Error messages and screenshots
Before fixing this:
![before](https://user-images.githubusercontent.com/8419404/117672893-e2d49180-b16f-11eb-9157-7f7f31e20cb8.png)

And after:
![after](https://user-images.githubusercontent.com/8419404/117672915-e9fb9f80-b16f-11eb-9ccd-fc410956e8db.png)

dsenalik:
I have a pull request #1204 which bypasses the Bp vs. CM problem by calling these **Map(s)**

spficklin:
Fixed with PR.  Thanks @dsenalik

PR #1204, opened 2021-05-11
Issue #1202 updates to sio__references type determination

dsenalik:
# Bug Fix

#

Issue #1202 

## Description
I am submitting some changes to better handle a few entity types for the sio__references field, specifically for featuremap, analysis, and project.

This goes in the code just after 
```
          // We want to add a 'type' and 'name' element to the values (at a
          // minimum) for each of the records.  Unfortunately, every base table
          // is different and there may not be an easy to identify name,
          // so... we'll do the best we can.
```
and I am trying to do it a little better! Unfortunately this involves special cases for some content types. I also added a default so it won't be possible any more to have "**(s)**" labels for unhandled content types.

Also a couple of spelling corrections.

Before
before

After
after

## Testing?
You need a publication page with references to a featuremap, analysis, or project to see these changes.


spficklin:
Hi @dsenalik. This looks like a good fix.  The only  weirdness is the word "Analysis(s)".   Which, I feel odd complaining about considering before it was just "(s)" on some items. I think maybe we should just take the "(s)" out and leave the singular form? What do you think?

dsenalik:
That would be my preference also!

dsenalik:
I added a commit to remove the "(s)" and appearance is now:
No-S

spficklin:
Thanks for the fix! it's merged.

Issue #1205, opened 2021-05-15
Bulk loader editing bug when deleting first record


dsenalik:
### Issue description
It is not possible to delete the first record when editing a bulk loader template, if attempted the following error appears
![CBL-delete-first-item](https://user-images.githubusercontent.com/8419404/118364038-75a07200-b55c-11eb-8dbf-3782b142e961.png)

#### Steps to reproduce
Select Delete on the first record, e.g.
![CBL-edit-template-box](https://user-images.githubusercontent.com/8419404/118363991-54d81c80-b55c-11eb-9b5f-10b8cc8f565a.png)

laceysanderson:
To my understanding, this issue was solved by the merged PR. If not please comment and I will re-open :-)

PR #1206, opened 2021-05-15
fix for issue #1205 bug, can't delete first record in template

dsenalik:
# Bug Fix
Issue #1205 

## Description
It is not possible to delete the first record when editing a bulk loader template. This is because the first record has an index of zero, and we test for "false" instead of isset in the code.
This simple fix corrects that

## Testing?
Try to delete the first the first record of any bulk loader template. You don't have to actually delete, a confirmation prompt will appear first.
CBL-edit-template-box

Before the fix the error is
CBL-delete-first-item

After the fix the expected result is
CBL-after-fix


dsenalik:
There is also an old commit to revert the insertion of two blank lines that I accidentally added in commit 0dd757b8ff37543e495131257e099f79154ff029
that snuck in. Might as well fix that too.

spficklin:
Thanks @dsenalik for the fix!

Issue #1211, opened 2021-06-12
Image field  is missing in views for organisms


spficklin:
### System information

* Tripal Version: 3.6
* Drupal Version: 7.x
* PostgreSQL Version: N/A
* PHP Version: 7.4

### Issue description

The Organism Image field is not showing up in Views, so it's not possible to create an organism summary view with the images.

spficklin:
Fixed as per PR.

PR #1212, opened 2021-06-14
Organism Image Doesn't appear in Views.

spficklin:
# Bug Fix

Issue #1211

## Description

Fields that don't store data in Chado are being excluded from the Drupal Views.  For example, if you wanted to make a listing of organisms on a site and include the organism image, you couldn't add the image as a field.  This is because there is an if statement check that excludes non Tripal Storage API fields from being included (see the code changes).   I've removed that if statement check.  

## Testing?

- Before pulling this fix create an Organism view and try to add the organism image to the view. It doesn't show up in the list.
- Pull this PR, clear the views cache, and now try to create the same view. The organism image shows up.

### DIFF
--- a/tripal/tripal.views.inc
+++ b/tripal/tripal.views.inc
@@ -30,7 +30,7 @@ function tripal_views_data() {
   $data = [];
   // Job Management System.
   tripal_views_data_jobs($data);
-  // Add all TripalEntity bundles. 
+  // Add all TripalEntity bundles.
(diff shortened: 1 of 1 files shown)


Ferrisx4 (approved):
Tested on a local dev version, seems to work as advertised.

spficklin:
Thanks @Ferrisx4 for testing.  I'll merge this in.

Issue #1216, opened 2021-07-02
Unable to edit certain cvterms with alternative dbxref accession naming scheme


dsenalik:
### System information

* Tripal Version: 7.x-3.5
* Drupal Version: 7.78
* PostgreSQL Version: 13.3 (Ubuntu 13.3-1.pgdg16.04+1)
* PHP Version: 7.3.28

### Issue description
I have some cv terms that I can't edit. Whe I try, I get the error
```Could not add term. Check Drupal recent logs for error messages.```
and there is nothing in the log.

I have tracked down why, and it is because the accession for the dbxref is constructed as
cvterm_id + colon + name
for example
![Screenshot_2021-07-02_12-42-41](https://user-images.githubusercontent.com/8419404/124311481-72c10700-db33-11eb-8b77-80bff8158b73.png)

There is code in tripal_chado.cv.api.inc that looks for an accession formatted as
db_name + colon + accession
for example ```TO:0000277```
and in this case separates into the two parts. This fails in my case because I have a numeric identifier before the colon instead of a db name. Why I have that is of course a good question.

A fix I propose for discussion is to modify the regex to ignore cases where the string before the colon is all digits.
```if (preg_match('/^.+?:.*$/', $id)) {```
becomes
```if (preg_match('/^[^:]*\D[^:]*:.*$/', $id)) {```

This should be safe because no database should have a name that is purely numeric.

**Comment by dsenalik** [comment]
This may not be the answer, I need to do more testing, I can't edit the TO:0000277 example either.
And a few dbxref accessions are URLs, those will also fail with the current code
```
 dbxref_id | db_id |                              accession                              | version | ...
-----------+-------+---------------------------------------------------------------------+---------+-
      2794 |    17 | https\://en.wikipedia.org/wiki/Allele_frequency                     |         | 
      2895 |    19 | https\://doi.org/10.1101/584664                                     |         | 
```
*Edit: But these are not linked to cv terms, so aren't relevant here

dsenalik:
While this issue exists, I am closing because an accession should not normally have a colon in it, so it is better to fix non-compliant accessions. If I find what caused it I will post an issue there.

dsenalik:
I found at least one source of these problematic accessions in the dbxref table, mainlab chado loader.
See Issue https://gitlab.com/mainlabwsu/mcl/-/issues/13 for details.

PR #1217, opened 2021-07-02
Issue #1216 Unable to edit certain cvterms

dsenalik:
# Bug Fix
Issue #1216 

## Description
This is a proposed fix for Issue #1216
It may be a significant change in that it may have unintended side effects, so would need proper testing. I can test with my own site, but I am unfamiliar with what other things may call this function.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.cv.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.cv.api.inc
@@ -836,9 +836,15 @@ function chado_insert_cvterm($term, $options = []) {
   if ($dbname) {
     $accession = $id;
   }
-  if (preg_match('/^.+?:.*$/', $id)) {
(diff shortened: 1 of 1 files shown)


dsenalik:
Yeah, this breaks the OBO importer test.

dsenalik:
I am updating this pull request to a version which will probably not break anything, but I don't think it is acceptable, and I'll probably close this.

This new version checks that the part before the colon is a valid db name. This imposes a chado query every time so is not ideal, though probably significant only when loading an entire ontology. And I still get an error when trying to edit a cvterm with accession ```TO:0000277``` because TO is a valid db name. Probably the real problem is I shouldn't have these types of accessions, I may just think about fixing them all.

dsenalik:
While this issue exists, I am closing because an accession should not normally have a colon in it, so it is better to fix non-compliant accessions. If I find what caused it I will post an issue there.

Issue #1221, opened 2021-07-26
Legacy content not accessible


spficklin:
### System information

* Tripal Version:  3.x
* Drupal Version:   7.x
* PostgreSQL Version:  N/A
* PHP Version:  N/A

### Issue description

If a user turns on the legacy support for Tripal v2 content types, anonymous users will always get an 'Access Denied' error when viewing any of the legacy pages.   This was reported in Slack by @cornellyujy

#### Steps to reproduce
Turn on the legacy support for the organism content type.  Create a new legacy organism page.  When not logged on go to the organism page just created. You will get an 'access denied' message.

spficklin:
This has been fixed as per the recently merged PR.

PR #1222, opened 2021-07-26
Fixes bug in access to legacy pages.

spficklin:
# Bug Fix

Issue #1221 

## Description

A bug exists that prevents any legacy content from being seen by an anonymous user.    The problem is that each legacy content type has it's own `hook_node_access()` function that checks the permissions for the user and for viewing it checks for a content type specific permission.  However, when we switched the legacy content types over to Tripal v3 we removed the older permission to allow Drupal to create them and Drupal doesn't use a content specific permission for node-based content types (which the legacy types are nodes).  This fix reverts the permissions for viewing content types to just `access content` as it should be. 

## Testing?

1. Make sure you have the "view published content" permission turned on.
2. Prior to pull this fix, enable the organism legacy module.  
2. Create a legacy organism page: Content > Add Content > Organisim
3. Log out and try to view the new organism page. You'll get "access denied".
4. Pull the code for this pull request
5. Now go to the organism page while still logged out and you should see the page.

### DIFF
--- a/legacy/tripal_analysis/includes/tripal_analysis.chado_node.inc
+++ b/legacy/tripal_analysis/includes/tripal_analysis.chado_node.inc
@@ -648,7 +648,7 @@ function tripal_analysis_node_access($node, $op, $account) {
       }
     }
     if ($op == 'view') {
-      if (!user_access('access chado_analysis content', $account)) {
+      if (!user_access('access content', $account)) {
         return NODE_ACCESS_DENY;
       }
     }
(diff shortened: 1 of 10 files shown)


spficklin:
@cornellyujy since you reported this bug can you test this pull request to make sure it fixes the problem?  

The fix is simple so only one review is needed.

cornellyujy:
Cool!!!
Thanks @spficklin , I have revised my codes among these files in legacy, and they all work well.

spficklin:
Thanks for testing @cornellyujy !

Issue #1214, opened 2021-06-24
Show Synonyms in mRNA Content Type


mario-lopez-c:
Hello! Do you know to show synonyms in the mRNA Content-Type? I mean, I have loaded into the synonym table some data and related it to the feature table through feature_synonym table, but I don't know how to publish those synonyms within the mRNA content type, I have only been able to publish the information that is within the featureprop table.
Thank you!

spficklin:
Hi @mario-lopez-c apologies for the slow reply.  So, after investigating this issue it seems we have a very simple bug in the code that prevented synonyms from appearing.   I will make a quick fix and it should resolve the problem for you.  If you could test the pull request for me to make sure its fixed on your site that would be great and should speed up getting this fix into the next release.  I'll tag you in the pull request.

spficklin:
Okay, the PR is submitted with the rfix.  @mario-lopez-c can you test it on your site to make sure you can now get synonyms?   Here's the link to the PR:  #1223

spficklin:
I believe this issue has been fixed in the recent PR.  I'm going to close out the issue. Please comment if there's still a problem.

mario-lopez-c:
Thank you so much! That worked! Thank you!

PR #1223, opened 2021-07-26
Missing Synonyms

spficklin:
#Bug Fix

Issue #1214

## Description

As reported by @mario-lopez-c the synonyms are not appearing on "feature" based pages, or any content type that supports synonyms. The reason is a return statement that must have been accidentally left in during previous testing. Removing the `return` fixes the issue and the synonym field is available.

## Testing?

1.  Go to (or create) any feature page (e.g. gene, mRNA,  etc.).  
2. You'll notice there is no field for synonyms or aliases for the content type.
3. Go to Structure > Tripal Content Types > mRNA (or whatever you are using) and click the 'Manage fields' link.   Notice there is no Synonyms field.
4. Click the 'Check for new fields` link at the top and it will not find a synonym field.
5. Pull the code from this PR.
6. Click the 'Check for new fields` link at the top and it will now find a synonym field!
7. Click the 'Manage Display' at the top and move the new Synonym field where you want it to appear on the page and save.
8. Return to the original page and you should now see a synonym field. You can edit the page and change/add a synonym.

### DIFF
--- a/tripal_chado/includes/tripal_chado.fields.inc
+++ b/tripal_chado/includes/tripal_chado.fields.inc
@@ -2760,7 +2760,7 @@ function tripal_chado_bundle_instances_info_linker(&$info, $entity_type, $bundle
       ],
     ];
   }
-  return;
+
   // SYNONYMS
   $syn_table = $table_name . '_synonym';
   if (chado_table_exists($syn_table)) {


spficklin:
This is such a simple fix I'm going to merge it without review.

mario-lopez-c:
Great! I'm sorry, I was absent from work for several days. Thank you very much for your help. I will test it! Thank you very much again!

Issue #1220, opened 2021-07-24
A security bug?


chunhuaicheng:
### System information

* Tripal Version: 3.4, 3.5, 3.6, and 7.x-3.x
* Drupal Version: 7.82
* PostgreSQL Version: 13.3
* PHP Version: 7.4.3

### Issue description

After editing a Tripal page (such as an Organism), the logged-in user gains privileges of the user who previously created the page.

#### Steps to reproduce
1. Log in as a user who is authorized to edit a Tripal page, such as a curator.
2. Edit a Tripal page that is created by a different user, such as an admin.
3. Click on 'Update' to save the page
4. Now you're logged in as the user who created the Tripal page. (the admin)

#### Error messages and screenshots

Step1: Log in as curator
![step1](https://user-images.githubusercontent.com/14333886/126852856-cdd75fa5-d5c9-4549-8f34-5223426d4730.png)

Step2: Edit a Tripal page
![step2](https://user-images.githubusercontent.com/14333886/126852858-80b0d60b-7894-4571-be10-8d770e7209db.png)

Step3: Save
![step3](https://user-images.githubusercontent.com/14333886/126852859-bf09868f-5859-4db4-a7d5-c728255787a7.png)

Step4: Now you are the admin
![step4](https://user-images.githubusercontent.com/14333886/126852861-226ad496-246e-4661-8982-85999f47dff6.png)

jcrabbwsu:
To add to the behavior described above:
When the authenticated user is also the author of the page, the steps described above result in the user being logged out.

spficklin:
I've added a PR (#1224) to fix this if anyone watching this thread wants to test.  Once this gets tested and merged in we'll make a new Tripal release to get this fix out.

risharde:
@spficklin Just tested your commit and it resolved the security issue on my side.

spficklin:
Issue resolved in dev branch with recent PR.

PR #1224, opened 2021-08-02
Security Fix for Editing Tripal Content

spficklin:
# Security / Bug Fix

Issue #1220

## Description

Please see the nice description on issue #1220 for how this security issue can be replicated.

Also, it seems that this switching users caused some warning messages to appear on the page that were not displayed before.  I'm not sure why the user switching hid those but this PR also adds a few checks to make sure those warnings don't appear.  They were simple warnings about arrays not having the keys. You'll see them in the code changes.  

## Testing?

1.  Pull this fix and retry the steps described in issue #1220. The user switching problem should be fixed.


spficklin:
@chunhuaicheng can you please test this PR to see if it resolves the issue?

spficklin:
This was tested by @risharde and he commented so on the issue rather than here. So, I'm going to merge. Thanks @risharde

Issue #1218, opened 2021-07-14
rapid_install failing to run becuase of the table db2cv_mview not existing


AlexsLaboratory:
### **System Info**

- OS: Ubuntu 21.04
- Tripal Version: 3.6
- Drupal 7.81
- PostgreSQL Version: 13
- PHP Version: 7.3

### **Steps to Reproduce**

1.  Install Drupal by downloading the [`drupal-7.81.tar.gz`](https://ftp.drupal.org/files/projects/drupal-7.81.tar.gz) file and moving it into the apache2 web root directory (`/var/www/html`)
2. Install `Drush` by following this [documentation](https://docs.drush.org/en/7.x/install/).
    - Make sure to install the `6.x.x` version as they deprecated a feature used to install `Tripal` in later versions
3. Install `Tripal` using the rapid install [documentation](https://tripal.readthedocs.io/en/latest/user_guide/install_tripal/rapid_install.html)

### **Issue Description**
When the install script runs the `db2cv_mview` table does not get created for some reason so it then tries to create a relation to it and fails to do so because it does not exist.

### **DB Error Message (short)**

```
PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "db2cv_mview" does not exist LINE 1: ...') as name FROM
chado.db DB INNER JOIN db2cv_mvie... ^: SELECT DB.name as short_name, DB.description, DB.url, DB.urlprefix,
SUM(DBCVM.num_terms) as num_terms, array_to_string(array_agg(DBCVM.cvname), ', ') as name FROM chado.db DB INNER 
JOIN {db2cv_mview} DBCVM ON DBCVM.db_id = DB.db_id WHERE DB.name = :name GROUP BY DB.name, DB.description, 
DB.url, DB.urlprefix ; Array ( [:name] => hydra ) in chado_query() (line 1790 of 
/var/www/html/drupal/sites/all/modules/tripal/tripal_chado/api/tripal_chado.query.api.inc).
```

### **Install Error Message (detailed)**

```
WD php: PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation   [error]
"db2cv_mview" does not exist
LINE 1: ...') as name      FROM chado.db DB       INNER JOIN db2cv_mvie...
                                                             ^:       SELECT
DB.name as short_name, DB.description, DB.url, DB.urlprefix,       
SUM(DBCVM.num_terms) as num_terms,       
[...]

PR #1225, opened 2021-08-05
#1218 add the mview  db2cv_mview on chado install

bradfordcondon:
# Bug Fix

Issue #1218

## Description
Site is missing an mview not defined in the installed schema.  was added via an update function but not install.

## Testing?
see issue for rapid install testing.

## Additional Notes (if any):
I'm pretty rusty, i think that chado install mview checks if the table exists first?  if it doesnt we'll need to add that.

### DIFF
--- a/tripal_chado/tripal_chado.install
+++ b/tripal_chado/tripal_chado.install
@@ -33,6 +33,8 @@ function tripal_chado_install() {
     // Fix the SOFP feature_property issue from the legacy feature_property.
     tripal_chado_fix_legacy_SOFP_7338();
 
+    module_load_include('inc', 'tripal_chado', 'includes/setup/tripal_chado.chado_vx_x');
+    tripal_chado_add_db2cv_mview_mview();
   }
 
   tripal_insert_variable('bundle_category', 'Bundles can be categorized to allow for grouping');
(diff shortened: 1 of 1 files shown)


Ferrisx4 (changes requested):
As suggested by @spficklin, moving these two new lines (36 and 37) above lines 33 and 34 ensures that the db2cv_mview is created before the `tripal_chado_fix_legacy_SOFP_7338()` function is run, which requires db2cv_mview to already exist.
I tested this slight change with the Rapid Installer script on a completely fresh database and it ran successfully.

spficklin:
Actually, I think the correct solution would be to not include these lines at all:

```
    module_load_include('inc', 'tripal_chado', 'includes/setup/tripal_chado.chado_vx_x');
    tripal_chado_add_db2cv_mview_mview();
```

But rather I think these lines:

https://github.com/tripal/tripal/blob/fe1f70dc607eb8108814e675f4d11abe453cdb6a/tripal_chado/tripal_chado.install#L735-L744

Should be wrapped in an if statement so that if the table doesn't exist the update to the mview won't be run.
```
  if (chado_table_exists('db2cv_mview')) {
  ...
  }
```

If the table doesn't exist it will be created in the Chado prepare function which creates it and that's where the table should be created rather than to fix this bug in the install.

PR #1233, opened 2021-10-31
20211031 pub importer

dsenalik:
# New Feature

## Description
This is a simple change to add a summary line to the publication loader. For example, formerly, the last line of output was

```INFO (PUB_IMPORT): Done.```

After this change, it will now look like

```INFO (PUB_IMPORT): Done. Publication importer summary: error=0 inserted=123 skipped=456 updated=0```

This summarizes elements stored in the ```$results``` array, which is otherwise only used if an email report is generated.


spficklin (approved):

spficklin:
I like this addition! Thanks for submitting.

Issue #1234, opened 2021-11-04
Does chado_associate_semweb_term() not work?


dsenalik:
### Issue description
I was trying to figure out why my call to chado_associate_semweb_term() did not work.

This function is defined in ```tripal/tripal_chado/api/tripal_chado.semweb.api.inc```
After some error checking, I find that there is this section:

```
    // Insert the record.
    $id = db_insert('chado_semweb')
      ->fields([
        'chado_table' => $chado_table,
        'chado_column' => $chado_column,
        'cvterm_id' => $term->cvterm_id,
      ]);
    if ($id) {
      return TRUE;
    }
    else {
      tripal_set_message('Failure associating term.', TRIPAL_ERROR);
      return FALSE;
    }
 ```

Please correct me if I'm not understanding this, but ```db_insert``` doesn't actually "insert" unless you include an ```->execute()``` at the end. Anyway, if I add that, things start working as I expect. I'll come up with a pull request #1235  to illustrate the change.

#### Steps to reproduce

Updating an existing entry already in the chado_semweb table works fine. The bug only strikes if it is a new table+column combination for the chado_semweb table.

I am testing this using /devel/php to execute the following (terms for the test are specific to our site)
```
    $sql = 'SELECT cvterm_id FROM {cvterm} WHERE name=:name AND' 
         . ' cv_id=(SELECT cv_id FROM {cv} WHERE name=:cvname)';
    $args = [ ':name' => 'name', ':cvname' => 'SIO' ];
    $results = chado_query($sql, $args);
    $obj = $results->fetchObject();
    $cvterm_id = (int)$obj->cvterm_id;
var_dump($cvterm_id);
    $semweb_term = (object) ['cvterm_id' => $cvterm_id];
    $result = chado_associate_semweb_term('fileloc', 'display_name', $semweb_term, true);
var_dump($result);
  $umm_what_gives = chado_get_semweb_term('fileloc', 'display_name', []);
var_dump($umm_what_gives);
```
Before patch from pull request
```
int(64753)
bool(true)
NULL
```

and nothing is added to the public.chado_semweb table

After patch from pull request #1235 
```
int(64753)
bool(true)
string(10) "SIO:000116"
```

and we get something added to the public.chado_semweb table.

And, the test for success now works, before it returned the db object which always evaluated to true.

spficklin:
Fixed with recent PR #1235

PR #1235, opened 2021-11-04
1234 tv3 chado associate semweb term

dsenalik:
# Bug Fix

Issue # 1234

## Description

Add missing ->execute() to db_insert

Description in linked issue #1234


dsenalik:
Something is wrong with the testing system, it fails when it doesn't find a cv table early on, before testing this code?

spficklin (approved):

spficklin:
You're right.  How odd we missed this.  I'm not sure how that function was working at all given that on a new install all of those mappings have to be added using that function.  But you're right, that db_insert() call needs an execute.

PR #1236, opened 2021-11-18
20211118 Corrections of several typos

dsenalik:
# Bug Fix

## Description
This is a collection of spelling errors or in one case a spurious tab character that I have noted. There should be no functional changes, except again for that spurious tab character possibly may affect the 'protocol' CV term in the 7338 updater.


spficklin (approved):

spficklin:
Thanks for these fixes!

Issue #1238, opened 2021-11-20
Pubmed publication importer invalid year breaks search


dsenalik:
### System information

* Tripal Version: 7.x-3
* Drupal Version: 7.82
* PostgreSQL Version: 13.2 and 13.3 (note that bug does not appear with 12.x)
* PHP Version: 7.3.28

### Issue description
The original issue was that a publication search with a year range specified would generate an error
```ERROR:  invalid input syntax for type numeric: " "```
This came from the SQL which I am including just for historical reasons
```
SELECT count(*) FROM (SELECT DISTINCT P.*  FROM chado.pub P
    INNER JOIN chado.cvterm CVT on CVT.cvterm_id = P.type_id
    LEFT JOIN chado.pubprop PP1 ON PP1.pub_id = P.pub_id AND PP1.type_id = '133'
    WHERE (NOT P.title = 'null')
    AND (lower(PP1.value)::text LIKE lower('%Senalik%'))
    AND (P.pyear ~ '....'
    AND to_number(P.pyear,'9999') >= '2018'
    AND to_number(P.pyear,'9999') <= '2018')
    ORDER BY P.pyear DESC, P.title ASC) as t1;
```

The cause of the error, however, is that a single publication in the database has, for the chado.pub pyear column
```Fall 2016```
and that triggers the non-helpful error message.

So the real issue to fix is in the pubmed publication importer.
The relevant PMID for this publication for testing is ```27860473```
The offending XML is
```<PubDate><MedlineDate>Fall 2016</MedlineDate></PubDate>```

The way the code is written currently, this date replaced the valid date already parsed, and "Fall " is included by the regex. The pull request fixes both of these issues.

PR #1239, opened 2021-11-20
For issue #1238 tv3 pubmed importer fix

dsenalik:
# Bug Fix

Issue #1238

## Description
Issue describes the problem. The fix here changes two things:
1. The medline date is not used to generate the chado.pub pyear value if a year was already obtained
2. The regex now extracts a four-digit value and all other text is removed, previously text on the left side was retained e.g. "Fall 2016"

## Testing?
PubMed id for the publication that triggers this is ```27860473```


spficklin (approved):

spficklin:
There's always a surprise from time to time in these data. Thanks for the fix.

Issue #1240, opened 2021-11-23
Tripal field sbo__relationship "is is" output for new cv term 


dsenalik:
### System information

* Tripal Version: 7.x-3.7
* Drupal Version: 7.82
* PostgreSQL Version: 13.4
* PHP Version: 7.2.24-0

### Issue description
For crazy reasons, I decided to create a CV term to link male sterile lines to their corresponding maintainer line. I called it ```is_a_maintainer_line_for```, and create appropriate relationship links. Works fine.

The issue is, however, that the ```sbo__relationship``` field displays

![Screenshot_2021-11-23_16-22-13](https://user-images.githubusercontent.com/8419404/143138609-174b3f46-c057-45f0-8272-595ff10e5024.png)

note that extra "is". This happens in ```sbo__relationship.inc``` in ```public static function get_rel_verb($rel_type)```

where some terms are hard-coded in as already valid verbs and don't get anything added, for example ```is_a_maternal_parent_of```

I propose a change to make any term starting with "is " or "has " be declared a valid verb as it is, this can then handle my new term and any similar terms. This will allow removal of eight lines of these special cases.

spficklin:
Fixed with recent PR.  Thanks @dsenalik

PR #1241, opened 2021-11-23
Issue #1240 sbo__relationship verb parsing

dsenalik:
# Bug Fix

Issue #1240 

## Description
This change moves the individual checks for various "is ..." "is a ..." and "has ..." cv terms to a single regex.

This will allow other "is" or "has" terms to display correctly in the sbo__relationship field.

Screenshot_2021-11-23_16-17-08

I also alphabetized the other terms "for tidiness"


spficklin (approved):

spficklin:
I think this is an appropriate fix for these types of terms.  Thanks for the fix.

Issue #1246, opened 2022-01-04
Alt text for AJAX spinner


Ferrisx4:
### System information

* Tripal Version: 3.x

### Issue description

AJAX spinner on certain pages is an animated gif. Per accessibility standards, all images are expected to have an alt text specified, even in cases where the image is decorative.
In this case, the image is to indicate loading of a paginated element.

We were initially inclined to use an empty alt text, `alt=""`, typically used for decorative images. From an accessibility standpoint, screenreaders may not find a value of 'loading' useful.
However, alt text is also useful in the case of missing images.

With this in mind, we have two options:
```html
alt=""
```
or
```html
alt="Loading..."
```

#### Error messages and screenshots
Screenshot of the WAVE tool scanning a Genome Assembly page

![](https://user-images.githubusercontent.com/5487172/131407869-f95a73e0-3ff6-44a4-9ada-fa650c28eb88.png)

PR #1247, opened 2022-01-06
Add alt text to the AJAX spinner

Ferrisx4:
# New Feature

Issue #1246

## Description

Add an alt text to a loading image "Loading content" for accessibility and if the image fails to load.

## Testing?

Use an accessibility tool such as WAVE or view the source code for a page such as an Analysis that has Cross References.


laceysanderson (approved):
Beautiful :-) Thanks!

Issue #1237, opened 2021-11-20
Automated testing is failing during "Install Tripal" phase


dsenalik:
## BUG/ERROR report
Automated testing is failing for everything during the "Install Tripal" phase with the following error
```
#15 /home/runner/.composer/vendor/drush/drush/drush.php(12): drush_main()
#16 {main}
Error: cution failed: PDOException: SQLSTATE[42P01]: Undefined table:    [error]
7 ERROR:  relation "cv" does not exist
LINE 1: SELECT * FROM cv CV
                      ^ in
/home/runner/work/drupal/includes/database/database.inc:2278
Stack trace:
#0 /home/runner/work/drupal/includes/database/database.inc(2278):
PDOStatement->execute(Array)
```

Reference: pull requests #1233, #1234, #1236

laceysanderson:
Looks like the error you mentioned is a secondary error and the original error is in the chado install phase [here](https://github.com/tripal/tripal/runs/4563106593?check_suite_focus=true#step:7:82).

> 2021-12-17 16:56:30: Calling: tripal_chado_install_chado(Install Chado v1.3)
> Creating 'chado' schema
> [site http://default] [TRIPAL ERROR] [TRIPAL_CHADO] FAILED. Line  2252, 0SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: box @ boxLINE 2: ... 'SELECT * from featureloc where boxquery($1, $2) @ boxrange...                                                             ^HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.:CREATE OR REPLACE FUNCTION featureloc_slice(bigint, bigint) RETURNS setof featureloc AS  'SELECT * from featureloc where boxquery($1, $2) @ boxrange(fmin,fmax)'LANGUAGE 'sql';
> Loading sites/all/modules/tripal/tripal_chado/chado_schema/default_schema-1.3.sql...

laceysanderson:
Wow that took a bit of digging but I finally found the issue:
<img width="661" alt="Screen Shot 2022-01-13 at 1 42 12 PM" src="https://user-images.githubusercontent.com/1566301/149401576-1989827c-6085-44e2-87b0-4880ba1d5c1d.png">

[...]

PR #1249, opened 2022-01-13
Fix automated test chado install errors

laceysanderson:
# Bug Fix

Issue #1237 

## Description
Recently our automated testing switched to PostgreSQL version 14. This brought out a bug that had been lurking for a very long time 🤪 Apparently Chado feature location comparison functions have been using old notation (`@` to indicate contained in or on for geometric set operations) that was [deprecated in version 8.2](https://www.postgresql.org/docs/8.2/functions-geometry.html) and just finally removed in version 14. As such, this PR updates to the new contained in operator, `<@` and thus automated testing passes again 😄 

## Testing?
The tests passing on this PR should be sufficient but you can also confirm that chado installs locally on your own version of PostgreSQL. (Note: the new operator has been available since 2006 and is in all supported versions of PostgreSQL).

## Additional Notes

I also added my SQL patch for the create_point errors to make it easier for us to help people in the future. It includes instructions for applying manually and is not currently referenced anywhere 🤷


dsenalik (commented):
"intented" should be "intended".
Otherwise awesome work, thanks for figuring this out!

guignonv:
I have a concern about featureloc_createpoint.patch.sql: I think all the functions reside in the chado schema and therefore, the search_path should be set to chado instead of public.
`SET search_path = chado;`

Then, this:
```
DROP INDEX IF EXISTS chado.binloc_boxrange;
CREATE INDEX binloc_boxrange ON chado.featureloc USING GIST (boxrange(fmin, fmax));
```
should not be schema-qualified (ie. no `chado.`). I think this is important to make it more easy to use the script with Chado instances in other schemas not using the "chado" name.

guignonv (changes requested):
- use `SET search_path = chado;`
- remove `chado.` prefix on lines 33 and 34.

laceysanderson:
Thank you both for the comments!

I think the featureloc patch needs more discussion based on @guignonv's comments so I have removed it, for now, to allow this fix to be merged sooner.

The main fix was the small change to the default schema SQL taking into PostgreSQL version changes.

Unless there are any concerns with the default schema SQL changes, I'm going to merge this come Monday assuming the tests pass which they should.

Issue #1252, opened 2022-02-24
Improve efficiency of chado_get_organism_select_options()


dsenalik:
### System information
* Tripal Version: 3.x

### Issue description
Our site CarrotOmics has 1,342 organisms in the chado.organism table. As a result, when editing an analysis page, or any other page using the ```obi__organism_linker``` field, it takes a significant time to get into edit mode. A large fraction of the time is used by ``chado_get_organism_select_options()``, and it is called twice by the widget. On my test system, each organism consumes about 15 ms, which adds up fast with that many organisms.

I am proposing a relatively simple fix which will be described in the linked pull request #1254

PR #1254, opened 2022-02-24
1252 tv3 get organism select

dsenalik:
# Bug Fix

Issue #1252

## Description

Modification to the ```function chado_get_organism_select_options()``` to increase efficiency. There should be no functional change.

The original code on my test system was taking about 15 ms per organism. Much of this is occupied by the call to ```chado_get_organism_scientific_name()``` which in turn is occupied in looking up the infraspecific rank with a call to ```chado_get_cvterm(['cvterm_id' => $organism->type_id])```. This is a slightly redundant call, because this information had already been looked up in the original sql. The changes here just save that information and pass it on. This reduces the time per organism from about 15 ms to about 6ms.

I did not modify the portion of the code where ```$syncd_only``` is true, intended for Tripal v2 sites, because I have no way to test that.

## Testing?
If you have as many organisms in the chado.organism table as CarrotOmics, you can test just by observing how long it takes to enter edit mode on any analysis page, assuming the ```obi__organism_linker``` is present.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.organism.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.organism.api.inc
@@ -156,6 +156,10 @@ function chado_get_organism_scientific_name($organism) {
         $rank = $organism->type_id->name;
       }
     }
+    // For db query where we explicitly supply infraspecific_type
+    elseif (property_exists($organism, 'infraspecific_type')) {
+      $rank = $organism->infraspecific_type;
+    }
     else {
       $rank_term = chado_get_cvterm(['cvterm_id' => $organism->type_id]);
       if ($rank_term) {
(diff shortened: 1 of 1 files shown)


spficklin:
Thanks @dsenalik do you mean testing on a feature page as the analysis page would not have an organism field by default.

dsenalik:
I did mean analysis, but any type of content with the ```obi__organism_linker``` field should work equally well, like genetic_marker.

spficklin:
Oh, I see. That field comes from the https://github.com/statonlab/tripal_manage_analyses module.  I'll install it and test it.

spficklin (approved):

spficklin:
Thanks @dsenalik I couldn't test performance because I didn't have a huge number of orgnanisms but functional tests showed no problems otherwise.

Issue #1138, opened 2020-12-13
Tripal field sbo__database_cross_reference timeout, can't handle many cross references


dsenalik:
## BUG/ERROR report
I thought it would be nice to link all NCBI sequence accessions to corresponding publications, using the ```pub_dbxref``` table. Well it works great, except for publications with many NCBI sequence accessions, the page can actually timeout, and even if not, the list is too long.

First, we would need a pager for the sbo__database_cross_reference field, but

Second, it takes way too long to render the page, both the call to ```chado_expand_var``` in ```load()```
as well as the formatter takes too much time also.

I could try to work on this field, but maybe there is a better way to do this? Maybe even a new tripal field specific for links to GenBank? Mainly I am wondering if anyone has any advice.

spficklin:
Hi @dsenalik, I think the solution to this problem is the same with a few other fields you're recently reported on (#1100). The solution was to not use the `chado_expland_var` function but to rewrite the `load` function to use strategic SQL.  I can try to take a look at it but it might take me a few days before I can get to it.

dsenalik:
I'm working on it. I have the SQL working, just need to add the pager.

dsenalik:
My pager appears to be working, rendering >3000 cross references is only 75 ms SQL query and store, 105 ms in view() YAY!

But I have a question, if there is only one cross reference (the more usual case) it now takes more screen space. What about only triggering the pager if there is > one page of cross references?
Before:
![before](https://user-images.githubusercontent.com/8419404/102377722-9a32f680-3f8a-11eb-8100-bfc2fb4c9788.png)

After:
![after](https://user-images.githubusercontent.com/8419404/102377741-9f904100-3f8a-11eb-9999-670673172f95.png)

laceysanderson:
Hi @dsenalik,

I personally like the consistency of the bullets so I would be hesitant to change the current behavior. Anyway, thank you for the PR! I have merged it which solves the issue reported so I'm going to close this issue. However, if you feel strongly regarding the additional white space feel free to submit another PR :-) 

Thank you! (Both for the bug report and the fix!)

dsenalik:
No, I am quite happy with the appearance, I was just wondering what others thought.
Thank you

PR #1139, opened 2020-12-17
Tripal field sbo__database_cross_reference timeout fix

dsenalik:
# Bug Fix

#

Issue #1138 

## Description
I thought it would be useful to link GenBank accessions with their corresponding publications, using the pub_dbxref table. Unfortunately this caused high page rendering times if there were many accessions, and if there are many accessions, a pager was needed.

A call to ```chado_expand_var()``` has been replaced by SQL to improve performance, and a pager has been added.

Changes also to ```tripal_chado/api/modules/tripal_chado.pub.api.inc``` for similar reasons of long page rendering time.

I am not full confident I did things the proper way, though, and would like comments.

Example of the field after these changes:
pullimage

PR #1257, opened 2022-03-02
Issue #1138 repair damage to sbo__database_cross_reference widget

dsenalik:
# Bug Fix

Issue #1138 and Pull request #1139

## Description

I am sorry, I vandalized the ```sbo__database_cross_reference``` widget and made it so that existing dbxref content is not appearing when editing a page. This happened with [Commit d57135a](https://github.com/tripal/tripal/commit/d57135a9dcc9c58ae80d2205e2e6a79a3b270cee) because I removed some values that the loader should pass on to the widget. This pull request repairs the damage.

## Testing?
Go to a page with this field, e.g. a publication with a PubMed dbxref, and edit. The existing dbxref should display in the widget, new one can be added, existing one deleted, etc.

## Additional Notes (if any):

While I am in here, I noted that this widget, like others, has a facility for rearranging items either through javascript dragging or by using an ORDER column. However, for this particular field, the changes are not saved. The basic reason is that the various ```xxx_dbxref``` tables do not have ```rank``` columns, so there is no place to store the ordering information. I investigated the option of just turning off this reordering functionality for just this widget, but drupal 7.x does not have any easy way to do this, although it has been proposed as a feature for later versions https://www.drupal.org/project/drupal/issues/2264739 so maybe in the future this could be done.

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__database_cross_reference/sbo__database_cross_reference.inc
+++ b/tripal_chado/includes/TripalFields/sbo__database_cross_reference/sbo__database_cross_reference.inc
@@ -129,6 +129,7 @@ class sbo__database_cross_reference extends ChadoField {
     $schema = chado_get_schema($field_table);
     $pkey = $schema['primary key'][0];
     $fkey_lcolumn = key($schema['foreign keys'][$base_table]['columns']);
+    $fkey_lcolumn_id = $entity->{'chado_record_id'};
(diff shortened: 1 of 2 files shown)


spficklin (approved):

spficklin:
Thanks @dsenalik for the fix.   I confirm that it does work. I'm able to add/remove new database cross  references on a Publication page and the code changes look good.

Issue #1258, opened 2022-03-18
Add option for radio-button instead of checkbox on tripal_get_term_lookup_form()


dsenalik:
## Feature Request/Discussion

* This feature does attempt to solve an existing problem with Tripal
* This feature is Not Urgent

### Description
In ```tripal/tripal/api/tripal.terms.api.inc``` line 594 ```function tripal_get_term_lookup_form()``` is used to select one cvterm from among several. I think the typical use case is to only select one term. This function currently uses checkboxes, and can return more than one term. An option to use a radio button instead of a checkbox would make it clear that only one option can be returned.

I am adding in a fix for a bug I found in here. CVs with long definitions are not getting expanded as intended. There is a separate commit to this pull request that adds a fix for this. The bug manifests as a notice of the form
```Notice: Undefined property: stdClass::$definition in tripal_get_term_lookup_form() (line 712 of .../sites/all/modules/tripal/tripal/api/tripal.terms.api.inc).```

### Your Specific Use Case
This would be incorporated into https://github.com/tripal/tripal_analysis_expression/pull/392 to make it clear that only one term can be returned.

### Generally Applicable
There may be a small number of other places this could be incorporated, _e.g._ in ```tripal/tripal/includes/TripalBundleUIController.inc``` line 605 only the first item is returned, so a radio button could be used here also
```
    $selected = tripal_get_term_lookup_form_result($form, $form_state);
    if ($selected) {
      $selected_term = $selected[0];
    }
```

### Additional information/screenshots
The bad thing about this is that we would be adding one more parameter to ```tripal_get_term_lookup_form()```, and it already has 12 parameters!

spficklin:
I agree, this is an important issue that needs fixing. The function that provides the form element is meant to only return a single selected item and if more than one is clicked it will be confusing for the end-user and may break some forms.

spficklin:
Issue fixed with PR merge.  Thanks @dsenalik

PR #1259, opened 2022-03-18
Issue #1258 add radio button option

dsenalik:
# New Feature

Issue #1258 

## Description
This adds one more parameter to ```tripal_get_term_lookup_form()``` which allows the use of radio buttons instead of checkboxes, to enforce returning only a single cv term.

Because this is a new parameter at the end, it should not affect existing code or usage, default is to retain the checkboxes as before.

I also updated the parameter documentation for the missing parameters.

The bad thing is that this adds another parameter to a function that already has 12 parameters.

## Testing?
Example from https://github.com/tripal/tripal_analysis_expression/pull/392
```
      // Add the lookup fieldset to the form, radio set to return only one term.
      tripal_get_term_lookup_form($form, $form_state, $default_term_name,
          $title, '', FALSE,  '', $delta,
          'tripal_biomaterial_loader_form_ajax_callback', 
          '', [], 0, TRUE);
```

Example
radio1

## Additional Notes (if any):
If this is acceptable, I can submit another pull request for places where the radio option should be set, _e.g._  ```TripalBundleUIController.inc```

### DIFF
--- a/tripal/api/tripal.terms.api.inc
+++ b/tripal/api/tripal.terms.api.inc
@@ -585,16 +585,28 @@ function tripal_get_vocabularies() {
  *   Indicates if this form element is required.
  * @param $field_name
  *   The name of the field, if this form is being added to a field widget.
- * @param  $delta
+ * @param $delta
  *   The delta value for the field if this form is being added to a field
  *   widget.
+ * @param $callback
+ *   Ajax callback function, defaults to
+ *     'tripal_get_term_lookup_form_ajax_callback'.
+ * @param $wrapper
+ *   Ajax wrapper ID.
+ * @param $validate
+ *   Widget validation configuration.
+ * @param $weight
+ *   The weight value for the field if this form is being added to a field
+ *   widget.
+ * @param $radio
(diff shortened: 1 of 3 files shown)


dsenalik:
I added a change found in https://github.com/tripal/tripal_analysis_expression/pull/392 where CVs with a long definition were not being expanded as intended. There was code to do this, but it was not working as intended. The call to chado_expand_var() is now moved inside the loop.

spficklin:
Hey @dsenalik. I updated the PR and removed the option to use a checkbox or a radio.  I can't think of any reason that form should return multiple terms. It's purpose is to select one so I think the default to a radio is perfect. Also, there were two other places where this type of form is used: editing a field type and changing the semantic web term for a column in a Chado table.  I changed those to radios too.   I tested all three and it works just fine.  This is a very simple fix so I'm going to go ahead and merge.

spficklin:
I'll merge when all check pass their tests.  Thanks for the PR and the other fixes (inline documentation and expanding the description of the term.

dsenalik:
Thanks! At least there isn't a 13th parameter now!

Issue #1260, opened 2022-03-22
Tripal field obi__organism infraspecific type no_rank displaying


dsenalik:
### System information

* Tripal Version: 7.x-3.8
* Drupal Version: 7.88
* PostgreSQL Version: 13.4
* PHP Version: 7.2.24-0

### Issue description

CarrotOmics.org makes extensive use of infraspecific nomenclature, e.g. subspecies, and the organisms without infraspecific nomenclature use the CV term ```taxonomic_rank:no_rank``` for the ```chado.organism``` table ```type_id``` column. This shows up in the output of the ```obi__organism``` field. A special case override is needed to mask "no_rank" from appearing in the field.

#### Steps to reproduce

Need: Chado v1.3 + organism with type_id taxonomic_rank:no_rank

View a page using this field, e.g. a stock page of Germplasm Accession with an organism without infraspecific nomenclature, but which has the ```no_rank``` CV term for the chado.organism table type_id column. Note that this does not happen entering an organism manually via web page, the type_id column in that case will be ```NULL```. We use a chado bulk loader for organisms.

#### Error messages and screenshots
Example
![obi__organism](https://user-images.githubusercontent.com/8419404/159573589-82eba0f3-9c88-4f6b-9927-9eb9b9f3ba84.png)

spficklin:
Fixed with PR #1261

PR #1261, opened 2022-03-22
Issue #1260 mask no_rank from appearing

dsenalik:
# Bug Fix

Issue #1260 

## Description
This resolves the problem described in issue #1260, basically the CV term taxonomic_rank:no_rank stored in the chado.organism table type_id column is masked from appearing as an infraspecific type when the ```obi__organism``` field is used.

## Additional Notes (if any):
This is just a regex used to remove " no_rank" from the generated label.

### DIFF
--- a/tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
+++ b/tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
@@ -124,19 +124,23 @@ class obi__organism extends ChadoField {
         return;
       }
       $string = $settings['field_display_string'];
+      // The infraspecific fields were introduced in Chado v1.3, include only if present.
+      // Special case conversion of the type_id 'no_rank' value to an empty string.
+      if (property_exists($organism, 'type_id')) {
+        if ($organism->type_id->name == 'no_rank') {
+          $organism->type_id->name = '';
+        }
+      }
       $label = chado_replace_tokens($string, $organism);
       $entity->{$field_name}['und'][0]['value'] = [
         $label_term => $label,
         $genus_term => $organism->genus,
         $species_term => $organism->species,
       ];
-      // The infraspecific fields were introduced in Chado v1.3.
+      // Again, only present in Chado v1.3+
       if (property_exists($organism, 'infraspecific_name')) {
-        $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = NULL;
+        $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = $organism->type_id->name;
         $entity->{$field_name}['und'][0]['value'][$infraspecific_name_term] = $organism->infraspecific_name;
-        if ($organism->type_id) {
-          $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = $organism->type_id->name;
-        }
       }
       $entity->{$field_name}['und'][0][$linker_field] = $organism->organism_id;
 
--- a/tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
+++ b/tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
@@ -124,23 +124,23 @@ class obi__organism extends ChadoField {
         return;
       }
       $string = $settings['field_display_string'];
-      // The infraspecific fields were introduced in Chado v1.3, include only if present.
-      // Special case conversion of the type_id 'no_rank' value to an empty string.
-      if (property_exists($organism, 'type_id')) {
-        if ($organism->type_id->name == 'no_rank') {
-          $organism->type_id->name = '';
-        }
-      }
       $label = chado_replace_tokens($string, $organism);
+      // Infraspecific type, if present but not used, can either be NULL,
+      // an empty string, or the term 'no_rank'. Special case processing
+      // so that this 'no_rank' term is not included in the label.
+      $label = trim(preg_replace('/ no_rank/', '', $label));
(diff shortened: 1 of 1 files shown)


tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
```diff
@@ -124,19 +124,23 @@ class obi__organism extends ChadoField {
         return;
       }
       $string = $settings['field_display_string'];
+      // The infraspecific fields were introduced in Chado v1.3, include only if present.
+      // Special case conversion of the type_id 'no_rank' value to an empty string.
+      if (property_exists($organism, 'type_id')) {
+        if ($organism->type_id->name == 'no_rank') {
+          $organism->type_id->name = '';
```
spficklin: Rather than change the `$organism->type_id->name` value how about just create a variable in this function for the infraspecific name that can get set and used below.  I'm not 100% sure if changing the value of the object means that when it is used in other fields if that change is persistent.

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
```diff
@@ -124,19 +124,23 @@ class obi__organism extends ChadoField {
...
-      // The infraspecific fields were introduced in Chado v1.3.
+      // Again, only present in Chado v1.3+
       if (property_exists($organism, 'infraspecific_name')) {
-        $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = NULL;
+        $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = $organism->type_id->name;
         $entity->{$field_name}['und'][0]['value'][$infraspecific_name_term] = $organism->infraspecific_name;
-        if ($organism->type_id) {
-          $entity->{$field_name}['und'][0]['value'][$infraspecific_type_term] = $organism->type_id->name;
-        }
       }
```
spficklin: I think we should leave this if statement here. Not everyone will set a 'no_rank' for the type and the type_id is allowed to be null.  If it null then the `$organism->type_id->name` value will create warnings.

spficklin (changes requested):

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
dsenalik: That's tricky, because right after that it is used in
```$label = chado_replace_tokens($string, $organism);```
to construct the label, which is where the visible displayed text is created.

spficklin: Hmmm..... Okay, well maybe then restore the value before exiting the function and just add a comment explaining what's going on?

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
dsenalik: Is it a problem to set these to NULL? Because infraspecific_name can be NULL too, and if it's a problem I should also check for that one.

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
dsenalik: To me it looks like organism is a local copy, so won't persist when load() returns.
```$organism = $record->taxon_id;```
Only if I set it directly in $entity would it persist.

Okay, I guess I am wrong...

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
spficklin: The if statement above (line 141) should check if the `infraspecific_name` is  NULL and short circuit the whole thing.  The internal if statement is to catch the case where type_id is NULL.   So, technically it is possible to have an infraspecific name but not indicate a type.

spficklin (changes requested):

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
spficklin: The if statement above (line 141) should check if the infraspecific_name is NULL and short circuit the whole thing. The internal if statement is to catch the case where type_id is NULL. So, technically it is possible to have an infraspecific name but not indicate a type.

spficklin: Oh, but I don't think there would ever be the case that the type_id is set and the infraspecific_name is not. However, Chado allows it :-/

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
spficklin: Another idea would be to remove the `no rank` text from the `$label` using a regular expression if it's there.

dsenalik: Lol. that was my original idea, but it seemed "wrong". It is a lot cleaner and simpler, though. I will push my extensive changes for the record, and then likely do that.
Too many bugs, give up, just do the simple regex fix.

dsenalik:
Okay, the change to the field is now simple and unlikely to break anything.

tripal_chado/includes/TripalFields/obi__organism/obi__organism.inc
dsenalik: Resolved by simple regex change

spficklin (approved):

spficklin:
Nice easy fix.  it's merged!  Thanks @dsenalik

Issue #1245, opened 2021-12-17
The newick phylotree loader does not link to organisms when loading "taxonomic" trees


dsenalik:
### Description

I've been working on the newick tree loader for a while, and this issue will be a placeholder for an eventual pull request, currently a work in progress.

A couple of issues I am working on here:
1. A lot of the job progress output is only viewable by running the drush job at the command line, so there needs to be some fixes for logging.
2. More log messages to show if leaf nodes were or were not associated with features.
3. A tree of type "taxonomy" (cv term "Species tree") should link to organisms instead of features. This does not happen because there is no code to look up organisms. This needs to be added. Pull request #1243 is going to be part of this.
4. There is a bug in the regular expression option because the option exists but is a null string, I added a fix for that.
5. A few typos.
6. The "Feature Name Regular Expression" does not work for taxonomy trees, but there is no reason it can't be made to work.
7. The ```operation__phylotree_vis_formatter``` field always says ```Click a species to view its species page.```, it should have a different message for non-taxonomic trees which link to a feature.
8. Taxonomy tree loader on rare occasions gets an incorrect taxid from NCBI, e.g. https://www.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=taxonomy&term=Torilis+arvensis+subspecies+neglecta returns 1924105 which is a bacterium.
9. I think there is a typo to be fixed, ```=``` instead of ```==``` in ```TaxonomyImporter.inc```
```
           if ($branch_set[$j]['name'] == $node['name'] and
            $branch_set[$j]['depth'] = $node['depth']) {
```
10. Implement the "to-do" in ```TaxonomyImporter.inc``` to use NCBI taxid linked to organism, if present, to avoid a query. This brings up an issue if we have a heterotypic synonym. In this case, we should use the site taxon name rather than the returned NCBI taxon name.
11. In ```function tripal_phylogeny_parse_newick_name``` an arbitrary bootstrap value is created (incrementing integers) for internal nodes and used for the 'name' field. If this is not actually used anywhere, a null string would seem more appropriate. This is not changed in the current pull request.

spficklin:
Issue fixed with PR #1262

PR #1244, opened 2021-12-17
1243 chado unabbreviate infraspecific rank

dsenalik:
# New Feature

Issue #1243

## Description
This commit adds ```function chado_unabbreviate_infraspecific_rank($rank)``` which is an inverse of existing ```function chado_abbreviate_infraspecific_rank($rank)```. This function will be useful for parsing organisms in text strings for determining an existing organism_id in the chado.organism table.

I also added ```cultivar``` as a new abbreviatable infraspecific type.

## Testing?
You can run this in /devel/php
```
$testarray = ['subsp', 'ssp', 'var', 'subvar', 'cv', 'f', 'subf', 'bogus'];
foreach ($testarray as $name) {
  print "\"$name\" becomes \"".chado_unabbreviate_infraspecific_rank($name)."\"\n";
  $name .= '.';
  print "\"$name\" becomes \"".chado_unabbreviate_infraspecific_rank($name)."\"\n";
}
```
Output
```
"subsp" becomes "subspecies"
"subsp." becomes "subspecies"
"ssp" becomes "subspecies"
"ssp." becomes "subspecies"
"var" becomes "varietas"
"var." becomes "varietas"
"subvar" becomes "subvarietas"
"subvar." becomes "subvarietas"
"cv" becomes "cultivar"
"cv." becomes "cultivar"
"f" becomes "forma"
"f." becomes "forma"
"subf" becomes "subforma"
"subf." becomes "subforma"
"bogus" becomes "bogus"
"bogus." becomes "bogus."
```

PR #1262, opened 2022-03-24
Newick Importer --Bug fixes + Enhancements

### LINKED ISSUE #1408: tv3g4 PHP8 errors in taxonomy loader
# BUG/ERROR report

## System information

* Tripal Version: 3.9dev
* Drupal Version: 7.94
* PostgreSQL Version: 13.3
* PHP Version: 8.0.27

### Issue description
Various warnings under PHP8 loading organisms with NCBI taxid numbers
```
Undefined array key "name_re" tripal_chado.phylotree.api.inc:317
Undefined array key "analysis_id" tripal_chado.phylotree.api.inc:322
Undefined array key 20 TaxonomyImporter.inc:1064
Undefined property: stdClass::$is_new TaxonomyImporter.inc:573
Undefined array key "job" tripal_chado.phylotree.api.inc:330
Undefined array key "name_re" tripal_chado.phylotree.api.inc:796
Undefined variable $value TaxonomyImporter.inc:1098  (hundreds of occurrences)
```

dsenalik:
# Bug Fix

Issue #1245

## Description

This pull request contains a number of bug fixes and enhancements to the Newick phylogenetic tree loader. This started as an attempt to get taxonomy trees to link to organisms, which had not been implemented for Newick loaded trees, but a number of issues showed up which needed fixing.

1. Fuller implementation of job logging to allow output to go to the "jobs" page and not just to drush output.
2. Additional log messages to show if leaf nodes were or were not associated with features or organisms.
3. Taxonomy trees _loaded from Newick files_ now link to organisms. A new function does the name to organism lookup ```chado_phylogeny_lookup_organism_by_name()``` which in turn calls ```chado_unabbreviate_infraspecific_rank()``` from pull #1244
4. Fixed the bug in the regular expression option.
5. Various typo fixes.
6. The "Feature Name Regular Expression"  has been expanded to also work on taxonomy trees loaded from Newick files.
7. The ```operation__phylotree_vis_formatter``` field now says ```Click a species to view its species page.``` only for taxonomy trees, it now displays ```Click a node label to view its page.``` for all other types of phylogenetic trees.
8. There is the existing option to load the tree with the loader job, or as a separate job. I flipped the logic so that unchecked is the default, and the default is to load everything at once. This option is not currently available to the drush implementation, so the change should not break anything. (```load_now``` becomes ```load_later```).
9. Some of the error messages have been enhanced to show the value that caused the error.
10. There's a thing added to ```tripal.module``` to make sure the legacy ```'file'``` value is present before adding it.
11. NCBI Taxonomy importer form adds options for tree name and root taxon. This allows a site to have a subset of the organisms in a tree, such as a separate tree for each family. I added an example image below for the new input form, and for a tree with the family as the root. The default for the tree name is unchanged from before for backwards compatibility, although really there should be a space added.
12. Queries to NCBI are now wrapped in a retry loop, to handle rare cases when an internet query goes bad. This is quite rare for me, but does happen.
13. On rare occasions a query to NCBI has no full match, but NCBI returns a partial match which yields an incorrect taxid. This is now checked for, and the invalid taxid discarded. Example, this returns a taxid for a bacterium:  https://www.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=taxonomy&term=Torilis+arvensis+subspecies+neglecta
14. Some conflicting comments as to whether an organism tree uses 'taxonomy' or 'organism' when loading has been changed to always use 'taxonomy'. Also this setting is now only passed through ```$options['leaf_type']``` consistently throughout.
15. Implemented the "to-do" of using stored NCBI taxid for site organisms, when present. Though this then can lead to a difference in site vs. NCBI taxon names. So...
16. In the case of heterotypic or homotypic synonyms, the taxon name stored in chado is used in preference to that returned from the NCBI taxid query, in order to keep the site consistent.
17. Phylotree leaf nodes can now link to stock in addition to feature and organism. Useful for loading newick tree files linked to germplasm accessions.
18. If internal tree nodes have labels, they will display with a mouse-over. Example image shown below.
19. PHP8 errors described in issue #1408
20. Infraspecific problem described in issue #1409


dsenalik:
Example images for item #11. Added two input fields for tree name, and root taxon.
20220331_loader

-----

Part of an example sub-tree with the root at the family level (Pittosporaceae).
20220331_pittosporaceae

dsenalik:
For item 18: Example image showing mouseover of interior node of a taxonomic tree
InteriorMouseover

dsenalik:
Item 3 on my list will need to be modified if pull #1328 is accepted, this implements a general api function to look up an organism from its name, and would replace most of the function ```chado_phylogeny_lookup_organism_by_name()``` I created in [tripal_chado.phylotree.api.inc](https://github.com/tripal/tripal/pull/1262/files#diff-e54bce40f43784e0216eeac80f73f3b36617b7ea5b20f12dc4bfd988190d3d6d)

dsenalik:
This pull request is now dependent on #1328 being merged first, so testing will fail until that happens.

spficklin:
@dsenalik can you provide a sample newick file I can use for testing and a set of instructions to test?

dsenalik:
@spficklin I actually created two basic automated tests for this loader, which previously had no tests. (Hopefully this can help with porting to tripal 4 eventually where we need full test coverage). Looks like they passed, yay!
The second test checks items 3 and 6 at the top of this pull request, to link directly to organisms and the regex applying to this also.
There are a lot of things I did here, and the hardest part of testing or writing a testing protocol is to create the organisms and features before you can test the loader. Hopefully this is a start...

spficklin (approved):

spficklin:
This looks great @dsenalik . Thanks for all your help improving it.

Issue #1266, opened 2022-04-11
sbo__relationship field needs to be aware of organism infraspecific nomenclature


dsenalik:
### System information

* Tripal Version: 7.x-3.8
* Drupal Version: 7.89
* PostgreSQL Version: 13.2
* PHP Version: 7.3.28

### Issue description
The tripal field sbo__relationship needs to be able to handle chado 1.3 infraspecific nomenclature if it is being used for organism relationships. Example shown here is Daucus carota subsp. sativus related to Daucus carota var. sativus

![sbo__relationship_example](https://user-images.githubusercontent.com/8419404/162789350-6d828727-f8c5-4e17-a2ce-887ea2056e03.png)

This will require a bit of special-case handling for the organism table, but the chado_get_organism_scientific_name() function could be used to get the full organism name.

#### Steps to reproduce
  - chado 1.3
  - an organism with an infraspecific name
  - a relationship in the chado.organism_relationship table
  - sbo__relationship field on the organism page

laceysanderson:
Fixed via merged PR.

dsenalik:
I'll reopen this as I have a second pull request #1396 to deal with full infraspecific support, abbreviations and all, dependent however on #1328 being merged first.

spficklin:
Fixed with PR #1396

PR #1267, opened 2022-04-12
Issue #1266 sbo__relationship field infraspecific nomenclature support

dsenalik:
# Bug Fix

Issue #1266

## Description
Changes in this pull request allow the ```sbo__relationship``` tripal field to support infraspecific nomenclature for organisms, which was added with chado v1.3

For the organism name, preference is given first to the abbreviation in the chado.organism table.
If no abbreviation is defined, and we have chado >=1.3, use the chado_get_organism_scientific_name() function to get the organism name.
Fallback is to the original genus + species method of constructing the name.

I also removed an extra space introduced into the relationship clause when the verb from get_rel_verb() is a null string by padding in that function only when there is a verb returned. This only appears visibly in web services.

## Testing?
Need chado 1.3, and one or more organisms with subspecies or other infraspecific name, and a relationship in the organism_relationship table, e.g.
20220412_example
20220412_webservices png


dsenalik:
I am trying to understand why testing is failing and so far can't figure it out.

dsenalik:
Ooooh that was an insidious hard-to-spot bug that the testing caught, but only after the first PHP8 bug was fixed.
All should be good now!

laceysanderson (commented):
Functional review:
❌ Seeing the full name w/ infraspecific in the subject/object drop down on the organism add page
✅ Seeing the full name w/ infraspecific in the subject/object drop down on the organism edit page
✅ Setting the formatter to "relationship statements", the statements are correct both when the subject and/or object organism has an infraspecific name with no abbreviation
❌  Setting the formatter to "relationship table by type", I'm seeing a weird bug (see below)

My relationship is set as follows:
Screen Shot 2022-12-01 at 12 33 44 PM
When I look at the Tripalus organism page (i.e. subject page), I see the table showing the relationship to L. culinaris as I expect. but when I look at the Lentil organism page (object page), the relationship table does not show Tripalus as it should but instead still shows Lentil as shown in this screenshot 😬 
Screen Shot 2022-12-01 at 12 32 44 PM

I'm also unable to add a relationship on the add organism page anymore. When I try with the form filled out as follows, I get a validation error:
> Please provide the object of the relationship.
Screen Shot 2022-12-01 at 12 40 17 PM
This is likely due to the dropping of the check for the record_id in the one if statement looking at the object.

dsenalik:
PHP8, empty tripal3 test site, adding a new organism, when select an empty ontology (i.e. no terms, ```project_relationship``` on an empty tripal3 site), I see this error from the ```schema__publication``` field, noting it so I don't forget
```Warning: Undefined array key "accession" in schema__publication_widget->form() (line 69 of /var/www/prova/drupal-7.92/sites/all/modules/tripal/tripal_chado/includes/TripalFields/schema__publication/schema__publication_widget.inc).```

Fixed with https://github.com/tripal/tripal/pull/1267/commits/ab2ac5f2c28c2f6785164376a9eef8aebda343de

**Comment by dsenalik** [comment]
The "weird bug" in https://github.com/tripal/tripal/pull/1267#pullrequestreview-1201494455 is a pre-existing bug. Most relationships are on content types named with just ```schema__name```. But others, such as organism, use something else, and ```schema__name``` does not exist. This is currently used in ```sbo__relationship_table_formatter.inc``` to determine if the current content page is subject or it is object, and for organisms this is undefined and fails so the page is always "subject". Under PHP8 you also get warnings that help track this down.
```Warning: Undefined property: TripalEntity::$schema__name in sbo__relationship_table_formatter->view() (line 183 of /var/www/prova/drupal-7.92/sites/all/modules/tripal/tripal_chado/includes/TripalFields/sbo__relationship_table/sbo__relationship_table_formatter.inc).```

The offending section of code is:
```
      // Determine if the subject or object is the current entity.
      $is_subject = TRUE;
      if ($object_name == $entity->schema__name['und'][0]['value']) {
        $is_subject = FALSE;
      }
```
I can make this check more generic and work on all content types with this simple fix:
```
      if ($entity->chado_record_id == $item['chado-'.$entity->chado_table.'_relationship__object_id']) {
```

dsenalik:
New issues noted in the "Relationship Table by Type" version:

The grammar of the cv term is not formatted as nicely as in the "Relationship Statements", e.g. 
The following are _Is Individual Of Population_ this Organism
because it uses these settings
20221203_table

dsenalik:
To finish off review responses, I can see infraspecific name in organism add page
add_page
so @laceysanderson can you please try that again?

The validation error may have been the same one I fixed in the chado_linker__contact field in pull #1319 so I will incorporate that fix here too for testing https://github.com/tripal/tripal/pull/1267/commits/5949490bdd9eb03e2a3895fe80c12e2d798469d4
and remember that the order of preference for the organism name is:
abbreviation → chado_get_organism_scientific_name() → Genus+species
so if you filled out abbreviation while creating the new organism, that is what needs to go in the box.

dsenalik:
The state of this pull request leaves organism relationships only fully working for Genus + species, infraspecific nomenclature only works if the Ajax callback sets the ```[id: xx]``` organism_id value. To look up an organism by name _e.g._ ```Tripalus domesticus subsp. chadoii``` is not easily done in just SQL because of the type abbreviations (for which there are api functions). I will propose a new api function for this lookup, and can then implement full infraspecific support for this field later.

I created Issue #1327 for this.

laceysanderson (commented):
Functional review:
✅ Seeing the full name w/ infraspecific in the subject/object drop down on the organism add page
✅ Seeing the full name w/ infraspecific in the subject/object drop down on the organism edit page
✅ Setting the formatter to "relationship statements", the statements are correct both when the subject and/or object organism has an infraspecific name with no abbreviation
✅ Setting the formatter to "relationship table by type"now works beautifully
✅ Testing the "empty cell means self" functionality on stock-based ADD + EDIT content types
✅ Testing the "empty cell means self" functionality on organism-based ADD + EDIT content types (tested both with + without infraspecific name in both forms and it works for me!)

Code Review
✅  Looks good! Thanks for all the spelling fixes and the automated testing!

Only noted caveat if if someone doesn't select from the autocomplete on add/edit of an organism relationship including the infraspecific name. On add this causes a nice validation error saying it couldn't find the subject. On edit it bypasses validation somehow and you end up with a circular relationship added:
This form was filled out when editing the Lens culinaris organism to add a relationship.
Screen Shot 2022-12-06 at 1 58 27 PM
After save, this is the organism page I was presented with. So validation error and a circular relationship of the type I had selected but not including the organism with the infraspecific name.
Screen Shot 2022-12-06 at 1 58 43 PM

dsenalik:
@laceysanderson this "caveat" you noted is almost certainly due to the lack of infraspecific support in one spot, where only genus and species are queried in the organism table. I can replicate exactly what you saw. I will be able to easily fix this if pull #1328 is approved.

laceysanderson:
Great, I thought that might be the case :-) Can we add a validation error to this PR for that one case just so it doesn't create an erroneous relationship?

dsenalik:
I have added this with https://github.com/tripal/tripal/pull/1267/commits/fa3e0a938da20bea8c0b606415866b4cfc1b695e  and it now prevents this from happening
nonexisticus
Note that I did not have to check for multiple matches, that is already caught elsewhere.

dsenalik:
I am putting this here so I remember next year :yawning_face: 
If you use the relationship field to link two derived content types, e.g. **Genome Assembly** which is derived from **Analysis**, then the underlying type is displayed, and both subject and object are clickable links. For example
20221230_relationship

laceysanderson (approved):
Now that @dsenalik addressed my concerns I feel this is ready to be merged :-) 

Regarding, the following comment, I don't think this is a big enough problem to worry about fixing in Tripal 3 :-) Thus it should not block merging of this PR.

> If you use the relationship field to link two derived content types, e.g. Genome Assembly which is derived from Analysis, then the underlying type is displayed, and both subject and object are clickable links.

PR #1279, opened 2022-06-09
Issue 1277 convert individual radio to radios

### LINKED ISSUE #1277: Radio buttons in function tripal_get_term_lookup_form() are not linked by name
### Issue description
The cvterm lookup form uses radio buttons to select a term. Right now, once checked, the button cannot be unchecked and multiple buttons can be checked.

This is defined by ```function tripal_get_term_lookup_form()``` in ```tripal/api/tripal.terms.api.inc``` and the form needs something like ```'#name' => $appropriate_unique_value,``` at about line 712. I'm not sure what that value should be, though. ```$delta``` is working, but is it guaranteed unique? Actually that doesn't work, the selected value is not returned.

#### Error messages and screenshots
radiobuttons

dsenalik:
# Bug Fix

Issue # 1277

## Description

Pull request #1259 converted checkboxes to radio buttons for CV term selection, since only one term should be returned. Inadvertently this introduced the problem of still checking multiple radio buttons since the form does not link them by name. And since each is an individual, once checked, they cannot be unchecked. 
assay
And at least in some cases with my testing, the selected item is not always returned as expected.

Changes here convert the existing individual "radio" fields to a single "radios" so that, as expected, only one item can be selected. This introduced a change in how the selected item is returned in the ```$values``` array, so the corresponding functions that get the returned items has been slightly modified.

Drupal 8 and above has the ability to add ```#description``` fields to each item in a "radios", but drupal 7 does not. However, a callback function is used to implement this, so the appearance should be unchanged by this pull request.

## Testing?

The semantic web form can be tested at ```/admin/tripal/storage/chado/semweb```, try changing any term to verify that the radio buttons allow only one term selected at a time, and returns the correct selected term.

Test the biomaterial loader at ```/admin/tripal/loaders/chado_biosample_loader```, you will need a biosample xml or csv to test this one. On the second screen of the form you can try selecting terms, and see that they are returned correctly on the third screen. A biosample does not have to be actually loaded for testing.


dsenalik:
Another test is to Add Tripal Content -> "File" (tripal_file module). Check that FILE TYPE selection and saving works.
Commit https://github.com/tripal/tripal/pull/1279/commits/c3d51f4c1f96e84b38bb3bb148e43dc1de5ec468 fixed a problem there.

dsenalik:
I have added another commit to this because of a problem with the code to automatically check the radio button if there is only one term to choose from. There is still a bug where when editing existing content, the selected term does not get auto-selected.

dsenalik:
Default value selection now works on editing. There was an issue if editing when there is more than one cv term with the same name, I now pass through the cvterm_id value using the same mechanism as the relationship field, e.g. "termname" can be passed in as "termname [id: 1234]" by the widget to disambiguate.

laceysanderson (approved):
✅ Quick code review looks good :-)
✅ The radios on the semweb form now work as expected when there is more then one term.
❔ I don't have access to the biomaterial loader at /admin/tripal/loaders/chado_biosample_loader. Is this added by an extension module?
✅ This form is also used when selecting the type for a new Tripal Content Type. I checked it there and it seems to work as expected (e.g. I searched for chromosome as the type, it returned 2 results in my site. I wasn't able to select more then one and on creation the content type had the term I selected.)
✅ This form is also used in the schema__additional_type field. This field type is used when there is a type_id on your table and it is not linked to the content type (e.g. you create a content type for all records in the stock table then this field will be added). I confirmed it functioned and saved as expected with the form changes :-)

This functionality seems pretty sound to me now :-) I believe this is ready for merging.

dsenalik:
The biomaterial loader at ```/admin/tripal/loaders/chado_biosample_loader``` is from the ```tripal_analysis_expression``` module.

laceysanderson:
Ok, since all core functionality around this module has been tested and is very sound, I'm going to go ahead and merge it. Thanks @dsenalik 🎉

Issue #1283, opened 2022-06-24
Vocabulary warnings shown to end-users


Ferrisx4:
### Issue description

Occasionally, a warning about missing details for a vocabulary is generated. This warning was enhanced in #500 by suggesting a fix.
This issue is not about the fact that the warning gets generated, but that it is shown to the end-user.

It would be a simple fix to make a check to see if the user is an admin or not.

#### Error messages and screenshots

![Cropped](https://user-images.githubusercontent.com/9593623/175560977-10a99c95-6087-47a5-8f6e-187082024c7b.png)

spficklin:
Fixed with recent PR

Issue #500, opened 2018-07-04
Could not find details about the vocabulary


laceysanderson:
### System information

* Tripal Version: 3.x-dev
* Drupal Version: 7.x
* PostgreSQL Version: 9.4.10
* PHP Version: n/a

### Issue description
When clearing the cache I was seeing a huge list of the following drupal_set_messages:
```
    Could not find details about the vocabulary: SUMMARY
    Could not find details about the vocabulary: PATO
    Could not find details about the vocabulary: OMIT
    Could not find details about the vocabulary: NCIT
    Could not find details about the vocabulary: FLOPO
    Could not find details about the vocabulary: OMIT
    Could not find details about the vocabulary: SEARCH
    Could not find details about the vocabulary: SUMMARY
    Could not find details about the vocabulary: SEARCH
    Could not find details about the vocabulary: NCIT
```

#### Solution
This was due to `db2cv_mview` missing ontologies I had recently added. Re-populating the materialized view at Administration » Tripal » Chado Schema » Materialized Views fixed the problem.

#### Steps to reproduce
1. Delete an entry out of db2cv_mview and clear the cache.
2. You should see a drupal_set_message telling you it could not find details about the vocabulary you deleted.

**Perhaps we should update the materialized view when we can't find a vocabulary? Or at least add to the message telling admin that updating db2cv_mview might solve the problem?**

spficklin:
This mview was supposed to be updated when the site was prepared but we had a bug and it didn't work.  It has been fixed in a previous PR. Also, anytime you load a new ontology it automatically updates the mview.  So, you shouldn't ever get out sync again unless you manually or programatically add a new vocabulary.  We could update those API functions to automatically update the mview.

laceysanderson:
In my case I was programmatically creating terms. I feel this will be more common though now that we need terms for each field. As such, I think we should update this materialized view in chado_insert_cvterm().

spficklin:
Coming back to this. If we update the mview each time we call chado_insert_cvterm it would get updated way too often.  There are 156 calls in the tripal_chado.semweb.inc file alone.  I think the solution may be that we need a way to schedule updates of mviews as per request #353.    Would that work for you?

laceysanderson:
Perhaps... However, I would want the ability to turn some mviews on for automatic update but not all. If we had that granularity and this was one was updated automatically by default then it could be a good solution. 

[...]

PR #1284, opened 2022-06-28
Check if user is admin before warning

Ferrisx4:
# Bug Fix

Issue #1283

## Description

The user would be warned if Drupal could not find details about a vocabulary, even if they are not an admin.

## Testing?
Testing is difficult due to the root of this issue mostly being solved by https://github.com/tripal/tripal/issues/500, although some sites still may encounter this infrequently. This message typically appears after a cache clear. If your site suffers from the warning, it should not be shown unless you are logged in as an admin.

## Additional Notes:

While this issue and PR address a specific case of warnings/messages for admins being shown to any user, it is not the only such case. As current Tripal 3 development is mostly over, it is and should be a consideration in Tripal 4 and beyond to ensure that messages only get shown to appropriate users.

### DIFF
--- a/tripal_chado/includes/tripal_chado.vocab_storage.inc
+++ b/tripal_chado/includes/tripal_chado.vocab_storage.inc
@@ -103,14 +103,9 @@ function tripal_chado_vocab_get_vocabulary($vocabulary) {
   $result = chado_query($sql, [':name' => $vocabulary]);
   $result = $result->fetchAssoc();
   if (!$result) {
-    // Only output this error to people who are admins
-    global $user;
-    // Is the user even logged in? If so, are they admin?
-    if ($user->uid && array_search('administrator', $user->roles)) { 
-      drupal_set_message("Could not find details about the vocabulary: $vocabulary. Note: " .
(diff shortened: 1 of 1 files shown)


dsenalik (changes requested):
Line 109 should be
``` if ($user->uid && ...```
otherwise I don't think the message will ever appear, an administrator will always have a uid.
I tested this with this change in the devel php screen and it seems to work as intended.

I will push this change to your branch.

spficklin (approved):

spficklin (changes requested):
We have an API function for targetting Tripal admin users.  Try this function: `tripal_set_message` instead of checking for the administrator user.

http://api.tripal.info/api/tripal/tripal_core%21api%21tripal_core.tripal.api.inc/function/tripal_set_message/2.x

spficklin (approved):

spficklin:
Thanks @Ferrisx4 !  The switch to use tripal_set_message means that the fix requested by @dsenalik wasn't also addressed.

Ferrisx4:
Yes, thanks!

Issue #1288, opened 2022-08-07
Some Web Services links for content types with a space not working


dsenalik:
### System information

* Tripal Version: 7.x-3.x current dev version
* Drupal Version: 7.91
* PostgreSQL Version: 13.3
* PHP Version: 7.3

### Issue description
An issue occurs if we have a content type that links to another content type with a space in the name, for example ```Germplasm Accession```. The content type is passed through ```urlescape()``` and this has the effect of changing spaces to ```+```.
Elsewhere in the code links are passed through a different mechanism, where spaces are replaced with underscores to detect the entity type. This means the links don't work.

#### Steps to reproduce
You need a tripal field that links to content type with a space in the name e.g. ```germplasm accession```, ```biological sample```, etc. then view the web services page. For example,
https://carrotomics.org/web-services/content/v0.1/organism/172912
![20220807_badlink](https://user-images.githubusercontent.com/8419404/183308122-7b267dfb-cc21-4ae7-92d2-1a759ead5b15.png)

The illustrated link gives this error message:
![20220807_errormessage](https://user-images.githubusercontent.com/8419404/183308132-cecdf306-f9e4-431f-a87a-264e92db3373.png)

but changing the ```+``` in the url to underscore makes it work properly:
![20220807_correcturl](https://user-images.githubusercontent.com/8419404/183308151-b4bd2a58-772a-4451-90cd-d3878082700b.png)

The proposed change is to duplicate the line of code from ```handleRequest()``` which replaces non-alphanumeric with underscore.
Do this prior to calling ```url_escape()```, see linked pull request for details.

dsenalik:
Fixed with PR #1289

PR #1289, opened 2022-08-07
Issue 1288 use the regex used by handleRequest()

dsenalik:
# Bug Fix

Issue #1288

## Description

The code in ```handleRequest()``` and elsewhere uses this method to "sanitize" the the bundle label
```$label = preg_replace('/[^\w]/', '_', $bundle->label);```

The simple change here uses the same code to generate the URL in ```function sanitizeFieldEntity()``` prior to calling ```urlescape()```

Screenshots in #1288

## Testing?
Was described in #1288. I tried to find the error occurring on other sites, but you need to have a content type containing a space in the name e.g. "Biological Sample" linked from an original content type by a tripal field, and so far have not found any examples.


spficklin:
Thanks Doug!

PR #1291, opened 2022-08-11
Issue 1088 sep__protocol check for null

dsenalik:
# Bug Fix

Issue #1088

## Description
Simple fix to avoid passing null to ```array_key_exists()``` as described in this comment https://github.com/tripal/tripal/issues/1088#issuecomment-1212360100

## Testing?
I see it editing an "Assay", not sure if related to another problem, but the fix should be harmless.


spficklin (approved):

spficklin:
Easy fix! Thanks Doug

Issue #1294, opened 2022-08-30
PHP8 error, fetchCol passed a non-integer parameter


dsenalik:
### Issue description
This is the same issue as 
https://github.com/tripal/tripal_elasticsearch/issues/277
but only shows up under PHP8

Specifically, ```fetchCol()``` accepts an integer parameter, but a string is passed.

Easy fix to line 211 of ```tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record.inc```

### Error message
TypeError: PDOStatement::fetchAll(): Argument #2 must be of type int, string given in PDOStatement->fetchAll() (line 2310 of .../includes/database/database.inc).

dsenalik:
P.S. I did a ```grep -nr "fetchAll('"``` and ```grep -nr 'fetchAll("'``` and did not find any other occurrences

spficklin:
Fixed with recent PR.

PR #1295, opened 2022-08-30
Issue 1294 fix non-integer parameter to fetchCol

dsenalik:
# Bug Fix

Issue #1294

## Description

```fetchCol()``` accepts an integer parameter, a string was passed causing the error noted in the linked issue. This only appears as an error with PHP8.


spficklin (approved):

spficklin:
Thanks Doug!

Issue #1300, opened 2022-09-26
Infraspecific nomenclature missing for FASTA and GFF3 loaders


dsenalik:
### System information

* Tripal Version: 7.x-3.9
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: 8.0.23

### Issue description
The FASTA loader does its own db query to get an organism list, and uses genus + species + (common name)
Therefore infraspecific names are not shown, leading to ambiguous selection. And if common name is not defined, empty parentheses are shown.
Example ```Daucus carota subsp. broteri``` displays as ```Daucus carota ()```
I propose to use the existing function ```chado_get_organism_select_options()```, although this will no longer include the common name. Any opinions about that? *Edit, common name is restored by pull #1311

#### Steps to reproduce
Site must have infraspecific nomenclature for at least one organism.
Go to the fasta loader form page at ```/admin/tripal/loaders/chado_fasta_loader``` and try selecting that organism.

dsenalik:
Same thing in GFF3 Importer

spficklin:
Thanks @dsenalik .  Fixed with PR.
@risharde can you make sure we get this fix in the GFF3 loader of Tripal v4?

PR #1301, opened 2022-09-26
Issue 1300 FASTA importer organism selector infraspecific nomenclature

dsenalik:
# Bug Fix
Proposed fix for issue #1300 

Issue #1300

## Description
Replace db query for organism selector with existing function ```chado_get_organism_select_options()``` so that infraspecific nomenclature is displayed. However, common name is no longer displayed.

## Testing?
Described in issue #1300


spficklin (approved):

spficklin:
Nice catch.  Thanks for the fix.

Issue #1302, opened 2022-09-27
local__source_data widget not respecting proper field length limits


dsenalik:
### System information

* Tripal Version: 7.x-3.9
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: 8.0.23

### Issue description

The widget for the tripal field ```local__source_data``` does not have the correct length limits for the "Data Source Name", "Data Source Version", and "Data Source URI" fields, they are defaulting to drupal defaults which do not match the schema values. Fix this by copying from the table schema, or for the URI field which has no limit, set it to a large value 8192. In Tripal4 this will become a textarea.

#### Steps to reproduce

Try entering a very long URI and observe it being truncated.

spficklin:
Fixed as per PR

dsenalik:
I'm reopening since there is a subsequent pull request on this branch

dsenalik:
This was completed with merged pull #1304

PR #1303, opened 2022-09-27
Issue #1302 get field length limits from table schema

dsenalik:
# Bug Fix

Issue #1302

## Description
Retrieves the table schema to determine proper length limits for textfields in the widget. For the "Data Source URI" which has no length limit, just set a large limit of 8192. In T4 this will change to a textarea.


spficklin (approved):

spficklin:
Thanks Doug for the quick fix!

dsenalik:
See pull #1304 where I later changed to textarea to eliminate length check for URI

PR #1304, opened 2022-09-28
Issue #1302 convert local__source_data URI to textarea

dsenalik:
# Bug Fix
Issue #1302

## Description
Update to previous pull #1303 to convert field to textarea and eliminate length limit issue entirely

### DIFF
--- a/tripal_chado/includes/TripalFields/local__source_data/local__source_data.inc
+++ b/tripal_chado/includes/TripalFields/local__source_data/local__source_data.inc
@@ -157,6 +157,11 @@ class local__source_data extends ChadoField {
     $sourceversion_term = chado_get_semweb_term('analysis', 'sourceversion');
     $sourceuri_term = chado_get_semweb_term('analysis', 'sourceuri');
 
+    // Need to expand sourceuri if it is >=250 characters long.
+    if (in_array('analysis.sourceuri', $analysis->expandable_fields)) {
+      $analysis = chado_expand_var($analysis, 'field', 'analysis.sourceuri', []);
+    }
+
     $entity->{$field_name}['und'][0] = [
       'value' => [
         $sourcename_term => $analysis->sourcename,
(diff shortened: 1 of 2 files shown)


Ferrisx4:
Are there any additional steps to get this to work? I have been testing this (with the initial change of setting the character limit of the textfield to 8192, as well as the recent pushes that switch it to a textarea with no size limit), and I am still unable to have Tripal display the value after setting a source URI that is 251 characters or longer. The value is stored in the database properly.

I noticed in our original conversation that we were experiencing slightly different issues, which is odd and may be a factor. My limit seems to be 250 while yours was 512.

I thought it might be a theme issue, so I switched from Garland 7.92 to Bartik 7.92 and the same issue persists.

The error I get on the page when viewing an analysis with a sourceuri with 251 or more characters:
```
    Notice: Undefined property: stdClass::$sourceuri in local__source_data->load() (line 164 of /var/www/drupal/sites/all/modules/tripal/tripal_chado/includes/TripalFields/local__source_data/local__source_data.inc).
    Notice: Undefined property: stdClass::$sourceuri in local__source_data->load() (line 168 of /var/www/drupal/sites/all/modules/tripal/tripal_chado/includes/TripalFields/local__source_data/local__source_data.inc).
```

dsenalik:
This value of 251 is ~~probably~~ from hitting the limit of ```chado_expand_var()```. From that function there is this comment

... tripal core provides a tripal_chado_exclude_type_by_default() which says that text fields are often very large and if they are longer than 250 characters then we want to exclude them by default

I did a similar fix with https://github.com/tripal/tripal/commit/b81881de93a94007195eddcf4c446390a6864e0e for the ```sbo__relationship``` field where long publication titles (more than 250 characters) were being excluded.

I will add this type of fix in.

Ferrisx4 (approved):
Thanks @dsenalik this looks good, the overlong URIs are now displayed correctly. Nice fix!

PR #1309, opened 2022-10-15
Changed array.

dapineda:
I have changed some arrays because PHP versions changed the way arrays are initialized.

Everything works exactly the same, just change the way arrays are used to adapt to the recommended PHP version and later.


laceysanderson:
Hello @dapineda, thank you for contributing to Tripal! Unfortunately this particular PR doesn't quite meet our requirements for a contribution. That said, we have decided to make this change from `array()` to `[]` syntax in our newest version being worked on [in our t4d8 repo](https://github.com/tripal/t4d8). If you would like to make a more meaningful contribution, we would welcome a PR changing this syntax throughout that repository :-) For an idea of magnitude that is about 336 instances according to grep. Alternatively, you're welcome to look into any of our other outstanding issues and see if they are something you would like to tackle.

PR #1311, opened 2022-10-16
Issue 1310 changes to chado_get_organism_select_options()

### LINKED ISSUE #1310: chado_get_organism_select_options() bug and optimization
### Issue description
1. The function ```chado_get_organism_select_options()``` I submitted some changes in pull #1095 two years ago but the SQL for tripal v2 was not right if there was chado 1.3 infraspecific nomenclature on the site, I added SQL code after the ```ORDER BY```, sorry. 
2. In pull #1300 this function is now used in place of custom SQL to get the organism list, but the previous behavior was to include common name in parentheses, that was lost. The pull request #1311 will add this back.
3. This function can be optimized a bit more, by only calling ```chado_get_organism_scientific_name()``` when necessary.

#### Steps to reproduce
I have been testing this mainly in the devel module php window with
```
$test = chado_get_organism_select_options(FALSE, TRUE);
var_dump($test);
```
with all possible combinations of the two parameters.

View result of adding common name back by going to the FASTA or GFF3 loader pages and see what's in the organism selector list.

dsenalik:
# Bug Fix

Issue #1310

## Description
A new parameter is added to ```chado_get_organism_select_options()``` so that an organism common name can be appended inside parentheses. This is to restore the behavior of including this, which was removed in the changes in pull #1300 
Bugs in tripal v2 sql fixed.
Optimization by only calling ```chado_get_organism_scientific_name()``` when infraspecific nomenclature is present.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.organism.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.organism.api.inc
@@ -182,60 +182,72 @@ function chado_get_organism_scientific_name($organism) {
 }
 
 /**
- * Returns a list of organisms that are currently synced with Drupal to use in
- * select lists.
+ * Returns a list of organisms to use in select lists.
  *
  * @param $syncd_only
  *   Whether or not to return all chado organisms or just those sync'd with
- *   drupal. Defaults to TRUE (only sync'd organisms).
+ *   drupal. Defaults to TRUE (only sync'd organisms). For Tripal v3 you want FALSE here.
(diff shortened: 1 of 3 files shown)


laceysanderson (approved):
Code review ✅  Looks good to me!
Thanks for this fix @dsenalik

laceysanderson:
Also did a functional review and it works great :-) This is ready to merge!

PR #1314, opened 2022-10-28
Issue 1313 field name typos fixed

### LINKED ISSUE #1313: Can't display the contact field on a biomaterial or assay page
### System information

* Tripal Version: 3.9
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: 8.0.24

### Issue description
The tripal field ```local__contact``` will display contact content type on a biomaterial page (as well as assay and arraydesign it seems from the code).

It works through the widget, can be edited there, but does not display on the biomaterial content page. This is because it is hidden and can't be unhidden as shown in this screenshot
contact-display

dsenalik:
# Bug Fix

Issue #1313 

## Description

This fixes two functional and two non-functional typos for the ```local__contact``` field that prevent it from displaying any content because the field cannot be un-hidden. Specifically, ```local_contact``` is changed to ```local__contact``` in two places.

### DIFF
--- a/tripal_chado/includes/TripalFields/local__contact/local__contact_formatter.inc
+++ b/tripal_chado/includes/TripalFields/local__contact/local__contact_formatter.inc
@@ -6,7 +6,7 @@ class local__contact_formatter extends ChadoFieldFormatter {
   public static $default_label = 'Contact';
 
   // The list of field types for which this formatter is appropriate.
-  public static $field_types = ['local_contact'];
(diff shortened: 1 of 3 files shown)


laceysanderson (approved):
Beautiful -Code review ✅ 
I also confirmed that using the "Find new fields" I was able to add the contact field to a project-based content type, edit it and display it on the page ✅ 

Thanks for the fix @dsenalik! This is ready to merge!

Issue #1272, opened 2022-04-25
chado_query array[] placeholder does not work as intended


risharde:
This issue was found while testing the GFF3 loader. Basically the GFF3 loader in some of it's functions sends an array containing values to th chado_query function. This is particularly noted for features. This causes an undesired error.

Example 
$array = [
   'value1',
   'value2'
]

chado_query('SELECT * FROM features WHERE feature_id IN(:array[])', array(
  ':array[]' => $array
));

To avoid this issue at the moment, the code has been converted to create a placeholder for each value instead of using the array[].

spficklin:
Toa clarify, this problem is buried in the `chado_query` function.  With Drupal 9 it is now required, as @risharde points out above, to specify square brackets in placeholders that have arrays.   

See details here:  https://www.drupal.org/docs/drupal-apis/database-api/static-queries#placeholder-arrays

The problem is that in Tripal 3 we added the construct to a query of surrounding Drupal tables with square brackets if we wanted to do a cross schema join.  If you follow Drupal 9 coding style our regular expression for finding brackets now breaks.  Here are the two places where it breaks:

https://github.com/tripal/t4d8/blob/9587d71bd4cdb9bf9b1d83d086da5c0befccf1eb/tripal_chado/src/api/tripal_chado.query.api.inc#L1823-L1825

https://github.com/tripal/t4d8/blob/9587d71bd4cdb9bf9b1d83d086da5c0befccf1eb/tripal_chado/src/api/tripal_chado.query.api.inc#L1866-L1875

So, this isn't a problem just with the GFF3 loader. It's a problem for any SQL statement that is passed to `chado_query` that needs to use a list of values in a placeholder.

dsenalik:
I looked at this issue and to me the fix could be quite simple, simply change the two regular expressions substituting ```+``` for ```*``` and now ```[]``` will not match

Testing:
To test I put this code in the devel module php window
```
$sql = "select * from [bogus] where xyzzy=:somestuff[]";
print ">>>Regex1 Before\n";
preg_match_all('/\[(.*?)\]/', $sql, $matches);
var_dump($matches);
print ">>>Regex1 After\n";
preg_match_all('/\[(.+?)\]/', $sql, $matches);
var_dump($matches);
print ">>>Regex2 Before\n";
preg_match_all('/\[(\w*?)\]/', $sql, $matches);
var_dump($matches);
print ">>>Regex2 After\n";
[...]

PR #1315, opened 2022-10-30
Issue 1272 regex fix for drupal 9 style array placeholders

dsenalik:
# Bug Fix

Issue #1272

## Description
Change regular expressions to not match for adjacent square brackets by substituting ```+``` for ```*```

## Testing?
Show in Issue #1272

## Additional Notes (if any):

One regex uses ```.``` and the other uses ```\w``` - is that significant?

### DIFF
--- a/tripal_chado/api/tripal_chado.query.api.inc
+++ b/tripal_chado/api/tripal_chado.query.api.inc
@@ -1751,7 +1751,7 @@ function chado_query($sql, $args = [], $options = []) {
     }
 
     // Now set the Drupal prefix if the table is surrounded by square brackets.
-    if (preg_match_all('/\[(.*?)\]/', $sql, $matches)) {
+    if (preg_match_all('/\[(.+?)\]/', $sql, $matches)) {
       $matches = $matches[1];
       $drupal_tables = array_unique(array_keys(drupal_get_schema()));
       foreach ($matches as $match) {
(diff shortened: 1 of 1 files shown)


dsenalik:
I realized I did not do a full test, this does not work, but works without the ```[]``` (pasted in to devel module php window)
Error is ```ValueError: PDO::prepare(): Argument #1 ($query) cannot be empty in PDO->prepare() (line 544 of .../includes/database/database.inc).```
```
$param = [14, 15];
$sql = "SELECT * FROM chado.pub WHERE pub_id IN (:param[])";
$args = [':param[]' => $param];
$results = chado_query($sql, $args);
$first = 1;
foreach ($results as $result) {
  if ($first) {
    $first = 0;
    foreach ($result as $key=>$value) {
      print $key . "\t";
     }
   print "\n";
  }
  foreach ($result as $key=>$value) {
    print $value . "\t";
  }
  print "\n";
}
```

dsenalik:
This is because the drupal ```function expandArguments(&$query, &$args)``` around line 816 of ```database/database.inc``` uses a regex that is not compatible with this scheme. Replicating the relevant section in the devel module php window, this is what happens
```
$query = "SELECT * FROM chado.bogus WHERE id IN (:array[])";
$key = ':array[]';
$new_keys = [];
$new_keys[$key . '_' . 0] = 'zero';
$new_keys[$key . '_' . 1] = 'one';
      $query = preg_replace('#' . $key . '\b#', implode(', ', array_keys($new_keys)), $query);
print "$query\n";
```
$query is now an empty string.
It could be fixed there, but since this is in drupal 7 I don't think it is worth fixing. I'll just add the "Address in tripal 4" tag and leave it at that.

PR #1316, opened 2022-10-30
Issue #1271 gff3 loader add organism to landmark query

### LINKED ISSUE #1271: GFF3 bug detected by Stephen
GFF3 loader

Line 2270:
$sql = "SELECT name, uniquename, feature_id FROM {feature} WHERE uniquename in (:landmarks)";
 
Stephen mentioned we need to perform a query that should include the organism ID here

referencing @spficklin

dsenalik:
# Bug Fix
Issue #1271

## Description
Adds organism to the query for landmarks in the gff3 loader, as described in issue #1271

### DIFF
--- a/tripal_chado/includes/TripalImporter/GFF3Importer.inc
+++ b/tripal_chado/includes/TripalImporter/GFF3Importer.inc
@@ -2283,6 +2283,10 @@ class GFF3Importer extends TripalImporter {
           $args = [':landmarks' => $names, ':organism_id' => $organism_id];
           $results = chado_query($sql, $args);
           while ($f = $results->fetchObject()) {
+            if (array_key_exists($f->uniquename, $this->landmarks)) {
+              throw new Exception(t('Duplicate landmark, "!uniquename" is ambiguous for this organism',
+                ['!uniquename' => $f->uniquename]));
+            }
             $this->landmarks[$f->uniquename] = $f->feature_id;
           }
         }
--- a/tripal_chado/includes/TripalImporter/GFF3Importer.inc
+++ b/tripal_chado/includes/TripalImporter/GFF3Importer.inc
@@ -2283,10 +2283,6 @@ class GFF3Importer extends TripalImporter {
           $args = [':landmarks' => $names, ':organism_id' => $organism_id];
           $results = chado_query($sql, $args);
           while ($f = $results->fetchObject()) {
-            if (array_key_exists($f->uniquename, $this->landmarks)) {
-              throw new Exception(t('Duplicate landmark, "!uniquename" is ambiguous for this organism',
-                ['!uniquename' => $f->uniquename]));
-            }
             $this->landmarks[$f->uniquename] = $f->feature_id;
           }
         }
--- a/tripal_chado/includes/TripalImporter/GFF3Importer.inc
+++ b/tripal_chado/includes/TripalImporter/GFF3Importer.inc
@@ -2259,8 +2259,10 @@ class GFF3Importer extends TripalImporter {
 
     $this->setItemsHandled(0);
     $this->setTotalItems($num_batches);
+    $organism_id = $this->organism->getID();
 
-    $sql = "SELECT name, uniquename, feature_id FROM {feature} WHERE uniquename in (:landmarks)";
+    $sql = "SELECT name, uniquename, feature_id FROM {feature} "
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Code review ✅ 
Functional review ✅ 

Looks good. One thing to note is that this still doesn't capture the full unique constraint of uniquename - type - organism so there is still a small chance of being unable to select the exactly correct feature.

One option would be to make the landmark type required and use that but to be honest, my GFF3's often have a mix of types for landmarks as genome assemblies are often a mix of contigs and pseudomolecules. Such a "fix" (requiring specifying the type) would actually break loading for these types of files 😬 We would almost need to have a whitelisted set of types that landmarks can be... which is not trivial to implement and a pain to specify when loading a file.

So yah, I thought I'd bring it up because that was my first thought when seeing this PR but I'm not actually sure we should go the extra step of requiring type for landmarks. What do you think?

dsenalik:
So if I understand, this could happen if for example on genome version 1 we had a feature on a scaffold for example with uniquename X, and genome version 2 has the feature with the same uniquename X in a chromosome. The feature table would allow this.
It does seem complicated to implement, and maybe not worth it if no one has this problem?
Certainly a loader option for the term (terms?) could be left blank for current behavior and not break anything, but how to get the list of valid terms for a select option? What if it is the first time loading a 'chromosome' and it is not in the feature table yet?
```
SELECT DISTINCT F.type_id, T.name FROM chado.feature F LEFT JOIN cvterm T ON F.type_id=T.cvterm_id ORDER BY T.name;
 type_id |            name             
---------+-----------------------------
     535 | CDS
     828 | chromosome
     841 | DNA
     596 | exon
     499 | gene
    1911 | genetic_marker
    1910 | heritable_phenotypic_marker
     403 | linkage_group
   51099 | marker_locus
     425 | mRNA
     394 | ncRNA
     534 | polypeptide
     540 | primer
    1269 | QTL
     486 | transcript
     721 | tRNA
(16 rows)
```

laceysanderson:
> So if I understand, this could happen if for example on genome version 1 we had a feature on a scaffold for example with uniquename X, and genome version 2 has the feature with the same uniquename X in a chromosome. The feature table would allow this.

Yes, that sounds about right. Alternatively you could expand the example like this to be a bit more clear:

> For example, picture two genome assemblies. Version 1 has a scaffold with the uniquename chr1 and then version 2 has a pseudomolecule with the uniquename chr1. This type mismatch can happen with different assembly software or just by judgment call if the second assembly chr1 is substantially bigger then the first version.
>
> Now if you have a GFF3 file with a collection of features located on "chr1", a bug would be reached since the importer wouldn't currently know which one to use.

I'm also not sure it's worth implementing -hence my question in slack to see if anyone had come across that situation by chance. If we were to implement it, I think we would want an optional advanced filter option. Maybe a textarea were the curator can enter as many landmark type names (i.e. cvterm.name) as they want. Then if it's filled in, we make sure the landmarks selected are one of those types? The where clause would be something like `WHERE type_id IN (SELECT cvterm_id FROM {cvterm} WHERE name IN ('chromosome','scaffold','contig')` if they entered the following list in the textarea:
```
chromosome
scaffold
contig
```

dsenalik:
There is already this box on the GFF importer form:
landmarktype
but the code only supports a single term and it would be confusing to use it for two different purposes.

I am looking at the code and what will happen in this case of duplicates is the query is made and one of them will be silently dropped when the SQL results are parsed (line 2296):
```
         while ($f = $results->fetchObject()) {
            $this->landmarks[$f->uniquename] = $f->feature_id;
         }
```
the later instince of any given uniquename will overwrite the first one. So we don't even get an error to alert us.

I think we should at least throw an exception if this happens so we know something went wrong.

laceysanderson:
Yes, we should definitely throw an exception at least 😬 

I also saw the landmark type box you show. It needs to stay limited to one term for when insert needs to happen so I think we need two separate filters. It would take too much thought I think to ensure the two filters with similar function do not cause confusion. I don't think it's worth it right now.

I would be content merging this if you add the exception throwing and we can revisit later if someone runs into the exception and needs a fix.

dsenalik:
LOL it blows up during testing because there are duplicate landmarks now!

dsenalik:
Further work on this problem will be in a new issue #1324 and pull request #1325

PR #1321, opened 2022-11-30
Issue 1320 null pub sql correction for GFF3 loader

### LINKED ISSUE #1320: GFF3Importer possible bug when checking for null Pub
I cannot explain unfortunately how to get this bug to show up since I'm actually testing the T4 version which uses most of the code from T3 but a quick inspection should show the issue

https://github.com/tripal/tripal/blob/48e2affdfc8f267018677b28d095f4da25415f74/tripal_chado/includes/TripalImporter/GFF3Importer.inc#L834

On this line, there seems to be no options/arguments for :name and :type_id so this chado_query function would error out

dsenalik:
# Bug Fix

Issue #1320

## Description
Fix SQL for detecting | inserting and returning the null publication.

## Testing?
See Issue #1320, you need to delete the existing null publication to fully test the insertion.

### DIFF
--- a/tripal_chado/includes/TripalImporter/GFF3Importer.inc
+++ b/tripal_chado/includes/TripalImporter/GFF3Importer.inc
@@ -822,26 +822,22 @@ class GFF3Importer extends TripalImporter {
     $select = ['uniquename' => 'null'];
     $result = chado_select_record('pub', ['*'], $select);
     if (count($result) == 0) {
+      // Insert the null pub.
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅  Code review -look good!
✅ Functional review -deleted null pub and then inserted a GFF3 without error, checked after and null pub was replaced

Great fix -worked perfectly. Thank you @dsenalik this is ready to merge!

Issue #1324, opened 2022-12-01
GFF3 Loader landmark lookup needs to check type_id


dsenalik:
### Issue description
In issue #1271 and pull request #1316 we have discovered that the landmark lookup may select the wrong feature if two features share the same `uniquename` and same `organism_id`. The other column of the unique constraint `type_id` is not checked.

#### Steps to reproduce

Currently this does not trigger any kind of error, the first step will be to throw an exception and then work from there. ~~This by itself is enough to get testing to fail.~~

#### Example showing when this is a problem:
Picture two genome assemblies for a single organism. Version 1 has a scaffold with the uniquename chr1 and then version 2 has a pseudomolecule with the uniquename chr1. This type mismatch can happen with different assembly software or just by judgment call if the second assembly chr1 is substantially bigger then the first version. If you imported these two genomes into chado you would end up with two records in the feature table with a uniquename of "chr1" and the same organism. This is allowed because the type_id is also included in the unique constraint.

Now if you have a GFF3 file with a collection of features located on "chr1", a bug would be reached since the importer wouldn't currently know which one to use. It seems like the current code will just throw away one of these "chr1" features and basically just pick one to use without throwing an error.

laceysanderson:
Work has been started on this in #~~1325~~ #1369 with a check throwing an exception. It currently breaks the automated tests showing this is definitely a problem that has gone unnoticed for awhile.

dsenalik:
The problem with automated tests was that I originally implemented the check incorrectly, it always failed.

No matter, it is all fixed now and ready for review.

spficklin:
Hi @dsenalik thanks for trying to tackle this issue, and for posting PR #1369.  Because I've been a bit negligent of the PR queue lately I hadn't noticed what you've done there.  @risharde and I were working on this very same issue in the Tripal v4 GFF3 loader.  So, I'm tagging him here.   This is the solution we came up with for the Tripal v4 loader.


Problems:
----------
1.  The findLandmarks() function has an SQL statement for finding a landmark
    that doesn't include the organism and the landmark type so this may cause
    an accidentally selection of an incorrect landmark, such as on for a 
    completely different species. 
2.  Landmark_type specified in the form currently applies to all landmarks but
    what if a GFF file has different types of landmarks:  chromosome, scaffold,
    contig, etc. Then the landmark_type will incorrectly give all them the same
    type. 
3.  When the landmark type is defined in the GFF file then the loader doesn't
[...]

PR #1325, opened 2022-12-01
Issue #1324 GFF3 Loader, handle duplicate landmark uniquename

dsenalik:
# Bug Fix

## Oops wrong branch, see #1369 instead

Issue #1324 

## Description

The issue is that the landmark lookup does not use all three unique constraints of the feature table, only ```uniquename``` and ```organism_id``` are used. Two features may share these and only differ by ```type_id```.

The first commit just adds an exception when we encounter a duplicate landmark uniquename for a given organism. This fails testing because several tests have duplicates.


dsenalik:
Here is a commit just to handle this PHP8 warning in Step 1 of 27:
```Undefined variable $cvterm GFF3Importer.inc:757```
This is because the code is trying to cache by the passed ```$type``` which might be a synonym, so needs the "true" cvterm name, but neglected to save this.

dsenalik:
Closed because I gave the branch the wrong name (1234 should have been 1324)
See new pull request #1369

Issue #1327, opened 2022-12-06
tripal_chado api function to look up organism from organism name


dsenalik:
## Feature Request/Discussion
* This feature does attempt to solve an existing problem with Tripal
* I will do it.
* This feature is Not Urgent

### Description
in ```tripal_chado/api/modules/tripal_chado.phylotree.api.inc``` there are api functions for working with organisms.
```function chado_get_organism_scientific_name()``` looks up an organism's name, I propose an inverse function to look up the organism given this scientific name.

Why? Infraspecific nomenclature! It is hard to do this in straight SQL because the infraspecific type in the organism table is stored as the id of the corresponding cv term, and plus this type may or may not be abbreviated ```subspecies``` or ```subsp.```
There is an api function for dealing with the abbreviation ```chado_unabbreviate_infraspecific_rank()```.

This could be a new api function, but it may be cleaner to incorporate into the existing function ```chado_get_organism()``` which currently supports genus+species but for infraspecific we need a type_id value.
This could be done by adding a new 'scientific_name' option and have special code to take care of the parsing. But I might expose that as a new function so both ways work, let's see...

### Your Specific Use Case
The ```sbo__relationship``` tripal field currently only supports genus+species lookup from the full organism name, and only suporrts infraspecific nomenclature through its ajax callback. With this new api function it would be relatively easy to implement full infraspecific nomenclature support.
This will also be used by the Newick importer, see https://github.com/tripal/tripal/pull/1262#issuecomment-1341547683

### Generally Applicable
This could also be useful to incorporate into data loaders.

laceysanderson:
I agree that this is a worthwhile addition!

We are finding both in core and in extension modules that support for the infraspecific name is not very good at all. This is likely since many module developers do not have that test case in their test databases since it was just added in Chado 1.3.

Having a easy to use function to get the organism_id and/or object based off the full scientific name (genus + species + infra type/name) and making it best practice to use that instead of adding the genus + species hack into our queries would make it a lot easier to ensure this support is consistent and wide-spread.

I like the idea of adding it to `chado_get_organism()` for sure! That said, often we only need the organism_id and `chado_get_organism()` does cause a performance hit so an additional specific function would be useful too.

Many of my extension modules need custom queries to make large datasets performant so a SQL function would be ideal. That way it can be incorporated into any query. I have some experience with this so perhaps we can work on that part together @dsenalik?

dsenalik:
[...]

PR #1328, opened 2022-12-06
Implement chado_get_organism_id_from_scientific_name

dsenalik:
# New Feature

Issue #1327 

## Description

This adds
1. A new api function ```chado_get_organism_id_from_scientific_name()``` which will look up and return an ```organism_id``` from a taxon name _e.g._ ```Tripalus domesticus subsp. chadoii```
2. A new identifier that can be passed to ```chado_get_organism()``` of ```scientific_name```, which allows the return of an organism given a taxon name (as in 1)
 

## Testing?
Since so far nothing is using this, test in the devel module php window
```
$s = 'Tripalus domesticus subsp. chadoii';  // A fake or a real organism name
$o = chado_get_organism(['scientific_name' => $s], []);
var_dump($o);
```

## Additional Notes:

This will first be used for the ```sbo__relationship``` widget on organism pages, see https://github.com/tripal/tripal/pull/1267#issuecomment-1339482302

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.organism.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.organism.api.inc
@@ -82,7 +82,7 @@ function chado_get_organism($identifiers, $options = []) {
   if (isset($identifiers['scientific_name'])) {
     $scientific_name = $identifiers['scientific_name'];
     unset($identifiers['scientific_name']);
-    $organism_ids = chado_get_organism_id_from_scientific_name($scientific_name, ['scientific_name']);
+    $organism_ids = chado_get_organism_id_from_scientific_name($scientific_name, $options);
     if (count($organism_ids) == 1) {
       $identifiers['organism_id'] = $organism_ids[0];
     }
@@ -212,10 +212,15 @@ function chado_get_organism_scientific_name($organism) {
  *   The organism name to be queried. Infraspecific type can be abbreviated.
  *
  * @param $options
- *   Specifies how to search organism table, one or more of the following can be used:
- *   $options = ['scientific_name', 'abbreviation', 'common_name', 'case_sensitive']
- *   Search is case insensitive unless the 'case_sensitive' value is specified.
- *   If not specified, search defaults to scientific_name only, case insensitive.
+ *   An array of options. The following keys are available:
+ *     - check_abbreviation: If TRUE and the $name did not match the
+ *         scientific name, then check the abbreviation.
+ *     - check_common_name: If TRUE and the $name did not match the
+ *         scientific name, then check the common name.
+ *     - case_sensitive: If TRUE then all searches should be case
+ *         sensitive. Default is FALSE.
+ *   If no options are specified, search is for a match of $name to
+ *     the scientific_name only, case insensitive.
  *
  * @return
  *   Array of matching organism_id values.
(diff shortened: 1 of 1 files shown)


laceysanderson (commented):
Thanks @dsenalik!

The logic flow and options in the new chado_get_organism_id_from_scientific_name() function could use some improvements to be more consistent with other API functions in Tripal. Specifically,

- Since the name specifies scientific name, that should be always checked first and if a match found then nothing else should be checked.
- The options should be changed as follows:
```php
/*
 * @param $options
 *     An array of options. The following keys are available:
 *        -  check_abbreviation: If TRUE and the $name did not match the scientific name then check the abbreviation.
 *        - check_common_name:  If TRUE and the $name did not match the scientific name then check the common name.
 *        - case_sensitive:  If TRUE then all searches should be case sensitive. Default is FALSE.
 */
```

dsenalik:
I made the suggested changes, thanks for the feedback, much appreciated.

A few of the tests I ran, but these depend on the organisms in the database
```
$s = 'Daucus carota subsp. sativus';
$o = chado_get_organism_id_from_scientific_name($s, []);
print 'default: '.count($o)."\n";

$s = 'D. carota';
$o = chado_get_organism_id_from_scientific_name($s, ['check_abbreviation']);
print 'abbreviation: '.count($o)."\n";

$s = 'carrot';
$o = chado_get_organism_id_from_scientific_name($s, ['check_common_name']);
print 'common name case insensitive: '.count($o)."\n";

$s = 'carrot';
$o = chado_get_organism_id_from_scientific_name($s, ['check_common_name', 'case_sensitive']);
print 'common name case sensitive: '.count($o)."\n";

$s = 'daucus carota subsp. sativus';
$o = chado_get_organism(['scientific_name' => $s], []);
print 'chado_get_organism default: '.($o?'found':'null')."\n";

$s = 'daucus carota subsp. sativus';
$o = chado_get_organism(['scientific_name' => $s], ['case_sensitive']);
print 'chado_get_organism case_sensitive: '.($o?'found':'null')."\n";

$s = 'carrot';
$o = chado_get_organism(['scientific_name' => $s], ['check_common_name']);
print 'chado_get_organism common_name: '.($o?'found':'null')."\n";
```
Results
```
default: 1
abbreviation: 1
common name case insensitive: 1
common name case sensitive: 0
chado_get_organism default: found
chado_get_organism case_sensitive: null
chado_get_organism common_name: found
```

dsenalik:
~~The Chado 1.2 way of handling infraspecific nomenclature was to put it all in the 'species' column. Should I add code to support this, what do people think?
For example genus = 'Daucus' species = 'carota subsp. sativus'~~
Edit: Never mind, it seems I forgot how the code works, this will be handled automatically.

spficklin (approved):

spficklin:
The code changes look fine.  @dsenalik responded to the requests that @laceysanderson made and I tested it with the following:

1. Added a new organism to to match that @dsenalik provided as an example.  
2. Ran the following drush command:
```bash
drush php-eval "\$s = 'Tripalus domesticus subsp. chadoii'; \$o = chado_get_organism(['scientific_name' => \$s], []); var_dump(\$o);"
```    
It worked as behaved.

Issue #1364, opened 2022-12-14
PHP8 Error in ogi__location_on_map field


dsenalik:
## Pull #1365 

* Tripal Version: 7.x-3.9
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: **8.0.23**

### Issue description

Another PHP8 error, the message is
```Warning: Undefined property: stdClass::$description in ogi__location_on_map->load() (line 310 of .../sites/all/modules/tripal/tripal_chado/includes/TripalFields/ogi__location_on_map/ogi__location_on_map.inc).```

### Testing

I appear to have never finished adding descriptions to all of my linkage maps. So if you have this case of a map without a description, the message above will appear in the log for Gene, mRNA pages. Fix is super simple.

laceysanderson:
Fixed -Thanks!

PR #1365, opened 2022-12-14
Issue 1364 check for description property

dsenalik:
# Bug Fix
Issue #1364

## Description

PHP8 warning from the ```ogi__location_on_map``` field if you don't have a description for your linkage map, as described in Issue #1364 
Super simple fix.


laceysanderson (approved):
Super simple fix approved and appreciated! <3

Issue #1366, opened 2022-12-14
bug in addGeneParent() and addProtein()


dsenalik:
### System information

* Tripal Version: 7.x-3.9
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: 8.0.23

### Issue description

Although this is a PHP8 warning, there is a bug causing it. The error message observed from a Gene page in this case
```Warning: Attempt to read property "name" on string in chado_get_fasta_defline() (line 783 of .../sites/all/modules/tripal/tripal_chado/api/modules/tripal_chado.feature.api.inc).```

The cause is that in
```tripal_chado/includes/TripalFields/data__sequence_record/data__sequence_record.inc```
in function ```addGeneParent()``` line 372 ```$fasta_defline => chado_get_fasta_defline($gene)```
```$gene``` is the result of a SQL query, not quite the object as expected by ```chado_get_fasta_defline()``` and so there is no name property, just the cvterm_id as a string.

A similar thing happens in function ```addProtein()``` at line 429 ```$fasta_defline => chado_get_fasta_defline($protein)```

See pull #1367 for how to fix this.

dsenalik:
Fixed with merged #1367

PR #1367, opened 2022-12-14
Issue 1366 addGeneParent addProtein

dsenalik:
# Bug Fix

Issue #1366 

## Description

This is a bug fix for Issue #1366, the problem is described there.
This fix is very simple, we can pass the missing type name as a parameter to ```chado_get_fasta_defline()```.

## Testing?
This is a warning that shows up only under PHP8, it shows up in the log viewing e.g. a Gene page.


laceysanderson (approved):
Code review looks perfect!

Confirmed that the new parameters are correct by looking at the function: https://github.com/tripal/tripal/blob/5a915c2c8ce1a83b032c5a86093a731b3275305d/tripal_chado/api/modules/tripal_chado.feature.api.inc#L739-L765

laceysanderson:
@4ctrl-alt-del this might impact your Tripal 4 field. Just bringing it to your attention in case :-)

PR #1369, opened 2022-12-20
GFF3 Importer: duplicate landmark uniquename bug

### LINKED ISSUE #1324: GFF3 Loader landmark lookup needs to check type_id

dsenalik:
# Bug Fix

## Description

The issue is that the landmark lookup for the GFF3 loader does not use all three unique constraints of the feature table, only ```uniquename``` and ```organism_id``` are used. Two features may share these and only differ by ```type_id```.

The first commit https://github.com/tripal/tripal/pull/1369/commits/895612170a230c3ba84ee7ddca61f401516b0e2f just adds an exception when we encounter a duplicate landmark uniquename for a given organism. This fails testing because ~~several tests have duplicates~~ it was implemented incorrectly.

## Testing?
1, 2. Load the same fasta file twice using the same organism but a different cv term each time, _e.g._ ```DNA``` and ```chromosome```. This will generate the duplicate landmark uniquenames.
3. Load a gff3 file, do not specify the "Landmark Type". On this branch it should fail with an error message like
```The landmark "XYZ_Chr1" is ambiguous for this organism. Please redo the importer job and specify a landmark type.```
4. Run it again specifying one of the two landmark types that you used. The file should load successfully.

### CODE REVIEW

**Comment by dsenalik** [comment]
commit https://github.com/tripal/tripal/pull/1369/commits/9eb3e5e4fede467b7a6d78953def8069e8977f1b is just to handle this PHP8 warning in Step 1 of 27:
```Undefined variable $cvterm GFF3Importer.inc:757```
This is because the code is trying to cache by the passed $type which might be a synonym, so needs the "true" cvterm name, but neglected to save this.

**Comment by dsenalik** [comment]
commit https://github.com/tripal/tripal/pull/1369/commits/6fb0c3e14433db85412bb9df959d894d5a21664b is just to handle this PHP8 warning in Step 5 of 27:
```Undefined array key "organism_id" GFF3Importer.inc:1662```

**Comment by dsenalik** [comment]
commit https://github.com/tripal/tripal/pull/1369/commits/9244688751311ad746384e15a68db03cf3fd3d50 is just to handle these PHP8 warnings in Step 26 of 27:
```
Undefined variable $sql GFF3Importer.inc:3013
Undefined array key "score" GFF3Importer.inc:3016
Undefined array key "score" GFF3Importer.inc:3017
```

**Comment by dsenalik** [comment]
Commit https://github.com/tripal/tripal/pull/1369/commits/507740e44c12d13c84e7eff63b5c3587c1e29afc implements the type_id to the landmark lookup, which is the purpose of this pull request. Let's see what happens with testing...

**Comment by dsenalik** [comment]
So to explain, my first commit https://github.com/tripal/tripal/commit/895612170a230c3ba84ee7ddca61f401516b0e2f implemented the check incorrectly, it would always fail. With this now corrected, testing does not trigger any errors.

**Comment by spficklin** [comment]
@dsenalik is this PR still relevant given the discussion on this issue https://github.com/tripal/tripal/issues/1324

**Comment by dsenalik** [comment]
@spficklin yes it is relevant - this is the tripal 3 version with the described issues fixed, and a new test included to test.

**Comment by spficklin** [comment]
Hey @dsenalik looking over the code it looks like we have some divergent solutions to the same problem between the Tripal v3 and Tripal v4 version of the GFF loader.  I think to decrease time it take to fix bugs in the future we should do our best to make sure the architecture (bones) of the loaders are as similar as possible.  

Would you be willing to adapt the code here to match what is in this PR #1459 .  @risharde indicated he would be happy to help with it or do the work if you want.  You've done a lot of work here and we wanted to know your preference.

**Comment by dsenalik** [comment]
Okay I am making the simple changes so there is less to dig through when doing a diff.

List of changes on the T4 side to make:
1. Cases like this where I changed messages to use substitution variables, e.g.
```
t3<       throw new Exception(t("Cannot open file: !file", ['!file' => $this->gff_file]));
t4>       throw new Exception(t("Cannot open file: " . $this->gff_file));
```
if you want me to fix all of those on the t4 branch let me know

2. code at t3 1632,1642 - this implemented a fix if there was more than one parent (that's in the title of #1305 😁)
```
1632,1642c1787,1796
<         foreach (explode(',', $feature['parent']) as $parent_name) {
<           if ($parent_name) {
<             if (!array_key_exists($parent_name, $this->proteins)) {
<               $this->proteins[$parent_name] = [];
<             }
<             if ($type == 'cds') {
<               $this->proteins[$parent_name]['cds'][] = $findex;
<             }
<             if ($type == 'protein' or $type == 'polypeptide') {
<               $this->proteins[$parent_name]['protein'] = $findex;
<             }
---
>         $parent_name = $feature['parent'];
>         if ($parent_name) {
>           if (!array_key_exists($parent_name, $this->proteins)) {
>             $this->proteins[$parent_name] = [];
>           }
>           if ($type == 'cds') {
>             $this->proteins[$parent_name]['cds'][] = $findex;
>           }
>           if ($type == 'protein' or $type == 'polypeptide') {
>             $this->proteins[$parent_name]['protein'] = $findex;
```

3. the organism list lookup should use the api function but is that ported yet? I may have to make an issue for that. #1483
```
325c345,351
<     $organisms = chado_get_organism_select_options(FALSE, TRUE);
---
>     $sql = "SELECT * FROM {organism} ORDER BY genus, species";
>     $org_rset = $chado->query($sql);
>     $organisms = [];
>     while ($organism = $org_rset->fetchObject()) {
>       $organisms[$organism->organism_id] = chado_get_organism_scientific_name($organism);
>     }
```

4. typos to fix:
provdied
is_obsolute

Otherwise @risharde please do what needs to be done to the T3 version please

spficklin:
Thanks @dsenalik I think we don't have to make them align perfectly. Just the solution to the landmark problems that Rish fixed and that you were working on too.  That was the big fix that you both were working on. The solution I posted on issue #1324 was what Rish implemented in Tripal v4.  It uses diffrent member variables and I think addresses a few more issues with landmarks than the origin issue pointed out.

laceysanderson:
Me + @dsenalik discussed this in the cofest today. The intricacies of the difference between the two solutions makes @spficklin's request a bit harder then we had hoped. As such and since there is a T3 release coming soon, I am going to do a functional test on this one and code review so that it can be merged.

Then afterwards, Doug has said he will add any small details from his solution missed in Rish's solution into an independent PR.

@risharde: would you be willing to look at the Tripal 3 GFF3 loader and and make a new PR that brings this solution in to match this important parts of yours? If there is some divergence that is conflicting (as compared to subtle differences where one or the other solution can be easily seen as better in that particular change), then lets all meet in next week's cofest and get it sorted out as a group.

laceysanderson (approved):
✅ My manual functional testing worked as expected 🎉 
✅ No major problems jumped out to me in code review.

I'm going to merge but something we should make sure is present in the 4.x version is an autocomplete for the landmark type.

risharde:
@laceysanderson yes, I'm attempting to do this (trying to port the changes I made wrt landmarks into the Tripal 3 GFF loader as a separate PR for review). Will report back once I have something working!

risharde:
@laceysanderson @dsenalik I tried to get this out before the meeting I think you guys were planning tomorrow. I ported the changes from TV4 GFF3 to TV3 in this branch: https://github.com/tripal/tripal/tree/1369-tv3-gff3-landmark-fixes-from-tv4-into-tv3 (and ran unit tests + adjustments)

@dsenalik I apologize but I commented out some of the changes you made in your implementation. There's a piece of code that is recording the seqlen and I commented that out as well - however if this is still applicable, please feel free to try to make it work on this branch. I have no problem if you work directly on the branch per say. I'll try to do pulls so I stay up to date with your changes. If you can get the seqlens code working again, I'll be happy to port your adjustments to Tripal 4's version as well. 

In the process, I fixed the tests so they seem to pass on my side for the moment. I also added seq ontology imported since it needed that to work with the default landmark type where applicable. 

Please feel free to test accordingly!

dsenalik:
Thanks so much for doing this @risharde it is much appreciated.

Issue #1370, opened 2022-12-22
PHP8 warning due to overzealous bundle_instances_info_alter functions


dsenalik:
### System information

* Tripal Version: 7.x-3.10dev
* Drupal Version: 7.92
* PostgreSQL Version: 13.3
* PHP Version: **8.0.26**

### Issue description
Under PHP8, I see this warning multiple times at the end of running the GFF3 loader
```
Undefined array key "field_name" tripal.entities.api.inc:683                                  [warning]
Undefined array key "field_name" tripal.entities.api.inc:691                                  [warning]
Undefined array key "field_name" tripal.entities.api.inc:693                                  [warning]
Undefined array key "field_name" tripal.entities.api.inc:695                                  [warning]
```

I will submit a pull request to fix it, but the reason it happens is that some modules implement the ```bundle_instances_info_alter()``` hook, and add in information in cases where the passed ```$info``` variable for the hook is empty, thereby creating new but incomplete records. ```tripal_file```, I'm looking at you. I see I patched another module ```tripal_analysis_blast``` sometime in the past. That should be fixed in each offending module by doing nothing if ```$info``` is empty, but we can also fix it in tripal in ```tripal.entities.api.inc``` by checking that we have the required information, and if not just skip over it. This all happens in ```tripal_tripal_cron_notification()``` which is where we are checking for any new fields that have not been added to content types yet.

#### Steps to reproduce
Run under PHP8, load a GFF3 file, and run the job with drush so you can see the output.

dsenalik:
Fixed with merged #1371

PR #1371, opened 2022-12-22
Issue 1370 skip if missing field_name

dsenalik:
# Bug Fix

Issue #1370

## Description

The problem is described in #1370
The fix here is a simple check that the passed ```$details``` has the required information, and if not, do nothing.


laceysanderson (approved):
Looks like a sound, simple fix :-) I don't have an easy way to test this but I think a code review is sufficient. Thanks @dsenalik

dsenalik:
Thanks @laceysanderson yes the fix is super simple so with your approval, I went ahead and merged

PR #1388, opened 2023-01-16
Tripal 4 Field: Sequence Residues

laceysanderson:
# Tripal 4 Core Dev Task

This PR is being resubmitted on behalf of @srobb1. She worked with @spficklin on this functionality during the PAG 2023 Cofest. It was originally submitted as https://github.com/tripal/t4d8/pull/333.

Issue #1414

## Description
This PR implements a specialized field and formatter for the chado feature.residues column.

## Known Issues

- The widget does not validate that the value entered is actually sequence. Chado doesn't either but it would be good to add this -maybe at a later date?
- There are no automated tests.

## Testing?

**Note: There are currently no automated tests for this PR since we have not determined a good testing scheme for Tripal Fields yet. We have decided we are okay to merge fields without tests for now.**

### Manual Testing

**Do all the following with this branch checked out.**

1. Create an organism for your testing site by going to Admin > Tripal > Content.

**Add the field**
2. Go to Admin > Tripal > Page Structure > Gene > Manage Fields
3. Delete the existing "Sequence" field by clicking the arrow beside edit and selecting "Delete". This is optional but does make it easier to ensure you are testing the field you think you are ;-p
4. Click "Add field" and then select "Chado Sequence Residues". If that is not an option then rebuild the cache on your site. Label it whatever you like.
5. Leave all the field storage settings the same and click "Save field settings".
6. Leave all the field settings the same and click "Save settings" to create the field.

**Create a Gene**
7. Go to Admin > Tripal > Content. Click "Add Tripal content" and then "Gene".
8. Choose and organism and enter an identifier. 

**Test Sequence field as you see fit.**
**I suggest the following test cases:**
- Create gene with no sequence
- Edit gene with no sequence to add one
- Create gene with sequence
- Edit gene with sequence and change it to something else
- Delete gene


laceysanderson (commented):
I reviewed the code in this field and it looks good ✅ 

I made the following small improvements:
- simplify the field name slightly to so it would be more descriptive (i.e. Chado Sequence Residues)
- ensured that the markup is only added when there are residues
- updated the display slightly as shown in the picture below.

Screen Shot 2023-03-06 at 5 15 08 PM

One thing to note: there is no validation that valid sequence is added in the residues field...

spficklin:
The field looks good. It works as described. It really needs some sequence validation. I will add an issue to remind us to do that, but will merge so we get our PR queue under better control.

PR #1396, opened 2023-01-24
Use new api function chado_get_organism_id_from_scientific_name

dsenalik:
# Bug Fix

Issue #1266 

## Description
This change implements full and complete infraspecific nomenclature support in the ```sbo__relationship``` widget, as mentioned in pull #1267 in this comment https://github.com/tripal/tripal/pull/1267#issuecomment-1341274120.

This change uses the new api function ```chado_get_organism_id_from_scientific_name()``` to handle the difficult task of looking up an organism from its scientific name, which is handled as a special case when the base table is ```organism```. This new api function is from pull #1328, which as of submission has not been merged yet, so this pull request is dependent on that one being merged. *it has been merged*

## Testing?
Testing is basically entering relationships for one organism to another, using real names, or incorrect ones, and observing the result.
Details of testing in this comment https://github.com/tripal/tripal/pull/1396#issuecomment-1428809144

### DIFF
--- a/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_widget.inc
+++ b/tripal_chado/includes/TripalFields/sbo__relationship/sbo__relationship_widget.inc
@@ -421,14 +421,22 @@ class sbo__relationship_widget extends ChadoFieldWidget {
         }
         // Otherwise we need to look it up using the name field determined in the
         // constructor for the current field. There may be more than one name field
-        // (e.g. organism: genus + species) so we want to check both.
+        // so we want to check all. Special case for organism where lookup is complicated.
         else {
-          $sql = 'SELECT ' . $fkey_rcolumn . ' FROM {' . $base_table . '} WHERE ' . implode("||' '||", $this->base_name_columns) . '=:keyword';
-          $subject = chado_query($sql, [':keyword' => $subject_name])->fetchAll();
-          if (count($subject) == 1) {
-            $subject_id = $subject[0]->$fkey_rcolumn;
+          if ($base_table == 'organism') {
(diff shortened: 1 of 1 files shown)


dsenalik:
I consider this finished and ready for review
### Testing
**1**  Have or create three organisms with infraspecific nomenclature, so much the better if they have the same infraspecies:
_Daucus carota_ subspecies _chadoii_
_Daucus carota_ varietas _chadoii_
_Daucus carota_ cultivar _chadoii_
**2**  Make sure sbo__relationship is present on organism pages, if necessary add at ```/admin/structure/bio_data```
**3** Go to the page of organism 1 and edit.
**4**  Any cv term could be used, let's use ```stock_relationship``` CV present in base tripal 3, and term ```is_a_selection_of```
**5a**  Leave left side blank, right side start typing and select the second organism, and save. This tests the normal workflow.
**5b**  Leave left side blank, right side start typing and select the third organism, remove the ``` [id: 123]``` and save. This tests the call to ```chado_get_organism_id_from_scientific_name()```, but should still work. Error before this pull request is ``` The object does not exist```.
**5c** Leave left side blank, right side enter an organism name that does not exist. This tests the error reporting. Expect: ```The object does not exist```
**6a-c** Do steps 5a-c but on the other side.
**7** Leave both sides blank to test for that error checking. Expect: ```Please provide the subject or object of the relationship.```
**8** Enter same organism on both sides to test for that error checking. Expect: ```The subject and the object in the relationship cannot both refer to the same record.```
**9** Enter two different organisms, but neither is the current entity, to test for that error checking. Expect: ```Either the subject or the object in the relationship must refer to this record.```

spficklin (approved):

spficklin:
Tested and works as described!  Code changes look good. Thanks @dsenalik

Issue #1401, opened 2023-01-31
tv3g3 PHP8 error with "Bulk update all titles"


dsenalik:
# BUG/ERROR report

## System information

* Tripal Version: 3.9dev
* Drupal Version: 7.94
* PostgreSQL Version: 13.3
* PHP Version: 8.0.27

### Issue description
What is happening is we are passing a job to a callback that does not expect it.
In file ```tripal/includes/TripalJob.inc``` function ```run()``` on this line
```
call_user_func_array($callback, $arguments);
```
the callback is ```tripal_update_all_urls_and_titles``` from file ```tripal/includes/tripal.bulk_update.inc```
which is ```function tripal_update_all_urls_and_titles($bundle_id, $update, $type) {```
and the arguments are
```
arguments="array(4) {
  ["bundle_id"]=>
  string(10) "bio_data_1"
  ["update"]=>
  string(21) "[local__abbreviation]"
  ["type"]=>
  string(5) "title"
  [0]=>
  object(TripalJob)#450 (8) {
    ["job_id":protected]=>
    NULL
    ["job":protected]=>
    object(stdClass)#451 (20) {
etc.
```
the fourth array element is ```$this```, appended at line 336 ```$arguments[] = $this;``` where, I think, we have checked for tripal v2 vs. v3, this is the v3 version.

#### Steps to reproduce
1. Need to run under PHP8.x
2. Go to organism content type /admin/structure/bio_data/manage/bio_data_1
3. Click the "Bulk update all titles" button
4. run the job with drush at the command line to see this output
```
2023-01-31 09:16:15: There are 1 jobs queued.
2023-01-31 09:16:15: Job ID 3953.
2023-01-31 09:16:15: Calling: tripal_update_all_urls_and_titles(bio_data_1, [local__abbreviation], title)
WD php: Error: Cannot use positional argument after named argument in TripalJob->run() (line 344 of          [error]
.../sites/all/modules/tripal/tripal/includes/TripalJob.inc).
Cannot modify header information - headers already sent by (output started at                                [warning]
.../sites/all/modules/tripal/tripal/api/tripal.jobs.api.inc:386)
bootstrap.inc:1551
Error: Cannot use positional argument after named argument in TripalJob->run() (line 344 of .../sites/all/modules/tripal/tripal/includes/TripalJob.inc).
Drush command terminated abnormally due to an unrecoverable error.                                           [error]
```

dsenalik:
[...]

PR #1402, opened 2023-01-31
tv3g3 php8 errors with Bulk update all titles

dsenalik:
# Bug Fix

## Description
The changes resolve warnings that show up under php8.x, where a job is passed to a callback that does not expect an additional (unnamed) parameter

## Testing?
1. Need to run under PHP8.x
2. Go to organism content type /admin/structure/bio_data/manage/bio_data_1 (or any other content type)
3. Click the "Bulk update all titles" button
4. run the job with drush at the command line and hopefully no errors now appear, see issue #1401 for the "before" output.
```
2023-01-31 13:38:56: There are 1 jobs queued.
2023-01-31 13:38:56: Job ID 3977.
2023-01-31 13:38:56: Calling: tripal_update_all_urls_and_titles(bio_data_4, study/[TripalEntity__entity_id], alias)
1/1 entities have been updated. Memory usage: 52,179,464 bytes.
Done.
```

### DIFF
--- a/tripal/includes/TripalJob.inc
+++ b/tripal/includes/TripalJob.inc
@@ -332,7 +332,7 @@ class TripalJob {
         if ($lastparam->getName() == 'job_id') {
           $arguments[] = $this->job->job_id;
         }
-        else {
+        elseif ($lastparam->getName() == 'job') {
           $arguments['job'] = $this;
         }
       }
--- a/tripal/includes/TripalJob.inc
+++ b/tripal/includes/TripalJob.inc
@@ -330,7 +330,7 @@ class TripalJob {
       if (count($refparams) > 0) {
         $lastparam = $refparams[count($refparams) - 1];
         if ($lastparam->getName() == 'job_id') {
-          $arguments[] = $this->job->job_id;
+          $arguments['job_id'] = $this->job->job_id;
         }
         elseif ($lastparam->getName() == 'job') {
           $arguments['job'] = $this;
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
Code-wise, this looks great 👍 Thanks @dsenalik 

Note: I don't have an environment up to quickly test this so I have done a code review only.

dsenalik:
This has a side effect, other callbacks will need to be updated
```
WD php: Error: Unknown named parameter $job in TripalJob->run() (line 342 of   [error]
.../sites/all/modules/tripal/tripal/includes/TripalJob.inc).
```
in this case the callback="chado_populate_mview"
and args=
```
array(2) {
  [0]=>
  string(2) "25"
  ["job"]=>
  object(TripalJob)#446 (8) {
    ["job_id":protected]=>
    NULL
    ["job":protected]=>
    object(stdClass)#451 (20) {
... etc.
```

dsenalik:
If I revert ```$arguments['job'] = $this;``` back to the original ```$arguments[] = $this;``` then this error goes away.

dsenalik:
Because there is no way I could ever know all of the functions that may be callbacks, and I assume other modules could use this too, I really can't change all callback parameters, it is best to do no harm by adding new stuff.
The only functional change is to now only pass the job object to the callback when the last parameter of the callback is ```$job```.
Because of this I removed the additional ```$job``` parameter to ```tripal_update_all_urls_and_titles()``` that I had added, because this change _should_ ensure that it is no longer passed to this callback or any callback not expecting it.

laceysanderson:
I agree with the new approach @dsenalik and your implementation of it looks good. I dismissed my code review from above and and we should probably wait for a functional review on this one. Thanks for being extra diligent -as always ❤️

spficklin (approved):

spficklin:
Thankd @dsenalik

Issue #1404, opened 2023-02-01
Improving maintainability as suggested by CodeClimate


laceysanderson:
# Core Tripal 4 Development Task

## Branch Name: tv4g0-issue1404-improveMaintainability

![Screen Shot 2023-02-01 at 11 08 44 AM](https://user-images.githubusercontent.com/1566301/216112638-812f1cfa-eb17-4c3a-ab21-aecaad9f4dcd.png)

According to Code Climate, we have 174 potential maintainability issues. The breakdown is as follows:

Severity
- 2 Critical
- 17 Major
- 155 Minor

Also, 160 are complexity issues and 14 are suggested duplication.

laceysanderson:
## Progress!
![image](https://user-images.githubusercontent.com/1566301/218167260-b66409d2-ffa5-44e7-81bf-f9d0d97b3129.png)

There are only 32 issues left to conquer! They are described more below.

### Too many return statements

- tripal.upload.api.php tripal_file_upload() has 8 return statements
- tripal_chado/src/api/ChadoSchema.php checkColumnType() has 7 return statements

### Cognitive Complexity

| Class                                | Method                      | Complexity Score |
|--------------------------------------|-----------------------------|------------------|
| ChadoUpgrader                        | prepareUpgradeTables        | 124              |
|                                      | processUpgradeQueries       | 34               |
| ChadoStorage                         | buildChadoRecords           | 107              |
|                                      | setPropValues               | 73               |
|                                      | validateTypes               | 52               |
|                                      | validateUnique              | 40               |
|                                      | insertValues                | 35               |
|                                      | updateValues                | 34               |
|                                      | setRecordIds                | 27               |
| GFF3Importer                         | parseGFF3Line               | 81               |
|                                      | findFeatures                | 36               |
|                                      | findSynonyms                | 26               |
| TripalDbx                            | parseTableDdlToDrupal       | 67               |
|                                      | parseTableDdl               | 38               |
| PersistentDatabaseSharedLockBackend  | cleanUnusedSharedLocks      | 50               |
| TripalFieldsManager                  | setFieldDefDefaults         | 44               |
| ChadoPreparer                        | addBaseTableSVFields        | 44               |
| TripalTerm                           | __construct                 | 43               |
| TripalEntity                         | preSave                     | 43               |
| OBOImporter                          | parse                       | 41               |
|                                      | findEBITerm                 | 40               |
|                                      | processTerm                 | 39               |
[...]

PR #1410, opened 2023-02-04
Improve Maintainability as suggested by CodeClimate

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1404 
*Issue not completely fixed by this PR*

## Description

1. Adjusted the limits used to measure maintainability for our project

Specifically,

- I expanded the max argument count to 8 since Drupal typically has a lot of arguments for dependency injection purposes and what not.
- I expanded the method complexity substantially as the limit felt unobtainable.
- I turned off the identical code detection as it was detecting service calls which should be identical and thus providing many false positives.

2. Expanded some if clauses with way too many criteria for better readability.

Specifically, I updated two identical if clauses in the ChadoSchema classes (both deprecated API class and new Tripal DBX ChadoSchema class). As you will see from the diff I basically just expanded a single if with way too many arguments to multiple if statements which spread out those same arguments. I feel this is much more readable and will be easier to debug --code climate agrees.

## Testing?
The changes to the CodeClimate yml do not need testing as they are just settings.
The ChadoSchema change should be caught by automated testing.
You can see the change in maintainability issues here if you are interested: https://codeclimate.com/github/tripal/tripal/compare/tv4g0-issue1404-improveMaintainability

So basically this PR just needs someone to ok the change in settings as reasonable :-)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,7 +4,7 @@ checks:
   argument-count:
     enabled: false
     config:
-      threshold: 4
+      threshold: 8
   complex-logic:
     enabled: true
     config:
@@ -16,7 +16,7 @@ checks:
   method-complexity:
     enabled: true
     config:
-      threshold: 7
+      threshold: 25
   method-count:
     enabled: false
     config:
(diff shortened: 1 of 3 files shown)


Ferrisx4 (approved):
These changes seem to make sense given the nature of Tripal.

spficklin (approved):

Issue #1393, opened 2023-01-19
Creation of content types & fields with YML


spficklin:
### Description

This issue is for the task of moving the ability to create content types and field out of the Chado prepare step and into YML files. This will make it easier for module developers to create new content types and attach fields to them without having to write code to automate the process.

As a side fix, as discussed in the codefest, we want to rename the fields to use CamelCase rather than the current naming system. There is no need, as there was in Tripal 3, for field names to have the CVterm identifier in the name.    I might as well fix this while I'm doing the YML infrastructure.

I'll use this issue to track both of these.

spficklin:
As per our discussion today we are going to add these considerations to this design:

- update fields to include Chado schema
- move content types and fields out of the prepare step and into install 
- Services that create content types and add fields should be called in the hook_install of the module creating those.
- Separate current content type YML into two separate YML files: one for content types and one for fields so they can be created independently (e.g. some modules will only add fields to existing content types).

laceysanderson:
Competed and finally merged! 🎉 Thank you for all your work on this @spficklin

Issue #1391, opened 2023-01-17
tv4g0 Some files have DOS style line endings


dsenalik:
## BUG/ERROR report

Pull #1392 

### System information

* Tripal Version: 4.x

### Issue description
I mentioned in this comment https://github.com/tripal/tripal/pull/1390#issuecomment-1385568215 that some files have DOS style line endings. I quote from the drupal coding standards:
_Files should be formatted with \n as the line ending (Unix line endings), not \r\n (Windows line endings)._

I am creating this issue to correct this.

dsenalik:
Resolved with merged #1392

PR #1413, opened 2023-02-06
Content Types and Fields defined by YAML

spficklin:
# Summary
Issue #1393 

## Description

This PR moves creation of Tripal content types out of the Tripal Chado prepare step and into Services that use YAML files. It partially addresses items in Issue #1393 .

### Changes
The following changes were made:
1.  Two new YAML files for content types and fields: 
     - `tripal_chado/config/install/tripal.tripal_content_types.chado_content_types.yml`
     - `tripal_chado/config/install/tripal.tripal_fields.chado_fields.yml`
2. The schema for these configuration files have been aded to the `tripal/config/schema/tripal.schema.yml` file.
3. New configuration entity classes were added. This allows the site admin to reload the default configuration settings, which probably won't ever need to happen but is useful for when we code new content types.
   - `tripal/src/Entity/TripalContentTypes.php`
   - `tripal/src/Entity/TripalFields.php`
5. New Service classes were created that allow for creation of content types and attaching their fields:
    - `tripal/src/Services/TripalContentTypes.php`
    - `tripal/src/Services/TripalFields.php`
6.  Functions for creating content types and attaching fields were removed from the Chado Prepare task and instead, these new service classes are used to create the content types from the YAML files. 
7. A new `tripal/api/tripal.api.php` file was created, and the function `tripal_load_configuration()` was added because we need a way to reload tripal provided configurations that may be present in extension module.   This function gets called in the `tripal_rebuild()` function which gets called anytime the cache is cleared for the site.
8. All the necessary interface classes, form classes and routing/links were added.
9. Content type machine names are now more human readable. In Tripal v3 they used to be the term namespace and accession.  The field names are also human readable now.  This makes them so much easier to use as string tokens.
10. Content type URLs and titles are now  set in the YAML file (but can still be changed in the interface).  This makes it easier to set a default by the content type creator.
      - FYI... During testing, I fixed a bug in the additional types field because it wasn't giving the infrasepcific type for the organism content.
12.  I updated the descriptions for each content type.  In Tripal v3 and until now, only the controlled vocabulary term description was used, but that wasn't really well tailored for the content type page listing.  The content type developers can now set this in the YAML file too.

### More to Do
This PR does not do the following suggested in issue #1391 :
- Support indicating the `chado` schema that the field should support.  I think we should not let the user specify that in the YAML... we'll have to rethink how to do that.
- Does not put the creation of the content types and fields into the install of the module. That will be in a future PR. This PR is just step one of the process of making the YAML files work.

### Functional Tests
This  PR is a placeholder because I'm having a hard time with funcitonal testing on my machine at the moment. At least this way I can see what's wrong with and try to fix it. So, feel free to do a funcitonal review, but the code review should wait until testing passes.  

## Testing?

1. To test content type creation
    - Start with a brand new installation of Tripal
    - Install Chado and run the Prepare step. You should see that each content type is created, then afterwrads each field that gets added to a content type is listed. 
    - Go to the Admin > Manage > Structure >Tripal Content Types. You should see all of the content types listed.
    - Click the `Organism` content type. You can see it still works and the page titles and url alias options have been updated to use the new field names from the fields YAML file.
    - You should be able to see the fields attached to the content type as well by going back a page and clicking the 'manage fields' link.
    - Try creating a new organism (or other content type) page. Go to Admin > Tripal Content and click the "Add Tripal Content" button.
       -  All of the content type descriptions have been tailored.
       -  Follow through with creation of a conteont type and it should still work.
2. To test configuration reload
    - You can refresh the YAML configurations (in case you made  change) by going to `Admin` > `Tripal` > `more` (in the header). Then click `Configuration` 
    - Pick any of the three configuration types: content terms, content types or field.  A very simple listing appears.  This is where all of the configuration files for each Tripal extension module will appear.  You can delete any configuration. If you then clear the Drupal cache with a `drush cr` then the configuration will return.  Normally, this is not common Drupal behavior. If you delete a configuration there is no way to get it back.  If you are a developer, this is really annoying because it means you hae to reinstall the entire site.  This feature will most likely only be used by developers.


laceysanderson (commented):
## Finally got some time to look into the failing tests:

```
Drupal\Tests\tripal\Functional\TripalRoutePermissionsTest::testTripalAdminPages
Behat\Mink\Exception\ExpectationException: Link with label Registration found.
```

This appears to be a valid failing. I checked the contents of the $html variable which shows the full page source in the test environment and it shows that no admin items are being loaded in the test environment on this branch. This is not the case on the main branch and interestingly they show up in the UI.

```
 2) Drupal\Tests\tripal_chado\Functional\ChadoStorageTest::testChadoStorage
Exception: The name is required when creating a TripalEntityType.
```

This is due to the `name` key being added to the TripalEntityType class in this branch. Why was this done? The Drupal standard is to use the label for the purposes that we might need a name.

```
Drupal\Tests\tripal_chado\Functional\Plugin\TripalField_chado_linker__propTest::testChadoLinkerPropField
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for tripal.bio_data.1 with the following errors: tripal.bio_data.1:synonyms variable type is NULL but applied schema class is Drupal\Core\Config\Schema\Sequence
```

This schema error seems to be due to the incomplete and outdated field test stub we started and not actually due to the new functionality. I've removed the field test for now.

## On another note

Why is there a new TripalContentTypes entity in this branch? The naming here feels confusing since we used TripalEntityType for the "Tripal Content Type" entity. I'm assuming this is related to the YML functionality but we may want to consider a different naming scheme here.

laceysanderson:
Notes from discussion with Stephen:

- The "name" was added to TripalEntityType in an effort to remove the bio_data_# assumption as our machine names. Synonyms were added at the same time to keep track of the Tripal 3 machine name of the same content types.
  - However, he was not able to fully get this working. As it's been awhile we don't completely remember why but think it's related to the id needing to be numeric. 
  - Anyway, we would like to support this in the future but are removing it for now in the interests of time to get this merged.
- TripalContentTypes entity is related to the config YML and actually is a configuration entity that stores a given YML scheme for Tripal Content Types in order to provide a UI.
  - I've added documentation to the tops of these classes to make that more clear.
  - We may also want to think about changing these names or namespaces to make it more clear of their purpose and to not confuse people contributing to core or looking for bugs.

Additionally now that the full test suite has completed I see that the schema issues were not isolated to the old field test as they seem to show up during the prepare test as well.

laceysanderson:
Update regarding the TripalEntityType::name. It turns out this already existed so it is not as simple as just removing it 😬 I've reverted my changed that tried to just remove it so I can start fresh with figuring out where this is at.

spficklin:
@laceysanderson is breaking this into smaller chunks. closing.

tripal/config/schema/tripal.schema.yml
```diff
@@ -81,172 +82,230 @@ field.storage.tripal_entity.*:
...
 
 field.field.tripal_entity.*.*:
   type: field.field.*.*.*
   mapping:
     settings:
       type: field.settings
       mapping:
+        max_delta:
+          type: integer
+          label: 'The maximum number of records to retreive for this field when querying'
```
dsenalik, line 110: *retrieve 😁

Issue #1400, opened 2023-01-30
Terms without Chado


spficklin:
We will be moving the creation of content types and fields into YML files as mentioned in issue #1393 .  But this means that if an extension modules tries to install content types without the tripal_chado module being enabled there won't be a place to store terms because currently the only backend we support is Chado, although, the IDSpace Vocabulary and Terms class support other backends.  

The goal is to allow modules to use Tripal services in their `hook_install` functions to create the content types, etc. needed but we nee this issue fixed before that can happen.

laceysanderson:
Fixed by merged PR

PR #1417, opened 2023-02-12
Deafult (non Chado) Vocabulary and IdSpace Plugin.

spficklin:
# Tripal 4 Core Dev Task

Issue #1400

## Description

This PR fixes issue #1400.  It adds two new plugins that implements the TripalVocabulary and TripalIdSpace for generic terms. It does not use Chado and therefore can be used for modules that don't want to use Chado but need terms.   It can also be used for testing core Tripal functionality in functional tests when terms are needed by Chado is not installed.  

This issue was not part of our Alpha 2 release plan, but when I was trying to get functional testing working for PR #1413 I needed to be able to create terms that were not stored in the Chado schema.  I could not use a Mock term because the content type and field services need to be able to lookup terms.  So, this was necessary. I'm providing it as a separate PR to make it simple to review rather than as part of PR #1413.

## Testing?

I think this PR only needs a code review.  Nothing uses these new plugins so there's no way to test. However, I did createa  full suite of functional tests for these new plugins. Also, the presence of these plugins allowed me to create some functional testing for the TripalTerm class independent of the tests done in the tripal_chado module.  

So, if this PR passes all tests, and a code review looks good I think it can be merged.  I will be using it for more functional testing in PR #1413, so if someting isn't quite right I can fix it there (but I did test it already and it works :-) ....

### DIFF
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -437,7 +437,7 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
   }
 
   /**
-   * Retrieve a tern record from tripal_terms table.
+   * Retrieve a term record from tripal_terms table.
    *
    * This function uses the db.name (IdSpace), cv.name (vocabulary)
    * and accession values to uniquely identify a term.
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -15,7 +15,7 @@ use Drupal\tripal\TripalVocabTerms\TripalTerm;
  */
 class TripalDefaultIdSpace extends TripalIdSpaceBase {
   /**
-   * A simple boolean to prevent Chado queries if the ID space isn't valid.
+   * A simple boolean to prevent queries if the ID space isn't valid.
    *
    * @var bool
    */
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -36,6 +36,12 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
    * {@inheritDoc}
    */
   public function getParent($child) {
+    // Don't get values for an ID space that isn't valid.
+    if (!$this->is_valid) {
+      return NULL;
+    }
+
+    $this->messageLogger->warning('The TripalDefaultIdSpace::getParent() function is currently not implemented');
   }
 
   /**
@@ -228,12 +234,14 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
    * {@inheritDoc}
    */
   public function destroy() {
+    $this->messageLogger->warning('The TripalDefaultIdSpace::destory() function is currently not implemented');
   }
 
   /**
    * {@inheritDoc}
    */
   public function removeTerm($accession) {
+    $this->messageLogger->warning('The TripalDefaultIdSpace::removeTerm() function is currently not implemented');
   }
 
   /**
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -49,8 +49,7 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
 
     // Make sure the description is not too long.
     if (empty($prefix)) {
-      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a urlprefix when calling setURLPrefix().',
-          ['@size' => 255, '@value' => $prefix]);
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a urlprefix when calling setURLPrefix().');
       return False;
     }
     if (strlen($prefix) > 255) {
@@ -217,12 +216,12 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
     if (!empty($name) AND (strlen($name) > 255)) {
       $this->messageLogger->error('TripalDefaultIdSpace: The IdSpace name must not be longer than @size characters. ' +
         'The value provided was: @value', ['@size' => 255, '@value' => $name]);
-      return;
+      $this->is_valid = FALSE;
+      return FALSE;
     }
 
     $this->is_valid = TRUE;
-
-    return $this->is_valid;
+    return TRUE;
   }
 
   /**
@@ -249,9 +248,8 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
 
     // Make sure the description is not too long.
     if (empty($description)) {
-      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a description when calling setDescription().',
-          ['@size' => 255, '@value' => $description]);
-        return False;
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a description when calling setDescription().');
+      return False;
     }
     if (strlen($description) > 255) {
       $this->messageLogger->error('TripalDefaultIdSpace: The description for the vocabulary ID space must not be longer than @size characters. ' +
@@ -437,9 +435,9 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
   }
 
   /**
-   * Retrieve a tern record from tripal_terms table.
+   * Retrieve a term record from tripal_terms table.
    *
-   * This function uses the db.name (IdSpace), cv.name (vocabulary)
+   * This function uses the IdSpace, vocabulary,
    * and accession values to uniquely identify a term.
    *
    * @param TripalTerm $term
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -1,0 +1,671 @@
+<?php
+
+namespace Drupal\tripal\Plugin\TripalIdSpace;
+
+use Drupal\tripal\TripalVocabTerms\TripalIdSpaceBase;
+use Drupal\tripal\TripalVocabTerms\TripalTerm;
+
+/**
+ * Default Implementation of TripalIdSpaceBase
+ *
+ *  @TripalIdSpace(
+ *    id = "tripal_default_id_space",
+ *    label = @Translation("Deafult Tripal IdSpace"),
+ *  )
+ */
+class TripalDefaultIdSpace extends TripalIdSpaceBase {
+  /**
+   * A simple boolean to prevent Chado queries if the ID space isn't valid.
+   *
+   * @var bool
+   */
+  protected $is_valid = FALSE;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    // Instantiate the TripalLogger
+    $this->messageLogger = \Drupal::service('tripal.logger');
+
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function getParent($child) {
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function setURLPrefix($prefix) {
+    // Don't set a value for an ID space that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // Make sure the description is not too long.
+    if (empty($prefix)) {
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a urlprefix when calling setURLPrefix().',
+          ['@size' => 255, '@value' => $prefix]);
+      return False;
+    }
+    if (strlen($prefix) > 255) {
+      $this->messageLogger->error('TripalDefaultIdSpace: The urlprefix for the vocabulary ID space must not be longer than @size characters. ' +
+          'The value provided was: @value',
+          ['@size' => 255, '@value' => $prefix]);
+      return False;
+    }
+
+    // Update the record in the Chado `db` table.
+    $conn = \Drupal::service('database');
+    $query = $conn->update('tripal_terms_idspaces')
+      ->fields(['urlprefix' => $prefix])
+      ->condition('name', $this->getName(), '=');
+    $num_updated = $query->execute();
+    if ($num_updated != 1) {
+      $this->messageLogger->error('TripalDefaultIdSpace: The urlprefix could not be updated for the vocabulary ID Space.');
+      return False;
+    }
+    return True;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setDefaultVocabulary($vocab) {
+    $retval = parent::setDefaultVocabulary($vocab);
+
+    // Don't set a value for an ID space that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // Make sure the description is not too long.
+    if (empty($vocab)) {
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a default vocabulary when calling setDefaultVocabulary().');
+      return False;
+    }
+    if (strlen($vocab) > 255) {
+      $this->messageLogger->error('TripalDefaultIdSpace: The default vocabulary must not be longer than @size characters. ' +
+          'The value provided was: @value',
+          ['@size' => 255, '@value' => $vocab]);
+      return False;
+    }
+
+    // Update the record in the Chado `db` table.
+    $conn = \Drupal::service('database');
+    $update = $conn->update('tripal_terms_idspaces');
+    $update = $update->fields(['default_vocab' => $vocab]);
+    $update = $update->condition('name', $this->getName(), '=');
+    $num_updated = $update->execute();
+    if ($num_updated != 1) {
+      $this->messageLogger->error('TripalDefaultIdSpace: The default vocabulary could not be updated for the vocabulary ID Space.');
+      return False;
+    }
+    return True;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function recordExists() {
+    $db = $this->loadIdSpace();
+    if ($db and $db['name'] == $this->getName()) {
+      return True;
+    }
+    return False;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function getTerm($accession, $options = [ ]) {
+
+
+    if (!$this->is_valid) {
+      return NULL;
+    }
+
+    $conn = \Drupal::service('database');
+    $query = $conn->select('tripal_terms', 'tt');
+    $query = $query->fields('tt');
+    $query = $query->condition('id_space', $this->getName(), '=');
+    $query = $query->condition('accession', $accession, '=');
+    $result = $query->execute();
+    if (!$result) {
+      return NULL;
+    }
+    $cvterm =  $result->fetchObject();
+
+    $term =  new TripalTerm([
+      'name' => $cvterm->name,
+      'definition' => $cvterm->definition,
+      'accession' => $accession,
+      'idSpace' => $this->getName(),
+      'vocabulary' => $cvterm->vocabulary ? $cvterm->vocabulary : $this->getDefaultVocabulary(),
+      'is_obsolete' => $cvterm->is_obsolete == 1 ? True : False,
+      'is_relationship_type' => $cvterm->is_relationship_type == 1 ? True : False,
+    ]);
+
+    // We need an IdSpace manager object to look up synonym an property types.
+    /** @var \Drupal\tripal\TripalVocabTerms\PluginManagers\TripalIdSpaceManager $idsmanager */
+    $idsmanager = \Drupal::service('tripal.collection_plugin_manager.idspace');
+
+    // Add in the synonyms
+    $synonyms = unserialize($cvterm->synonyms);
+    foreach ($synonyms as $tuple) {
+      $synonym_name = $tuple[0];
+      $synonym_type = $tuple[1];
+      $syn_type_term = NULL;
+      if ($synonym_type) {
+        list($syn_idspace_name, $syn_accession) = explode(':', $synonym_type);
+        /** @var \Drupal\tripal\Plugin\TripalIdSpace\TripalDefaultIdSpace $syn_idspace */
+        $syn_type_idspace = $idsmanager->loadCollection($syn_idspace_name);
+        $syn_type_term = $syn_type_idspace->getTerm($syn_accession);
+      }
+      $term->addSynonym($synonym_name, $syn_type_term);
+    }
+
+    // Add in the property.
+    $properties = unserialize($cvterm->properties);
+    foreach ($properties as $tuple) {
+      list($prop_term_id, $rank, $prop_value) = $tuple;
+      list($prop_idspace_name, $prop_accession) = explode(':', $prop_term_id);
+      /** @var \Drupal\tripal\Plugin\TripalIdSpace\TripalDefaultIdSpace $syn_idspace */
+      $prop_type_idspace = $idsmanager->loadCollection($prop_idspace_name);
+      $prop_type_term = $prop_type_idspace->getTerm($prop_accession);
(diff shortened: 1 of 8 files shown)


spficklin:
Dang... I named the branch wrong :-/

Also, I forgot.  I did fix a bug in the ChadoIdSpace plugin where it was using the default vocabulary when it pulled out terms from Chado rather than what was in the 'cv' table.  The deafult isn't always the same as  the vocabulary for a term, so this could have caused problems.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+   * {@inheritDoc}
+   */
+  public function isValid() {
+
+    // Make sure the name of this collection does not exceeed the allowed size in Chado.
+    $name = $this->getName();
+    if (!empty($name) AND (strlen($name) > 255)) {
+      $this->messageLogger->error('TripalDefaultVocabular: The vocabulary name must not be longer than @size characters. ' +
+          'The value provided was: @value',
+          ['@size' => 255, '@value' => $this->getName()]);
```
laceysanderson: I think you also need to set $this->is_valid to FALSE within this IF statement? Otherwise won't it always return valid regardless?

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  public function setURL($url) {
+    // @todo there may be a problem in the future if we are able to
+    // associate borrowed terms with a vocabulary.  If we
+    // add the ID space of borrowed terms to a vocabulary then
+    // setting the URL will be incorrect for those ID spaces.
+
```
laceysanderson, line 140: I'm feeling really sad about the above comment. It seems there was something lost in translation between me + Josh. The original design for this API that me + Josh came up with does support borrowed terms. But when I went looking for it, this seems to have been changed at some point and I didn't catch it :-( You can see more about the original design here: https://tripaldoc.readthedocs.io/en/latest/design/vocab/file-structure.html and https://tripaldoc.readthedocs.io/en/latest/design/vocab.html.

This was intended to work by having multiple terms with the same ID Space and different vocabularies. 

For example, say you have the following term in it's primary vocabulary:

- Name: person
- ID Space: ABC
- Accession: 12345
- Vocabulary: The Core Ontology for People

This term was then hypothetically borrowed by the "Components of a Research Group Ontology". This would result in an additional term object in Tripal as so:

- Name: person
- ID Space: ABC
- Accession: 12345
- Vocabulary: Components of a Research Group Ontology

You would be able to tell that the term was borrowed because you could get the default vocabulary of the ID Space and see that it is different then the id space of this particular term.

laceysanderson (commented):
Ran out of time to complete my review but I thought I would submit what I've noticed so far for further contemplation.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: I copied this code from the ChadoVocabulary class and updated for the table I created in the public schema of Drupal.  So this problem is not unique to this class, but also the ChadoVocabulary class.  I think this should be added as a new issue and it should be addressed for both the TripalDefaultVocabulary and ChadoVocabulary classes.

spficklin: Can we fix this later so we can get this merged in so I can get the testing done for PR #1413?

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Yes. I'll fix that.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
laceysanderson: Ok, I've created an issue here: https://github.com/tripal/tripal/issues/1421 for this so we don't forget.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+ * Default Implementation of TripalIdSpaceBase
+ *
+ *  @TripalIdSpace(
+ *    id = "tripal_default_id_space",
+ *    label = @Translation("Deafult Tripal IdSpace"),
+ *  )
+ */
+class TripalDefaultIdSpace extends TripalIdSpaceBase {
+  /**
+   * A simple boolean to prevent Chado queries if the ID space isn't valid.
```
laceysanderson:
```suggestion
   * A simple boolean to prevent queries if the ID space isn't valid.
```

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+    // Instantiate the TripalLogger
+    $this->messageLogger = \Drupal::service('tripal.logger');
+
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function getParent($child) {
+  }
```
laceysanderson, line 45: We should let someone know that this isn't supported in case they do try to use this implementation.

```suggestion
    $this->messageLogger->warning('TripalDefaultIdSpace: Relationships between terms are currently not supported by this implementation.');
    return NULL;
  }
```

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+  public function setURLPrefix($prefix) {
+    // Don't set a value for an ID space that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // Make sure the description is not too long.
+    if (empty($prefix)) {
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a urlprefix when calling setURLPrefix().',
+          ['@size' => 255, '@value' => $prefix]);
```
laceysanderson: Shouldn't provide tokens when they are not used.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+   */
+  public function isValid() {
+
+    // Make sure the name of this ID Space does not exceeed the allowed size in Chado.
+    $name = $this->getName();
+
+    if (!empty($name) AND (strlen($name) > 255)) {
+      $this->messageLogger->error('TripalDefaultIdSpace: The IdSpace name must not be longer than @size characters. ' +
+        'The value provided was: @value', ['@size' => 255, '@value' => $name]);
+      return;
```
laceysanderson:
```suggestion
      $this->is_valid = FALSE;
      return FALSE;
```

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+
+    $this->is_valid = TRUE;
+
+    return $this->is_valid;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function destroy() {
```
laceysanderson, line 236: Shouldn't this be implemented? At a minimum it should have a message warning anyone who tries to use this implementation that it's not implemented.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function removeTerm($accession) {
```
laceysanderson, line 243: Shouldn't this be implemented? At a minimum it should have a message warning anyone who tries to use this implementation that it's not implemented.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+
+    // Don't set a value for an ID space that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // Make sure the description is not too long.
+    if (empty($description)) {
+      $this->messageLogger->error('TripalDefaultIdSpace: You must provide a description when calling setDescription().',
+          ['@size' => 255, '@value' => $description]);
```
laceysanderson: Don't add parameters if they're not used.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+          }
+        }
+      }
+    }
+
+    return $terms;
+  }
+
+  /**
+   * Retrieve a tern record from tripal_terms table.
```
laceysanderson:
```suggestion
   * Retrieve a term record from tripal_terms table.
```

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
```diff
@@ -0,0 +1,671 @@
...
+   *
+   * This function uses the db.name (IdSpace), cv.name (vocabulary)
+   * and accession values to uniquely identify a term.
```
laceysanderson, line 449: This should be updated as it's specific to the chado implementation.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
laceysanderson: I saw below that the getChildren is implemented... why isn't this one? it looks like it would be relatively easy to grab the parents since they are stored directly.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
+<?php
+
+namespace Drupal\tripal\Plugin\TripalVocabulary;
+
+use Drupal\tripal\TripalVocabTerms\TripalVocabularyBase;
+
+/**
+ * Chado implementation of the TripalVocabularyBase.
```
laceysanderson:
```suggestion
 * Default implementation of the TripalVocabularyBase.
```

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    // Instantiate the TripalLogger
+    $this->messageLogger = \Drupal::service('tripal.logger');
+  }
+
+  /**
+   * Loads a Vocabulary record from Chado.
```
laceysanderson: Not from Chado...

laceysanderson: The description below is also specific to the chado implementation.

laceysanderson: Also in-line comments in the method are chado specific and do not make sense.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+   */
+  public function getLabel() {
+    $cv = $this->loadVocab();
+    return $cv['label'];
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function recordExists() {
```
laceysanderson, line 79: Shouldn't this be implemented? At a minimum it should have a message warning anyone who tries to use this implementation that it's not implemented.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+   *   The namespace of the vocabulary.
+   */
+  public function getNameSpace() {
+    return $this->getName();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function destroy() {
```
laceysanderson, line 121: Shouldn't this be implemented? At a minimum it should have a message warning anyone who tries to use this implementation that it's not implemented.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+    // associate borrowed terms with a vocabulary.  If we
+    // add the ID space of borrowed terms to a vocabulary then
+    // setting the URL will be incorrect for those ID spaces.
+
+    // Don't set a value for a vocabulary that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // This value goes to the Chado `db.url` column, so check its size
```
laceysanderson: Again, chado specific.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    // If we don't have any ID spaces then there is no URL.
+    $id_spaces = $this->getIdSpaceNames();
+    if (count($id_spaces) == 0) {
+      $this->messageLogger->error('TripalDefaultVocabulary: Cannot set the URL when no ID spaces are present for the vocabulary.');
+      return False;
+    }
+
+    // Update the record in the Chado `db` table for the URL for all ID spaces.
```
laceysanderson: Again, chado specific comment.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+  public function setLabel($label) {
+    // Don't set a value for a vocabulary that isn't valid.
+    if (!$this->is_valid) {
+      return False;
+    }
+
+    // Make sure the label is not too long.
+    if (empty($label)) {
+      $this->messageLogger->error('TripalDeafultVocabulary: You must provide a label when calling setLabel().',
+          ['@size' => 255, '@value' => $label]);
```
laceysanderson: Don't pass in parameters that are not used.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+
+
+    // Update the record in the Chado `cv` table.
+    $chado = \Drupal::service('database');
+    $update = $chado->update('tripal_terms_vocabs');
+    $update = $update->fields(['label' => $label]);
+    $update = $update->condition('name', $this->getName(), '=');
+    $num_updated = $update->execute();
+    if ($num_updated != 1) {
+      $this->logInvalidCondition('TripalDeafultVocabulary: The label could not be updated for the vocabulary.');
```
laceysanderson: Why did you use logInvalidCondition when everyone else you use the tripal logger? Including in the same case in the ID Space.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+        'idspaces' => serialize([]),
+      ]);
+      $query->execute();
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function addIdSpace($idSpace) {
```
laceysanderson, line 242: Shouldn't this also check that the ID Space already exists in the ID Space table?

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
```diff
@@ -0,0 +1,262 @@
...
+    $update = $conn->update('tripal_terms_vocabs');
+    $update = $update->fields(['idspaces' => serialize($idspaces)]);
+    $update = $update->condition('name', $this->getName(), '=');
+    $update->execute();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function removeIdSpace($idSpace) {
```
laceysanderson, line 264: Do we need to check here if this ID Space is attached to any other vocab and if not then remove it from the ID Space table?

tripal/tripal.install
```diff
@@ -153,9 +153,182 @@ function tripal_schema() {
...
+/**
+ *
+ */
+function tripal_tripal_terms_schema() {
+  // @todo: Storing parents and synonyms in a text field will make it
+  // slow to lookup terms by those names.  Since this table is just to
+  // provide some sort of default storage so that Tripal can work without
+  // Chado it may not be that big of a deal.  But if there is a case
+  // where this table is used for storing massive numbers of terms it
+  // will need to be optimized.
```
laceysanderson, line 255: It eally wouldn't be that much more work to use a prop-like table and relationship table, would it? Maybe we should just do that in case someone tries to use Tripal without Chado. This does seem to be likely as not everyone is a fan of Chado.

laceysanderson (changes requested):
Yikes, that ended up being more comments then I had intended 🙈 Even though this is meant as a default implementation for those who don't want to use Chado and our main focus right now is using it in testing... I do think we should ensure we are providing a good template for people to learn from and understand.

Also, I didn't get the chance to look at the testing 🙈

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: This has been fixed. Since I borrowed this code from the ChadoVocabulary Plugin that class had the same problem, so I fixed both.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: This is not implemented in the ChadoIdSpace plugin either.    We need to implement it in both.  For now can we just add an issue to remind us to do both?

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: Fixed

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: Yikes... Fixed in both the TripalIdSpace and the ChadoIdSpace classes.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: It's not implemented in the ChadoIdSpace either... I wasn't sure if we should or not... I wanted to have a conversation about that...

spficklin: Can we move that to an issue to get this PR merged?

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: Same thing as before. It's not implemented in the ChadoIdSpace class either.

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: Fixed

tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Yes it should. Fixed.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Same coment as with the ChadoVocabulary.destroy() method.  It's not implemented there either

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Fixed... also a problem in the ChadoVocabulary class too... so fixed there as well.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: Yes it should! Great catch.

spficklin: Fixed

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
spficklin: I don't think so.  This function is just for managing assocaitions with the vocabulary. If we had it remove the IDspace from the system that might be unexpected.

tripal/tripal.install
spficklin: I agree, we should. My time is just so limited I went this route.   I can add it to an issue to fix.

spficklin:
Thanks @laceysanderson for the fantastic review.  My apologies for not proofing it better.  I should have caught most of those minor issues.  Related to the larger ones. I created an issue here:  https://github.com/tripal/tripal/issues/1426

My hope is that as long as all of the testing passes, and you are satisfied with the code fixes that we can merge this so I can move on with PR #1413 (goal of an Alpha 2 by end of the month).   Since many of the issues you called out this class on exist in the Chado implemenation I figured we can fix both at the same time.

tripal/src/Plugin/TripalVocabulary/TripalDefaultVocabulary.php
laceysanderson: Ok, sounds reasonable :-) I'm resolving.

tripal/tripal.install
laceysanderson: Ok, fair enough!

laceysanderson:
Regarding the unimplemented functions: I'm content with moving their implementation to the focus of issue #1426 :-) That said, I do think we should add a quick message logger warning now to all those functions reporting that they are not implemented. This will keep us accountable and ensure that people developing extension modules are told if they try to use them!

spficklin:
10-4 I'll add a message to the logger for the unimplemented functions.

spficklin:
Okay, warnings have been added to unimplemeneted functions.  Let me know if there's anything else.

spficklin:
All functional tests are passing!  Unless there are any other comments it looks like it's ready for merging.

laceysanderson (approved):
Awesome! Changes look good -Thanks @spficklin  :-) I agree; this is ready to be merged! 🎉

spficklin:
thanks @laceysanderson again, great review :-)

PR #1423, opened 2023-02-16
tv3g4 Issue 1422 fix publish update progress

dsenalik:
# Bug Fix

## Description
Progress updates on publishing were not happening if more than one batch of 100 entities was published, because the internal batch variable was being used.

## Testing?
Load a whole genome gff3 file with thousands of genes, and then publish mRNA for example. Now the progress is updated
```
INFO (PUBLISH_RECORDS): There are 14336 records to publish.
There are 14336 records to publish.
Percent complete: 0%. Memory: 54,711,184 bytes.
Percent complete: 1%. Memory: 55,774,848 bytes. Duration: 0.08 mins
Percent complete: 2%. Memory: 55,774,848 bytes. Duration: 0.18 mins
Percent complete: 3%. Memory: 55,774,848 bytes. Duration: 0.27 mins
...
Percent complete: 99%. Memory: 55,774,848 bytes. Duration: 8.78 mins
Percent complete: 100%. Memory: 55,774,848 bytes. Duration: 8.88 mins
INFO (PUBLISH_RECORDS): Successfully published 14336 mRNA record(s).
Successfully published 14336 mRNA record(s).
```

### DIFF
--- a/tripal_chado/api/tripal_chado.api.inc
+++ b/tripal_chado/api/tripal_chado.api.inc
@@ -240,7 +240,6 @@ function chado_publish_records($values, $job = NULL) {
     try {
       // Keep track of how many we've published so far
       // before clearing our chunk,
-      $num_actually_published = $num_actually_published + $i;
       $i = 0;
       while ($record = $records->fetchObject()) {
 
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Looks good 👍

PR #1429, opened 2023-02-18
tv3g9 issue1428 Check for null in tripal_chado_field_storage_load()

dsenalik:
# Bug Fix

## Description
In ```function tripal_chado_field_storage_load()``` do not try to expand a variable in a null record.

## Testing?
Testing is done for us by automated testing. Before this change, this line showed up in testing results:
```[site http://127.0.0.1] [TRIPAL ERROR] [TRIPAL_CHADO] Missing $object argument to the chado_expand_var function from tripal_chado_field_storage_load (line: 426)```

With this pull, this line no longer appears.


laceysanderson (approved):
I agree with your reasoning and the fix looks good 👍 
Thanks for the fix @dsenalik

PR #1431, opened 2023-02-27
Tripal 4 Fields: Chado Linker Fields / Chado Contact Linker Field

dsenalik:
# New Feature

## Updated Description
Convert the `ChadoContactDefault` field to dual-mode action! It will now handle contacts referenced in the base table, as it currently does, as well as being able to handle contacts referenced by a linker table. This will allow similar functionality as is provided by the Tripal 3 [Chado Linker Fields](https://github.com/tripal/chado_linker_fields) module, but without needing special linking fields.

This field should be one of the simpler "linker" fields, since all of the linker tables are simple, they do not contain additional columns. Some of the other content type linker tables also contain `type_id`, `rank`, both, or other columns.

Note that the CV term for this field in Tripal 3 was `local:contact` - here it is changed to a real term [NCIT:C47954](https://www.ebi.ac.uk/ols4/ontologies/ncit/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCIT_C47954) 

The following chado base tables present in Tripal 4 have a foreign key to contact:
 * study
 * arraydesign - here the foreign key column is `manufacturer_id`
 * stockcollection - this is not a default contact type

The following chado linker tables for contact exist for these content types present defined by default in Tripal 4:
 * feature_contact (gene, mRNA, QTL, SequenceVariant, GeneticMarker, Phenotypic Marker)
 * featuremap_contact (genetic_map, physical_map)
 * library_contact (dna_library)
 * project_contact (project)

## Updated Testing procedure
1. Build a new docker on this branch, since we need a new CV term for this field.
2. Create several contacts for testing. For at least one, leave the contact type blank, to test the type display, and another leave the description blank.
3. Also create an organism to test feature content types later.

### Base Table Testing
This is described in PR #1599 

### Linker Testing
1. Create a project, or any of the other linker content types listed above, and specify one contact. (dna library is missing some fields, issue #1509 so can't currently be tested) Confirm that it displays. If it has a type, it won't display yet.
2. Modify the field settings for project to display type and description. Go to `/admin/structure/bio_data/manage/project/fields` and for contact select Edit. For the token string enter this 
Full_token_string

3. Edit the project and change the contact to a different one. Bonus points: do `select * from chado.project_contact;` to confirm the chado information is changed
4. Edit the project and add a second contact.
5. Edit the project and change the order of contacts. (the order is stored by drupal but is not saved in chado)
6. Edit the project and delete one of the contacts (change it to `-- Select --`)
7. If you add the `null` contact, it should display as `Unknown` 
project_1
8. repeat for a `feature` content type _e.g._ Gene, and a featuremap content type _e.g._ Physical Map

### Test the form for adding the field
1. Remove the field from one of the content types, and then manually add back the `Chado Contact` field, to test the form and ajax for base table selection, and cardinality. For example on the Project content type.
2. Go to `/admin/structure/bio_data/manage/project/fields` and delete the Contact field.
3. Now click on + Create a new field
4. Add back a "Chado Contact" field

5. Select the base table. This is where the ajax is triggered

6. The linking method should be populated with the only possibility in this case. Set cardinality to unlimited.

7. Finish defining the field and save with a token string suitable for testing
re-add-2
the comment https://github.com/tripal/tripal/pull/1431#issuecomment-1809188933 talks about the token string portion now added to this form
9. Create a new project with a contact, just to confirm this works.
re-add-3

10. Test similarly on a content type with the contact_id in the base table, such as Study. Here the link will appear as


dsenalik:
When manually adding this field to a content type, we need to specify the base table, but in addition, we need to determine the linker table. In Tripal 3 some modules define new linker tables, and in some cases we have both, for example ```stock_feature``` and ```feature_stock``` (from mainlab_tripal module) so we need a mechanism to allow the site administrator to select the desired linker table.
Commit https://github.com/tripal/tripal/pull/1431/commits/02fc94d6a8bba6bf7692a52c773cd13cb8b83fd2 adds this table to the settings form, and in addition I have added an ajax callback to limit the tables to only those with both the base table and the object table (i.e. contact) in the table name. Of course if you don't have a linker table, the second field will not have anything in it and you won't be able to finish adding the field.


---

---

dsenalik:
Some notes for myself, commit https://github.com/tripal/tripal/pull/1431/commits/91ad98d63e59069f751001fe0aca890df32a1c85 adds functions to generate lists of base tables and linker tables. Basically, starting from the object table (contact), look for all tables with a foreign key to the primary key of contact (contact_id), and if there is one, see if there is a second different foreign key. If so, this becomes a candidate for the list of base tables. There are a few odd cases, such as the study table contains both a contact_id and pub_id, so in effect study is a linker table from pub to contact. And because study is not really a linker table, the dbxref shouldn't be there either.
```
Add base table cvterm from linker table contactprop
Add base table feature from linker table feature_contact
Add base table featuremap from linker table featuremap_contact
Add base table library from linker table library_contact
Add base table nd_experiment from linker table nd_experiment_contact
Add base table project from linker table project_contact
Add base table pubauthor from linker table pubauthor_contact
Add base table cvterm from linker table stockcollection
Add base table dbxref from linker table study
Add base table pub from linker table study
```

spficklin:
This PR should have the same issue with the `store_link` action.  We need PR #1463 merged before we can fully fix this.  It has the same problem as the synonym linker (PR #1447 )

dsenalik:
This PR now updates `ChadoFieldItemBase.php` to allow the parent class function `getBaseTables()` to be used on linker fields such as this one, and adds a new function to the parent class `getLinkerTables()`

These would be used by any of the other linker fields yet to be created, so logically this should go in the parent class.

dsenalik:
This field is ready for review. See the updated testing procedures at the top of this PR.

dsenalik:
This field now uses a "token string" to configure what information to display, here is an example for displaying all three
token_string

Example output
new_formatter

The default token string is `[name]` to match Tripal 3 behavior. 

Validation for missing tokens is done
validate1

as well as validation for invalid tokens.
validate2

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoContactFormatterDefault.php
```diff
@@ -27,15 +27,49 @@ class ChadoContactFormatterDefault extends ChadoFormatterBase {
...
+      $list[$delta] = $displayed_string;
+    }
+
+    // Only return markup if the field is not empty.
+    if (!empty($list)) {
+
+      // If only one element has been found, don't make into a list.
+      if (count($list) == 1) {
+        $elements[0] = [
+          "#markup" => $list[0]
```
laceysanderson, line 67: Shouldn't we not be using `#markup` anymore as mentioned in #1686?

laceysanderson (commented):
Manual Testing:
- Checked functionality on the following content types: ✅ study, ✅ array_design, ✅ genetic map, ✅ sequence variant, ✅ project and it worked as expected
- ✅ Ensured field was auto added to the above content types
- ✅ Ability to add/edit/delete contact and rearrange when more than one is supported
- ✅ Ability to change formatting of the contact via tokens
- ✅ not able to leave token string empty or to include random text/tokens only
- ✅ Property field still works as expected including recognizing empty values after changes to chadostorage.
- 🙈 Manually adding two contact linker fields to the same content type causes swapping of records... Did we want to support this?

Comments:
It wasn't intuitive to me that the settings for display be on the fieldtype edit form... I would expect them to be a cog on the manage display page and associated only with the formatter.

Code Review:
✅ Looks good! Although as mentioned in the comments: I think it would be better to move the token form out of the field type settings and into the formatter settings.

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoContactFormatterDefault.php
dsenalik: '#markup' is fine, it is just '#type' => 'markup' that does not do anything

dsenalik:
I can confirm that if you have two contact fields, you see this swap behavior.
I presume that if publish was working, and you published, then both fields would display both.
You can set them both the same, but if you then edit the second one there is no change, if you edit the first one, both change.

There is trouble testing this with other fields, which brought up issue #1687 - but not directly relevant to this PR

There will probably be general chaos if we allow duplicate fields right now. I propose preventing it by removing existing fields from the select list when you manually add a field. If we want to support this then somebody who wants it will need to do it 😁

I'll see how hard this is to implement (removing from the select list), but I think since it would apply to all fields it should be a separate PR (tbd)

laceysanderson:
Publish would fix this but only until the first time you edit the entity. I agree though that we just don't support it for now. I'm on my way home now but maybe just an issue mentioning this is not supported for double-hop linker fields. It is supported for the property field and we could support it for these fields too by using a table alias specific to the field 🤔 but we can discuss in the other issue and not worry about it for this PR.

dsenalik:
Thanks for your suggestion @laceysanderson 
I have moved the token string setting to the formatter. This is the first case in Tripal where we have one, so this can become the first example of how to add a setting and validation function.
It looks like this
display_settings_1

and if you click on the gear
display_settings_2

Validation is performed when you click on either the "Update" or the "Save" button
display_settings_3

dsenalik:
Issue #1688 created to discuss having two copies of a field

laceysanderson (approved):
Retested the token format now it is moved into the formatter and it works perfectly and is very intuitive! Thanks @dsenalik <3 I think this is ready to merge now!

PR #1433, opened 2023-02-28
BUG: organism field cannot be added manually to gene page

dsenalik:
# Bug Fix
## Description
We cannot add the ```obi__organism``` field to a gene page.
This change copies the code from ```chado_linker__prop``` to check for the case where the ```$base_table``` is not yet defined, which happens when first adding the field to a content type.

## Testing?
Add ```obi__organism``` to a gene content type. 
Before this change you will get a WSOD with the message ```The website encountered an unexpected error. Please try again later.```
After this change you should be able to add it.

## Other notes
I hate when I screw up the branch name. When will I learn...


dsenalik:
Commit https://github.com/tripal/tripal/pull/1433/commits/debf3a3508219a8fd5d15429e756a78f1bc62c9d is to add an exception when trying to add this field to a content type that does not support it, _e.g._ project.
Is this appropriate? Is it the correct way to do this? _This proposal is moved to Issue #1440_ 
Not supported

laceysanderson (approved):
This is the exact same piece of code we see in most fields so I think it's safe to say ✅ for code review 🤪 

I did a quick functional test and confirmed I get the WSOD when adding an organism reference to a gene on the 4.x branch. When I switch to this branch I am able to add a new organism reference field ✅ 

I just updated this branch as it had become out of date since other PRs were merged today. Once the tests pass again, this can be merged 🎉 

Thanks @dsenalik

dsenalik:
Awesome, thank you @laceysanderson

Issue #1436, opened 2023-03-04
Warnings generating chado_schema-1.3.yml (formerly foreign_key problems)


dsenalik:
# BUG/ERROR report

### Issue description
In Issue #1432 This comment https://github.com/tripal/tripal/issues/1432#issuecomment-1454840026 reports that the schema for project and at least one other table are missing the foreign keys.

#### Steps to reproduce
Look at ```tripal_chado/chado_schema/chado_schema-1.3.yml```
these lines, there are no foreign key definitions
```
   4888 project:
   4889   description: "Standard Chado flexible property table for projects."
   4890   fields:
   4891     project_id:
   4892       type: serial
   4893       not null: TRUE
   4894     name:
   4895       type: character varying
   4896       not null: TRUE
   4897       size: 255
   4898     description:
   4899       type: text
   4900       not null: FALSE
   4901   unique keys:
   4902     project_c1: name
   4903   primary key: project_id
   4904   referring_tables: assay_project, nd_experiment_project, project_analysis, project_contact, project_dbxref, project_feature, project_pub, project_relationship, project_stock, projectprop
```
The table right after ```project_contact:``` has foreign keys. I don't know how this file is generated, but I suspect it needs to be rebuilt.
For today's (20230304) docker the table does have foreign keys. In an earlier docker it did not have foreign keys.
```
sitedb=> \d project
                                           Table "chado.project"
   Column    |          Type          | Collation | Nullable |                   Default                   
-------------+------------------------+-----------+----------+---------------------------------------------
 project_id  | bigint                 |           | not null | nextval('project_project_id_seq'::regclass)
 name        | character varying(255) |           | not null | 
 description | text                   |           |          | 
Indexes:
    "project_pkey" PRIMARY KEY, btree (project_id)
    "project_c1" UNIQUE CONSTRAINT, btree (name)
Referenced by:
    TABLE "assay_project" CONSTRAINT "assay_project_project_id_fkey" FOREIGN KEY (project_id) REFERENCES project(project_id) DEFERRABLE INITIALLY DEFERRED
    TABLE "nd_experiment_project" CONSTRAINT "nd_experiment_project_project_id_fkey" FOREIGN KEY (project_id) REFERENCES project(project_id) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
[...]

PR #1438, opened 2023-03-06
Tv4g1 issue1436 warnings generating chado schema 1.3.yml

dsenalik:
# Bug Fix

Issue #1436 

## Description
Removes the unused version variables which cause php warnings, update the example usage now that tripal 4 has moved back to the main repo.
```
 [warning] Undefined variable $arguments generate_chado_schema_yaml.php:15
 [warning] Trying to access array offset on value of type null generate_chado_schema_yaml.php:15
```


laceysanderson (approved):
Clean, simple fix -looks good to me! Assuming the tests pass (which there is no reason this PR would break anything) then this can be merged <3

PR #1441, opened 2023-03-09
Tv4g9 issue1439 assert object has attribute

dsenalik:
## Description
PHPUnit has deprecated the use of ```assertObjectHasAttribute()``` and a number of other assertions that operate on class/object properties as described here: https://github.com/sebastianbergmann/phpunit/issues/4601

There was already a ```assertIsObject()``` for all of these cases, so we now do a ```assertIsTrue(property_exists($obj, 'key'))```

In addition to the changes listed by Lacey in the issue, there is one instance of ```assertObjectNotHasAttribute()``` that is updated.

## Testing?
Code review and automated testing should be adequate to test these changes.


laceysanderson (approved):
Looks beautiful! <3 Thanks @dsenalik!

PR #1443, opened 2023-03-10
BUG: Tripal 3 OBO Importer checks for file when none specified in local file field

dsenalik:
# Bug Fix

## Description
A typo in the form validation error message, and the error is caused by an incorrect checking of file existence when no file has been specified, that is, when the "Local File" box is left empty.

Fix the form_set_error typo, and skip check if $uobo_file is not set.

## Testing?
The error likely only happens under PHP8.
Go to the OBO Vocabulary loader,
Under "Ontology OBO File Reference" select "Sequence Ontology"
"Local File" should be empty, leave it empty.
Click the "Update Ontology Details" button.

Error message before this fix
```TypeError: count(): Argument #1 ($value) must be of type Countable|array, string given in form_set_error() (line 1660 of /var/www/prova/drupal-7.92/includes/form.inc).```


spficklin (approved):

spficklin:
Thanks @dsenalik for the fix!

Issue #1444, opened 2023-03-10
tv4g4 OBO importer validation error


dsenalik:
# BUG/ERROR report

### Issue description
This issue is to port the bugfix from pull #1443 to the tripal 4 version. But there may be additional validation problems...

#### Steps to reproduce

Tripal -> Data Loaders -> OBO Vocabulary Loader
Select "The Sequence Ontology"
Make both "Remote URL" and "Local File" blank
Click "Update Ontology Details" button
There should be a validation error, but the blank values are saved.
Now put ```xxx``` for "Local File" and click "Update Ontology Details" button
There should be a validation error that the file does not exist, but again it is accepted.

dsenalik:
Also this is in the log
```Deprecated function: trim(): Passing null to parameter #1 ($string) of type string is deprecated in Drupal\tripal_chado\Plugin\TripalImporter\OBOImporter->formValidate() (line 482 of /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php) ```
~we can copy the ```array_key_exists``` version from tripal 3 to fix this.~
Let's fix this with the null coalescing operator

laceysanderson:
Fixed by merged PR -thanks @dsenalik 🎉

PR #1446, opened 2023-03-10
 BUG: Tripal 4 OBO Importer checks for file when none specified in local file field

dsenalik:
# Bug Fix
## Description
This ports over the changes in pull #1443 which deal with typos in the OBI Importer validation.
In addition to those changes, an incorrect triggering element name is corrected, and one that is not present in the form is removed. This error meant that some of the validation was never performed.

Also we can't pass a null to ```trim()``` so fix this with the null coalescing operator ```??```

## Testing?
Described in issue #1444

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php
+++ b/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php
@@ -474,12 +474,12 @@ class OBOImporter extends ChadoImporterBase {
     $public = \Drupal::database();
 
     $obo_id = $form_state->getValue('obo_id');
-    $obo_name = trim($form_state->getValue('obo_name'));
-    $obo_url = trim($form_state->getValue('obo_url'));
-    $obo_file = trim($form_state->getValue('obo_file'));
-    $uobo_name = trim($form_state->getValue('uobo_name'));
-    $uobo_url = trim($form_state->getValue('uobo_url'));
-    $uobo_file = trim($form_state->getValue('uobo_file'));
+    $obo_name = trim($form_state->getValue('obo_name')??'');
+    $obo_url = trim($form_state->getValue('obo_url')??'');
+    $obo_file = trim($form_state->getValue('obo_file')??'');
+    $uobo_name = trim($form_state->getValue('uobo_name')??'');
+    $uobo_url = trim($form_state->getValue('uobo_url')??'');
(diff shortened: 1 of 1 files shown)


dsenalik:
There is a code climate complexity issue, I thought I would tackle that but I found an underlying issue preventing validation from happening, the default triggering element is `op`. I'll convert to draft until that is resolved

dsenalik:
Issues fixed: 
 - Correct for the triggering element name returned for the default submit button, `add_new_obo` was wrong, `op` is the name for the default submit.
 - Validation in the case where a URL or a file specified, but no vocabulary name, and then "Import OBO File" is checked, previously would show "Job 'Import OBO File' submitted.", now we get the proper validation error.
 - To deal with code climate complexity, two bits of code that were duplicated two times in `formValidate` are moved to two new private functions `private function getVocabID()` and `private function formValidateFile()`.

laceysanderson (changes requested):
✅ Code Review: code looks good 👍 
Functional Review: 
 - ✅  the expected validation error now shows up 👍 
 - ✅  we can still update / delete / add an ontology 👍 
 - ❌  we can't import an existing ontology :-( 

Steps to reproduce:
1. Go to the ontology importer
2. Select and ontology from the "Use a saved OBO reference" fieldset
3. Without changing anything, click the "Import OBO File"

I expect to see a loading job submitted but instead I get the following:
image

dsenalik:
I see what is happening, it's too late today to think, but I will fix it, thanks for the thorough testing @laceysanderson

dsenalik:
@laceysanderson I think that now I tried every combination that I could think of, the validation should now be correct for all cases. Unfortunately the validation function got a bit more complex, let's see if it still passes code climate...

dsenalik:
Code climate thinks it is too complex, but the validation has a lot of if-then statements, which is kind of necessary for validation. I feel that if the function was broken up more, it would actually become more complex, so I think it is okay as it is considering the inherent complexity of the form. 🙈

laceysanderson (approved):
I agree that this is a case to ignore code climate for now. 

Maybe one day in the future we can make this form a bit more streamlined for even make it two separate forms (one for managing ontologies and another for loading them -may be good for controlling permissions as well 🤷 ) but that's out of scope for now :-) 

✅ Code review looks good 👍 
✅ Functional testing is also good -I checked all the combinations I could think of and validation worked as expected :-)

PR #1447, opened 2023-03-10
DRAFT Tripal 4 Field: Synonym Field

srobb1:
created chado synonym field.

The only way I could figure out to make it work is to add a new storage property 'linked_table'. the store_id and store_link only use the base table, but i needed to use synonym.

i am not sure if i needed to include all the elements that i created in the widget, but it is working! 

-- hmm for some reason, i ended up with my sequence field files here. i will try to clean up


srobb1:
ok. i got rid of the sequence length and md5 field.

tripal_chado/src/Plugin/Field/FieldWidget/ChadoSynonymWidgetDefault.php
```diff
@@ -105,12 +105,22 @@ public function massageFormValues(array $values, array $form, FormStateInterface
...
     }
 
-    // Remove any empty values that aren't mapped to a record id.
+    //Remove any empty values that aren't mapped to a record id.
     foreach ($values as $val_key => $value) {
-      # if i need to change the format for name to be sgml, i can do it here
-      $values[$val_key]['synonym_sgml'] = $value['name'];
-      $values[$val_key]['syn_type_id'] = $type_id;
-      $values[$val_key]['pub_id'] = $pub_id;
+      if ($value['value'] == '' and $value['record_id'] == 0) {
```
dsenalik, line 110: ``` if ($value['value'] == '' and $value['record_id'] == 0) {```
should be
``` if ($value['name'] == '' and $value['record_id'] == 0) {```
otherwise the whole thing gets deleted and this is why we can't save the synonym

File: tripal_chado/src/Plugin/Field/FieldWidget/ChadoSynonymWidgetDefault.php [hunk 2]
```diff
@@ -105,12 +105,22 @@ public function massageFormValues(array $values, array $form, FormStateInterface
...
+        unset($values[$val_key]);
+      }
+    }
+
+    foreach ($values as $val_key => $value) {
+      if (isset($values[$val_key]['name'])){
```
dsenalik, line 116: ```      if (isset($values[$val_key]['name'])){```
can be just 
```      if ($value['name']){```
to fix the multiple not working. Because name exists (isset is true) but is just a null string on the last unused field when multiple is enabled.
But this is still not enough to get multiple working...

spficklin:
I'm closing this PR because I submitted new one for this field as PR #1650 .  I explain on that PR why I created a new PR rather than just update this one.

Issue #1450, opened 2023-03-16
Page Title Format - deprecated function str_replace


dsenalik:
# BUG/ERROR report

### Issue description
```Deprecated function: str_replace(): Passing null to parameter #2 ($replace) of type array|string is deprecated in Drupal\tripal\Entity\TripalEntity->replaceTokens() (line 387 of /var/www/drupal9/web/modules/contrib/tripal/tripal/src/Entity/TripalEntity.php) ```

This comes from the **Page Title Format** for gene, which by default is
```[TripalEntityType__label], [bio_data_8_data_0842], [bio_data_8_obi_0100026], [bio_data_8_schema__additionaltyp]```

Two of the fields in this example will be NULL, so one of the error messages comes from ```[bio_data_8_obi_0100026]``` (organism, probably shouldn't be NULL but is) and one comes from ```[bio_data_8_schema__additionaltyp]``` (none present)

A simple fix for this particular error.

#### Steps to reproduce
 1. Create an organism (I filled in something for every field)
 1. Create a gene with this organism, enter a **Name** and **Identifier** and for **Sequence Length** enter a zero (different bug to be fixed elsewhere) and save.
 1. Edit the gene, change nothing, and save again.
 1. Look in the log and find two copies of the error message shown above.

dsenalik:
Fixed with merged #1452

PR #1452, opened 2023-03-16
Tv4g1 issue1450 deprecated str replace

dsenalik:
# Bug Fix

## Description
A one-line change to avoid passing a NULL to ```str_replace()``` as described in issue #1450. Procedure to see the error shown there can be used for testing.


laceysanderson (approved):
Clean simple fix -thanks @dsenalik

PR #1457, opened 2023-03-23
Tv3g9 issue1456 drupal7.95 fix automated testing - Abandoned, wrong starting branch

dsenalik:
# Bug Fix

## Description
Try to find what is causing the error message described in the issue when running testing under the current drupal version 7.95
Commit xxx : separate the module enables to know exactly which one is causing it

## Testing?
Testing is with automated testing.


dsenalik:
I started from the wrong branch! Let's start over

PR #1319, opened 2022-11-22
DRAFT Tripal 3 chado linker fields

dsenalik:
# New Feature

## Description

This branch will add several new tripal fields to link between content types as described in issue #1318:

| proposed cvterm | field name |
| --- | --- |
| [NCIT:C25391 (Analysis)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C25391)<sup>3,4</sup> | chado_linker__analysis |
| [NCIT:C60819 (Assay)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C60819)<sup>4</sup> | chado_linker__assay |
| [NCIT:C43376 (Biologically-Derived Material)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C43376)<sup>4</sup> | chado_linker__biomaterial |
| local:contact | chado_linker__contact (**existing**) |
| [NCIT:C73619 (Feature)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C73619)<sup>4</sup> | chado_linker__feature |
| EDAM [data:1274 (map)](https://edamontology.github.io/edam-browser/#data_1274) | chado_linker__featuremap |
| [NCIT:C14250 (Organism)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C14250) | chado_linker__organism |
| [NCIT:C47885 (Project)](https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C47885) | chado_linker__project |
| [NCIT:C48288 (Stock)]( https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C48288)<sup>4</sup> | chado_linker__stock |
[...]

PR #1458, opened 2023-03-23
tv3g9-issue1456-separate-installs-for-debugging

dsenalik:
# Bug Fix

## Description
Try to find what is causing the error message described in the issue when running testing under the current drupal version 7.95
Commit https://github.com/tripal/tripal/pull/1458/commits/601c3042886ddfb29e43a25c58a5e0d831bbafb3 : separate the module enables to know exactly which one is causing it

## Testing?
Testing is with automated testing.


dsenalik:
How embarassing, this is a problem with #1319, the cv ncit is added with update 7332 but terms are added before that is run. Fix it there not here

PR #1459, opened 2023-03-26
Tripal 4 - GFF3 Loader Unit tests and fixes

### LINKED ISSUE #1379: Undefined function error in gff loader: Tripal 4
## BUG/ERROR report
When loading a GFF I get an undefined function error if "Start Line Number" field is blank.  Error does not occur if I put "1" in the field.  

### System information
Version: t4d8 docker image

### Issue description
Error: Call to undefined function Drupal\\tripal_chado\\Plugin\\TripalImporter\\form_set_error() in /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php on line 506 

#### Steps to reproduce

minimal test gff:

##gff-version 3
chr1_h1	.	chromosome	1	339755681	.	.	.	ID=chr1_h1;Name=chr1_h1
chr2_h1	.	chromosome	1	270168396	.	.	.	ID=chr2_h1;Name=chr2_h1
chr3_h1	.	chromosome	1	133856195	.	.	.	ID=chr3_h1;Name=chr3_h1
chr4_h1	.	chromosome	1	55278592	.	.	.	ID=chr4_h1;Name=chr4_h1

risharde:
@laceysanderson @spficklin 

- This code has the latest GFF3 loader modifications like using the new Tripal database api (Tripal DBX) for most db work.
- It also contains approximately 48 checks within a single test function utilizing the GFF3 loader and the fixture files included.
- I also tried to clean up comment code I used while working on this branch and the debug print_r statements I used as well.
- I merged in the latest 4.x and fixed a minor merge conflict.


laceysanderson:
I'm excited for the progress!

It looks like the tests are failing in our environment due to an assumption somewhere that that chado schema is named `chado`.

```
1) Drupal\Tests\tripal_chado\Functional\GFF3ImporterTest::testGFFImporterSimpleTest
The gff_strand.gff produced an 
      exception which should not happen since it is a valid file. SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "chado.featureloc" does not exist
LINE 1: SELECT * FROM chado.featureloc fl 
                      ^: SELECT * FROM chado.featureloc fl 
        LEFT JOIN chado.feature f ON (fl.feature_id = f.feature_id)
        WHERE uniquename = :uniquename LIMIT 1; Array
(
    [:uniquename] => FRAEX38873_v2_000000010
)

Failed asserting that false matches expected true.

/var/www/drupal9/vendor/phpunit/phpunit/src/Framework/Constraint/Equality/IsEqual.php:96
/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php:923
/var/www/drupal9/vendor/phpunit/phpunit/src/Framework/TestResult.php:728
```

It looks like there are just a few places in your test where you hardcoded `chado.` instead of `1:`. For example, see this line https://github.com/tripal/tripal/pull/1459/files#diff-f17a9509ef0886e4038bdb267c189059eda6c47a1ee8722cc37226bc1fe84906R894.

risharde:
Thanks Lacey! Good catch, will try to fix as soon as possible! Thank you
for the heads up!

risharde:
Updated those calls to chado, awaiting the results of the docker tests

dsenalik:
I was curious to try this, sadly it didn't work on my file. Here are all of the "bad" lines from the output
```
 [warning] Undefined array key "add_only" GFF3Importer.php:553
 [warning] Undefined array key "update" GFF3Importer.php:554
 [warning] Undefined array key "uniquename" GFF3Importer.php:1296
 [error]  ERROR: The "Target" attribute is incorreclty formatted for the feature "" 
 [warning] Undefined array key 0 GFF3Importer.php:548
 [warning] Trying to access array offset on value of type null GFF3Importer.php:548
 [warning] Undefined array key "add_only" GFF3Importer.php:553
 [warning] Undefined array key "update" GFF3Importer.php:554
```
Some of these warnings are fixed in pull #1369

risharde:
I had similar issues getting the FASTA loader ported over just about last
week Doug, I'll look into these, but I won't get around to this until
likely next week - thank you for providing the error lines! Sorry for the
disappointment on getting it to run!

I'm guessing this is PHP 8 related

dsenalik:
It is php 8 related indeed.
I wanted to provide a test gff3, but I can only replicate most of my problems on a test docker with a bunch of fields removed, on a normal one there are only two warnings, and the data is loaded successfully. So you can ignore the rest of my error messages, it's just these two:
```
 [warning] Undefined array key "add_only" GFF3Importer.php:553
 [warning] Undefined array key "update" GFF3Importer.php:554
```

And for what it's worth, the incorrect target attribute was `NA` but I don't have any Targets, it's something somehow related to damage I did to that docker version.

risharde:
Thanks @dsenalik appreciate the clarification! So to be specific, this is form related and so it might be a good idea for me to consider a form test in the process so I'll try to explore that one!

dsenalik:
I pushed the simple fix for this with null coalesce operator, and enhanced the error message should it ever strike again, I hope you don't mind 😬.

tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php
```diff
@@ -0,0 +1,1333 @@
...
+      $gff3_importer->create($run_args, $file_details);
+      $gff3_importer->prepareFiles();
+      $gff3_importer->run();
+      $gff3_importer->postRun();
+    } 
+    catch (\Exception $ex) {
+      $message = $ex->getMessage();
+      $has_exception = true;
+    }
+
```
laceysanderson, line 248: We want an assert here to make sure that $has_exception is true when we expect it to be true. That should be outside the catch section.

tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php
```diff
@@ -0,0 +1,1333 @@
...
```
laceysanderson: We want an assert here to make sure that $has_exception is false since here the GFF3 importer should allow commas in the ID. That should be outside the catch section.

laceysanderson: A quick check that the feature with the comma in the id was inserted would also be really good here.

tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php
```diff
@@ -0,0 +1,1333 @@
...
+      $gff3_importer->create($run_args, $file_details);
+      $gff3_importer->prepareFiles();
+      $gff3_importer->run();
+      $gff3_importer->postRun();
+    } 
+    catch (\Exception $ex) {
+      $message = $ex->getMessage();
+      $has_exception = true;
+    }
+    $this->assertEquals($has_exception, false, "The GFF3 loader should reverse the values automatically but somehow produced an exception which is an error");
```
laceysanderson, line 349: I would add an addition assert here that the start and end are now what we expect to make sure the GFF3 importer fixed them appropriately.

tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php
```diff
@@ -0,0 +1,1333 @@
...
+      $has_exception = true;
+    }
+    $this->assertEquals($has_exception, true, "Should not complete when there 
+      is invalid phase value (in this case a number > 2) but did not throw 
+      error which should have happened.");
+
+    
+    /**
+     * Test that when checked, when phase is specified for CDS
+     * TODO: Maybe create a test for phase being a . for a CDS which should in theory fail?
```
laceysanderson: >TODO: Maybe create a test for phase being a . for a CDS which should in theory fail?

I don't actually think this should fail 🤔 it would mean the GFF3 file doesn't make biological sense but it does still match the format.

tripal_chado/tests/src/Functional/Plugin/GFF3ImporterTest.php
```diff
@@ -0,0 +1,1338 @@
...
+      $gff3_importer->prepareFiles();
+      $gff3_importer->run();
+      $gff3_importer->postRun();
+    } 
+    catch (\Exception $ex) {
+      $message = $ex->getMessage();
+      $has_exception = true;
+    }
+    $this->assertEquals($has_exception, false, 'Unescaped tag should not throw an error but did.');
+
```
laceysanderson:
```suggestion
  // @TODO check that the feature with the comma in the ID was inserted properly.
```

laceysanderson (approved):
Looks good -Thanks @risharde 🎉

risharde:
Like yay!!!

Issue #1449, opened 2023-03-14
ChadoStorage changes for records linked through a linker table


dsenalik:
## Tripal 4 Feature Request
* This feature does attempt to solve an existing problem with Tripal
* This feature is "semi-urgent" 😀

### Description
Sofia and I are working on different tripal 4 fields with something in common: linker tables.
Basically, from the base table (e.g. feature) there is a linker table (e.g. feature_synonym) that links to one **or more** records in another "base table", "linked_table", "object table" or whatever we should call it (in this example the table is synonym).

When adding a record for the field, we need to do two things (in this order): 
1. UPDATE or INSERT a record for the synonym in the chado.synonym table (my field won't have to do this)
2. (DELETE +) INSERT a record in the linker table (feature_synonym) that links feature to synonym.

We each have similar updated versions of ChadoStorage.php  #1447 #1431 (I copied from Sofia😀) but this is not yet working. I'm creating this issue so others know about this and hopefully can make suggestions.

When we have something reasonable to propose, maybe we can make a pull request just for ChadoStorage.php from this issue and all fields can use it?

### More details

We are defining the synonym table with ```store_id``` and two ```store``` actions. We can't use ```store_pkey``` because we need to make sure the synonym is stored before the linker table, and base tables get handled first. What this means is that there are two "base tables" (feature and synonym). Importantly, there can be more than one "base table" record for synonym, whereas for the "real" base table there is only one. Currently ```store_id``` is hardcoded with a delta of ```[0]``` so only the first synonym works.

Updates to follow...

dsenalik:
The current crackpot idea is to introduce a new action ```store_oid``` to store object base table id, along with storing delta values in the base table array.
https://github.com/tripal/tripal/pull/1431/commits/4ce27d5736eb8a48b23b7bf987c4123172a472a2

dsenalik:
Today's suffering involves the fact that two different records might use the same synonym. Like the same gene from two different assemblies. This adds a whole new nightmare of complexity, because we need to check if the synonym we want to insert already exists, otherwise it fails because of the unique constraint. 
I added a ```function selectOrInsertChadoRecord()``` and it does not make me happy, way too complicated, I have to convert fields into conditions, so much could go wrong. Well I'll commit it anyway 💩 https://github.com/tripal/tripal/pull/1431/commits/b9eea9059ebd46e36dd510218f2a29c5fff1801f

laceysanderson:
[...]

Issue #1398, opened 2023-01-25
Tripal 4 property field can't handle two properties at the same time


dsenalik:
### System information

* Tripal Version: tv4α docker container

### Issue description
I'm trying to understand fields, but playing with an existing field is not behaving like I would expect. One property is leaking over into another.

#### Steps to reproduce
1. Create a project to play with
2. Go to ```/admin/structure/bio_data``` and for Project select "Manage Fields"
3. +Add Field
4. Select field type "Chado Property" and label it "Library", "Save and continue"
5. Base table select project
6. Number of values set to Unlimited, "Save field settings"
7. Controlled Vocabulary Term select ```Library (local:library)```, "save settings"
8. Repeat steps 2 through 7, this time use a DIFFERENT cv term, I used "Book" CV term ```Book (TPUB:0000004)```
9. Go to the entity page for the project, and click "Edit"
10. Under "Library" enter "National Ag Library", and Save
11. Now click "Edit" again
12. Under "Book" enter "Harry Potter and the Tripal Daemon", and hit save
13. This is the result. The second property did not save into chado.
![tv4](https://user-images.githubusercontent.com/8419404/214701543-41d7b199-f030-4b2f-b137-1153a7f7a303.png)
14. Edit again and add a second "Book"
![tv4-2](https://user-images.githubusercontent.com/8419404/214702033-70530383-29cd-47cb-9fee-047111363e97.png)

15. Edit again, fix the Book names, and add a second "Library", and save
![tv4-3](https://user-images.githubusercontent.com/8419404/214702065-c9bfa256-28f2-4607-80ae-1dd8beddd345.png)
```
sitedb=> select * from projectprop;
 projectprop_id | project_id | type_id |        value        | rank 
----------------+------------+---------+---------------------+------
              4 |          2 |    2938 | Harry Potter 2      |    1
              5 |          2 |     113 | National Ag Library |    0
              6 |          2 |     113 | Library of Congress |    1
(3 rows)
```

dsenalik:
This is not directly the cause, but something I found.
In ```tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php``` the variable ```$chado_table``` is not defined when processing a join, and gets a value from whatever was processed previously, I think this line needs to be added near line 659:
```
  if ($action == 'join') {
**  $chado_table = $prop_storage_settings['chado_table']; // ** add this line
    $chado_column = $prop_storage_settings['chado_column'];
[...]

PR #1463, opened 2023-03-30
Expanding ChadoStorage to better support linker tables

laceysanderson:
# Tripal 4 Core Dev Task 

Issue #1449 

## Description
This PR was started when working with @srobb1 and @dsenalik on Tripal 4 Fields for linking tables. It was becoming clear that the current ChadoStorage implementation was not meeting the needs here. Most notably these were a few of the core restrictions:

1. store_id made assumptions that it would only be used for the base table of the field.
2. we were not able to make a "double hop" query with two joins through a linker table (e.g. feature > feature_synonym > synonym)
3. we were having trouble only inserting if the record didn't already exist (i.e. select or insert functionality)
4. records added to chado independently of the fields were not being added to the field display (e.g. adding properties via an importer then refreshing a Tripal content page wasn't displaying these properties, described within #1398)

## Design Discussion

Originally we thought we needed to fix these problems in order to support this functionality:

 - store_id assumes base table
 - store_link doesn't join
 - not able to detect new records added to chado outside of the UI
 - store_link doesn't support double-hops including insert or select of base record at the right side of the relationship.

We have since realized a couple of things:

1. detecting records added outside the UI is the job of publishing, not chadostorage.
2. using joins here might incur a performance hit and is not needed once we have good publishing
3. handling of the base record at the right side of a double-hop relationship may incur more complexity to chadostorage than it's worth... instead we should let the widget handle insert/select of this record and chado storage handle the linking table.

This PR embodies these new design ideas. Specifically, the changes are:

 - remove the assumption of there being a single base table that is the same across fields.
 - introduce left/right table notation to the store_link for better describing of the relationship
 - backwards compatible check for store_link to support old style properties with this action as much as possible.
 - extra documentation in chadostorage to separate out and explain the actions.
 - fixes to ensure the record_id for core tables is set for secondary tables as early as possible.
 - Kernel tests focused on property-style fields and linker-style fields.
 
## Testing?

There is automated testing in this PR to test the specific use cases. This PR does not have specific fields yet so you cannot test manually.

Tests:
  - ChadoLinkerPropertyDefaultTest: Tests the new and old style of store_link for property-style fields
  - ChadoLinkerTableTest: Tests the new style of store_link for double-hop style linker tables (e.g. feature_synonym). Currently only tests insert as this test is not yet complete.


dsenalik:
Excellent! 🎉 It sounds like this will fix the problem I noted in this comment https://github.com/tripal/tripal/issues/1398#issuecomment-1452279615 of not detecting records entered directly into the db.

dsenalik:
Testing with the `chado_linker__prop` field, it looks to me like the `store_link` property for this field should become
before
```
       new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id',  $link_term, [
         'action' => 'store_link',
         'chado_table' => $prop_table,
         'chado_column' => $prop_fk_col,
       ]),
```
after
```
      new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id',  $link_term, [
        'action' => 'store_link',
        'left_table' => $base_table,
        'left_table_id' => $base_pkey_col,
        'right_table' => $prop_table,
        'right_table_id' => $prop_fk_col,
      ]),
```
@laceysanderson If this seems correct (I tested, before there were errors, now it works) I could add this change to this branch. We will have to update `schema__additional_type` also!

dsenalik:
A minor news update, with the test procedure described in issue #1398, I still see that error on this branch at step 13
Then at step 14 I get an error
The item cannot be saved because the following values are missing. Name.

spficklin:
The changes to `store_link` in the current state of this PR break creation of content types that use the prop tables to store the types.  I figured it would be better to fix it here rather than try to fix it on the publish issue #1515 .  I think we just need to fix the problems with the ChadoStorage than I can go back to publishing. I'm going to take a stab at it.

spficklin:
First comment on one of the issues mentioned above.  The issue is this:  

- You add a Chado Property field to the gene content type for comments
- You create a gene and you add two comments "Note1" and "Note2".
- You then add a comment, "Note 3",  outside of Tripal (e.g. via an SQL statement)
- The page doesn't pick up the new "Note 3" comment that was added outside of Tripal.

The fact the "Note 3" doesn't load is not a bug in ChadoStorage that needs fixing.  That's because when the values are added to the field, a corresponding record is added to the field table in Drupal (e.g. `tripal_entity__field_comment` table).  So "Note 1" and "Note  2" both get a record in the Drupal field table.   Publishing in Tripal 4 is different from Tripal 3 because entities aren't tied to a back-end, the field data is actually getting published, not the entity.    So, if the ChadoStorage class returns values that are not published (e.g. "Note 3") then it's doing more than it should.  It should only return what is instructed by the $values array that gets passed into the `findValues()` function and that $values array gets created by what is present in the Drupal field table--and that table only knows about two records: "Note 1" and "Note 2".  If we want "Note 3" to be loaded then we need to add  new record to the Drupal field table for that new value.  Then it will show up on the page.  That's "publishing" of the field values.   

I think the correct solution for this is not to adjust the ChadoStorage class but do one of the following:

1.  Make the site admin publish new field data anytime data is loaded.  Perhaps this could be part of the TripalImporter class.
2. Add a tab at the top of each page, like we had in Tripal v3 that let the site admin reset the cache.  We could have a similar tab that would let the site admin republish (or find new values) for the fields on the page.  
3. Add a cron job that runs a "publish" for existing entities to pick up new data values.

spficklin:
Another item to consider.... I'm not sure we need to update the `store_link` action when there are linker tables like `feature` <-> `feature_synonym` <-> `synonym`.   I started to try and work on the ChadoStorage to support this and I realized I can make the synonym field work without making changes to ChadoStorage.  

The issue is that the synonym must exist first in the synonym table.  So if it doesn't, we want to create it.  This creates some real complexity for the ChadoStorage class because it needs to figure out that if a user is adding a new synonym that doesnt exist yet, that it needs to first add that synonym before it can add the linker code.  

We can easily get around this if the field widget does the insert of the synonym first and then set the ID for it in the property that represents the linker.  I have made a working version of tje field in #1447 that does not require any changes to ChadoStorage to work.  I think given the complexity of making ChadoStorage smart enough to do this, that it's really not to hard to make the field do a little housekeeping before an insert/udate.   My suggestion is to leave ChadoStorage as is.   

I haven't yet made a PR for the synonym field because I have one bug with multiple instances. Once I fix that, then I'll make a push to PR #1447 and then you can review it and see how I got around the problem.

laceysanderson:
🤯 I feel like an important missing piece just dropping into place! <3

laceysanderson:
After discussing with @spficklin today, we have decided that Chado Storage is becoming too complicated and more importantly -more complicated then it needs to be.

The plan moving forward is:
- ChadoStorage will support the following cases
    - insert/update columns in the base table (e.g. stock)
    - insert/update columns accessed through a single-hop (e.g. stockprop)
    - select across multiple hops for read-only access
    - read-only access for the base-table (e.g. stock)
- ChadoStorage will NOT:
    - find unpublished records from Chado. For example, if data is imported or pre-existing in chado, then you need to publish first. This will not be done by ChadoStorage.
    - insert/update records in a table accessed through a linker table (e.g. feature => feature_synonym => synonym). In this case the widget will be expected to add/select the synonym record and then add the id to the property. ChadoStorage will manage the feature => feature_synonym part through.

I will be updating this PR to match this new design.

spficklin (approved):

Issue #1432, opened 2023-02-28
$base_schema_def['foreign keys']  is not returning an array when adding a new Chado Organism Reference


srobb1:
# BUG/ERROR report

## System information
Mac with Docker Desktop 4.17.0 (99724) 

### Tripal Version:
4.x
tripalproject/tripaldocker:latest
289cbfd0aa8c

### Issue description
When I try to add a new field, Chado Organism Reference, to a gene content type I get this error.
```
Message | TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in array_keys()
(line 74 of /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/obi__organism.php)
#0 /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/obi__organism.php(74): array_keys(NULL)
```

this is line 74 in /tripal/tripal_chado/src/Plugin/Field/FieldType/obi__organism.php
```$base_fk_col = array_keys($base_schema_def['foreign keys']['synonym']['columns'])[0];```

the `$base_schema_def['foreign keys']` is not returning an array, it is returning nothing.

#### Steps to reproduce
Home  -> Administration -> Structure -> Tripal Content Type -> Edit Gene -> Manage Fields -> Add Field -> Chado Organism Reference

**Comment by dsenalik** [comment]
Lacey was asking me yesterday why I had this block in my field

```
    // If we don't have a base table then we're not ready to specify the
    // properties for this field.
    if (!$base_table) {
      $record_id_term = 'SIO:000729';
      return [
        new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $record_id_term, [
          'action' => 'store_id',
          'drupal_store' => TRUE,
        ])
      ];
    }

```

and it is because I copied it from the `chado_linker__prop` field, because when adding my field I had a similar error because there was no base table defined.

If I add a similar block to this field, it progresses beyond this error and allows me to add this field.

i.e. after this line
```
   $base_table = $settings['base_table'];
```
I added this
```
    // If we don't have a base table then we're not ready to specify the
    // properties for this field.
    if (!$base_table) {
      $record_id_term = 'OBI:0100026';
      return [
        new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $record_id_term, [
          'action' => 'store_id',
          'drupal_store' => TRUE,
        ])
      ];
    }
```

[...]

PR #1467, opened 2023-04-04
Tripal 4 bugfix for  schema__additional_type field

### LINKED ISSUE #1466: Fix from pull #1433 applies to other fields also
# BUG/ERROR report

### Tripal Version: 4.alpha1-dev (today's version)
* Drupal Version: 9.5.4-dev (Tv4 docker)
* PostgreSQL Version: 13.9
* PHP Version: 8.1.16

### Issue description
The problem noted in issue #1432 and fixed in pull #1433 applies to another field, `schema__additional_type`

#### Steps to reproduce
 1. Go to `/admin/structure/bio_data`
 1. For `Biological Sample` select `Manage fields`
 1. `+ Add field`
 1. Select `Chado Type Reference`
 1. For `Label` enter anything
 1. `Save and continue`
 1.  This error is reported:
`Error message
There was a problem creating field BioType: Cannot create a StorageProperty object without a property formatted term:`

dsenalik:
# Bug Fix
## Description
Applies to field `schema__additional_type`. The change is to make sure `$base_table` and `$type_table` are both defined before continuing to define all the properties for the field. See also issue #1432 and pull #1433 for a similar change on the `obi__organism` field.

I added null coalescing to `$type_table` and `$type_column` because these will be undefined between steps of manually adding the field, causing chaos otherwise, and the partially completed field can't be removed. This simple fix prevents that.

## Testing?
 1. Go to `/admin/structure/bio_data`
 1. For `Biological Sample` select `Manage fields`
 1. `+ Add field`
 1. Select `Chado Type Reference`
 1. For `Label` enter anything
 1. `Save and continue`
 1.  This error is reported before this change:
`Error message
There was a problem creating field BioType: Cannot create a StorageProperty object without a property formatted term:`
and after, the field can be added.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/schema__additional_type.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/schema__additional_type.php
@@ -68,8 +68,20 @@ class schema__additional_type extends ChadoFieldItemBase {
     // Get the Chado table and column this field maps to.
     $storage_settings = $field_definition->getSetting('storage_plugin_settings');
     $base_table = $storage_settings['base_table'];
-    $type_table = $storage_settings['type_table'];
-    $type_column = $storage_settings['type_column'];
+    $type_table = $storage_settings['type_table'] ?? '';
+    $type_column = $storage_settings['type_column'] ?? '';
+
+    // If we don't have a base table then we're not ready to specify the
+    // properties for this field.
+    if (!$base_table or !$type_table) {
+      $record_id_term = 'SIO:000729';
+      return [
+        new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $record_id_term, [
+          'action' => 'store_id',
+          'drupal_store' => TRUE,
+        ])
+      ];
+    }
 
     // Get the the connecting information about the base table and the
     // the table where the type is stored.  If the base table has a `type_id`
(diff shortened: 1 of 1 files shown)


dsenalik:
I changed this pull request to draft because testing this field on a gene content type, and trying to edit the gene, I see these errors and a WSOD

`InvalidArgumentException: Property term_name is unknown. in Drupal\Core\TypedData\TypedDataManager->getPropertyInstance() (line 196 of /var/www/drupal9/web/core/lib/Drupal/Core/TypedData/TypedDataManager.php).`

`Warning: Undefined array key "type_table" in Drupal\tripal_chado\Plugin\Field\FieldWidget\schema__additional_type_widget->formElement() (line 37 of /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldWidget/schema__additional_type_widget.php) `

dsenalik:
I withdraw this comment and remove the DRAFT designation. These errors only occur under the `tv4g1-issue1449-chadostorage-linkertables` branch.

spficklin (approved):

spficklin:
Looks good to me!  I recognize this fix.    I wonder though if we need a better solution then to put this if statement repeatedly in multiple fields.... I'll leave this unmerged to get the attention of @laceysanderson for comments. But I think it's ready for merging.

laceysanderson:
Yes, I noticed this block is repeated everywhere which also gives me pause. I say we merge this PR and open an issue for discussion about how we can go about fixing this. I'm not deep enough in fields yet myself to know why this is necessary 🤔 Ideally, if possible we would ensure tripalTypes is not called if the info is not yet instantiated (i.e. if $base_table is empty). It might be as easy as moving this if into the TripalField or ChadoField base classes.

dsenalik:
For further discussion see #1469

PR #1470, opened 2023-04-06
Tripal 4 issue1468 standard naming for tripal fields

### LINKED ISSUE #1468: Rename fields using standard tripal 4 naming scheme

## Pull #1470

Tripal fields in core no longer use a name based on a cv term as in Tripal 3
Link to [Naming convention documentation](https://tripaldoc.readthedocs.io/en/latest/dev_guide/module_dev/fields.html#naming-convention)
So, a few fields currently defined using this old method need to be renamed as follows:

`chado_linker__prop` ⟹ `ChadoLinkerPropertyDefault`
`obi__organism` ⟹ `ChadoOrganismDefault`
`schema__additional_type` ⟹ `ChadoAdditionalTypeDefault`

This will apply to both file names and internal `id` and function names

### LINKED ISSUE #1481: Tripal 4 additional type field bug with "term_name"
# BUG/ERROR report

### Issue description
Field `schema__additional_type`  cannot be manually added

#### Steps to reproduce
1. Fire up a tripal 4 docker
2. Create an organism
3. Create a gene
4. We will add the field `schema__additional_type` to gene content type, with **cardinality unlimited** and using any random cv term. While adding, just before defining this term there is `Warning: Undefined array key "type_table" in Drupal\tripal_chado\Plugin\Field\FieldWidget\schema__additional_type_widget->formElement() (line 37 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldWidget/schema__additional_type_widget.php). `
5. Edit the gene created in step 3 and enter something for the new field, and save. The following error appears
`InvalidArgumentException: Property term_name is unknown. in Drupal\Core\TypedData\TypedDataManager->getPropertyInstance() (line 196 of core/lib/Drupal/Core/TypedData/TypedDataManager.php). `

For the first warning, I am seeing the widget table `formElement()` being called twice, the second time with less "information", and it appears likewise in the FieldType
20230417fw
The reason for this is that the first case is the default field already on the gene content type `bio_data_8_schema__additionaltyp` and the second case is from the newly added field. What happens is that `type_table` and `type_column` are not defined because the field is lacking the capability to do this.

dsenalik:
# Tripal 4 Core Dev Task

## Description
This pull renames Tripal fields to use the new [naming convention](https://tripaldoc.readthedocs.io/en/latest/dev_guide/module_dev/fields.html#naming-convention).
Because of this, it is very destructive 💣 to an existing docker as files for three fields will now suddenly be gone, and chaos will ensue. As a result, this should not be merged until just before we create the alpha2 release and a new docker. This is the only reason it is labelled DRAFT.

### Method A - use an existing docker container (not recommended 😬)
Testing is difficult because we have removed code for existing fields in the docker. If we run unmodified, we break everything!
1. Before checking out this branch, we need to delete these three fields from all content types. I listed what content types they were on with
`\dt tripal_entity*additional*` (5, 7-9, 10, 12-15, 17-24)
`\dt tripal_entity*obi_0100026*` (8)
`\dt tripal_entity*linker*` (none)
and then manually deleted them from each at e.g.
`/admin/structure/bio_data/manage/bio_data_5/fields`
1. remove obi_0100026 (organism) from 8 (gene)
1. also remove the uo_unit field from 11 (Physical Map) and 16 (Genetic Map)
1. also remove the Infraspecific type from 1 (Organism)
1. and remove both efo__substrate_type and ncit__technology_pla from 27 (Array Design)
1. Rebuild cache
1. Checkout this branch
1. Rebuild cache
1. put the "Chado Organism Reference" field back on gene, base table feature of course (I had to set a default organism, maybe a different bug)
1. put a "Chado Type Reference" field on gene, we want to replace the field deleted so use gene (SO:0000704)
1. put a "Chado Property" field on gene, I used "RNA" as the term but any should do
1. now go to a gene page and edit and put values for all three fields
The item cannot be saved because the following values are missing. Type_id. 
Apparently replacing the type field, even with the same term, doesn't work. Confirmed that this happened exactly the same before changes in this pull request.

### Method B - create a new docker container
1. `git clone https://github.com/tripal/tripal tripal-1468`
1. `cd tripal-1468`
1. `git checkout tv4g1-issue1468-standard-naming-for-tripal-fields`
1. `sudo docker build --tag=tripaldocker:testing-1468 --build-arg drupalversion="9.5.x-dev" --file tripaldocker/Dockerfile-php8.1-pgsql13 ./`
1. `sudo docker run --publish=80:80 -tid --name=testing-1468 --volume=$(pwd):/var/www/drupal9/web/modules/contrib/tripal tripaldocker:testing-1468`
1. `sudo docker exec testing-1468 service postgresql restart`
1. Now see if things worked. Go to http://localhost and login
1. Create some various content pages at http://localhost/bio_data/add : organism analysis gene 
1. Add fields to gene at http://localhost/admin/structure/bio_data<br>
1. "Chado Organism Reference" (again you need to set a default organism)
1. "Chado Property"
1. "Chado Type Reference" - see this link in the comments https://github.com/tripal/tripal/pull/1470#issuecomment-1529723386 - for "Chado Base Table" select `feature` and for "Type Table and Column" select `featureprop → type_id`, and set cardinality to `Unlimited`. And then for the "Controlled Vocabulary Term" I selected `DNA (SO:0000352)`
1. Go to the gene and edit, adding values for the new fields and confirm they "stick" and even better make sure they are in the appropriate chado table in the db.


spficklin:
Hey @dsenalik just checking... is this still a draft or can we review it?

dsenalik:
This is still a draft because I still cannot manually add an additional type field to a gene. I have the selector for the type table and column working, but it doesn't "save". This is not a new problem with this pull request, it was a problem before I renamed the fields, see https://github.com/tripal/tripal/issues/1481, but I would like to fix it so that you can actually test the thing.

dsenalik:
I am removing the draft designation on this pull request.
I have updated the `ChadoAdditionalType` field to include form elements to select the required table and column.
I am now able to add this field to a `Gene` content type.
AT-1-add-field

Once the base table is selected, an AJAX callback generates a list of tables and columns with appropriate foreign keys.
AT-2-ajax

To make the user interface as simple as possible, these are selected as a combined element table plus column. The most common use case, where the type_id column is in the base table, is presented at the top  of the list, otherwise they are in alphabetical order.
AT-3-table-plus-column

An example of using the widget to add multiple (made-up) gene types
AT-4-widget

laceysanderson (approved):
Perfect <3 👍 This is ready to merge once tests pass... we just have a merge queue going from a productive morning ;-p

laceysanderson (approved):
This works beautifully on my end ✅ You'll see I made a new commit. This moves your changes that used to be to prepare into the new YML structure. This screenshot shows it worked as these were automatically attached:
Screenshot 2023-05-08 at 5 14 01 PM

PR #1474, opened 2023-04-09
Tripal 4 Issue 1472 $form_state bugfix

### LINKED ISSUE #1472: Tripal 4 code review bug in OBO importer
# BUG/ERROR report

### Issue description
I noticed this in the code, but didn't encounter an error until I copied and pasted it elsewhere. Error is
`Error: Cannot use object of type Drupal\Core\Form\FormState as array`

Line 464 of `tripal_chado/src/Plugin/TripalImporter/OBOImporter.php`:
```
$form_state['rebuild'] = TRUE
```
should be
```
$form_state->setRebuild(True);
```

#### Steps to reproduce
Not sure, this is found with code review 🧐

dsenalik:
# BUG/ERROR report

### Issue description
I noticed this in the code, but didn't encounter an error until I copied and pasted it elsewhere. Error is
`Error: Cannot use object of type Drupal\Core\Form\FormState as array`

Line 464 of `tripal_chado/src/Plugin/TripalImporter/OBOImporter.php`:
```
$form_state['rebuild'] = TRUE
```
should be
```
$form_state->setRebuild(True);
```

#### Steps to reproduce
Not sure, this is found with code review 🧐

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php
+++ b/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php
@@ -461,7 +461,7 @@ class OBOImporter extends ChadoImporterBase {
         \Drupal::messenger()->addMessage(t("The vocabulary @vocab has been added.", ['@vocab' => $obo_name]));
       }
       else {
-        $form_state['rebuild'] = TRUE;
+        $form_state->setRebuild(True);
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
That's a clear bug and such a small, clear change that it doesn't need to be tested.

This is a change between D7 and D9 so this line just got missed during the upgrade. There may be more as upgrading form API stuff required very few changes making it easy to miss the ones that it does need.

Issue #1477, opened 2023-04-11
Phpunit workflows not running on PRs from a forked repo


laceysanderson:
# BUG/ERROR report

### Issue description

Something seems to have changed recently with github actions. In the past collaborators have been able to submit a PR from a forked version of our repo and our workflows would all run in this repo to validate the PR. This is needed because we can't merge PRs until they pass certain required checks.

However now no workflows are run on PRs from forked repositories. [This GitHub doc](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) says we can approve these workflows to run. However, in practice we can only approve running of the assert branch name workflow and not our automated testing.

Based on [this GitHub discussion](https://github.com/orgs/community/discussions/25012), we can change our action to pull_request for the phpunit workflow to enable it to run in our repo for PRs triggered by forked repos... however, I believe this will only be triggered when the PR is first made and not on any additional commits which improve or fix things in the PR...

I have seen this in both @srobb1 PR #1419 and @pdtouch PR #1476. I was able to merge @pdtouch's PR but only after forcing override of branch protections which is not ideal.

The weird thing is I could have sworn this used to work. Specifically, I remember the phpunit workflow running on #1419 when it was first created and again after Sofia added commits. It's only now after I resolved the merge conflicts that it won't run.

#### Steps to reproduce

1. Fork the tripal/tripal repo
2. make a small change on a branch and submit a PR
3. The automated testing workflows will not run:
 
<img width="837" alt="Screen Shot 2023-04-10 at 7 00 30 PM" src="https://user-images.githubusercontent.com/1566301/231028246-ebb85a19-47b5-42ba-99e4-8bf146ef7842.png">

pdtouch:
Lacey,
Is it possible that other code in my repo have bugs that is causing this ?
I have incomplete ChadoLocalSourceData field in my pdtouch repo.
Thanks
Sukha

laceysanderson:
I don't think so @pdtouch bugs should never stop workflows from running unless they are in the workflow yml files and even then -it is still triggered but will fail. Also only the branch you made a PR with will affect tests on that PR so you should be good 👍 

I submitted PR #1478 as I think I've found the problem 🤞

laceysanderson:
[...]

PR #1478, opened 2023-04-11
Change Automated PHPunit testing workflow to be triggered on pull requests

laceysanderson:
This change is triggered by Issue #1477.
It turns out the automated testing has never run on pull requests from forked repositories. After much reading I suspect this is because it only runs on pushes to local branches. Since the branch for these PRs from forked repositories is not local, this workflow is not triggered.

Thus I added the pull_request action which will be triggered for all pull request opening and whenever a commit is pushed to any pull request branch even if it's in a fork -or at least that is what the docs imply.

Anyway, once I confirm it runs I'm going to merge it because I am 100% sure this will not hurt nor will it open us up to security concerns (this workflow does not have access to secrets when run by a fork triggered by this action and is read-only).

See here for more: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/ALL-phpunit.yml
+++ b/.github/workflows/ALL-phpunit.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches-ignore:
       - 9.x-4.x
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   run-tests:


laceysanderson:
Turns out this approach causes double the runs within a PR because it triggers both pull_request and push 🤔

laceysanderson:
Turns out there is no way to stop the duplicate runs without a third party action 😒 This seems badly designed GitHub.

dsenalik:
Although isn't this how the Tripal 3 testing is now? Duplicate runs for push and pull if submitted by a maintainer 🤔

PR #1479, opened 2023-04-11
Update Dockerfile-php8.1-pgsql13

pdtouch:
Corrected line 208

#

Issue #

PR #1480, opened 2023-04-11
Update Dockerfile-php8-pgsql13

pdtouch:
Quotes and backslashes were fixed in line 208. The current version of the dockerfile throws error at this line and does not execute this line as expected.

#

Issue #

### DIFF
--- a/tripaldocker/Dockerfile-php8-pgsql13
+++ b/tripaldocker/Dockerfile-php8-pgsql13
@@ -205,7 +205,7 @@ RUN service apache2 start \
 ## Configuration files & Activation script
 RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
   && mv /app/tripaldocker/default_files/000-default.conf /etc/apache2/sites-available/000-default.conf \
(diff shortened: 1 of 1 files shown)


dsenalik:
I'm sorry I didn't notice this yesterday! This new version
```
echo "\$settings['trusted_host_patterns'] = [ '^localhost$', '^127\.0\.0\.1$', \$_SERVER['SERVER_NAME'] ];"
```
does indeed generate the desired result in the settings file:
```
$settings['trusted_host_patterns'] = [ '^localhost$', '^127\.0\.0\.1$', $_SERVER['SERVER_NAME'] ];
```
and I verified that this is correct by executing this in the `/devel/php` window:
```
echo  $_SERVER['SERVER_NAME'];
```
which returns my server name.

dsenalik:
I'm going to blatantly copy Lacey's comments from yesterday and merge this  and #1479 right away because it was actually broken.

✅ Code review: ok see https://github.com/tripal/tripal/pull/1480#issuecomment-1504274837 👍 
✅ Automated testing: the automated testing passes showing that the dockerfile builds perfectly with these changes and that the site is still functional with them.

Thus this is ready to be merged! No manual testing needed :-)

pdtouch:
Thanks for the confirmation.

PR #1484, opened 2023-04-19
Tv4g4 issue1483 organism lists for select

### LINKED ISSUE #1483: Port pull #1311 to Tripal 4

## Pull #1484

Port changes from pull #1311 to Tripal 4 so that we can use in the GFF3 loader and others to replace the sql query for a sorted organism list (including by infraspecies), e.g. change in the gff3 loader:
```
<     $sql = "SELECT * FROM {organism} ORDER BY genus, species";
<     $org_rset = $chado->query($sql);
<     $organisms = [];
<     while ($organism = $org_rset->fetchObject()) {
<       $organisms[$organism->organism_id] = chado_get_organism_scientific_name($organism);
<     }
---
>     $organisms = chado_get_organism_select_options(FALSE, TRUE);

```

dsenalik:
# Tripal 4 Core Dev Task 

# Port of pull #1311 from Tv3 to Tv4 

## Description
This is a port of pull #1311 which postdated when Tripal 4 was started, it involves the `chado_get_organism_select_options()` api function, changes to better support infraspecific nomenclature, and it replaced code in various importers. This function is used to populate an organism select on an importer form.

The first parameter is a legacy from Tripal 2, but is left for the sake of porting modules to Tripal 4. ~It will now trigger an exception if passed as True (i.e. Tripal2), it must be passed as False.~ It will be a potential future "published_only" parameter.

## Testing?
In devel php window, each of :
```
$list = chado_get_organism_select_options();
$list = chado_get_organism_select_options(TRUE);
$list = chado_get_organism_select_options(TRUE, FALSE);
$list = chado_get_organism_select_options(TRUE, TRUE);
```
should trigger an exception.
```
$list = chado_get_organism_select_options(FALSE, FALSE);
var_dump($list);
```
should return an array without common name included
```
$list = chado_get_organism_select_options(FALSE, TRUE);
var_dump($list);
```
should return an array with common name included (this was for compatibility with how some importers work)
Screenshot_2023-04-19_17-30-38

### DIFF
--- a/tripal_chado/src/api/tripal_chado.organism.api.inc
+++ b/tripal_chado/src/api/tripal_chado.organism.api.inc
@@ -177,47 +177,57 @@ function chado_get_organism_scientific_name($organism) {
 }
 
 /**
- * Returns a list of organisms that are currently synced with Drupal to use in
- * select lists.
+ * Returns a list of organisms to use in select lists.
  *
  * @param $syncd_only
- *   Whether or not to return all chado organisms or just those sync'd with
- *   drupal. Defaults to TRUE (only sync'd organisms).
+ *   Legacy Tripal 2 parameter, no longer supported. Must be FALSE.
+ *
+ * @param $show_common_name
+ *   When true, include the organism common name, if present, in parentheses.
  *
  * @return
- *   An array of organisms sync'd with Drupal where each value is the organism
+ *   An array of organisms where each value is the organism
  *   scientific name and the keys are organism_id's.
  *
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
I did a thorough code review and it looks good and covers the same cases as the original T3 PR.

Also, you hit 100% test coverage!!! 🎉 <3 

Since this is a small PR, was tested in another PR for Tripal 3 and has perfect test coverage, I've decided it can be merged without a manual test :-) 

Thanks @dsenalik

Issue #1318, opened 2022-11-11
Tripal fields don't exist for all of the linker tables, e.g. link analysis and project


dsenalik:
## Feature Request/Discussion

* This feature does attempt to solve an existing problem with Tripal
* I am open to developing or collaborating on an extension module if this is not a good fit for Tripal Core
* This feature is Not Urgent

### Description
Correct me if I'm wrong, but right now in core tripal, there is not a field to link between project and analysis or even assay. I wrote one for CarrotOmics, but is there a reason this field does not already exist, the table ```project_analysis``` exists, so why not a field for it?

### Your Specific Use Case
Here is an example of linking a project to an analysis and assays
https://www.carrotomics.org/Project/291804
and each of those then has a link back to the project. 

### Generally Applicable
I "overloaded" the field so a project also links to assays for example, and that's why I never finished the widget. Is there a possibility of a more general "linker" field that will just use all linker tables in chado? Sort of like the ```sio__references``` field does for publications? However that field doesn't have a widget so couldn't be used to add links.

### Testing
See https://github.com/tripal/tripal/pull/1319#issuecomment-1329543545

spficklin:
Linking to this discussion about the analysis -> project link:  https://github.com/tripal/tripal/issues/804

laceysanderson:
Yes, I use the project tables A LOT. There are a number of these project-focused linker tables in Chado but not for all my purposes. As such I've created the following additional ones using Tripal Custom Tables: project_feature and project_stockcollection. We group all our data via project although this is usually done via Tripal Data Importers and then used in our custom searches. We don't have a field to display these connections on the page unfortunately as we just didn't get around to it.

For the beautiful generic linker fields we have in Tripal, they generally just don't get created unless someone asks and/or has time to help make the field 🙈 I do not see a reason not to have a such a field though and we would also benefit from it so I'm happy to see / review/ help as I'm able :-) Even if it's just focused on display for now, it would be a great addition to core in my opinion! <3

dsenalik:
Okay, what is the best CV term for the field, then. I used ```local__project_analysis``` for example on CarrotOmics, is that reasonable or is there a better term?

laceysanderson:
[...]

PR #1487, opened 2023-04-21
Issue 1318 Add autocompletes for upcoming chado_linker module

dsenalik:
# New Feature

## Description
Most of the changes from issue #1318 pull #1319 will be moved to a new module [chado_linker_fields](https://github.com/tripal/chado_linker_fields). 
However, a few changes needed by the new linker fields are best handled in core tripal:
1. api functions for autocompletes in forms e.g. `function chado_autocomplete_analysis()`
2. some text in the project api that was copied from stock api was removed.
3. typo and PHP8 bugfix in `chado_linker__contact` field
4. a tiny bit of css to accomodate a max_height setting for fields

I did not implement the max_delta and other settings in core tripal because nothing there will use it, and to make this pull request reasonably "safe". In tripal 4 it will be in core, though.

## Testing?
1. Will have to be code review as the functions calling the new api functions are not there yet. These api changes are very very similar to existing ones for other content types, so I hope code review is sufficient. If you really want to test, use the branch from #1319 (now closed) and see the autocompletes in all their glory
2. the text removed from the project api you can see came from the stock api, doesn't apply to project.
3. this is the same as another fix somewhere, I'll have to look up which...
4. this css will be unused in core tripal I hope that's okay.

### DIFF
--- a/tripal_chado/api/modules/tripal_chado.analysis.api.inc
+++ b/tripal_chado/api/modules/tripal_chado.analysis.api.inc
@@ -17,6 +17,36 @@
 
 
 /**
+ * Used for autocomplete in forms for identifying analyses
+ *
+ * @param $string
+ *   The string to search for.
+ *
+ * @return
+ *   A json array of terms that begin with the provided string.
+ *
+ * @ingroup tripal_analysis_api
+ */
+function chado_autocomplete_analysis($string = '') {
+  $items = [];
+  $sql = "
+    SELECT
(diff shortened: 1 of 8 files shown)


laceysanderson (approved):
✅ Code review looks good
✅ Tested each autocomplete path with a test query without error.

This is ready to merge! 🎉 
Thanks @dsenalik <3 

PS. I did add back in the routing you had in the other PR and the include statements for the API but no other changes were needed.

PR #1506, opened 2023-05-08
Remove default select from chado_get_organism_select_options()

### LINKED ISSUE #1505: Remove built-in "Select an organism" option from chado_get_organism_select_options()

Remove the built-in `Select an organism` from the `chado_get_organism_select_options()` api function. 
The "Drupal" way would be for a form to add this itself with `#empty_option` if it is desired.

dsenalik:
# Tripal 4 Core Dev Task

Issue #1505

## Description
This removes the built-in `Select an organism` from the `chado_get_organism_select_options()` api function.
The "Drupal" way would be for a form to add this itself with `#empty_option` if it is desired.

## Testing?
Currently only used in the gff3 loader, there were two select an organism top selects, now there should only be one.


spficklin (approved):

spficklin:
Looks great. I'm going to merge without waiting for tests

Issue #1502, opened 2023-05-01
All PHP files should end in `.php` for Drupal 9/10


laceysanderson:
PR #1498 brought to my attention that all the api files in tripal_chado end in `.inc` as was the custom in Drupal 7. This has changed in recent versions of drupal. Now all PHP files should end in `.php` with no exceptions.

This issue is a reminder that we need to change all the api files in tripal_chado: https://github.com/tripal/tripal/tree/4.x/tripal_chado/src/api

It seems like tripal/src/api is following the new convention ✅

spficklin:
This will cause problems for our extension modules when they have included those files in the past.  We just need to make sure our upgrade instructions for extension modules include this change.

dsenalik:
I see three other `.inc` files, should I change them?
```
./tripal/templates/tripal_entity.page.inc
./tripal/tripal.views.inc
./tripal_chado/tripal_chado.views.inc
```

laceysanderson:
It looks like the views.inc files need to stay as is for some unknown reason... It looks like the autodiscovery for these hook filled files has not changed yet so it needs to stay in order to be found by view. Just not sure why they haven't upgraded it 🤔 

but I don't see a reason not to switch `tripal_entity.page.inc` => `tripal_entity.page.php`.

dsenalik:
Pull #1524 will rename this one remaining file, and if it works okay this issue can be closed

laceysanderson:
Fixed!

PR #1512, opened 2023-05-11
Tripal 4 rename some remaining .inc files to .php - issue1502

dsenalik:
# Bug Fix

## Description
@laceysanderson has noted: All the api files in `tripal_chado` end in `.inc` as was the custom in Drupal 7. This has changed in recent versions of drupal. Now all PHP files should end in `.php` with no exceptions.

## Testing?
Let's see if automated tests pass.
Note @spficklin 's message in issue #1502 "This will cause problems for our extension modules when they have included those files in the past. We just need to make sure our upgrade instructions for extension modules include this change."


laceysanderson (approved):
Thanks @dsenalik

PR #1519, opened 2023-05-19
Adds a Dockerfile for PHP 8.2 and testing for Drupal 10.1.x

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1455 

## Description

This PR adds GitHub workflows, Dockerfile and README grid options for PHP 8.2 and Drupal 10.1.x.

It also switches our workflows that generate the compatibility badges to using the test-tripal-action for better maintainability.

NOTE: This PR does not fix any incompatibilities with Tripal and PHP 8.2 or Drupal 10.1.x. It just set's us up to be aware of them.

## Testing?
Confirm that the readme is rendering: https://github.com/tripal/tripal/tree/tv4g9-issue1455-upgradeDockers#compatibility
Screenshot 2023-05-19 at 4 52 54 PM

Check that the workflows which are failing are doing so in the automated tests, not in building the dockerfile.
- [Drupal 10.1.x on PHP 8.1](https://github.com/tripal/tripal/actions/runs/5028652337) has Errors: 3, Failures: 5 in automated tests - docker builds ✅ 
- [Drupal 10.1.x on PHP 8.2](https://github.com/tripal/tripal/actions/runs/5028652339) has Errors: 4, Failures: 5 in automated tests - docker builds ✅ 
- [Drupal 10.0.x on PHP 8.2](https://github.com/tripal/tripal/actions/runs/5028652342) has Errors: 1 in automated tests - docker builds ✅

PR #1521, opened 2023-05-21
Tripal 4 - Fix all the PHP 8.2 deprecations

dsenalik:
# Tripal 4 Core Dev Task

## Description
Pull request #1519 adds Drupal 10.1 and PHP 8.2 which are initially producing numerous deprecation notices.
The deprecations all seem to be for [PHP 8.2: Dynamic Properties are deprecated](https://php.watch/versions/8.2/dynamic-properties-deprecated)

A random example from `tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php` where the dynamically created property is `$this->messageLogger`
```
  /**
   * {@inheritdoc}
   */
  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
    parent::__construct($configuration, $plugin_id, $plugin_definition);

    // Instantiate the TripalLogger
    $this->messageLogger = \Drupal::service('tripal.logger');

  }
```

This pull request fixes these by declaring the variable in the parent object

n.b. The GFF loader does it differently, for example
```
      if ($result === FALSE) {
        \Drupal::messenger()->addError('Invalid replacement string.');
      }
```
and maybe these should be changed for consistency.

## Testing?
We will have to wait for #1519 to be merged, then merge 4.x with this branch and confirm that at least the deprecation notices go away. There are some other unrelated errors to be fixed


laceysanderson (approved):
Your code fix is perfect 👍 And the tests are no longer showing these messages although they are still failing as expected -one step at a time. Thanks @dsenalik 🎉 

Regarding:

> The GFF loader does it differently, for example
> 
>       if ($result === FALSE) {
>         \Drupal::messenger()->addError('Invalid replacement string.');
>       }
> 
> and maybe these should be changed for consistency.

We will definitely want to change `\Drupal::services` usage within classes to the way things are done in this PR. Not just for consistency (although we all know how much I love consistency ;-) ) but also because of Drupal standards and that dependency injection will make testing easier and allows for better decoupling and more obvious dependencies. See #1522 for the issue to do this update.

PR #1526, opened 2023-05-25
Update docker build workflow to match new testing grid.

laceysanderson:
# Tripal 4 Core Dev Task

## Description
This PR literally updated the build docker workflow to match the phpunit testing grid so that we actually have pre-made dockers for the same versions we are testing now 🙈 

## Testing?
As long as the build docker workflow passes and the docker images are created, this can be merged.


dsenalik (approved):
Looks great, I shall merge forthwith...
(I'm glad I'm not the only one to mess up a branch name 😂)

Issue #1522, opened 2023-05-23
Remove use of \Drupal from Class settings in favour of dependency injection


laceysanderson:
# Core Tripal 4 Development Task

**Not Yet started -anyone can tackle any one of these :-)**

## Branch Name: tv4g0-issue1522-dependencyInjection

There are a number of places in the Tripal 4 codebase where we use `\Drupal` static methods (e.g. `\Drupal::service('tripal.logger')` ). According to the Drupal documentation, **we should only be using these static methods in procedural code and not in classes**.

> Generally, code in Drupal should accept its dependencies via either constructor injection or setter method injection. However, there are cases, particularly in legacy procedural code, where that is infeasible. This class acts as a unified global accessor to arbitrary services within the system in order to ease the transition from procedural code to injected OO code.
> 
> This class exists only to support legacy code that cannot be dependency injected. If your code needs it, consider refactoring it to be object oriented, if possible. When this is not possible, for instance in the case of hook implementations, and your code is more than a few non-reusable lines, it is recommended to instantiate an object implementing the actual logic.
https://api.drupal.org/api/drupal/core!lib!Drupal.php/class/Drupal/10

### Needs Method 1 Implemented
- [ ] tripal/src/TripalDBX classes
- [ ] tripal/src/Services/TripalJob: date.formatter, tripal.logger
- [ ] tripal/src/Services/TripalPublish: tripal.logger, entityTypeManager, tripal.storage, entity_field.manager, plugin.manager.field.field_type

### High Priority
- [x] tripal/src/TripalVocabTerms,tripal/src/Plugin/ (TripalIdSpaceTripalVocabulary/TripalDefaultVocabulary)
- [ ] tripal/src/TripalField/TripalFieldItemBase
- [ ] tripal/src/TripalImporter/TripalImporterBase
     - [x] tripal_chado/src/TripalImporter/ChadoImporterBase
- [x] tripal/src/TripalStorage

# Dependency Injection

*If someone has time to add this in the tripaldoc documentation that would be really helpful!.*

There are two main ways to do dependency injection and which you use depends on the type of class you are needing the services in.

## Method 1: Services use arguments in services.yml + set in __constructor

[Here is a controller example](https://www.drupal.org/node/2122195#s-accessing-other-services) from Drupal.org.

## Method 2: Controllers, Plugins, Forms + Blocks use ContainerInjectionInterface + set in create

[...]

PR #1527, opened 2023-05-29
Use dependency injection in Chado Vocal + ID Space classes

laceysanderson:
# Tripal 4 Core Dev Task  

Issue #1522 

## Description

This PR updates the Chado Vocal + ID Space classes to use dependency injection for its database connection + Tripal logger. This is needed to help fix some tests failing in Drupal 10.1 and also just ensures that we are testing on the test chado schema not the production one ;-) 

This uses the Symfony\Component\DependencyInjection\ContainerInterface to support dependency injection in a plugin. By implementing this interface, you need to also implement its static create() function to provide the services from the container... however, our collection interface already had a non-static create function. I resolved this by renaming the collection create to createRecord() which I believe better describes its purpose.

## Testing?

I added testing the dependency injection to the automated tests for Vocal, ID Space and term. This means if automated testing passes then we should be good.

You cannot easily test this manually...

### DIFF
--- a/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
+++ b/tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php
@@ -364,7 +364,7 @@ class TripalDefaultIdSpace extends TripalIdSpaceBase {
   /**
    * {@inheritDoc}
    */
-  public function create() {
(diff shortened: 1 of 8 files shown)


laceysanderson:
Merging as these changes have great test coverage (98%) and are difficult to test manually.

PR #1540, opened 2023-06-07
tripal 3 issue 1538 local__contact widget ajax error fix

### LINKED ISSUE #1538: Tripal 3 ajax error in local__contact widget
# BUG/ERROR report

### Tripal Version: 3.10
* Drupal Version: 7.92, 7.97
* PostgreSQL Version: 13.11, 15.3
* PHP Version: 8.0.28, 8.1.18

### Issue description
Ajax error adding a relationship to a biomaterial triggers an error in the local__contact widget. No error on other content type such as organism. Hypothesize that this is because there is no local__contact field on organism, but there is one on biomaterial.

`TypeError: array_key_exists(): Argument #2 ($array) must be of type array, string given in local__contact_widget->form() (line 41 of .../sites/all/modules/tripal/tripal_chado/includes/TripalFields/local__contact/local__contact_widget.inc).`

I think this is an old bug, but never showed up until #1313 allowed the field to even be used.

It's very likely a PHP8 thing, too, which would not show up under 7.x, but I don't think it is worth the time to verify that.

#### Steps to reproduce
  * Edit a biomaterial that has a contact
  * Select any Vocabulary for the Relationship field. Ajax error appears.

dsenalik:
# Bug Fix

## Description
I have made a small change in the interests of not breaking anything, and of not spending too much time on Tripal 3. I suspect the widget could be looked at more closely, but this will wait for the Tripal 4 version.

When the widget is called, `$items[0]` will look like this
```
$items[0]="array(2) {
  ["value"]=>
  array(4) {
    ["schema:additionalType"]=>
    string(14) "Research Group"
    ["schema:name"]=>
    string(8) "USDA ARS"
    ["schema:description"]=>
    string(71) "United States Department of Agriculture - Agricultural Research Service"
    ["entity"]=>
    string(19) "TripalEntity:177904"
  }
  [0]=>
  string(2) "10"
}
```
but when it is called by Ajax from the relationship field, it looks like this
```
$items[0]="array(3) {
  ["value"]=>
  string(8) "USDA ARS"
  ["chado-biomaterial__biosourceprovider_id"]=>
  string(2) "10"
  ["name"]=>
  string(8) "USDA ARS"
}
```
note that `['value']` is no longer an array in this case!

My very minimal fix is just to check that we have an array before calling `array_key_exists`. I hypothesize this was not flagged under PHP 7.x.

Manual testing confirms that the relationship field now works on a biomaterial page.

## Testing?
Edit a relationship on a biomaterial page that also has a contact.

### DIFF
--- a/tripal_chado/includes/TripalFields/local__contact/local__contact_widget.inc
+++ b/tripal_chado/includes/TripalFields/local__contact/local__contact_widget.inc
@@ -38,7 +38,7 @@ class local__contact_widget extends ChadoFieldWidget {
     // If the field already has a value then it will come through the $items
     // array.  This happens when editing an existing record.
     if (count($items) > 0 and array_key_exists($delta, $items)) {
-      $name = array_key_exists($name_term, $items[$delta]['value']) ? $items[$delta]['value'][$name_term] : $name;
+      $name = (is_array($items[$delta]['value']) and array_key_exists($name_term, $items[$delta]['value'])) ? $items[$delta]['value'][$name_term] : $name;
       $record_id = tripal_get_field_item_keyval($items, $delta, 'chado-' . $field_table . '__' . $pkey, $record_id);
       if ($field_table == 'biomaterial') {
         $contact_id = tripal_get_field_item_keyval($items, $delta, $linker_field, $contact_id);


laceysanderson (approved):
Wow, that's crazy. Code review looks good -small intuitive fix. I did not get the chance to do a manual test.

dsenalik:
I'm working on this field for Tripal 4 so will keep this in mind in case it still exists there (although I doubt it, fields are much different now)

Issue #1539, opened 2023-06-07
Test Directory Structure


laceysanderson:
# Core Tripal 4 Development Task

## Branch Name: tv4g9-issue1539-testDirCleanup

Drupal/PHPUnit has a very specific  recommendation for directory structure and namespaces as described here: https://www.drupal.org/docs/automated-testing/phpunit-in-drupal/phpunit-file-structure-namespace-and-required-metadata

This issue assigns a task to fix our current test suites to match these expectations :-) 🙈

PR #1541, opened 2023-06-07
Restructure Kernel Tests

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1539 

## Description

**This only affects Kernel tests at this point. No existing functional or unit tests have been touched in this PR.**

4.x currently has Kernel and Functional tests mixed together in the tests/src/Functional directory. This does not match [Drupal standards](https://www.drupal.org/docs/automated-testing/phpunit-in-drupal/phpunit-file-structure-namespace-and-required-metadata) and makes it difficult to see the difference between the two test types.

This PR moves existing Kernel tests and the ChadoKernelTestBase class into tests/src/Kernel where they belong. I also improved the biotaskbase test using closures to allow direct testing of the protected variables. Additionally I found a bug in the ChadoUpgrader and fixed the task tests so the class name matched the file name.

## Testing?
This only impacts automated tests so if tests pass then we are good to go.

A quick code review of the biotaskbase tests would be helpful but I don't think it need be required.

### DIFF
--- a/tripal/tests/src/Kernel/TripalDBX/ConnectionTest.php
+++ b/tripal/tests/src/Kernel/TripalDBX/ConnectionTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Drupal\Tests\tripal\Functional\TripalDBX;
+namespace Drupal\Tests\tripal\Kernel\TripalDBX;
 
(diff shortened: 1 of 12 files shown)


laceysanderson:
There are better PRs to review than this one and it doesn't really need it so I'm just merging ;-)

PR #1549, opened 2023-06-13
Dependency injection for Chado Storage + Storage Property Types/Values

laceysanderson:
# Tripal 4 Core Dev Task 

Issue #1522 

## Description

This Pr updates chadostorage and the associated storage properties to use dependancy injection for the chado database, logger and id space manager. This is needed for the kernel tests for chado storage so that I can mock the terms associated with storage property types. It's also just good practice to use dependency injection -see issue #1522 for more justification on this front.

## Testing?

Automated testing should pass as is since this is a backwards compatible change.

For manual testing, we just want to confirm that fields that worked before this PR still work. I confirmed by manually adding an organism and a gene.

### DIFF
--- a/tripal/src/TripalStorage/StoragePropertyBase.php
+++ b/tripal/src/TripalStorage/StoragePropertyBase.php
@@ -2,10 +2,19 @@
 
 namespace Drupal\tripal\TripalStorage;
 
+
 /**
  * Base class for a Tripal storage property type or value.
  */
 class StoragePropertyBase {
+
+  /**
+   * The ID Space Plugin manager service.
+   * Note: Used to confirm the term passed in exists and to retrieve it.
+   *
+   * @var Drupal\tripal\TripalVocabTerms\PluginManagers\TripalIdSpaceManager
+   */
+  protected $idSpaceService;
 
   /**
    * Constructs a new Tripal storage property base object.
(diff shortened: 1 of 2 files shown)


dsenalik (approved):
I tested fields manually
Added germplasm accession, then edited and changes were saved, deleted, all worked as expected.
Tested contact same way, ok.
Tested organism, ok.
I will create an issue - unrelated to this pull request - some fields can't be saved because they are missing organism_id (DNA Library, Genetic Marker, may be others) = Issue #1550

laceysanderson:
Awesome! Thanks for the manual check @dsenalik ❤️ 🎉 

> I will create an issue - unrelated to this pull request - some fields can't be saved because they are missing organism_id (DNA Library, Genetic Marker, may be others) = Issue TBD

For these, that is likely because that field was missed in the yaml file. To fix/check in the case, look at [the yaml file defining all fields for our content types](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml) and check that the organism_id field is included for those specific content types.

PS. This is also the way you will make sure your new fields are auto-attached to the content types. It's a simple matter of adding them to this particular yaml for the content types they pertain to.

Issue #1546, opened 2023-06-09
Kernel tests for Chado Storage


laceysanderson:
# Core Tripal 4 Development Task

## Branch Name: tv4g9-issue1546-chadoStorageTests

I am developing very use-case specific kernel tests for chado storage to ensure that we are meeting all the needs. This will also help compartmentalize the tasks left to do to upgrade chado storage to meet all our needs.

laceysanderson:
Current state of things is described fully in the PR referenced above.

# Summary from PR:
This PR adds a new Test Trait to make it easier to write Kernel tests for Chado Storage. As mentioned in the linked issue, the plan is to write a full suite of Chado Storage kernel tests for each field. This PR lays the ground work for that and adds an example test for the property field.

NOTE: When looking at the property test you will notice that the data provider would provide two test datasets except that the second one is commented out. This is because the second one currently fails due to https://github.com/tripal/tripal/issues/1398.

## ChadoLinkerPropertyDefault Field

tripal_chado/tests/src/Kernel/Plugin/ChadoStorage/ChadoLinkerPropertyDefaultTest.php

The $fields array describes the hypothetical fields and properties to test. For this test, there are two property fields whose chado storage properties match the form currently used in this field. It also provides a single "other" field which contains properties to meet the unique constraint for feature.

The dataProvider provideFieldExpectations() currently provides one set of expectations which includes the values for a single property field and the other field. As noted above, there is a second set of expectations which is commented out currently that allows us to test multiple property fields attached to the same content type... However, it is not currently being used in tests due to Issue https://github.com/tripal/tripal/issues/1398.

#### testInsertValues()

Currently tests ChadoStorage addTypes(), getTypes(), insertValues() methods for expectation sets provided by the data provider. This is written/tested to work once we uncomment the second expectation set.

I still have the following tests to implement for this field but will do so in another PR as this one was already becoming substantial.

- Create Values in Chado using ChadoStorage when they violate unique constraint.
- Load values in Chado using ChadoStorage when they don't yet exist.
- Load values in Chado using ChadoStorage after we just inserted them.
- Update values in Chado using ChadoStorage after we just inserted them.
- Update values in Chado using ChadoStorage when they don't actually exist.
- Ensure property field picks up records in Chado not added through field.

laceysanderson:
With PR #1584, we now have a full test suite for a single property field. The next step are: 
- to resolve bugs known bugs in the property field
- to document this new API for testing the chado interaction of fields
[...]

PR #1552, opened 2023-06-20
New infrastructure for Kernel Testing Chado Storage + Property Field Tests

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1546 

## Description

This PR adds a new Test Trait to make it easier to write Kernel tests for Chado Storage. As mentioned in the linked issue, the plan is to write a full suite of Chado Storage kernel tests for each field. This PR lays the ground work for that and adds an example test for the property field.

NOTE: When looking at the property test you will notice that the data provider would provide two test datasets except that the second one is commented out. This is because the second one currently fails due to #1398.

This PR also fixes a very small bug in chado storage where `delete if empty` assumed that the property key was always the same as the chado column. This is an invalid assumption so the fix is to use the chado column there instead. See 8ed9e720137a4a4c93c35c57b8efe7f08835e4c2.

Additionally, you will notice that I moved the getCvtermId() helper method from ChadoBrowserTestBase to ChadoTestTrait. This is a backwards compatible change as ChadoBrowserBase uses ChadoTestTrait and was needed to ensure ChadoKernelTestBase would have access to this helper function.

## Testing?

There is no way to manually test this.

#### ChadoLinkerPropertyDefault Field
*tripal_chado/tests/src/Kernel/Plugin/ChadoStorage/ChadoLinkerPropertyDefaultTest.php*

The `$fields` array describes the hypothetical fields and properties to test. For this test, there are two property fields whose chado storage properties match the form currently used in this field. It also provides a single "other" field which contains properties to meet the unique constraint for feature. 

The dataProvider provideFieldExpectations() currently provides one set of expectations which includes the values for a single property field and the other field. As noted above, there is a second set of expectations which is commented out currently that allows us to test multiple property fields attached to the same content type... However, it is not currently being used in tests due to Issue #1398.

##### testInsertValues()
Currently tests ChadoStorage addTypes(), getTypes(), insertValues() methods for expectation sets provided by the data provider. This is written/tested to work once we uncomment the second expectation set. 

Specifically it tests:
- For each newly created property type described in $fields the result is an object that is an instance of the StoragePropertyTypeBase class (within createPropertyTypes).
- At the end of createPropertyTypes() the number of properties created matches the number of properties expected based on the data provider (within createPropertyTypes).
- ChadoStorage getTypes() returns an array with the same number of entries as the array we passed into addTypes() (within addPropertyTypes2ChadoStorage).
- We were able to create mock field config objects for use with the insertValues() (within createDataStoreValues)
- For each newly created property value based on the expectations from the data provider, we were able to create an object of type StoragePropertyValue with no default value set (within createDataStoreValues).
- That for each field, we had the expected number of values in our data store values array after creating the property values above (within createDataStoreValues).
- That at the end of createDataStoreValues we have the expected number of fields in our data store values array (within createDataStoreValues).
- That we were able to use getValue() on each property value object with a default value described in the expectations array to retrieve the same value we set using setValue() (within setExpectedValues).
- That we were able to call insertValues() with our prepare data store values array without it returning an error.
- That the base feature record described by the other field now exists in chado with the values we expect
- Checks that each value from the expected values array matches exactly one record in the featurepop table. This is selected based on the unique key and then it's check the value is what we expect.

#### Notes:

I still have the following tests to implement for this field but will do so in another PR as this one was already becoming substantial.

- Create Values in Chado using ChadoStorage when they violate unique constraint.
- Load values in Chado using ChadoStorage when they don't yet exist.
- Load values in Chado using ChadoStorage after we just inserted them.
- Update values in Chado using ChadoStorage after we just inserted them.
- Update values in Chado using ChadoStorage when they don't actually exist.
- Ensure property field picks up records in Chado not added through field.


laceysanderson:
All automated tests pass.
Since this PR only touches automated testing and no one has had time to review for 7 days, I'm going to merge this in the interests of keeping things moving.

Issue #1553, opened 2023-06-21
Chado project autocomplete


reynoldtan:
## Tripal 4 Feature Request

* This feature [does not] attempt to solve an existing problem with Tripal
* I [am] open to developing or collaborating on an extension module if this is not a good fit for Tripal Core

* This feature is [Not Urgent]

### Description

Provide a controller to autocomplete project name. The controller can limit number of suggestions to show in an autocomplete field and a type_id option to limit/restrict projects to only match project type set in projectprop.type_id field. 

### Your Specific Use Case

![Screen Shot 2023-06-21 at 1 32 32 PM](https://github.com/tripal/tripal/assets/15472253/f03c1eeb-61d7-4047-b750-65bdd6a525b7)

Autocomplete field that requires project name to be entered.

dsenalik:
I added an autocomplete for project for Tripal 3, but of course it looks like this is implemented differently in Tripal 4. I also suspect we will need autocompletes for a lot more things than just project!

PR #1488, opened 2023-04-22
Tripal 4 -  FASTA loader

risharde:
Working version of FASTA loader with included tests / fixtures - up for review. Seems to pass most of the automated tests.

Some tested in regard to the form for specifying additional fields would be nice since I only did a basic test with the Citrus sinensis files from the Tripal 3 documentation.

@spficklin @laceysanderson

PR #1554, opened 2023-06-23
Project Autocomplete Controller

reynoldtan:
#

Issue https://github.com/tripal/tripal/issues/1553

## Description

This is a CHADO-focused autocomplete for projects. Check the issue for more motivation for this PR.

This PR will create a route that can be used as a value to the #autocomplete_route_name when creating an autocomplete form element. It includes a controller which follows the template provided by Tripal Term Autocomplete and queries CHADO using Tripal DBX to suggest projects based on user input.

## Testing?

This PR includes automated tests which will test:
1. Insert test projects.
2. The class handler to evaluate number of suggestions returned and if desired project was part of the suggestions.
3. Support classes getProjectName and getProjectId to resolve project name (given a project id) and project id (given a project name), respectively.
4. Invalid inputs to all methods.

Manual Testing.
There is no place in Tripal Core that needs this autocomplete. An example form element would look this:

```
 $form['my_autocomplete'] = [
   '#type' => 'textfield',
   '#autocomplete_route_name' => 'tripal_chado.project_autocomplete',
   '#autocomplete_route_parameters' => ['type_id' => 0, 'count' => 5]
 ];

// Where type_id is the type id number used to create types in projectprop table to restrict results to only match this id. 
// In this example it is set to 0 to return all projects regardless of type.
```

Since the autocomplete returns human-readable output (project name) and not the project id, there is an additional method that will fetch the project id of a given project name.

```
 $project_id = ChadoProjectAutocompleteController::getProjectId($project_name);
```

To set default value (#default_value property) to fields, where stored value is the project_id, a method to resolve project id to project name is provided by the method getProjectName().

```
$project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
```


dsenalik:
I made a docker with this branch, and set up testing as follows.
1. Create some test projects, I created these 
Test Projects
2. Create the projectprop table entries, I picked some random cvterm_id values, they are not significant (I think)
```
insert into chado.projectprop (project_id, type_id) values (1, 10), (2, 10), (3, 10), (4, 11);
INSERT 0 4
sitedb=> select * from chado.projectprop;
 projectprop_id | project_id | type_id | value | rank 
----------------+------------+---------+-------+------
              1 |          1 |      10 |       |    0
              2 |          2 |      10 |       |    0
              3 |          3 |      10 |       |    0
              4 |          4 |      11 |       |    0
(4 rows)
```
3. Put the form field somewhere to test. I added it to `tripal/src/Form/TripalImporterForm.php` because that's easy to get to: Tripal → Data Loaders → OBO Vocabulary Loader
At line 47 I inserted this
```
$form['my_autocomplete'] = [
   '#type' => 'textfield',
   '#autocomplete_route_name' => 'tripal_chado.project_autocomplete',
   '#autocomplete_route_parameters' => ['type_id' => 10, 'count' => 5],
   '#title' => 'Test autocomplete title',
   '#description' => 'Test autocomplete description',
];
```

Unfortunately when I test this with a single letter, nothing happens, it sticks at Loading... and nothing happens
StuckRunning

There are three errors in the log

Warning: Undefined array key "function" in Drupal\Core\Utility\Error::getLastCaller() (line 150 of /var/www/drupal9/web/core/lib/Drupal/Core/Utility/Error.php) 

Symfony\Component\HttpKernel\Exception\ControllerDoesNotReturnResponseException: The controller must return a "Symfony\Component\HttpFoundation\Response" object but it returned null. Did you forget to add a return statement somewhere in your controller? in () (line 98 of /var/www/drupal9/web/core/lib/Drupal/Core/EventSubscriber/EarlyRenderingControllerWrapperSubscriber.php).

Warning: Undefined array key "function" in Drupal\Core\Utility\Error::getLastCaller() (line 150 of /var/www/drupal9/web/core/lib/Drupal/Core/Utility/Error.php) 

Only when you enter a second letter does the autocomplete kick in (this is by design in the code to wait until two letters are entered). 

If I comment out at line 75 in `tripal_chado/src/Controller/ChadoProjectAutocompleteController.php`
```
//    if ($response) {
      return new JsonResponse($response);
//    }
```
then there are no errors generated in the log.

AB

And if I enter "BC" I get only one project "ABC", as expected, not the "BCD" project with a different type_id. This is the correct desired behavior.

An issue, though, is the annoying spinning circle if you only enter 1 letter, or if there is no match to the letters you entered, if I enter "XYZ" for example. But I'm not sure how that normally works...

dsenalik:
I figured it out, in addition to the change above, in `tripal_chado/src/Controller/ChadoProjectAutocompleteController.php` line 32 change
```
    // Array to hold matching project records.
    $response = null;
```
to
```
    // Array to hold matching project records.
    $response = [];
```

tripal_chado/src/Controller/ChadoProjectAutocompleteController.php
```diff
@@ -0,0 +1,133 @@
...
+   *   Default to 5 items.
+   *   Must be declared in autocomplete route parameter ie. [count => 5].
+   * 
+   * @return Json Object
+   *   Matching project rows in an array where project name
+   *   is both the value to the array keys label and value.
+   */
+  public function handleAutocomplete(Request $request, int $type_id = 0, int $count = 5) {
+    // Array to hold matching project records.
+    $response = null;
```
dsenalik: change to $response = [];

tripal_chado/src/Controller/ChadoProjectAutocompleteController.php
```diff
@@ -0,0 +1,133 @@
...
+              'label' => $record->name  // Value shown in the list of options.
+            ];
+          }
+        }
+      }  
+    }
+    
+    if ($response) {
+      return new JsonResponse($response);
+    }
```
dsenalik: remove the `if ($response)`, always want to return a response
`return new JsonResponse($response);`

dsenalik (changes requested):

dsenalik:
@reynoldtan can you please make the same change to initialize `$response = [];` in `tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php` line 30. It seems like a lot of work to create a whole separate issue for such a minor change, I feel we can add it in here since it is the exact same change.
Thanks

Actually we may instead fix it in pull #1488, see the comment here https://github.com/tripal/tripal/pull/1488#pullrequestreview-1509624678 the first two points are relevant to the CV autocomplete.

reynoldtan:
Hi everyone, thanks for the review and as recommended, I have revised autocomplete handler with the correct initial value for the $response variable and removed the condition of the return value. I have also removed the requirement to begin autocomplete when 2 characters have been entered.

dsenalik (approved):
Thanks

laceysanderson (approved):
Thanks to @dsenalik thorough functional review, I just did a code review on this one.

✅  code review looks good

Thanks @reynoldtan ❤️ 🎉 
This one is ready to merge too

PR #1562, opened 2023-07-05
Tripal4 issue1561 fix parallel message

dsenalik:
# Bug Fix

## Description
This is a simple correction to the tripal jobs message where `--parallel` does not take an argument, so the message should reflect that.


laceysanderson (approved):
No testing needed 😆

Issue #1573, opened 2023-07-07
Tripal 3 PHP 8.1 deprecation notices


dsenalik:
# BUG/ERROR report

### Tripal Version: 7.x-3.10
Drupal Version: 7.98

### Issue description
Various deprecation notices with PHP 8.1 - easy to fix
`
Deprecated function: Optional parameter $form_state declared before required parameter $mode is implicitly treated as a required parameter in require_once() (line 26 of .../sites/all/modules/tripal/tripal_bulk_loader/tripal_bulk_loader.module).

Deprecated function: Optional parameter $job_id declared before required parameter $node_type is implicitly treated as a required parameter in require_once() (line 24 of .../sites/all/modules/tripal/legacy/tripal_core/tripal_core.module).

Deprecated function: Optional parameter $input declared before required parameter $form_state is implicitly treated as a required parameter in require_once() (line 30 of .../sites/all/modules/tripal/legacy/tripal_core/tripal_core.module).

Deprecated function: Optional parameter $input declared before required parameter $form_state is implicitly treated as a required parameter in require_once() (line 30 of .../sites/all/modules/tripal/legacy/tripal_core/tripal_core.module).

Deprecated function: Optional parameter $analysis_id declared before required parameter $property is implicitly treated as a required parameter in require_once() (line 15 of .../sites/all/modules/tripal/legacy/tripal_feature/tripal_feature.module).

Deprecated function: Optional parameter $feature_id declared before required parameter $property is implicitly treated as a required parameter in require_once() (line 15 of .../sites/all/modules/tripal/legacy/tripal_feature/tripal_feature.module).

Deprecated function: Optional parameter $analysisfeature_id declared before required parameter $property is implicitly treated as a required parameter in require_once() (line 15 of .../sites/all/modules/tripal/legacy/tripal_feature/tripal_feature.module).

Deprecated function: Optional parameter $do_contact declared before required parameter $do_update is implicitly treated as a required parameter in require_once() (line 21 of .../sites/all/modules/tripal/legacy/tripal_pub/tripal_pub.module).
`

#### Steps to reproduce
Sorry, this is kind of a difficult procedure to see this, see the alternative procedure below.

Running under PHP 8.1, update Drupal to 7.98 and then run database updates
![drupalupdate](https://github.com/tripal/tripal/assets/8419404/c0763ea6-4c3d-4e08-a5fa-efa63ba714ae)

Then check the log
[...]

PR #1574, opened 2023-07-07
Tripal 3 Fix php8.1 deprecations

dsenalik:
# Bug Fix

## Description

This fixes a few deprecations that trigger warnings under PHP8.1, they are of the form
`function do_something($param1, $param2 = 'default', $param3) {`
where `$param3` is a required parameter, but `$param2` is set up as optional.
Under PHP8.1 this is flagged as being impossible.

In some cases I have removed the default value, and in others I have added one, as seemed most appropriate.
Many are for legacy functions, some of which are deprecated, so nothing is likely to break from these changes.

One functional typo in a deprecated legacy function was fixed, `$do_udpate` to `$do_update` 😆 https://github.com/tripal/tripal/pull/1574/commits/e51407f53009f841e8c4b15f9f484f3d269f42d6
And my testing was showing 
`Deprecated function: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in chado_get_feature_sequences() (line 496 of .../sites/all/modules/tripal/tripal_chado/api/modules/tripal_chado.feature.api.inc).`
so I fixed that too https://github.com/tripal/tripal/pull/1574/commits/09d0b04e2388b9ff8c25765135ca87c25a2aabcf

## Testing
I described testing in issue #1573

### DIFF
--- a/legacy/tripal_core/api/tripal_core.chado_nodes.api.inc
+++ b/legacy/tripal_core/api/tripal_core.chado_nodes.api.inc
@@ -1079,9 +1079,8 @@ function chado_cleanup_orphaned_nodes($table, $nentries = 25000,
  *
  * @ingroup tripal_legacy_chado_node_api
  */
-function chado_cleanup_orphaned_nodes_part($table, $job_id = NULL, $nentries,
+function chado_cleanup_orphaned_nodes_part($table, $job_id, $nentries,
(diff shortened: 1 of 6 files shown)


laceysanderson (approved):
Looks good! Thank you for the fix @dsenalik. These are such simple changes that I don't feel a functional review is needed so this is ready to merge!

Issue #1576, opened 2023-07-11
Tripal 4 tripal logger string substitution not done in some cases


dsenalik:
# BUG/ERROR report

### Tripal Version: 4.alpha1-dev docker
* Drupal Version: 10.1
* PostgreSQL Version: 13.11
* PHP Version: 8.2.6

### Issue description
The tripal logger object provides capability for string substitutions. In some cases this is not performed.
The `messageStr()` function is responsible for the substitutions, we just need to make sure it is called for all cases, e.g.
```
-      $this->logger->error($message);
+      $message_str = $this->messageString($message, $context);
+      $this->logger->error($message_str);
```

#### Steps to reproduce
  1. In devel module php window execute this
```
$message_opts = [
  'drupal_set_message' => TRUE,
];
\Drupal::service('tripal.logger')->error(
  'This next text should be substituted: "@fail".', 
  ['@fail' => 'testpassed'], 
  $message_opts
);
```
  2. Observe substitution occurs correctly here
![Screenshot_2023-07-11_09-18-05](https://github.com/tripal/tripal/assets/8419404/29a51c59-e6e5-48ac-b5da-3f3edeccc74f)

  3. Go to the log, observe substitution does not occur
![Screenshot_2023-07-11_09-19-07](https://github.com/tripal/tripal/assets/8419404/a315aedb-224b-4c4c-8aab-90290cb7201e)

dsenalik:
Resolved with merged PR

PR #1577, opened 2023-07-11
Tripal 4 tripal logger string substitution not done in some cases

dsenalik:
# Bug Fix

## Description
The tripal logger object provides capability for string substitutions. In some cases this is not performed.
The messageStr() function is responsible for the substitutions, we just need to make sure it is called for all cases, e.g.
```
-      $this->logger->error($message);
+      $message_str = $this->messageString($message, $context);
+      $this->logger->error($message_str);
```

## Testing?
See issue #1576 for how to reproduce this error. With these changes, the error in the log should have the placeholder substituted.


laceysanderson (approved):
✅  Code review looks good 👍 

Functional review:

I tested each of the following using this code sample in the execute PHP development area:
```php
$message_opts = [
  'logger' => TRUE,
  'drupal_set_message' => TRUE,
];
\Drupal::service('tripal.logger')->notice(
  'This next text should be substituted: "@fail".', 
  ['@fail' => 'testpassed'], 
  $message_opts
);
```
Then I checked that it had the right status type and the substitution was made both in the Drupal set message and in the watchdog log.

✅ notice()
✅ info()
✅ error()
✅ warning()
✅ emergency()
❓ alert() did substitute as expected but is a warning in the drupal_set_message and a error in the watchdog log 🤔 
✅ critical()
❓ debug() doesn't actually show a drupal_set_message or add to the log which is not expected based on the code.
✅ log()
Note: If you forget to pass the status as the first parameter to log before the message, you do not get a helpful error 🤦‍♀️ 
```
Warning: Array to string conversion in Drupal\tripal\Services\TripalLogger->log() (line 607 of modules/contrib/tripal/tripal/src/Services/TripalLogger.php).
Drupal\tripal\Services\TripalLogger->log('This next text should be substituted: "@fail".', Array, Array) (Line: 5)
eval() (Line: 107)

Drupal\Core\Logger\LoggerChannel::log(): Argument #2 ($message) must be of type Stringable|string, array given, called in /var/www/drupal9/web/modules/contrib/tripal/tripal/src/Services/TripalLogger.php on line 615
```

I'm approving since the uncovered bugs are not specific to these changes 🤕

dsenalik:
1. **alert() did substitute as expected but is a warning in the drupal_set_message and a error in the watchdog log** 
It is a warning because `MessengerInterface` only has three types, `TYPE_STATUS`, `TYPE_ERROR`, and `TYPE_WARNING` so in `log2Message()` we have to select what we think fits best. Because it is between Emergency and Critical in severity, Alert needs to be an Error not a Warning. https://github.com/tripal/tripal/pull/1577/commits/b67f1d96f0cc0e94d43ffa8d2e7e8225f6182196 and let's list in order of severity https://github.com/tripal/tripal/pull/1577/commits/d7048fab005775a370721e0439ff43ee4449b7e7
```
    else if (in_array($level, ['alert', 'warning'])) {
      $status = \Drupal\Core\Messenger\MessengerInterface::TYPE_WARNING;
```

2. **debug() doesn't actually show a drupal_set_message or add to the log which is not expected based on the code.**
First I found a bug in `log2Message()` - here debug was omitted and so it was not handled.
Then I found a bug in `debug()` where it will only do anything if environment variable `TRIPAL_DEBUG` is set. I call this a bug because the comment reads
```
    // Get the backtrace and include in the error message, but only if the
    // TRIPAL_DEBUG environment variable is set.
```
~~so I will change that code to match the comment. https://github.com/tripal/tripal/pull/1577/commits/df7570198fa8c3123efc738c010f1c7bbb1c86e1~~
Because this is Tripal 3 behavior, it will not be changed now, but become a separate issue #1581 

3. **Note: If you forget to pass the status as the first parameter to log before the message, you do not get a helpful error**
I am not going to fix this 🙈, and the reason is because this is code that already handles error messages. Where do you send the error if you are already in the function that may be printing an error? And anyway we never can get as far as printing an error to begin with, because of the message you noted. While not obvious, I actually don't think it is that hard to dig down to figure out what went wrong, so this will remain unfixed.

Issue #1440, opened 2023-03-06
Tripal 4 You can add a field to a content type that does not support it


dsenalik:
# BUG/ERROR report

### Issue description
We will need a mechanism to prevent adding a field to a content type that does not support it.
I reported in these comments https://github.com/tripal/tripal/issues/1432#issuecomment-1454840026 and https://github.com/tripal/tripal/issues/1432#issuecomment-1455117899 and I posted issue #1436, but it is not a problem with the yaml file but with the field not validating that the field can be added to all base tables.

So briefly, we need to gracefully display an error and clean up the broken field if the base table selected when adding a field does not support the field.

#### Steps to reproduce
My example was adding the ```obi__organism``` field to ```project``` and getting this error:

```Warning: Undefined array key "foreign keys" in Drupal\tripal_chado\Plugin\Field\FieldType\obi__organism::tripalTypes() (line 89 of /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/obi__organism.php) ```

For ```obi__organism``` we could add something like the following, but is this correct?
```
    if (isset($base_schema_def['foreign keys']['organism']['columns'])) {
      $base_fk_col = array_keys($base_schema_def['foreign keys']['organism']['columns'])[0];
    }
    else {
      throw new \Exception(t('base table @base_table does not support adding this field',
        ['@base_table' => $base_table]));
    }
```
![proposal](https://user-images.githubusercontent.com/8419404/223208819-07050ddb-9e2e-47cc-b042-5f37c5fd4548.png)

spficklin:
Working on this on branch `tv4g1-issue1440-limit_fields`

spficklin:
I think I've found a way to fix this issue.  Configuration entities allow any module to add [3rd party settings](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Config!Entity!ThirdPartySettingsInterface.php/interface/ThirdPartySettingsInterface/8.2.x).  Since our content types extend the base classes of configuration entities that means we can add 3rd party settings to any content type.   So my suggestion is that we use this third party settings interface to add a setting that only the Tripal Chado module will use that basically maps a chado base table to each content type.  For example, The YAML for defining a content type would look like this:

```yaml
id: 'general_chado'
label: 'General Content Types (Chado)'
description: 'More generalized content types based on Chado, which all Tripal sites should have.'
content_types:

    -   label: Organism
        term: OBI:0100026
[...]

PR #1578, opened 2023-07-12
Tripal 4 ChadoAnalysisDefault field

dsenalik:
# Tripal 4 Core Dev Task

## Description

This is the initial implementation of the ChadoAnalysisDefault field.

There are only two content types with an `analysis_id` column that will be able to use this field:
  * `phylotree`
  * `quantification`

## Testing?
I am testing it on the phylogenetic tree content type, but we don't have a dbxref field yet, so right now you can't manually enter a tree. To be able to test, I have to temporarily do two things to allow me to create a phylotree record:
 1. `ALTER TABLE chado.phylotree ALTER COLUMN dbxref_id DROP NOT NULL;`
 2.  In `tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php` line 1048 I temporarily add a `return;` to disable validation.

Then I can create a tree, and test adding and changing an analysis record.


dsenalik:
A thing I still need to do is figure out how to log an error if you attempt to use the field on a content type that does not support it, _i.e._ there is no `analysis_id` column in the base table.

That's the same as issue #1440 mentions.

(See below, this is added)

dsenalik:
This is ready for review, I have added base table validation, which might even be redundant because now only valid base tables are presented for selection, i.e. those with a foreign key to analysis.

laceysanderson:
Mw and @dsenalik talked about this field in cofest yesterday.

We agreed that the following very helpful method is generic enough to be added to the Chado Field Type base class:

```php
  /**
   * Return a list of candidate base tables. We only want to
   * present valid tables to the user, which are those with
   * a foreign key to the specified object table.
   * The list of tables is returned in an alphabetized list
   * ready to use in a form select.
   *
   * @param string $object_table
   *   The Chado table being linked to (object).
   */
  protected function getBaseTables($object_table);
```

I will wait to do a code/functional review until the above is completed.

We also talked about whether this field should also handle the linking table variant for analysis linking. We think likely yes but will continue to contemplate. In the meantime this PR still stands as is since ChadoStorage is not ready to support linking tables yet. If we decide to do this in the future, it is easy to add a PR adding to this field to support the other variant.

dsenalik:
I have moved the function to the parent class and tested, it appears to still work correctly 🎯
This is ready for review.

tripal_chado/src/Plugin/Field/FieldWidget/ChadoAnalysisWidgetDefault.php
```diff
@@ -0,0 +1,73 @@
...
+    $elements = [];
+    $elements['record_id'] = [
+      '#type' => 'value',
+      '#default_value' => $record_id,
+    ];
+    $elements['analysis_id'] = $element + [
+      '#type' => 'select',
+      '#options' => $analyses,
+      '#default_value' => $analysis_id,
+      '#placeholder' => $this->getSetting('placeholder'),
```
laceysanderson, line 66: Where does the placeholder get set?

laceysanderson:
✅ Code Review looks good except for the one question above.
✅ Added chadostorage automated testing for phylotree + quantification use cases.

Now I just need to do a functional review using the phylotree example :-)

dsenalik:
> Where does the placeholder get set?

Oh my. That is in several other fields also, but I have no clue what it is for. I won't be around much any more this week, but will invesitgate next week.

laceysanderson (approved):
Added code so this field and the dbxref_id are added automatically.
After this I was able to do a functional test using the phylotree content type and it all worked beautifully!

tripal_chado/src/Plugin/Field/FieldWidget/ChadoAnalysisWidgetDefault.php
laceysanderson: Resolving this in favour of a new issue to look at this in all fields it occurs in.
https://github.com/tripal/tripal/issues/1600

PR #1584, opened 2023-07-21
Full Test Suite for ChadoStorage: Single Property Field

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1546

## Description
This PR expands upon #1552 in a couple of ways:

**1. It provides a simplified interface for easier testing of other fields**

We have moved to needing to call a single method (`chadoStorageTestInertValues`) with a simplified values array. This single method tests the following for you: 

- creation of property types and property values, 
- adding the types to ChadoStorage, 
- creating a field config mock and data storage array, 
- calling ChadoStorage->insertValues().

Then you simply query chado to ensure the expected records are present.

**2. Support for testing ChadoStorage->loadValues and ChadoStorage->updateValues()**

Using the same type of values array as above, you can call `chadoStorageTestLoadValues` and `chadoStorageTestUpdateValues` to test the same stack of functionality as `chadoStorageTestInsertValues`.

## Testing?

This is another automated test only PR. As such there is no manual testing.


spficklin:
This looks great! I think it will make testing of fields easy.  I had to merge in the 4.x branch and so once the automated testing is done we can merge this in.

spficklin (approved):

laceysanderson:
Awesome, Thanks for the review! I'm working on documentation in https://github.com/tripal/tripal_doc/tree/tripalissue1546-chadoStorageKernelTests so that others can implement this in their fields.

Issue #1579, opened 2023-07-12
Tripal 4 importer logging enhancement


dsenalik:
# Core Tripal 4 Development Task

This issue is a place to put some cleanup of some of the logging and error messages with any of the importers.

## Branch Name: tv4g4-issue1579-logging

Dependent on #1585 (now merged)

laceysanderson:
Completed with PR merge

dsenalik:
Sorry, reopening, because this issue is a placeholder for some log messages I wanted to (but haven't) worked on.
1. Removing commented-out Tripal 3 versions
2. Log calls that don't use variable substitution
_e.g._
`$this->logger->error("Failed to update feature '$name'");`
should be
`$this->logger->error("Failed to update feature '@name'", ['@name' => $name]);`

This will take a slow tedious examination of all the importers - that's why I haven't done it yet 🙈

Notes for myself from https://drupalize.me/blog/how-log-messages-drupal-8
**Placeholder types**

&commat;variable — Use this style of placeholder for most use-cases. Special characters in the text will be converted to HTML entities.

%variable — Use this style of placeholder to pass text through drupal_placeholder() which will result in HTML escaped text, then wrapped with <em> tags.

&colon;variable — Use this style of placeholder when substituting the value of an href attribute. Values will be HTML escaped, and filtered for dangerous protocols.

PR #1585, opened 2023-07-22
Cvterm autocomplete needs filter by cv

### LINKED ISSUE #1571: Cvterm autocomplete needs filter by cv

## Pull #1585

The [Cvterm Autocomplete](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php) needs to be able to filter by cv. This was [mentioned as a todo in the code](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php#L17-L18) as it is needed for multiple importers. An example of this is the FASTA importer for which the drop-down will only select sequence ontology terms. In it's current implementation, the cvterm autocomplete used on the fasta importer allows you to pick any term and then the validation fails if it is not a sequence ontology term. Ideally you would only be able to select valid terms 😜 

## Approach

We need to change the autocomplete parameters in routing to include the cv_id. This should be added after count and it should be optional since some cvterm autocompletes do want access to all terms.

https://github.com/tripal/tripal/blob/4.x/tripal_chado/tripal_chado.routing.yml#L141-L148
```yaml
tripal_chado.cvterm_autocomplete:
  path: 'chado/cvterm/autocomplete/{count}'
  defaults:
    _controller: '\Drupal\tripal_chado\Controller\ChadoCVTermAutocompleteController::handleAutocomplete'
    _title: 'Chado Vocabulary Term Autocomplete'
    _format: json
  requirements:
    _permission: 'access content'
```

Next you will change the handleAutocomplete() method to handle an optional cv_id.

https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php#L13-L73

The remaining methods in the autocomplete controller should be good as is since they depend on an already chosen term and do not need to be restricted thus.

Once you finish the above, you will want to add additional asserts to the cvterm test to make sure this functionality works in the automated setting and that it continues to work ;-)

You'll also want to update the FASTA importer to restrict to the sequence ontology. This then provides a way to manually test in the form. You can also manually test using the route itself.

dsenalik:
# Tripal 4 Core Dev Task

## Description
The [Cvterm Autocomplete](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php) needs to be able to filter by cv. See Issue #1571 for full description.

This pull request implements this.

There is a change I have made. The autocomplete formerly had a wildcard character on both ends, but this means some short CVterms cannot be selected since too many terms match! For example, to select `RNA` we are presented with
RNA
To resolve this I have removed the wildcard from the beginning of the string. **This matches Tripal3 behavior**, e.g. 
`$records = chado_query($sql, [':str' => $string . '%']);`

Now you will see:
RNA2

Automated testing is updated to reflect this.

The internal match behavior is still available to the user if needed by starting with a `%` character when typing, an automated test does this for project.

## Testing?
I have implemented this in the FASTA and GFF3 loaders, so use that for testing, such as the **Sequence Type** field in the FASTA loader.
A good term to test with is `is_a` because it is in both the sequence ontology and in the taxonomic_rank ontology, and another term starts with this `local:is_analysis`. Only the `is_a (SO:is_a)` term should appear for the FASTA importer.
is_a

We can also test the case where the user enters a term and the autocomplete is not allowed to append its ` (db:accession)`
try typing in `is_a` and don't select the autocomplete option. It should still work. Equivalent code to test this
```
use Drupal\tripal_chado\Controller\ChadoCVTermAutocompleteController;
$cv_autocomplete = new ChadoCVTermAutocompleteController();
$type = 'is_a';
$cvterm_id = $cv_autocomplete->getCVtermId($type, 'sequence');
var_dump($cvterm_id);
```
Prints `"int(205)"`


dsenalik:
I did this same change of removing the wild card at the beginning to the project autocomplete also 🔨

dsenalik:
I did pull #1585 so that I could use that for the newick loader form 😁

laceysanderson:
Great, Thanks @dsenalik! 

@reynoldtan can you take a look at this and try it out in our module? Then submit a review on here so I can get this merged in faster? Thanks!

dsenalik:
When using this in a loader, we also need to call `getCVtermId()` on the returned CV term in both validate and submit - so until I do that for the Fasta and GFF loaders, this is draft status...

dsenalik:
While incorporating the `getCVtermId()` function, I found more issues with the loaders that need to be fixed, such as validation not tagging the offending field. These will be put in issue #1579 as they are not directly related to this PR.

Now both the Fasta and GFF loaders can be used to test the CV term select functionality.

reynoldtan:
Tested cv id parameter of 0 (default) to do a full cvterm term search and specific cv id for a more refined search - works in both cases. A minor change I would like to suggest is to re-order the parameter definition in the routing to match that of project autocomplete where the count parameter is the last parameter.

```
routing for project autocomplete:
path: 'chado/project/autocomplete/{type_id}/{count}'   

```

I also misspelled Controller in this line.
https://github.com/tripal/tripal/blob/e7a20967bb21646bb9b6cc7da89e8cac77675821/tripal_chado/src/Controller/ChadoCVTermAutocompleteController.php#L14

Thanks!

dsenalik:
@laceysanderson you suggested to put the type_id as the second parameter in the routing (because it would not always be used).
I did notice the discrepancy with the order for project, but didn't say anything 🙊.
What do you think of changing it to match project? I think consistency is a good thing.
Or maybe project should be the one changed?

dsenalik:
@reynoldtan @laceysanderson  I have gone ahead and switched the order of the cvterm autocomplete parameters to match project with https://github.com/tripal/tripal/pull/1585/commits/5214246f2ebe81388523631d493237fa0516cb4b
If by chance we wanted the other order, I can revert that commit, but if it is good then this is ready

dsenalik:
Random note, in the lookup function to get cvterm_id from a term name, I changed the `LIMIT 1` into an exactly-one-match. This would most likely only be important if a user manually typed in the name and didn't let the autocomplete do its job, which it is possible to do. And if there was not an explicit CV specified (`sequence`) as there is with the FASTA and GFF3 loaders.

dsenalik:
@laceysanderson all you need to confirm here is that you approve of this order for the parameters (or you want them swapped) `chado/project/autocomplete/{type_id}/{count}`

@reynoldtan do you want to do a final test?

laceysanderson:
Ah, thanks for pulling me back to this one @dsenalik.

We definitely want `chado/cvterm/autocomplete/{count}/{cv_id}`
and project should be changed to `chado/project/autocomplete/{count}/{type_id}`
for consistency and to ensure both {cv_id} and {type_id} are optional parameters.

dsenalik:
Thanks for the feedback @laceysanderson , I have swapped the parameters

laceysanderson (approved):
Functional test works with this most recent change. Code review also looks good! Thanks @dsenalik

Issue #1565, opened 2023-07-05
REVEAL: Remove `bio_data` assumption from content types


laceysanderson:
# Core Tripal 4 Development Task

## Current Issue Description

We have used bio_data in the URLs and a couple of other places in our Tripal Content. However, this is not really necessary and not ideal for researcher friendliness 🙈 

In #1413 Stephen started to remove this assumption, but it was not fully completed and may have been the change that broke the routing in that PR (we were never able to confirm this though).

Specific places to remove Bio_data from:
- content type machine name. 
    - It would be easier if these were named in a human readable way (e.g. organism instead of bio_data_1) to make it easier for extension modules to alter specific content types. You should not rely on the specific bio_data_### as this is dependant upon order of creation. As such you currently should look up the content type using other means or look up the bio_data_# mapping first. This can be error prone...
    - tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml and tripal_chado/config/install/tripal.tripalentitytype_collection.default_chado.yml
    - TripalFieldCollection.php and TripalEntityTypeCollection.php
- Routing / URLs
    - the URL within the structure: edit, manage field, manage display, etc.
 
## Background

The background for this issue is a cautionary tale: Once upon a time a PR was made, #1413, with so many great changes for Tripal 4. This PR was focused on moving content type + field creation into YAML files for the good of all. This was a large task in and of itself but as such tasks have a habit of doing... it picked up many additional bug fixes and features as it grew. The developer was testing locally both manually and running the automated tests they developed just as they should but forgot to push out to github as often as recommended 😨 This caused the developer to miss a test failure in unrelated code within their beautiful PR 😱 As the PR had become an inter-related collection of fixes by the point it came up for review and the unrelated failing test was found... it became a nightmare to debug and though the developer and their friends tried, they could not find the bug 😭 🗿  This meant that the PR had to be redone with careful testing and often github pushing to ensure the bug did not resurface... This processed birthed PR #1475 and this collection of issues.

tl;dr;
- PR #1413 included a large number of fixes but needed to be closed due to a test failure in unrelated code which could not be fixed
- PR #1475 brought back the core functionality the PR originally meant to cover (Content Types + Fields created via YAML) but could not bring back all of the fixes from the original PR.
- This issue has been created to mark the place of one of these fixes to be brought back.

## Process:
- Look at PR #1413 and try to find the code that made this particular fix. 
- Create a new branch from 4.x using the branch name above
[...]

PR #1588, opened 2023-07-24
Remove 'bio_data' from Tripal

spficklin:
# Tripal 4 Core Dev Task

Issue #1565 

## Description

The goal of this PR is to remove the use of the word `bio_data` in Tripal. As part of Tripal3 it was used as an equilvalent of `node` but for Tripal content types.   The goal was to use the bundle name (e.g. organism, analysis, contact, etc.) instread of `bio_data`.   After digging through the code I found this wasn't easily doable.  I got part of the way there, but hit on problems when creating brand new content items.  So, instead, I opted for using `tripal` instead of `bio_data` and I enabled the automatic creation of URL aliases using the `url_format` setting for content types.  So, all tripal entities will get a default URL of something like `tripal/1`, `tripal/2`, etc. but with the URL alias will appear to the user as something like `organism/1`, `project/2`, etc.  

The best benefit for this change is that content type are no longer referred to as `bio_data_1` and `bio_data_2` but rather by a more human-readable name (e.g. `organism`, `contact`, etc.).

## Testing?

There are two items to test.
1. Entity Creation: To  test this, simply update your Drupal installation to include this branch, install Drupal and create tripal an organism, project, analysis... whatever you like.  After creation, they should have URL aliases appropraite for the content type and you should be able to get to the same entity by changing the URL to have `tripal\<num>` where `<num>` is the same ID in the original URL (e..g `organism\1`)
2. Content Type Creation:  Now that we no longer use `bio_data` for the name of content types, you can create a new content type and it will use as a name a lowercase version of the label you give it.  Go to Admin > Structure > Tripal Content Types  and click the `Add Tripal Content Type` button at the top.  You should be able to create a new content type. Notice the machine name in the form (to the right of the label field), it should be a lowercase version of whatever you type in.  Any non alphanumeric values get converted to underscores.

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -162,29 +162,29 @@ tripal.tripalentitytype_collection.*:
                  type: string
                  label: "Category"
                  nullable: false
-               # name:
-               #   type: string
-               #   label: "A machine readable content type id (a.k.a., the bundle ID or Entity Type ID)"
-               #   nullable: false
+               name:
+                 type: string
+                 label: "A machine readable content type id (a.k.a., the bundle ID or Entity Type ID)"
+                 nullable: false
                help_text:
                  type: string
                  label: "Text used to describe to the end-user what this content type is used for"
                  nullable: false
-               # url_format:
-               #   type: string
-               #   label: "A tokenzied string for the URL of the content type. Field names in square brackets can be used as tokens."
-               #   nullable: true
+               url_format:
+                 type: string
+                 label: "A tokenzied string for the URL of the content type. Field names in square brackets can be used as tokens."
+                 nullable: true
                # title_format:
                #   type: string
                #   label: "A tokenzied string for the title of the content type. Field names in square brackets can be used as tokens."
                #   nullable: true
-               # synonyms:
-               #   type: sequence
-               #   label: "Synonyms"
-               #   nullable: true
-               #   sequence:
-               #      type: string
-               #      label: "Synonym name"
+               synonyms:
+                 type: sequence
+                 label: "Synonyms"
+                 nullable: true
+                 sequence:
+                    type: string
+                    label: "Synonym name"
 
 ## For TripalFieldCollection functionality
 ## focused on using YML files to describe TripalFields
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -162,7 +162,7 @@ tripal.tripalentitytype_collection.*:
                  type: string
                  label: "Category"
                  nullable: false
-               name:
+               id:
                  type: string
                  label: "A machine readable content type id (a.k.a., the bundle ID or Entity Type ID)"
                  nullable: false
--- a/tripal/config/schema/tripal_entity_type.schema.yml
+++ b/tripal/config/schema/tripal_entity_type.schema.yml
@@ -1,4 +1,4 @@
-tripal.bio_data.*:
+tripal.content_type.*:
   type: config_entity
   label: 'Tripal Content Type'
   mapping:
--- a/tripal/config/schema/tripal_entity_type.schema.yml
+++ b/tripal/config/schema/tripal_entity_type.schema.yml
@@ -3,11 +3,8 @@ tripal.content_type.*:
   label: 'Tripal Content Type'
   mapping:
     id:
-      type: integer
-      label: 'ID'
-    name:
       type: string
-      label: 'Name'
+      label: 'ID'
     label:
       type: label
       label: 'Label'
--- a/tripal/js/tripal.js
+++ b/tripal/js/tripal.js
@@ -116,7 +116,7 @@
    */
   TripalAjaxField.prototype.attach = function () {
     $.ajax({
-      url     : baseurl + '/bio_data/ajax/field_attach/' + this.id,
+      url     : baseurl + '/tripal/ajax/field_attach/' + this.id,
       dataType: 'json',
       type    : 'GET',
       success : this.setFieldContent.bind(this)
@@ -212,7 +212,7 @@ function tripal_navigate_field_pager(id, page) {
   jQuery('#' + id + '-spinner').show();
   jQuery.ajax({
     type   : 'GET',
-    url    : Drupal.settings['basePath'] + 'bio_data/ajax/field_attach/' + id,
+    url    : Drupal.settings['basePath'] + 'tripal/ajax/field_attach/' + id,
     data   : {'page': page},
     success: function (response) {
       jQuery('#' + id + ' .field-items').replaceWith(response['content']);
--- a/tripal/src/Controller/TripalEntityUIController.php
+++ b/tripal/src/Controller/TripalEntityUIController.php
@@ -43,7 +43,7 @@ class TripalEntityUIController extends ControllerBase {
       $bundles[$category]['members'][] = [
         'title' => $entity->getLabel(),
         'help' => $entity->getHelpText(),
-        'url' => Url::fromRoute('entity.tripal_entity.add_form', ['tripal_entity_type' => $entity->getName()]),
+        'url' => Url::fromRoute('entity.tripal_entity.add_form', ['tripal_entity_type' => $entity->id()]),
       ];
     }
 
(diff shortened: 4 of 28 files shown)


dsenalik:
@spficklin this error in your automated tests
`
2) Drupal\Tests\tripal\Functional\TripalRoutePermissionsTest::testTripalContentPages
Exception: The name is required when creating a TripalEntityType.
`
is a validation I just added in pull #1583 - when I was adding the name I thought it was missing here too https://github.com/tripal/tripal/blob/e7a20967bb21646bb9b6cc7da89e8cac77675821/tripal/src/Entity/TripalEntityType.php#L243-L245

If by chance it is valid to not have a name for an entity, then you can remove these three lines on your branch and my apologies 😦. If, on the other hand it is your test that is in error, you can fix that and I will be happy that it was caught 😀.

laceysanderson:
I marked this as a draft PR until the tests are passing.

spficklin:
@dsenalik the validation check you added is correct.  The problem with the testing is something else.  Digging into it also made me realize that we don't need a `name` parameter for the content types. Now that we no longer need to keep a numeric ID around (i.e. for the bio_data_x machine name) we no longer need a separte Id and Name.

dsenalik:
@spficklin I was curious about why `bio_data` needed to be changed. It is not a big deal, this may be the only example I have, but I used a bio_data link to access some content and if it changes later when we update Tripal it would break the link.
bio_data-link

spficklin:
Hey @dsenalik .  In Tripal v3 `bio_data_x` was used all over the place, in field names, as content types and it was really annoying to have to use a lookup table when you were digging around in the code or tables in postgres.    It made the code and data hard to follow.  Also, there was no consistency between the bio_data numbers and the content types after a site had prepared the initial content types. New types would not have consistent types across sites.   The changes in Tripal v4 allow us to use a human-readable name for the content types which is so much easier for field naming and looking through all those tables that Drupal creates for fields.    So, removing it is more convenience for us rather than anything related to functionality..

laceysanderson:
Yes, there have been bugs in extension modules pop-up where they made an assumption on a specific bio_data_# always being a given content type until one site it wasn't 🙈 Also as Stephen mentioned, it just makes coding extension modules more difficult than it needed to be when trying to target specific content types.

For your link, the first iteration had a synonym key which was intended to provide alias' back to the old bio_data_# to keep it intact. Since the synonym was in the YAML, you could override it with the specific bio_data_# for your site and ensure links like the one you showed do not break 💞

dsenalik:
Maybe there is some confusion, I was not referring to the `bio_data_#` but rather to the generic `/bio_data/entity#` in a URL. Such as `/Project/1234` is equivalent to `/bio_data/1234` in a URL, and it looks like this is what is changing to `/tripal/1234` (though there are some differences to this rule when editing, and in the web services)

The alias option is very desirable, because what if you put the link in a paper that now can't be changed.

I fully agree that looking at `bio_data_#` tables in the DB is always confusing 😕

spficklin:
@dsenalik oh... for the URLs you're right, that's a change.   I just throught I'd get rid of it all together and sicne with Tripal v3 it should devault to using the URLs aliases I made the assumption that changing this wouldn't be a bad idea for published URLs.  But you're right, if someone published URLs with the bio_data in them they won't work anymore.  We could, as part of the site upgrade process, add as a secondary URL alias the old bio_data URLs from the Tripal v3 sites....

spficklin:
Aside from the naming of the branch which I guess I munged, this PR is ready for review!  All tests are passing.

dsenalik (commented):
As far a code review, this seems good.

For testing, I built a new docker on this branch with 
`docker build --tag=tripaldocker:testing-1565 --build-arg drupalversion="10.1.x-dev" --file tripaldocker/Dockerfile-php8.2-pgsql13 ./`

I created three content types and the URL created was as expected
```
/organism/1
/gene/2
/genetic_marker/3
```

I verified that the tripal URLs worked
```
/tripal/1
/tripal/2
/tripal/3
```

I verified that the bio_data URLs did not work
```
/bio_data/1
/bio_data/2
/bio_data/3
```

I created a new content type "Primer" under the "Genetic" category using the CV term "DNA primer (SO:0000112)"
**The machine name was not populated**
And if I save
`The website encountered an unexpected error. Please try again later.
Drupal\Core\Entity\EntityMalformedException: The entity does not have an ID. in Drupal\Core\Config\Entity\ConfigEntityStorage->save() (line 246 of core/lib/Drupal/Core/Config/Entity/ConfigEntityStorage.php).`

spficklin:
Whoops!  Sorry about that @dsenalik   I borked  the form when I made fixes and forgot to check it.  You should be able to create those new entity types now.  I did notice you can't add fields to new entity types, but I don't think that's a problem with this PR, but should e added as a new issue. I'll do that.    It's ready for testing that last bit again.

dsenalik:
I tested your changes, and it is now working as expected. I entered "Primer" as the content type, and the machine name was populated with "primer". I then clicked the edit link and changed it to "primer1" successfully.
primer1

dsenalik:
I would like to discuss, maybe at the next cofest, changing the `bio_data` in the URL to `tripal` though. I feel this particular change could be a potential problem for existing sites or modules.

spficklin:
@dsenalik Sure, we can wait to merge it if you want to chat on it.   I thnk though we can accommodate it, if we make sure to add a second URL alias, when upgrading a Tv3 site to Tv4 to set the `bio_data/x` URL alias.  That means both the new `/tripal/xxx` and previous `bio_data/yyyy` and the prefered `organism/xxx` will all still work.   We did something similar for the upgrade from Tv3 to Tv4.

spficklin:
Oh, I should add. I didn't like setting it to `/tripal/xxxx` but I hated `/bio_data/yyyy`. I just wanted to get rid of it.  I suspect @laceysanderson might feel the same way?  Anway, if we're okay preserving `/bio_data/xxx` as a URL alias for upgraded sites then maybe we could come up with something better than `/tripal/xxxx`?   Drupal uses `/node/` for their content types.  Maybe we could come up with something better.

dsenalik:
I agree that `tripal` is probably the most logical choice. And I can see you had a bad experience with `bio_data` 😆 Here's an idea, what about a single letter? `/t/1234` - although these are not URLs you are going to type in (usually) so it really makes no difference. This could potentially make debugging harder though, how can you grep for just `t`?
If we have and document the `bio_data` alias method for upgrading sites, then I am happy with this. I will mark my review as approved and let the will of the PMC prevail 😎

dsenalik (approved):
Working as intended 🎉

spficklin:
I do like '/t/xxxx`.  It makes the URL shorter, it could also be `/data/xxxx`. I like that better than `/bio_data/xxxx`.  

> I will mark my review as approved and let the will of the PMC prevail 😎

Deferring to @laceysanderson then to decide if we wait for further discussion or merge.  @laceysanderson , if you vote to change the URL from `/tripal/xxxx` I can fix it here before we merge.

laceysanderson:
To be honest I like any of `bio_data/xxxx` , `tripal/yyyy`, `data/zzzz` so I may not be the best tie-breaker 🤪 . Of these I might lean slightly towards `data/zzz` since it might be easier to search for if there is a problem as Tripal is literally everywhere in the code :zany_face:!

I think that we just let Stephen pick one rather than wait for discussion as this is a minor enough change to not need a lot of discussion. All I ask is that we create an issue to remind us to add the bio_data redirect in the upgrade code!

dsenalik:
Okay @spficklin you get to decide! If I can, tomorrow I will create the issue for bio_data redirect, as I am out thursday-friday.

dsenalik:
And don't forget `/t/xxx` or `/d/yyy` or `/🥕/zzz` as other options

spficklin:
Okay. After thinking on it, I decided to just use `bio_data/xxxx`.  That way we don't have to create more work for ourselves by creating those aliases.  No reason to make more work for everyone if I'm the only one who really dislikes it.  Lacey makes a good point that it's easier to find where these links are used in the code if we don't use `tripal/xxxx`.  

I've updatd the code and running tests, and  i did another round of manual testing.  If all the automated tests pass I'll go ahead and merge.

Issue #1461, opened 2023-03-27
Tripal 4 Pub Importer


spficklin:
# Core Tripal 4 Development Task

## Branch Name: tv4g4-issue1461-pub_importer

The design for porting the Pub importers from Tripal v3 to Tripal v4 was discused by @dsenalik , @laceysanderson and @spficklin  and is as follows:

In summary we will create:
1.  A new plugin type:  `TripalPubParser`
3. Three new TripalPubParser Implementations
     - `TripalPubParserAGL`
     - `TripalPubParserPMID`
     - `TripalPubParserBibTex`
4.  An implementation of the `TripalImporter` called `ChadoPubImporter`

The `TripalPubParser` class will perform the following
1.  Provide a form for collecting all information the parser needs to obtain and parse publication data. This includes search strings, URLs, files, etc
2.  A `run` function that will accept an input list of arguments that will cause the plugin to retrieve pub data, parse it and return it in an array structure that is the same for all pub importers.  

The `ChadoPubImporter` will need to talk to the `TripalPubParser` plugins in the following way:
1. For the importer form
    - First it will provide a dropdown of all available `TripalPubParser` implementations. The user will select the type of parser then want to setup/use.
    - After the user selects the parser, the `ChadoPubImporter` will retrieve the appropriate form and display it to the end-user.  The user will provide all of the necessary information to the form. 
    - Somehow (not sure how just yet), the parser plugin will need to give to the `ChadoPubImporter` the list of arguments needed for its `run` function using the values provided by the user in it's form.  
    - This information provided in the previous step will be summarized in a table and stored persistently in a the `public.tripal_pub_import` table (form Tripal v3) so that in the future anyone who returns to the pub importer can select the same importer information to re-run the importer job.
2. For submitting the importer job
   - After all parsing run information is collected, and saved in the `public.tripal_pub_import` then when the importer is submitted for running it will then call the appropriate run function for the parser.

dsenalik:
My working notes about the files for this importer

## INSTALLATION

FILE: `tripal/tripal.install`
🠚 define the new table `tripal_pub_import`

## MENU ITEMS
### JOB RUNNER
FILE: `tripal_chado/src/Plugin/TripalImporter/ChadoPubImporter.php`
🠚 Plugin for importer that shows up in menu
🠚 defines importer form, lists existing importers
namespace `Drupal\tripal_chado\Plugin\TripalImporter`
[...]

PR #1591, opened 2023-07-26
Tv4g4 issue1513 newickloader

dsenalik:
# Tripal 4 Core Dev Task

## Description

This pull request will
  * Implement the Newick file importer
  * Tests for the Newick importer
  * add kernel tests for the phylotree API functions that require the Newick file importer

I have implemented this because it was easier than doing the publication loader (#1461), and hopefully I now understand a bit more.

1. I have implemented this as a more generic "tree" loader with plugins for different tree file formats. But right now there is only one format supported, Newick, which is all Tripal 3 supported. There is a plugin for Newick, but it currently doesn't do much of anything, but if we add other formats in the future this will be super helpful.
2. Currently a draft so more to do...

## Testing?
to be added

## Tree Loader Form
TreeLoaderForm


dsenalik:
There is much to consider and discuss for loaders, so for now I am closing this to unclutter the PR queue, and will let @risharde create the newick loader 👍

PR #1596, opened 2023-07-31
Always print debug messages

dsenalik:
# Bug Fix

## Description

The tripal logger inherited a toggle from Tripal 3 to enable and disable debug messages. To enable, an environment variable `TRIPAL_DEBUG` needed to exist.

Our preference for Tripal 4 is to always print debug messages.

## Testing?
In devel module php window execute this
```
$message_opts = [
  'drupal_set_message' => TRUE,
];
\Drupal::service('tripal.logger')->debug(
  'This next text should be substituted: "@fail".', 
  ['@fail' => 'testpassed'], 
  $message_opts
);
```
You should see this
debug1

And in the log you should see this
debug2


dsenalik:
The documentation pull request https://github.com/tripal/tripal_doc/pull/40 assumes that this pull request is merged, i.e. always print debug messages, so it does not mention the old environment variable `TRIPAL_DEBUG`

laceysanderson (approved):
Beautiful <3

Issue #1550, opened 2023-06-14
Tripal 4: Implement some of the missing core fields and add to content types


dsenalik:
# BUG/ERROR report

### Tripal Version: 4.alpha2.dev docker
* PHP Version: 8.1.18

## branches
* tv4g1-issue1551-Add-chado-boolean-field
* tv4g1-issue1414-ChadoAnalysisDefault
* tv4g1-issue1414-ChadoContactDefault

### Issue description
This issue was originally named **Tripal 4: Some content types are lacking some fields for NOT_NULL table columns** but it turns out the major need is to create a number of missing chado fields.

The original description: Some of the fields don't work because they are missing a field to enter the organism.
Try to create a new "Genetic Marker" entity.
Unique Name="Marker 1"; Sequence="ACTG"; Sequence Length=4

`Deprecated function: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in Drupal\Core\Entity\ContentEntityForm->flagViolations() (line 265 of core/lib/Drupal/Core/Entity/ContentEntityForm.php). `
...
`The item cannot be saved because the following values are missing. Organism_id.`

Fixing the missing organism field is all that is needed for this, as the field `ChadoOrganismDefault` exists.

See https://github.com/tripal/tripal/issues/1550#issuecomment-1595138368 for the fields we need to add

laceysanderson:
For these, that is likely because that field was missed in the yaml file. To fix/check in the case, look at [the yaml file defining all fields for our content types](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml) and check that the organism_id field is included for those specific content types.

PS. This is also the way you will make sure your new fields are auto-attached to the content types. It's a simple matter of adding them to this particular yaml for the content types they pertain to.

dsenalik:
**To-Do List**
I tested **all** content types manually, the following need organism added (of the feature types, only **Gene** is okay)
  * **Genetic Marker**  _fixed with #1555_
  * **Phenotypic Marker**  _fixed with #1555_
  * **Sequence Variant**  _fixed with #1555_
  * **QTL**  _fixed with #1555_
  * **mRNA**  _fixed with #1555_

**DNA Library** has an additional missing field:  _fixed with #1555_
`The item cannot be saved because the following values are missing. Organism_id, Is_obsolete.`
**Phylogenetic Tree** needs a different field added:
[...]

PR #1599, opened 2023-08-09
Tripal 4 ChadoContactDefault field

dsenalik:
# Tripal 4 Core Dev Task

## Description

This is the initial implementation of the ChadoContactDefault field.

There are only two tables with a `contact_id` column that will be able to use this field:
  * `stockcollection` (but is not a default content type)
  * `study`

and in addition the `arraydesign` table has a `manufacturer_id` which maps to the contact table

Note that before this PR, we can't create a study because `contact_id` has a NOT NULL constraint in the study table. This PR includes an update to the YAML as described in issue #1550 so that it is easy to test.

## Dependency
As of https://github.com/tripal/tripal/pull/1599/commits/591868140958fe6776e885856bd8cf979956e629, this PR is dependent on pull #1604. ~It cannot be merged until that one is merged.~ it has been merged

## Testing?
I will test on both the Study and Array Design content types. Currently you can't create a study because the `contact_id` is NOT NULL. This new field will fix that! Similar problem for Array Design.

1. Create two contact records, one a `Person (TCONTACT:0000003)` and one a `Company (NCIT:C54131)` (though types do not really matter)
contacts

2. Now create a study, and specify the person contact.
study1

3. It should appear similar to this
study2

4. Create an array design, and specify the company contact as the "Manufacturer". "Platform Type" is also a required field, just enter any random CV term there, I used "DNA".
ad1

5. Content should be saved and appear something like this
ad2

If you happen to notice that that "Platform Type" is blank, that is actually issue #1398 but not relevant to this PR


dsenalik:
I am hoping this field is done.
* the field is added to the two content types in the YAML with https://github.com/tripal/tripal/pull/1599/commits/f823b9ded70518f7489870b82c8941eac89813e5 , manufacturer is put at the top of the form to match Tripal 3
* I have added an ugly special case handling with https://github.com/tripal/tripal/pull/1599/commits/9d2c9683b24abaf4877591b507a5f7c73116a313 because the `manufacturer_id` column in the `arraydesign` table is the column that corresponds to the `contact_id`. I don't know if there is a more elegant way to do this.
_Update_ one special case removed with https://github.com/tripal/tripal/pull/1599/commits/2ff984f3581e2e1d6ad66380fd4de6dfc22c22dc
* I have added testing following @laceysanderson 's tests for the ChadoAnalysisDefault field. I have started to genericize it just a bit so there is less work using this as a base for creating tests for another field, I am using `base` and `linked` in the sql instead of `study` and `contact`, just one fewer thing to change.
* I added an additional assertion in `tripal_chado/tests/src/Traits/ChadoStorageTestTrait.php` with https://github.com/tripal/tripal/pull/1599/commits/e91cf2a9abbda069c54af19b3e0687c5490f5557 to catch my own stupid mistakes.
* typo `intstrument`

dsenalik:
After discussion with @laceysanderson we decided to make the `null` contact more user friendly with https://github.com/tripal/tripal/pull/1599/commits/2b0f3d8330769d4b77ee83fcebb7cb9d4054415e
This is the contact with name `null` which is defined by chado, which will have `contact_id==1`
For the user-facing select list presented by this field, and in the formatter, it will be changed to `Unknown`.
The same can be done for the publications, where there is a null publication defined (future project).
Ideally this would be a configurable setting, but we will not implement that now.

dsenalik:
Example of select list now, the Unknown should sort to the top
Select
And sort is now case insensitive, before all contacts starting with upper case would come before those with lower case.

laceysanderson (approved):
Code review looks perfect including the tests!
Functional review also works perfect after a bit of back and forth during cofest!

I tested on both study and array_design:
- Can set it to the null contact
- Can set it to an existing contact
- If I leave it at select it won't let me submit as it's required
- updating works great to both another existing contact and the null contact
- validation for required works on update as well as create

PR #1601, opened 2023-08-13
Switched back to bio_data links

spficklin:
# Bug Fix

Issue #1565

## Description

It seems I forgot to push the last commit to change the `tripal/xxx` URLs back to `bio_data/xxx`.  :-/ This should have been part of PR #1588. I'm not sure how this happened because I saw the PR redo the testing.  How would that have happened if I hadn't pushed?  I must have confused myself.  So,  this is that last commit.  I should have let someone test it again.  So, I'll leave this open for testing before merging.

## Testing?

- Entity Creation: To test this, simply update your Drupal installation to include this branch, install Drupal and create tripal an organism, project, analysis... whatever you like. After creation, they should have URL aliases appropraite for the content type and you should be able to get to the same entity by changing the URL to have tripal\<num> where <num> is the same ID in the original URL (e..g organism\1)
- Content Type Creation: Now that we no longer use bio_data for the name of content types, you can create a new content type and it will use as a name a lowercase version of the label you give it. Go to Admin > Structure > Tripal Content Types and click the Add Tripal Content Type button at the top. You should be able to create a new content type. Notice the machine name in the form (to the right of the label field), it should be a lowercase version of whatever you type in. Any non alphanumeric values get converted to underscores.


dsenalik (approved):
I merged 4.x into this branch

As for PR#1588, I built a new docker on this branch with 
`docker build --tag=tripaldocker:testing-1565 --build-arg drupalversion="10.1.x-dev" --file tripaldocker/Dockerfile-php8.2-pgsql13 ./`

I created three content types and the URL created was as expected
```
/organism/1
/gene/2
/genetic_marker/3
```

I verified that the bio_data URLs worked
```
/bio_data/1
/bio_data/2
/bio_data/3
```

I verified that the tripal URLs did not work
```
/tripal/1
/tripal/2
/tripal/3
```

I created a new content type "Primer" under the "Genetic" category using the CV term "DNA primer (SO:0000112)"
Success!

If I try to add a primer record, however, it only provides the fields "Title" (grayed out) and "Authored by"
Well, that is because it is then up to me to then add all required fields, but this is not relevant to this PR so I am stopping here on testing that.

dsenalik:
I'm going to go ahead and merge this now so that we can continue with other stuff 😀

spficklin:
Thanks @dsenalik !

PR #1603, opened 2023-08-14
Tripal 4 gff3 loader use cvterm lookup during run

### LINKED ISSUE #1602: Bug in GFF Loader
# BUG/ERROR report

### Tripal Version:
* Drupal Version:  9.x
* PostgreSQL Version:  15.1
* PHP Version:  8.1

### Issue description

The GFF3 loader works but has two issues that will cause problems.

1.  When submitting the form, if the user enters the 'Default Landmark Type' and allows the autocompleteo work then it will submit the value with the ontolgoy term Id (e.g. "chromosome (SO:0000340)").  This breaks the loader as it seesm to want just the term name (e.g. "chromsome").  If you just put the term name it works fine.  
2.  The alignment species target drop-down does not have an empty value, therefore, it defaults to the first species in the list.   I think it should be empty at first so that the user can set it or not.  

#### Steps to reproduce

1. Simply use the automplete when setting the landmark type.  If you download the MSU rice genome it doesn't have the chromosome as features so you have to enter the landmark type to force it to creat them.  You can downloa this from here:  http://rice.uga.edu/pub/data/Eukaryotic_Projects/o_sativa/annotation_dbs/pseudomolecules/version_7.0/all.dir/ (Download the `all.gff3` file).
2. For the alignment target that just something to observe in the form.

dsenalik:
# Bug Fix

## Description
1.  When submitting the form, if the user enters 'chromosome' for the 'Default Landmark Type' and allows the autocomplete to work then it will submit the value with the ontology term Id (e.g. "chromosome (SO:0000340)").  This breaks the loader as it seems to want just the term name (e.g. "chromsome").  If you just put the term name it works fine unless the name is ambiguous (i.e. exists in two different CVs)
2.  The alignment species target drop-down does not have an empty value, therefore, it defaults to the first species in the list.  It should be empty at first so that the user can set it or not.

## Testing?
1.  Create two organisms and an analysis
2.  I am using the `tripal_chado/tests/fixtures/gff3_loader/gff_Citrus_sinensis-orange1.1g015632m.g.gff3` file for this test.  Load this file.
3. For "Default Landmark Type" enter `chromosome` and select from the autocomplete `chromosome (SO:0000340)`
4. For "Target Organism" note that the dropdown starts at "- Select -", select your other test organism
5. For "Target Type", select `mRNA` and select the autocomplete option `mRNA (SO:0000234)`
6. Press "Import GFF3 file" button
7. `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal9/web`

## Changes
Commit 001a0f9f is a one line addition that fixes item 2
Commit 973515e0 incorporates the cvterm lookup. The nice thing is that is simplifies the code a fair amount.

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
+++ b/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
@@ -435,6 +435,7 @@ class GFF3Importer extends ChadoImporterBase {
         different species then the organism must be specified using the 'target_organism=genus:species'
         attribute in the GFF file."),
       '#options' => $organisms,
+      '#empty_option' => t('- Select -'),
     ];
 
     $form['targets']['target_type'] = [
--- a/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
+++ b/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
@@ -207,9 +207,9 @@ class GFF3Importer extends ChadoImporterBase {
   private $landmark_types_type_ids = [];
 
   /**
-   * The results object for the landmark type cvterm.
+   * The cvterm_id for the landmark type cvterm.
    */
-  private $landmark_cvterm = NULL;
+  private $landmark_cvterm_id = NULL;
 
 
   /**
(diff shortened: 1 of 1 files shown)


risharde:
@dsenalik @spficklin Ran the loader using Doug's instructions and all went well. Also run the GFF3 tests and they passed as well. I have one question since I'm not sure what the target organism: I noticed all features get created for my first organism (which was Citrus Sinensis). I see no features for the target organism (organism id 2). Does this sound normal? Just want to make sure we don't have any issue there!

spficklin:
Wow that was a quick resolution to the issue!  Thanks @dsenalik . @risharde the target is for alignments which are part of the GFF3 specification:  http://gmod.org/wiki/GFF3#Alignments.   The GFF3 file doesn't provide a way to specify which organism the aligned sequences are from.  So this is a Tripal-specific addition that we made to help with that.

Issue #1594, opened 2023-07-31
Better Debugging of ChadoStorage for Field Development


laceysanderson:
# Core Tripal 4 Development Task

## Branch Name: tv4g2-1594-debuggingChadoStorage

As we have been developing more fields, it's becomming apparent that ChadoStorage is too much of a black box at this point. For example, in cases where you've defined the properties as expected based on documentation but they still don't work --you just don't get your data... There is no way for you to see the queries created by chadostorage, get an idea of how chadostorage is processing your properties, etc. This basically means there is no way to debug your properties without adding debugging statements into chado storage yourself. This is not friendly to non-core developers... and since many extension module developers will be developing fields we need to fix this!

## Design

Create a debugging ChadoStorage trait to contain debugging specific functions. Where possible I will also put functions in a debugging TripalStorage trait if they are not specific to ChadoStorage.

Fields will be able to implement a debugging flag in the annotation and doing so will turn on the debugging trait. You can also turn on the debugging trait using chadostorage directly. Output will be both to the screen and to the watchdog log using the TripalLogger debugging level.

### Debugging Info to Provide

- [x] Report the complete SQL that chadostorage would execute.
- [x] Show simplified reporting of the contents of the property types, field definition and property values so you can confirm they match what you expected.
- [ ] Give hints when things are not present but are expected. This will be organically figured out in buildRecords() and may or may not be feasible. I will need to think on this one. We may be able to tell developers that if they provide a single store_link that they should also have additional properties expected for this type of link such as the store_pkey.

laceysanderson:
Ok, I'm calling this done for now. It doesn't make suggestions but it is already very helpful IMO and prints a lot 😬 so I think I'll leave it here.

You can use the branch now, I'm just waiting to make a PR until I have time to document how to use it -for now it's bedtime!

If you are excited to try this before the PR is made, debugging is enabled for a field through the field settings UI. 
For example, for the organism field on the gene content type, you would 
1. go to Admin Toolbar > Tripal > Page Structure > Gene > Manage Fields. 
2. Then beside the organism field you would clicked edit. 
3. The field edit page now has a simple "Enable Debugging" checkbox. Select this and save. 
4. Now whenever you create, edit, load gene pages you will see a lot of debugging output. 
5. To disable once done, just uncheck the debugging checkbox and save.

Here is an example of loading a gene page:
[...]

PR #1604, opened 2023-08-14
Better Debugging of ChadoStorage for Field Development

laceysanderson:
# Tripal 4 Core Dev Task

Issue #1594 

## Description

As we have been developing more fields, it's becomming apparent that ChadoStorage is too much of a black box at this point. For example, in cases where you've defined the properties as expected based on documentation but they still don't work --you just don't get your data... There is no way for you to see the queries created by chadostorage, get an idea of how chadostorage is processing your properties, etc. This basically means there is no way to debug your properties without adding debugging statements into chado storage yourself.

This PR adds a checkbox to the field settings form to enable debugging information for a field. This allows developers who are working on fields to target a specific field instance (i.e. their field attached to a specific content type with a specific set of settings) and receive the following debugging information to give them insight into how chado storage is interpreting the properties they gave them. This will help with detecting bugs both in fields being developed and in chadostorage itself!

Debugging information made available:

- The status of chado storage at the beginning of the process including the following information for each attached field instance:
  - field name, type and definition
  - field settings
  - the number of properties
  - an array with the following information for each property type:
    - the class name with namespace
    - the term attached
    - the storage settings set in the field
- The property values at the beginning of the process
- An array describing the records to be created. This is the result of buildChadoRecords and is used directly by chadostorage not generated for debugging. This is aggregated across all fields.
- The primary key values of the so-called base records as understood by chado storage. This is aggregated across all fields.
- The actual queries being executed by chado storage. There will be the record key and delta in the label. Unfortunately the placeholder values are not included due to Drupal bugs documented in the code.
- The values after the process is complete. These will be the values made available to the widget and formatter classes.

screenshot

## Notes

- This debugging information is posted for load, insert, update, validate (+ delete/find once implemented)... 
  - Inserting results in both a load and insert
  - update though is particularly long with multiple loads and the update. This is how Drupal works for some reason 🤷‍♀️ 
  - there can be a lot of information which could be overwhelming
- Since multiple fields make up a single chado record, if you enable debugging for a single field you actually see information for all fields on that content type. I felt this was needed to truely debug
- The content listing page might time out if there are too many entries with debugging turned on as that triggers a full load for every entity on the page.
- The field debugger uses dpm so devel is required to use it 🤷‍♂️ we can't use html in drupal_set_message and I felt formatting was necessary for this to be usable at all.

## Testing?

1. Create a docker image and container on this branch. [See these instructions for how to do that.](https://tripaldoc.readthedocs.io/en/latest/contributing/docker-for-testing.html#testing-an-unmerged-pull-request)
2. Choose a content type and field to debug (e.g. gene, organism field)
3. Go to Tripal > Page Structure > Gene > Manage Fields
4. Click on edit beside the field you want to debug
5. Select the "Enable debugging" checkbox and save the field settings
6. Create a gene page by going to Tripal > Content > Add Tripal Content > Gene and filling out the form
7. When you click save you would expect to see debugging information.
8. You can then refresh the page and confirm you see debugging information for the load.
9. Then edit the page and confirm you see debugging information for the load multiple times and the update.


pdtouch:
This works as described perfectly well. It is very very helpful in a way that makes development and debugging of Tripal fields much easier. It exposes the underlying web Drupal/Tripal GUI elements, databases, tables, columns and values involved in the field before and after entering content through the GUI which have pretty much been black boxes, making debugging very very difficult.

Tested 3 modified fields simultaneously using the Debug Field(s) option : 
Sequence Residues, Sequence Length and Sequence Checksum for a new Gene content.

```
Debugging has been enabled for field_sequence_checksum field.
Debugging has been enabled for field_sequence_length field.
Debugging has been enabled for field_sequence_residues field.
``` 

The field variables, arrays including the database/storage operations are visible in the Debug area for all the 3 fields helping better in the debugging process. 

Thank you very much @laceysanderson !

dsenalik:
Some debug output is "leaking through" without debugging activated, _e.g._ going to `/admin/content/bio_data` (I have an organism and two genes)


and debugging is not yet turned on


Or adding a new gene


Ah, the `printHeader()` function doesn't check whether it should print or not, can we just add this?
```
  if ($this->has_fields2debug === FALSE) {
     return;
   }
```

dsenalik (approved):
This is beautiful 🥰
Code looks good and debugging output will be super duper helpful for field development.

My only suggestion is to make the variable name|array key `debug` a bit more descriptive, how about `field_debug` to correspond to the `field_debugger` naming? That way it is more clear in the code what it is for. Just a thought

spficklin:
Thnks @dsenalik and @pdtouch for the reviews!

Issue #1414, opened 2023-02-06
Tripal 4 Fields -upgrading core fields


laceysanderson:
# Core Tripal 4 Development Task

## Branch Name: tv4g1-issue1414-[name of field you are working on]

This issue is providing an organization spot to upgrade all the core Tripal fields. Please comment here with the field you are going to upgrade and create a branch using the naming schema above to contain your work. When done, create a PR for your field.

Here is the documentation for [Tripal Fields](https://tripaldoc.readthedocs.io/en/latest/dev_guide/module_dev/fields.html)

## Fields in order of complexity

- [x] [data__sequence_checksum](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/data__sequence_checksum/data__sequence_checksum.inc) Pull #1419 closed #1607 merged
- [x] [data__sequence_length](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/data__sequence_length/data__sequence_length.inc)  Pull #1419 closed #1607 merged
- [x] [operation__analysis](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/operation__analysis/operation__analysis.inc)  Pull #1578 merged
- [x] [local__contact](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/local__contact/local_contact.inc)  Pull #1599 merged
- [x] [sep__protocol](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/sep__protocol/sep__protocol.inc) Pull #1743 merged
- [x] [uo__unit](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/uo__unit/uo__unit.inc) Pull #1680 merged
- [x] [local__source_data](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/local__source_data/local__source_data.inc) Pull #1504 merged
- [x] [data__sequence_coordinates](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/data__sequence_coordinates/data__sequence_coordinates.inc)  Pull #1700 merged
- [x] [operation__analysis](https://github.com/tripal/tripal/blob/7.x-3.x/tripal_chado/includes/TripalFields/operation__analysis/operation__analysis.inc) Pull #1578 merged
[...]

PR #1419, opened 2023-02-14
Tripal 4 Fields: Sequence Length + Sequence Checksum

srobb1:
# Tripal 4 Core Dev Task

Issue #1414 
### Tripal: 4.x

This is now ready for review however, it is dependant on #1388 as it includes the field created in that PR and makes some specific changes related to these two fields.

@srobb1 could you please fill out this template? Before reviewing a PR it needs to be well described and have some information on how to test it. Thanks!

PR #1607, opened 2023-08-15
Tripal 4 Fields: Sequence Length + Sequence Checksum

pdtouch:
Code is cleaned up, removing unnecessary classes and code, added feature to not show checksum if sequence is not provided. Tested functionality

Issue #1414 

tv4g1-issue1414-data__sequence_length

## Instructions for Testing

1. Create an organism
2. Go to Tripal > Page Structure > Gene > Manage Fields
3. Delete the existing "Sequence" (bio_data_8_data_2044) and "Sequence Length" (bio_data_8_data_1249).
4. Add the following fields in this order:
    - Chado Sequence Residues
    - Chado Feature Sequence Length
    - Chado Feature Sequence Checksum
5. Then create a gene:
    - Go to Tripal > Content, click "add Tripal Content" and then "Gene"
    - Select an organism, enter an identifier, and then fill in "Sequence" in specific ways to test the field.

**How to add a field:**
On Tripal > Page Structure > Gene > Manage Fields

1. Click "Add field"
2. Select the field name from the "Add a new field" dropdown and enter a label. Click "Save and continue".
3. On the next form, select "feature" as the Chado Base Table and click "Save Field settings".
4. Then on the final page, leave the defaults and click "Save settings".

**Note: There is no widget for the sequence length and sequence checksum fields. Instead the value is calculated based on the sequence residues and then displayed by the formatter.**


laceysanderson:
## Notes

- This PR is dependant upon #1604. It cannot be merged until that one is first.
- This PR includes the work of @srobb1 done in PR #1419. As such we are closing #1419 in favour of this one.
- Testing instructions for this PR are in #1419 
- We will want to ensure the new fields are automatically attached via the YAML.

Thanks @pdtouch <3

tripal_chado/src/Plugin/Field/FieldType/ChadoSequenceChecksumDefault.php
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Plugin implementation of Default Tripal field for sequence data.
+ *
+ * @FieldType(
+ *   id = "chado_sequence_checksum_default",
+ *   label = @Translation("Chado Feature Sequence Checksum"),
+ *   description = @Translation("A chado feature sequence md5 checksum"),
+ *   default_widget = "chado_sequence_checksum_widget_default",
+ *   default_formatter = "chado_sequence_checksum_formatter_default"
+ * )
```
dsenalik, line 18: Add `*   cardinality = 1` here

dsenalik (changes requested):
I followed the testing steps, and the field worked perfectly.
- [x] Entering a simple sequence


displays 

and I verified the correctness of this checksum
`echo -n cat | md5sum`
`d077f244def8a70e5ea758bd8352fcd8  -`

- [x] Changing the sequence updates the length and checksum correctly.

- [x] A long sequence is correctly wrapped to the same length as for Tripal 3.

- [x] If I add extra white space, it is removed when I save. I just checked, and this is what Tripal 3 does, so I guess this is correct.

- [x] I configured a default value, and it showed up correctly when creating a new gene

- [x] In my container I am seeing the field debugger output, but I never turned that on. I'm not sure yet if that's a problem with this PR or not. Checking it here, and reference on PR#1604 in this comment https://github.com/tripal/tripal/pull/1604#issuecomment-1682760409


Changes requested:
- [x] The cardinality for this field should be set permanently to 1. This field will only apply to the feature table, and you can only have one sequence. You can add
`*   cardinality = 1` I have a separate comment showing where

- [x] In ChadoSequenceWidgetDefault.php you use a variable `$rec_id_val` but it is never set in this function, something might be wrong here

- [x]  Figure out why testing is failing

laceysanderson:
The testing is failing on this branch due to the inclusion of the code from PR #1604. It has been fixed in that PR so please update this branch as well.

laceysanderson:
Tests are now passing!
Should just be the two remaining comments in @dsenalik review:

> The cardinality for this field should be set permanently to 1. This field will only apply to the feature table, and you can only have one sequence. You can add `cardinality = 1` I have a separate comment showing where.

> In ChadoSequenceWidgetDefault.php you use a variable $rec_id_val but it is never set in this function, something might be wrong here

dsenalik (approved):
Approved with suggested changes added.

PR #1590, opened 2023-07-25
Entity Publishing

spficklin:
# Tripal 4 Core Dev Task

Issue #1515 

## Description

This PR introduces basic publishing to Tripal 4. Currently, a publish job submitted in this PR will:

- Create new Tripal Entities of the requested content type that do not already exist.
- Integrate data added to Chado outside of content forms (i.e. via importers or pre-existing) for a SUBSET OF FIELDS (see below)
- It works both for (1) content page created through UI then new data added to chado later and (2) data added and no entity exists yet.
- It can and should be re-run when new data is added to existing content pages outside of the UI.

**NOTE: Publish in Tripal 4 is meant to create entities (same as Tripal 3) but ALSO to sync data in chado with fields (new in Tripal 4). Data added to Chado outside the content add/edit form will not show up in fields (i.e. on the content page) without it being published.**

## Caveats

It has currently only been tested for feature-based content types although it will be tested with other content types before it is elevated above DRAFT status.

It will only work for fields where all store properties are on the base table. It DOES NOT yet work for linker tables, property fields, etc. This will be added in a future PR.

It only works for the following content types: 
- Gene: 50
- mRNA: 75
- Organism: 6
- Contact: 7
- Analysis: 7 (two are of type Genome Assembly and 2 are of type genome annotation)
- Project: 2 (both of type Genome Project)
- Protocol: 3
**NOTE: it should also work for the germplasm types but I haven't tested them yet.**

### Automated Testing

Publish Form Test
- Complete. Tests the form using a mock storage backend, ensures that the backend can add form elements to the form, checks form validation and submit.

FindValues() Test
- Complete and working for the base table fields. Tests publishing of 50 genes.

### Manual Testing

1. Prepare a site with test data [according to this GIST](https://gist.github.com/laceysanderson/7c57bd173d354e432eb83f12360936b6).
3. Go to Admin > Tripal > Content and click on "Publish Tripal Content"
4. Choose "Chado Storage" as the data backend and then select a content type. Click publish to submit the Tripal job.
Screenshot 2023-12-05 at 5 09 15 PM

5. Run the Tripal job on the command line and ensure there are no errors. Also check that it found the number of records you imported into chado.
6. Go to Admin > Tripal > Content and check that the new entries have been added to the content listing.
7. View one of the content pages and ensure it has the values you added to chado.
8. Edit the content page and ensure there are no errors and it updates chado.
9. Make a change to chado directly using SQL.
[...]

Issue #1520, opened 2023-05-20
Tripal 4 - Fix PHP 8.2 dynamic property deprecations


dsenalik:
# Core Tripal 4 Development Task

## Branch Name: tv4g0-issue1520-php8.2-deprecations

Pull #1519 adds PHP 8.2 to automated tests. This reveals some new deprecation notices, but most are related to [PHP 8.2: Dynamic Properties are deprecated](https://php.watch/versions/8.2/dynamic-properties-deprecated)

A random example from `tripal/src/Plugin/TripalIdSpace/TripalDefaultIdSpace.php` where the dynamically created property is `$this->messageLogger`
```
  /**
   * {@inheritdoc}
   */
  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
    parent::__construct($configuration, $plugin_id, $plugin_definition);

    // Instantiate the TripalLogger
    $this->messageLogger = \Drupal::service('tripal.logger');

  }
```

laceysanderson:
Fixed by PR. Confirmed with @dsenalik that he is making separate issues for the other issues.

dsenalik:
Reopening because I found another one, in `tripal/tripal.install` line 905 ` $messenger->addMessage('Removed PostgreSQL schema clonging functions');`

I did a recursive grep for other similar instances and did not find any.

laceysanderson:
Good catch -merged and re-closed 🙃

dsenalik:
There are more! This one is observed when creating a new organism `/bio_data/add/bio_data_1`

`Deprecated function: Creation of dynamic property Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::$connection is deprecated in Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace->__construct() (line 83 of modules/contrib/tripal/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php). `

Fix this one with Pull #1543

dsenalik:
Reopening because of another deprecation `Creation of dynamic property Drupal\tripal_chado\ChadoCustomTables\ChadoMview::$custom_table is deprecated`, see https://github.com/tripal/tripal/pull/1610#issuecomment-1688842412

Deal with this in pull #1617

dsenalik:
Running all automated tests locally on PHP8.2 to find any other instances:

[...]

PR #1610, opened 2023-08-20
TripalTokenParser Class

### LINKED ISSUE #1609: Tripal Field Token Replacement Service

## Branch Name: tv4g1-issue1609-field_token_service

In the process of working on publishing of content (#1515), the publish service needs to set titles of entitites. It does so using the title setting that has tokens representing fields.  The tokens get replaced with the corresponding field values.  Initially, I was was writing my own code to replace tokens for titles but I noticed I was getting different titles from when the same entity is created manually.  In the PMC meeting on Friday we chatted on this and decided a service should be added that would provide a single source for replacing field value tokens in a string.

spficklin:
# Tripal 4 Core Dev Task

Issue #1609

## Description

This PR adds a new `TripalTokenParser` Service.  This is needed because tokens in strings (e.g. `[organism_genus] [organism_species] [organism_infraspecific_type] [organism_infraspecific_name]`) need to get replaced to set entity titles and URLs.  This happens in two places: 1) during manual creation of an entity via the Drupal GUI and 2) during automatic bulk publishing of entities (#1590 ).  In order to prevent two different implementations of token replacement using field values, this class provides a single service for doing so.   This PR includes a full set of functional tests for every function of this new class.  

In the process of creating this class I also came across a bug that needed fixing, a code correction that I thought needed to be made, and I added a new tripal default storage class -- all related to this PR.

- The bug was related to the `chado_additional_type_default` field.  It was not setting in the widget values for the `term_name`, `accession`, and other properties that the field requires.  This didn't cause a problem for saving, but it meant that values for those properties were not available for token parsing.  This resulted in organism titles missing the infraspecific type name.  
- I created a new `TripalDefaultStorage` class because during testing, I needed to save an entity that has fields without using Chado.  Initially I was going to try to create a Mock instance of the plugin, but I couldn't figure out how to make that happen as Drupal automatically picks up on the plugins. The `TripalDefaultStorage` plugin is basically an empty shell because it ends up using the Drupal default SQL storage system for saving values.  If someone knows how to create a mock plugin I can remove it.  However, there may be a benefit to keeping it, as it would allow folks to store values in Drupal (this happens sometimes) and allow for searching using the `TripalStorage::findValues()` just like any other TripalStorage method (just needs to be implemented).
- In the process of creating the `TripalDefaultStorage` class, I realized that some of the functions dealing with adding property types to the storage class should be in the `TripalStorageBase` implementation and not in the `ChadoStorage` impelmentation so I moved them.  
- The term and accession for the `organism_infraspecific_type` field was incorrect. I fixed it.

## Testing?

- To test this, make a brand new entity (e.g. organism page). 
  - The title should get automatically updated to match tthe deafult title string for the content type.   
  - The URL should get set to an alias that includes the content type name in it (e.g. `/organism/1`) 
- If you want to play around with changing the tokens, go to admin > structure > Tripal Content Types and select the content type for the entity you made previously.
  - Change the default title to use other fields, non-existing fields, etc, 
  - Return to the entity, edit and save it.  The title or URL should change to match the changes you made in the content type settings. 
  - Note:
    - Keep in mind that tokens that don't have true field won't get replaced.  
    - Tokens for fields that do exist but that don't have a value are simply replaced with an empty string.


dsenalik:
I built a docker on this branch
`docker build --tag=tripaldocker:testing-1609 --build-arg drupalversion="10.0.x-dev" --file tripaldocker/Dockerfile-php8.2-pgsql13 ./`
and I was just using this to test something else, but there are a number of similar deprecation notices after running
`docker exec --workdir=/var/www/drupal9/web/modules/contrib/tripal 1609 phpunit --group 'Tripal Chado Preparer' --debug`
such as
`Test 'Drupal\Tests\tripal_chado\Functional\Task\ChadoPreparerTest::testChadoPreparer' started
PHP Deprecated:  a:5:{s:11:"deprecation";s:106:"Creation of dynamic property Drupal\tripal_chado\ChadoCustomTables\ChadoMview::$custom_table is deprecated";s:5:"class";s:59:"Drupal\Tests\tripal_chado\Functional\Task\ChadoPreparerTest";s:6:"method";s:17:"testChadoPreparer";s:15:"triggering_file";s:93:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/ChadoCustomTables/ChadoMview.php";s:11:"files_stack";a:12:{i:0;s:93:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/ChadoCustomTables/ChadoMview.php";i:1;s:92:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Services/ChadoMviewsManager.php";i:2;s:83:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Task/ChadoPreparer.php";i:3;s:83:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Task/ChadoPreparer.php";i:4;s:83:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/src/Task/ChadoPreparer.php";i:5;s:104:"/var/www/drupal9/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Task/ChadoPreparerTest.php";i:6;s:50:"/var/www/drupal9/web/sites/simpletest/TestCase.php";i:7;s:50:"/var/www/drupal9/web/sites/simpletest/TestCase.php";i:8;s:68:"/var/www/drupal9/vendor/phpunit/phpunit/src/Framework/TestResult.php";i:9;s:50:"/var/www/drupal9/web/sites/simpletest/TestCase.php";i:10;s:19:"Standard input code";i:11;s:19:"Standard input code";}} in /var/www/drupal9/vendor/symfony/phpunit-bridge/Legacy/SymfonyTestsListenerTrait.php on line 284`

spficklin:
@dsenalik sounds like we need to open an issue on that?

dsenalik:
Reopening #1520 yet again to deal with this. This is not a problem with this PR so won't be fixed here. We don't even do automated testing on PHP 8.2 yet!

dsenalik (changes requested):
I did the suggested test procedure on Drupal 10.0 PHP 8.2
All steps were successful:
- I added `([organism_common_name])` and it was replaced correctly "(Common false tripal) Tripalus bogusii subspecies fakus"
- I added a bogus field `[fake]` and it was passed through unmodified as expected "[fake] Tripalus bogusii subspecies fakus"
- I added a field with no value `([organism_comment])` and it returned an empty string "() Tripalus bogusii subspecies fakus"
- good catch on the `organism_infraspecific_type` CV term fix.
Sorry about bringing up that PHP8.2 issue here, I forgot we don't do tests on that yet, but I often make "bleeding edge" dockers to find that kind of stuff.

However, I thought I would try on another content type, I used Phylogenetic Tree. I entered an analysis first, then the tree **with no comment**, then changed the token string to `[phylotree_name] [fake] [phylotree_analysis] ([phylotree_comment])`

This may be a problem elsewhere, if so then I can change this to Approve.
```
 Deprecated function: preg_replace(): Passing null to parameter #2 ($replacement) of type array|string is deprecated in Drupal\tripal\Services\TripalTokenParser->replaceTokens() (line 179 of modules/contrib/tripal/tripal/src/Services/TripalTokenParser.php).

Drupal\tripal\Services\TripalTokenParser->replaceTokens(Array) (Line: 280)
Drupal\tripal\Entity\TripalEntity->replaceTokens('[phylotree_name] [fake] [phylotree_analysis] ([phylotree_comment])', Object) (Line: 133)
Drupal\tripal\Entity\TripalEntity->setTitle('[phylotree_name] [fake] [phylotree_analysis] ([phylotree_comment])') (Line: 40)
Drupal\tripal\Form\TripalEntityForm->save(Array, Object)
call_user_func_array(Array, Array) (Line: 114)
Drupal\Core\Form\FormSubmitter->executeSubmitHandlers(Array, Object) (Line: 52)
Drupal\Core\Form\FormSubmitter->doSubmitForm(Array, Object) (Line: 597)
Drupal\Core\Form\FormBuilder->processForm('tripal_entity_phylotree_edit_form', Array, Object) (Line: 325)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 580)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 163)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 74)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 106)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 51)
Drupal\Core\StackMiddleware\StackedHttpKernel->handle(Object, 1, 1) (Line: 692)
```

dsenalik:
Simplest case to reproduce on phylotree content type
One token: `[phylotree_analysis]`
It looks like the analysis name is not getting looked up, it is coming back as NULL
after line 178 of `tripal/src/Services/TripalTokenParser.php`:
tree

So, does the analysis widget need fixing like you did for the additional type field?

spficklin:
Yeah, it looks to me like the same problem.  The field widget needs to set the name property.  Maybe we need to have a built-in check somewhere to alert field developers of this because this is going to continue to be a problem....  I'll add an issue

dsenalik (approved):
Well the analysis field I wrote so I can fix it for issue #1616, so I'll mark this as approved

dsenalik:
On second thought, we should at least fix `tripal/src/Services/TripalTokenParser.php` so it at least doesn't generate this error if it hits a NULL. I'm going to push a simple fix since I have the environment to test it.

With https://github.com/tripal/tripal/pull/1610/commits/6006b2aacb5a5cf5c4b1b049d2c35a0fba16115a you see an unsubstituted `[phylotree_analysis]` instead of all the error messages. Not correct, but it no longer blows up your site.

laceysanderson (changes requested):
Comments on 
> I created a new TripalDefaultStorage class because during testing, I needed to save an entity that has fields without using Chado. Initially I was going to try to create a Mock instance of the plugin, but I couldn't figure out how to make that happen as Drupal automatically picks up on the plugins. The TripalDefaultStorage plugin is basically an empty shell because it ends up using the Drupal default SQL storage system for saving values. If someone knows how to create a mock plugin I can remove it. 

I'm pretty sure I know how to do the mock such that Drupal will automatically detect it. I'll take a stab at it today.

> However, there may be a benefit to keeping it, as it would allow folks to store values in Drupal (this happens sometimes) and allow for searching using the TripalStorage::findValues() just like any other TripalStorage method (just needs to be implemented).

The default TripalField implementation with PR #1612 fix allows Fields to be created which use the Drupal Storage directly if a TripalStorage plugin is not provided. This feels easier to maintain and I think satisfies the above requirement as it would be searched/indexed the same as any Drupal field.

spficklin:
@laceysanderson sounds good. Thanks for jumping in.  Sounds like we need to hold off on merging this, then until #1612 is merged?

laceysanderson (commented):
> @laceysanderson sounds good. Thanks for jumping in. Sounds like we need to hold off on merging this, then until https://github.com/tripal/tripal/pull/1612 is merged?

Perhaps yes, I'm actually taking a moment to think about that PR based on Dougs recent review and some thoughts that have come up while looking into testing on this PR.

On another note:
@spficklin why is there such a change in ChadoStorage? It looks like some of your changes from publishing might have leaked over into this PR?

spficklin:
HI @laceysanderson those changes to ChadoStorage are in reference to this change described above:

> In the process of creating the TripalDefaultStorage class, I realized that some of the functions dealing with adding property types to the storage class should be in the TripalStorageBase implementation and not in the ChadoStorage implementation so I moved them.

So basically, those functions that got removed from ChadoStorage I put into the TripalStorageBase class because they are not unique to ChadoStorage, they should be available to any TripalStorage implementation.  Since I created the TripalDefaultStorage class I realized it needed those functions too if we were to keep it and later expand it.   But that changes is not necessary for this new TripalTokenParser class to function properly.

laceysanderson:
## Removing the hold on this PR

With #1625 being merged, we now have a DrupalSqlStorage implementation as discussed in some of the comments above. I then updated this PR to remove the Tripal Default Storage and updated the tests to use DrupalSqlStorage.

As such, this PR is now completely focused on the token functionality with no changes to the storage :-) 

### Now ready to be re-reviewed!

laceysanderson (approved):
Manual Testing:
✅ Create organism using default title template + not providing infraspecies (title showed as `Tripalus databasica` and left out the unfraspecies as expected)
✅ Create organism using default title template + provide infraspecies (title showed as `Tripalus databasica subspecies postgresquelus` as expected)
✅ Create new chado-focused field and update title template to use it (performed as expected; replaced empty when no value + replaced correctly when value set)
❌ Create multi-value chado field and fill in multiple values. If one value supplied it replaces as expected but it multiple are, it leaves the token in the title (hoped for either comma-separated, or even choose delta 0).
❌ Create single value Drupal field (plain text).... token never replaced...

So there are two cases that I had hoped it would support that it doesn't but I'm still marking this as approved since I don't think adding this support will change the structure of the service. I suggest we add a new issue mentioning that these cases are not yet supported and come back to it later.

Especially as I did not find any problems with replacing single-value chado fields on any content type and this will be the main case for this service.

spficklin:
Thanks for the testing and fixes!  I created the issue that @laceysanderson requested here:  #1639

PR #1622, opened 2023-08-29
Tripal 3 delete stray "delete"

### LINKED ISSUE #1621: Tripal 3: text left over from debugging
### Tripal Version

3.10

### Drupal Version

na

### PostgreSQL Version

na

### PHP Version

na

### General Description

In `tripal_chado/includes/tripal_chado.fields.inc` there is a bit of text `delete` plus a tab character that should not be there on line 1772
```
        'term_name' => 'delete  Sequence length',
```
This was added June 2018 so it has been there a while 😂

### Steps to reproduce

This is mainly a code review bug

### Error messages

```shell
no error messages
```

### Screenshots

no screenshots

dsenalik:
Super simple fix to remove stray `delete` plus tab character in the sequence length field
Issue #1621


laceysanderson (approved):
Good simple fix 👍

Issue #1614, opened 2023-08-22
Tripal Github Issue Template Forms - Feature Request


Ferrisx4:
## Tripal 4 Feature Request

* This feature does not attempt to solve an existing problem with Tripal
* This will not need to be an extension module.

* This feature is not urgent

### Description

Github now offers nice forms for issue templates which can be constructed with yml files. These forms make it much easier to create new issues without having to muck through the markdown to fill out the categories (for example, the ones below).

I propose migrating these github templates to the new form format.

### Additional information/screenshots

Branch: tv4g0-issue1614-issue_template_forms

### Progress
 - [x] Bug Report
 - [x] Core Development Task
 - [x] Discussion/Question
 - [x] Tripal 4 Feature Request (this one!)
 - [x] Pull request template (.md, because .yml not possible :frowning_face: )

Ferrisx4:
Forms can be previewed on Github without merging with main.
To do so:
 1. Navigate to the latest commit that changes the particular form of interest.
 2. Click the three dots to "View File".
A read-only preview of the form will be displayed, and the code can be viewed here as well.

Ferrisx4:
I realized that I used the 4.x to branch off of, whereas the current issue templates are in 7.x-3.x.

@laceysanderson and @spficklin your advice? If we're happy with the new forms, I can switch to the 7.x-3.x as the base for this change and simply replace the existing .md templates with the .yml ones.

I set the title in each yaml file to have `(Form)` in it to distinguish it from the original, if we're simply going to replace, I can remove that before doing so.

spficklin:
I was wondering if perhaps we shoul switch the current master branch of Tripal to 4.x since we are no longer providing updates to v3.x unless there are major security or bugs.

laceysanderson:
Yes, I agree highly to switching the major branch to 4.x! It would alleviate a lot of pain points development-wise and I think it would send a better message to people using the repo that 3.x is being phased out in favour of 4.x and that progress is being made :-)

spficklin:
Who shall do the honors :-)

dsenalik:
I like the new forms!
Just one comment, I think the error messages and the screenshots should not be required, not every issue will have them.

Ferrisx4:
Thanks @dsenalik, I agree with this. I may have gotten carried away with that tag while copy-pasting the form elements...

dsenalik:
Fixed with merged PR #1623

dsenalik:
[...]

PR #1623, opened 2023-09-01
Make some fields not required

Ferrisx4:
Fixes some fields that shouldn't be required, suggested by @dsenalik: https://github.com/tripal/tripal/issues/1614#issuecomment-1697855771


dsenalik (approved):
Thank you!

laceysanderson (approved):
I agree with these changes :-)

PR #1624, opened 2023-09-02
Tripal 3 issue1573 php8.1 deprecation

dsenalik:
## Fix another PHP deprecation notice in Tripal 3

This is a simple patch and typo fix for this comment for issue #1573 
https://github.com/tripal/tripal/issues/1573#issuecomment-1703839125

Error message:
`Deprecated function: str_replace(): Passing null to parameter #2 ($replace) of type array|string is deprecated in chado_replace_tokens() (line 432 of .../sites/all/modules/tripal/tripal_chado/api/tripal_chado.api.inc`

(Triggered on https://www.carrotomics.org/accession/291092 )
This came from the `obi__organism` field for a germplasm accession without an `infraspecific_name` in the `organism` table, _i.e._ it is null.


dsenalik:
Another one:
`Deprecated function: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in tripal_pub_importer_setup_page() (line 238 of .../sites/all/modules/tripal/tripal_chado/includes/loaders/tripal_chado.pub_importers.inc).`

This happens when no publications match a search while testing an importer. Commit https://github.com/tripal/tripal/pull/1624/commits/381dfb63d7305a9d52e1363a22f5149c700fca84

dsenalik:
I am going to assume this is another PHP deprecation, since it has been int the code forever
`Error: Undefined constant "STDOUT" in tripal_bulk_loader_progress_bar() (line 1110 of .../sites/all/modules/tripal/tripal_bulk_loader/includes/tripal_bulk_loader.loader.inc).`

This constant is only available to the cli version of PHP, but everywhere else in this function `echo()` is used, so just use that instead. Commit https://github.com/tripal/tripal/pull/1624/commits/17b991e463ae284a318839d507a31d6412096192

laceysanderson (approved):
Another beautiful, simple fix -thanks @dsenalik 

Regarding the change to the bulk loader, that should only be called from the daemon/drush/job-launcher and thus is executed on the command-line. That said, I love the consistency of switching to echo 👍

PR #1630, opened 2023-09-09
Importer file validation improvements

### LINKED ISSUE #1629: Tripal Importer class doesn't validate if a remote filepath exists
### Tripal Version

4.x

### Drupal Version

10.0.11-dev

### PostgreSQL Version

The tripaldocker version

### PHP Version

8.1.22

### Branch

tv4-issue1629-importer-file-validation
Pull #1630 

### General Description

I am trying to create a custom Tripal Importer and during testing, I entered something random for the remote path:

Screen Shot 2023-09-08 at 3 29 06 PM
And the job was submitted successfully. 

Here is the relevant section of the annotation for my new importer class:
```
file_upload = True,
file_load = True,
file_remote = True,
file_required = True,
```

Let me know if any more information or context would be helpful!

### Steps to reproduce

I checked if this issue can be reproduced using one of the existing Tripal Importers. Steps I did to repeat this for the FASTA Importer:

1.   I modified "file_remote" and "file_required" to be True in `tripal/tripal_chado/src/Plugin/TripalImporter/FASTAImporter.php`
2. Cleared the cache using `drush cr` to have the Remote path option show up in the form
3. Created an organism and an analysis, if none existed already, and selected these in the dropdowns in the form
4. Entered the same nonsensical `lalalalalalalalala` as before as the Remote path and submitted the form

Interestingly, I get an error this time, but my job also gets submitted:

Screen Shot 2023-09-08 at 4 09 21 PM

### Error messages

_No response_

### Screenshots

_No response_

dsenalik:
# Bug Fix

## Description
For all tripal importers, adds validation for remote file (URL) field.
Adds local and remote file fields to FASTA and GFF3 importers to match Tripal 3

## Testing?
Go to either or both of the FASTA and GFF3 importers and try to enter an invalid URL

Example of something that is not valid, no http prefix
invalid1

Example of input that is formatted correctly, but is not a valid web site
invalid2


tripal/src/Form/TripalImporterForm.php
```diff
@@ -256,16 +256,43 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+        // Validate a correct format remote URI to make sure it can be accessed.
+        else {
+          $headers = @get_headers($file_remote);
+          if (($headers === false) or (!is_array($headers)) or (!strpos($headers[0], '200'))) {
+            $form_state->setErrorByName('file_remote', t('The Remote Path provided cannot be accessed.'
+              . ' Check that it is correct.'));
+          }
+          // Some internet providers will insert content with a valid '200' response code for invalid
+          // URIs that redirects to their own search page, such as AT&T "DNS Error Assist".
+          // Try to detect this by rejecting 'http-equiv="refresh"' in the first 1000 bytes
```
laceysanderson: I cannot think of a good, repeatable way to test this 🤔 and would be tempted to remove it... Alternatively, we could test the the downloaded header indicates the right file format for the importer but this feels a bit finicky as well.

laceysanderson (commented):
✅ Code Review: All looks good except the one bit I pointed out which you were also unsure of.

I also added automated testing for the validate/submit of the importer form :-) 
✅ The automated testing showed that the functionality worked as I expected with only a small update required for file_local validation.

tripal/src/Form/TripalImporterForm.php
dsenalik: I also feel very uncertain about this bit of code, as you note, it is nearly impossible to test, and may not handle any situation other than my specific internet provider. 
And yes, each importer could check the header when the job is run, that is probably much safer and reliable.

I shall remove it 💣

carolyncaron (approved):
Thanks Doug for these changes, my functional testing for both the FASTA and GFF3 importers shows that there are checks in place for the remote filepath that it is a valid URL. 👍 

I did find something else peculiar during testing, but after talking with Lacey about it, she feels it would be ideal to make a separate issue for it...

Now that the GFF3 and FASTA importers allow for multiple file input methods, I tried submitting the form with multiple filled in at the same time. For example, I can upload a file AND provide a valid remote path (or a working URL at the very least) but an improper server path, and I get an error that the file cannot be found on the system. I think this is expected behavior; we still want to inform the user that an input method is incorrect. But, if I provide only _valid_ multiple inputs, the job gets submitted successfully with no indication to the user what has occurred - did one method take priority over the others? Were they ALL submitted at once? (Lacey says the latter!) This does become even more complicated when we consider that cardinality can be set for each importer, so we will still want to allow for multiple files in the base importer class.

dsenalik:
My initial thought is that we can also catch this in the validation, make sure not more than one is specified. I'll see what I can do...

dsenalik:
This ended up being a small reordering of the validation, but if you specify more than one method, you will now see a proper validation something like this:
more_than_one

Note that the FASTA and OBO importers currently do their own missing file validation (GFF doesn't), so there may be two messages. This is because in Tripal 3 the importer did this validation. I don't think having two messages is that bad...

no-fasta

no-obo

dsenalik:
I just reread your comment about cardinality. I'm not sure that is supported now, files from the form are strings not arrays. Anyone have any more insight?

laceysanderson:
Cardinality is set in the annotation for an importer, then the [importer base form configures the html5 upload element to support this cardinality here](https://github.com/tripal/tripal/blob/4.x/tripal/src/Form/TripalImporterForm.php#L84):

```php
      $form['file']['file_upload'] = [
        '#type' => 'html5_file',
        '#title' => '',
        '#description' => 'Remember to click the "Upload" button below to send ' .
            'your file to the server.  This interface is capable of uploading very ' .
            'large files.  If you are disconnected you can return, reload the file and it ' .
            'will resume where it left off.  Once the file is uploaded the "Upload ' .
            'Progress" will indicate "Complete".  If the file is already present on the server ' .
            'then the status will quickly update to "Complete".',
        '#usage_type' => 'tripal_importer',
        '#usage_id' => 0,
        '#allowed_types' => $importer_def['file_types'],
        '#cardinality' => $importer_def['cardinality'],
      ];
```

The html5 element passes multiple files through as a string of `fid` concatenated together + separated by `|`. You can [see that here](https://github.com/tripal/tripal/blob/4.x/tripal/src/TripalImporter/TripalImporterBase.php#L202-L212). 

In the [create of the importer object](https://github.com/tripal/tripal/blob/4.x/tripal/src/TripalImporter/TripalImporterBase.php#L185-L230), all files submitted by all types are compiled into a single `files` array with the `file_path` indicating the local path for each file. Multiple are supported here and this is where the string version from the upload would be separated into it's component managed files.

The base importer class also uses prepareFiles to download and remote files so that the path to a local version is available above in the `files` array.

Then the run command simply has to handle the fully processed files array and can depend on a local path always being available.

So cardinality is already supported but I don't think it has ever been validated in any Tripal version 🤔 😬

dsenalik:
I tested changing the cardinality on the importers. What happens is that it breaks, and you can no longer upload a file. And the "Server path" and "Remote path" fields do not support more than one either.
If we want an importer to support more than one file, I think this should be a separate issue (#1635), but since no importers currently do that (?) then let's call this good for now.
Fasta_cardinality_unlimited

laceysanderson:
🙈 Good to know. Ok, I'm good with creating an issue saying that importers don't currently support multiple files and then moving forward with this PR including your validation restricting to a single file. 

Although, it might be good to add a bit in this PR that shows a warning at the top of the importer if cardinality > 1 is specified and hard-codes `1` for the file upload.

dsenalik:
How about this?
cardinality_error_message

laceysanderson:
Much Better 👍 Once tests pass I think this is ready to merge :-) 🎉

PR #1636, opened 2023-09-14
Striving towards Drupal 10.1 Compatibility

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR fixes some of the known compatibility issues with Tripal and Drupal 10.1 (both on PHP 8.1 + 8.2).

### Fixed

The following tests were failing and are now passing based on the changes in this PR.

- ChadoLinkerPropertyDefaultTest::testInsertValuesForSingleField
- ChadoContactDefaultTest::testArrayDesignBaseTableFieldCRUD
- ChadoContactDefaultTest::testStudyBaseTableFieldCRUD
- ChadoAnalysisDefaultTest::testQuantificationBaseTableFieldCRUD
- ChadoAnalysisDefaultTest::testPhylotreeBaseTableFieldCRUD
- ChadoPreparerTest::testChadoPreparer
- ChadoVocabTermsTest::testTripalVocabularyClasses
- ChadoStorageTest::testChadoStorage
- testFindVersion
- testPrepareTestChadoSimulation

Specifically, the following changes were addressed:

1. Drupal 10.1 implementation of Connection::tableExists()made assumptions that did not work for our multi-schema setup. Instead I overrode it in Tripal DBX to ensure that case is supported. This was actually involved in being unable to detect the chado version as that code will check the existence of specific tables if there is no record in the chado installations table which is the case in our testing environment.
2. getVersion was not always returning a string which caused some failures.
3. ChadoSchema was making some assumptions about version when it was not specified that I removed.

### Outstanding

The following tests are not fixed by this PR and will be addressed in a later PR.

- testTripalDbxSchemaPrefixInfo
- testTripalDbxSchemaScenario1
- testPerformTaskInstaller

## Testing?

There is quite good coverage of these changes in the automated testing.
That said, manual testing for such an integral part of the system is always important.

For this PR testing is more general as these changes affect querying of chado.
I suggest testing:
1. Create, edit, delete Tripal content
2. Run a Tripal Importer (GFF3 is a good choice as it has a lot of queries)

Note: There is already manual-like testing of installing Chado and preparing it as these are done when the docker is built.


laceysanderson:
I believe the most recent change will additionally fix the table prefixing in Drupal 10.1. The fix is to implement getPrefix as based on [this changelog](https://www.drupal.org/node/3260849) Drupal 10.1 has removed support for tablePrefix() in favour of getPrefix() since per table prefixing has been removed.

spficklin (approved):

spficklin:
Works great!  I tested creating and deleting content and loading a GFF3 file containing 880K features. All worked great. I love to see the "All checks passed".   Since the tests also show this is working as intended. I'm going to merge.

PR #1638, opened 2023-09-15
Fixed testing issue

spficklin:
This is a very simple PR to fix a bug in testing that I let through by merging the newick importer.

It's just a renaming of a variable. If you look at the files changed there are some white spaces  removed... my IDE does that when saving.


laceysanderson (approved):
Looks good and tests now pass 👍

PR #1640, opened 2023-09-18
TripalImporter support for use_button is false

### LINKED ISSUE #1580: Tripal Importer Plugin Submit Button
# BUG/ERROR report

I have an implementation of the Tripal Importer plugin and noticed that even with a **false** value set to **use_button** definition the form will render with a submit button (Please see screenshot below).

Screen Shot 2023-07-18 at 11 50 39 AM (2)

Screen Shot 2023-07-18 at 11 50 39 AM

### Tripal Version:
* Drupal Version: 9.5.10-dev 
* PostgreSQL Version: 13.11
* PHP Version: 8.1.20

#### Steps to reproduce
1. Create an instance of Tripal Importer Plugin
2. Set the annotation definition **use_button** to False

 
#### Error messages and screenshots

```
no error message
```

laceysanderson:
# Bug Fix

## Description

Simple PR ensuring that the TripalImporter only adds it's own submit button if the specific importer has not indicated `use_button` = FALSE. This allows importers to take control of the submit button if needed (e.g. multi-page forms).

## Testing?
I also added an automated test to confirm this works!

That said, if you want to confirm yourself manually, just edit and existing importer and add the following into the annotation (or replace this line if it already exists).

Then refresh the importer page, scroll to the bottom and confirm that there is no submit button 🙃 Check an unaltered importer who doesn't define use_button in the annotation and ensure they do have a button (default set in annotation class).

```
 *    use_button = False,
```

For example,
Screenshot 2023-09-18 at 5 12 00 PM

### DIFF
--- a/tripal/src/Form/TripalImporterForm.php
+++ b/tripal/src/Form/TripalImporterForm.php
@@ -129,7 +129,7 @@ class TripalImporterForm implements FormInterface {
 
     // We should only add a submit button if this importer uses a button.
     // Examples of importers who don't use this button are mutl-page forms.
-    if (array_key_exists('use_button', $importer_def) AND $importer_def['use_button'] != FALSE) {
+    if (array_key_exists('use_button', $importer_def) AND $importer_def['use_button'] !== FALSE) {
       $form['button'] = [
         '#type' => 'submit',
         '#value' => $importer_def['button_text'],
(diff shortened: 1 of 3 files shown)


dsenalik (approved):
Code review looks good.
I "turned off" the button for the Fasta importer by adding
` *    use_button = False,`
and it indeed was no longer on the form.
I changed it to 
` *    use_button = True,`
and the button returned, to test explicitly to specify to use a button rather than rely on the default.

PR #1650, opened 2023-09-23
Synonym Field

spficklin:
# Synonym Field

## Description

This PR is a replacement of PR #1447 that was submitted by @srobb1 .  My plan was to make changes to Sofia's PR but the PR is tied to Sofia's forked version which I can't push to and it seems we've made so many changes to the 4.x branch I was having a hard time dealing with merge conflicts.  So, instead, I created a new branch, and copied in the field classes with my edits. I set @srobb1 as the author of the commit so she will get credit for her contributions.  

## Testing?

This field is attached to the following content types: gene, QTL, genetic marker, sequence variant, heritable phenotypic marker, and DNA library. This field would also be attached to a cell line content type if Tripal created one.

Make sure to create an organism first as all these content types require one.

Then create and edit pages of these content types with various combinations.

### DIFF
--- a/tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml
@@ -987,7 +987,34 @@ fields:
             default:
               region: content
               weight: 90
-
+              
+    -   name: gene_synonym
+        content_type: gene
+        label: Synonyms
+        type: chado_synonym_default
+        description: Alternate names, aliases or synonyms for this record.
+        cardinality: 0
+        required: false
+        storage_settings:
+            storage_plugin_id: chado_storage
+            storage_plugin_settings:
+                base_table: feature
+                syn_linker_table: feature_synonym
+                syn_linker_fkey_column: feature_id
+        settings:
+            termIdSpace: schema
+            termAccession: alternateName
+        display:
+          view:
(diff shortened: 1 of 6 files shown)


laceysanderson:
Made a small change to switch the storage plugin settings for linker table / fkey column to be more generic for use with other linking tables and added an issue that we should design a context sequence so that developers don't have to do this set every time they need something passed into their field from the field creation step. This should fix the automated testing.

laceysanderson:
This PR is getting closer to being ready for review! There is currently one small, funny bug in that all automatically added fields are set with their allowed number of values to 0 not unlimited:
Screenshot 2023-09-27 at 4 50 29 PM

This is a bug in the field collection service that needs to be fixed to allow addition of fields with unlimited cardinality.

laceysanderson:
This is now ready for review 🎉 I updated the original issue description with some instructions for testing :-)

Ferrisx4 (approved):
Tested this new field in the following ways:
 - Created new gene with no synonym
 - Created new gene with 1 synonym
 - Changed field setting for gene to only allow 2
   - Field reflected this setting by only allowing 2 to be added and no errors.
 - Created new mRNA with 1 synonym
 - Edited same mRNA to add 2 more synonyms, no issues.

Everything seems to be working as advertised and looks good

PR #1655, opened 2023-09-26
Ability to Disable TripalImporter submit button

### LINKED ISSUE #1652: Disable submit button in Importer

No

No

### Is this feature urgent?

Yes

### Description

We have implemented the Tripal Importer Plugin in our module using Accordion layout where each header is a stage and each stage has to be completed in order to proceed to the next stage. We would like to be able to disable the Submit button on importers based on variable in the form state until a user is in a particular stage.

image

### Your specific use case

n/a

### Generally Applicable

n/a

n/a

laceysanderson:
# Tripal 4 Core Dev Task

## Description

The Tripal Importer form takes complete control over the file and submit elements in such a way that Importer implementations (e.g. GFF3 loader) cannot modify these elements.

In some cases, this is problematic -especially for the submit button. For example, if an importer wants to do in page AJAX validation before allowing submit such as validating the file via a Tripal Job, developers of the importer have no way to stop submission of the form via the submit button. Additionally, Drupal updates the DOM after ajax submission by appending unique codes to the IDs which makes targetting these ids via javascript alone difficult. Another example if so a multi-stage form or a form using an accordion.

This PR adds two pieces of functionality to allow TripalImporter implementations to disable the submit button.
1. A new property in the annotation `submit_disabled`. When set to true for an importer, the submit button will be disabled when a user first loads the form.
2. A key in the form_state storage `disable_TripalImporter_submit` that when true the submit button is disabled and when false, the submit button is enabled. Since this is in the form_state it can be set via AJAX or via the form itself.

An example of using these two additions is in the importer attached to the testing section. This example has `submit_disabled: True` in the annotation. The form on first load shows the submit button disabled and a selection of form elements. AJAX is triggered once the user selects something from the select list which rebuilds the entire form. On rebuild, the form sees the select list element has a value and sets the form_state storage `disable_TripalImporter_submit` to FALSE. Since TripalImporterForm adds the button after it pulls in the implementation form elements, it sees the form state set to FALSE and renders the button as enabled.

Here is a code snippet of changing the value of the form state. This is done using the Storage of the form which stores random information about the state. This ensures we are following Drupal Standards :-)

```php
$storage = $form_state->getStorage();
$storage['disable_TripalImporter_submit'] = FALSE;
$form_state->setStorage($storage);
```

Also this change is backwards compatible as when I added `submit_disabled` to the annotation, I set a default of FALSE.

### Check Backwards compatibility.
Load the existing importers and ensure the submit button is not disabled ;-)

### Check current Functionality.
1. Download my [example importer class](https://github.com/tripal/tripal/files/12730176/TestImporter.txt) and change the file ending to php. Place this class in the `tripal/src/Plugin/TripalImporter` directory.
2. Clear the cache. This is needed for new importers to be found. You may need to clear the cache 2X if the routing isn't picked up the first time.
3. Go to the importer listing Tripal > Data Loaders > Test Disable Button
4. Ensure the import button is disabled!
5. Select anything from the drop down list.
6. Ensure that the button is now enabled! Also enjoy the reply to your selection <3

### DIFF
--- a/tripal/src/Form/TripalImporterForm.php
+++ b/tripal/src/Form/TripalImporterForm.php
@@ -130,10 +130,20 @@ class TripalImporterForm implements FormInterface {
     // We should only add a submit button if this importer uses a button.
     // Examples of importers who don't use this button are multi-page forms.
     if (array_key_exists('use_button', $importer_def) AND $importer_def['use_button'] !== FALSE) {
+
+      // We will allow specific importers to disable this button based on the state of the form.
+      // By default it is enabled.
+      $disabled = FALSE;
+      // But if they set the storage to indicate we should disable it then we will.
+      $storage = $form_state->getStorage();
+      if (array_key_exists('disable_TripalImporter_submit', $storage)) {
+        $disabled = $storage['disable_TripalImporter_submit'];
+      }
       $form['button'] = [
         '#type' => 'submit',
         '#value' => $importer_def['button_text'],
         '#weight' => 10,
+        '#disabled' => $disabled,
       ];
     }
 
--- a/tripal/src/Form/TripalImporterForm.php
+++ b/tripal/src/Form/TripalImporterForm.php
@@ -134,7 +134,12 @@ class TripalImporterForm implements FormInterface {
       // We will allow specific importers to disable this button based on the state of the form.
       // By default it is enabled.
       $disabled = FALSE;
-      // But if they set the storage to indicate we should disable it then we will.
+      // Unless the annotation says it should be disabled by default..
+      if (array_key_exists('submit_disabled', $importer_def) AND $importer_def['submit_disabled'] === TRUE) {
+        $disabled = TRUE;
+      }
+      // But if they set the storage to indicate we should disable/enable it
+      // then we will do whatever they say ;-).
       $storage = $form_state->getStorage();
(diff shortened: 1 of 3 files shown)


dsenalik (approved):
As far a functional testing, adding this to the FASTA importer definition, or the OBO importer definition
` *    submit_disabled = True,`
and the submit button is grayed out.

Putting this in the form() function re-enables the button:
```
$storage = $form_state->getStorage();
$storage['disable_TripalImporter_submit'] = FALSE;
$form_state->setStorage($storage);
```

Then I tested the listed "TestImporter" steps,
this is the most fun test in a long time!

reynoldtan:
Tested submit_disabled definition with True value and Import button rendered grayed and disabled:
Screen Shot 2023-09-26 at 3 26 01 PM

a False value and the button rendered enabled and clickable
Screen Shot 2023-09-26 at 3 25 21 PM

Tested the same definition programmatically using the code snippet

```
$storage = $form_state->getStorage();
$storage['disable_TripalImporter_submit'] = FALSE; // Or TRUE;
$form_state->setStorage($storage);
```

and the Import button rendered enabled/disabled accordingly.

AJAX select test works as expected. I chose: **Powerful but challenging** 👍👍👍

reynoldtan (approved):

laceysanderson:
Thanks for the quick reviews! <3

PR #1644, opened 2023-09-21
Tripal 4 importer logging updates

dsenalik:
# Bug Fix

## Description

 * logger warnings and errors should all use variable substitution, for example
`$this->logger->error("Regular expression for the feature name finds nothing. Line $i.");`
becomes
```
$this->logger->error("Regular expression for the feature name finds nothing. Line @line.",
  ['@line' => $i]
);
```

 * A number of updates from Drupal messenger to setErrorByName in validation functions. The primary problem is that messenger allows the import job to proceed. setErrorByName will highlight the faulty field and require edit and then resubmission. For example
`\Drupal::messenger()->addError(t("You must provide a regular expression to identify the sequence unique name"));`
becomes
`$form_state->setErrorByName('re_name', t('You should not specify a regular expression for name if you are matching 
to sequence unique name'));`

 * Remove some variables used by Tripal 3 but now in the object header comment definitions. https://github.com/tripal/tripal/pull/1644/commits/c12382d9f4f25b2a2cebdba7630fcd4ba06a63d6

 * A few typos fixed

 * A few bugs fixed seen in the code review. Some of these are backported to Tripal 3 in Issue #1643 pull #1645

## Testing?
Manual testing involves trying to run importers and leave fields out or enter bogus values in fields.
 * Fasta importer select "Name" but put a regex in "Unique name" or vice versa. 
Before these changes, there was an error, but the job was submitted anyway.
fasta-regex0

After this PR it will be validated thusly:
fasta-regex1
fasta-regex2

 * enter non-existing cv term names (type in a value and don't use the ajax list provided) and verify error is shown, e.g. for the GFF3 importer:
CV1
CV2

PR #1657, opened 2023-09-26
Removes duplicated importer metadata

### LINKED ISSUE #1359: Importer Base Class -duplicated information
When reviewing the GFF3 importer I noticed we have duplicated information at the top of the class:
1. In the docblock annotation
```php
/**
 * GFF3 Importer implementation of the TripalImporterBase.
 *
 *  @TripalImporter(
 *    id = "chado_gff3_loader",
 *    label = @Translation("Chado GFF3 File Loader"),
 *    description = @Translation("Import a GFF3 file into Chado"),
 *    file_types = {"gff","gff3"},
 *    upload_description = @Translation("Please provide the GFF3 file."),
 *    upload_title = @Translation("GFF3 File"),
 *    use_analysis = True,
 *    require_analysis = True,
 *    button_text = @Translation("Import GFF3 file"),
 *    file_upload = True,
 *    file_load = False,
 *    file_remote = False,
 *    file_required = False,
 *    cardinality = 1,
 *    menu_path = "",
 *    callback = "",
 *    callback_module = "",
 *    callback_path = "",
 *  )
 */
class GFF3Importer extends ChadoImporterBase {
```
2. As class member variables
```php
  /**
   * The name of this loader.  This name will be presented to the site
   * user.
   */
  public static $name = 'Chado GFF3 File Loader';

  /**
   * The machine name for this loader. This name will be used to construct
   * the URL for the loader.
   */
  public static $machine_name = 'chado_gff3_loader';

  /**
   * A brief description for this loader.  This description will be
   * presented to the site user.
   */
  public static $description = 'Import a GFF3 file into Chado';

  /**
   * An array containing the extensions of allowed file types.
   */
  public static $file_types = ['gff', 'gff3'];

  /**
   * Provides information to the user about the file upload.  Typically this
   * may include a description of the file types allowed.
   */
  public static $upload_description = 'Please provide the GFF3 file.';

  /**
   * The title that should appear above the upload button.
   */
  public static $upload_title = 'GFF3 File';

  /**
   * Text that should appear on the button at the bottom of the importer
   * form.
   */
  public static $button_text = 'Import GFF3 file';
```

This is likely due to the merging of our Tripal 3 classes with the Drupal 9 annotation model.

### TASKS: 
- Check that all uses of the annotation-set data uses the annotation method and not the member variable (i.e. in the TripalImporterForm using `$importer_def = $importer_manager->getDefinitions()[$plugin_id];` is the correct way)
- Remove the class member variables that are already set in the annotation for the base class, OBOImporter and GFF3Importer
- Test that nothing broke when those member variables were removed
- Document this in ReadtheDocs in a new page describing how to upgrade importers from Drupal 7/Tripal3

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This is a pet peeve of mine and I needed an easy task ;-) 

When the importers were upgraded, the kept the original variables used in our Tripal 3 classes but then also added these properties to the annotation. This resulted in duplicated information. The upgrade was done well though so only the annotation values are being used which is correct ✅ This PR removes the old variables to reduce confusion and ensure there doesn't end up being confusing mismatches ;-) 

Additionally I set defaults for the annotation values where it made sense to do so which allows these to be left out of the annotation of importers happy to use the defaults.

## Testing?

All these forms have 100% automated test coverage of the elements controlled by the annotation so this should not even need manual testing.

That said, if you would like to be thorough, simply starting up a fresh docker and loading each of the importers provided by core to ensure the form elements appear as expected:
 - check that the text from the annotation is on the form
 - check that the form elements for file upload, analysis and submit are still present

### DIFF
--- a/tripal/src/TripalImporter/Annotation/TripalImporter.php
+++ b/tripal/src/TripalImporter/Annotation/TripalImporter.php
@@ -45,8 +45,7 @@ class TripalImporter extends Plugin {
    *
    * @var array
    */
-  public $file_types;
-
+  public $file_types = ['txt'];
 
   /**
    * Provides information to the user about the file upload.
@@ -77,7 +76,7 @@ class TripalImporter extends Plugin {
    *
    * @var bool
    */
-  public $use_analysis;
+  public $use_analysis = TRUE;
 
   /**
    * If the $use_analysis value is set above then this value indicates if the
@@ -85,7 +84,7 @@ class TripalImporter extends Plugin {
    *
    * @var bool
    */
-  public $require_analysis;
+  public $require_analysis = FALSE;
 
   /**
    * Indicates whether the base importer should add a submit button or not.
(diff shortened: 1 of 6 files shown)


dsenalik:
lol @laceysanderson I thought the same thing and was removing some of them in PR #1644 - but you were more thorough!

PR #1619, opened 2023-08-28
Sequence length + Md5sum fields

pdtouch:
#

Issue #1509

## Description
Modified the tripal_chado tripal_collections YAML file to include new fields into the gene page bundle.

## Testing?
- To test the YAML you need to build a new docker on this branch
```
n=1509
git clone https://github.com/tripal/tripal tripal-$n
git checkout tv4g1-1509-sequence-fields
docker build --tag=tripaldocker:testing-$n --build-arg drupalversion="10.1.x-dev" --file tripaldocker/Dockerfile-php8.2-pgsql13 ./
docker run --publish=8082:80 -tid --name=${n} --volume=/home/admin/git/tripal-$n:/var/www/drupal9/web/modules/contrib/tripal --volume=/home/admin/docker_prepare:/usr/local/bin/docker_common tripaldocker:testing-$n
```
- Log in to http://localhost:8082
- Create an organism record
- Create a gene record and include a sequence e.g. `actg`
- The sequence should appear along with sequence length and sequence MD5 checksum

PR #1663, opened 2023-10-05
ChadoStorage should support loading base values

### LINKED ISSUE #1662: ChadoStorage read value from base table
### Task Description

As found in #1619, we need the ability to read a value from the base table. Currently you have the choice of a variety of store* actions for the base table and join for linked tables but no read on the base table.

This is needed for fields which do not handle editing of a base table column, as well as, fields which add additional functionality or specialized displays for a specific column in addition to another field which handles editing.

### Branch Name

tv4g2-issue1662-loadBaseValues

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

As found in https://github.com/tripal/tripal/pull/1619, we need the ability to read a value from the base table. Currently you have the choice of a variety of store* actions for the base table and join for linked tables but no read on the base table.

This is needed for fields which do not handle editing of a base table column, as well as, fields which add additional functionality or specialized displays for a specific column in addition to another field which handles editing.

This PR:
 - adds support for a new `read_value` action which allows you to indicate you want to read a value from the base table
 - Fixes a bug for the join action which assumed you would have a base_table store property directly before defining the join property.
 - Adds automated tests for the read_value action.
 - Deprecated the join action in favour of the read_value action. Now you can provide read_value for any value you do not want to edit... if it's on the base table define chado_table and if it's on another table define `path`.

### Automated Tests

- Ensure that a property with the read_value action has the value set
- Ensure that a property with the read_value action can't change the value
- That two fields accessing the same chado column do not conflict
    A. both read_value action for the same column
    B. one read_value and one store for the same column

### Manual Testing

1. Download [the following Field](https://github.com/tripal/tripal/files/12821857/ChadoReadFeatureNameTypeItem.txt) and save it in the `tripal_chado/src/Plugin/Field/FieldType` folder. Change the file suffix to `.php`.
2. Go to Admin > Tripal > Page Structure > Gene > Manage Fields and add a field of the type "Chado Read Feature Name Field Type".
3. Create an organism record
4. Create a gene record. Any value entered into the widget for the field you created in step 2 should not be saved but rather display the same as the name field for the gene.
5. Edit the gene record and change the name but not the value of your field. Ensure the value of your field is updated to the new name.
6. Edit the gene record and change the value of your field but not the name field. Ensure the value of your field is reset to the name and the value you entered is not saved.

This PR now also updates the old join action to read_value. To test that, you will want to test the AdditionalType, Analysis, Contact and organism fields to ensure they still act the same way they did before. 

Note: There are already automated tests for these fields.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeDefault.php
@@ -157,19 +157,19 @@ class ChadoAdditionalTypeDefault extends ChadoFieldItemBase {
     // This field needs the term name, idspace and accession for proper
     // display of the type.
     $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'term_name', $name_term, 128, [
-      'action' => 'join',
+      'action' => 'read_value',
       'path' => $type_table . '.' . $type_column . '>cvterm.cvterm_id',
       'chado_column' => 'name',
       'as' => 'term_name'
     ]);
     $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'id_space', $idspace_term, 128, [
-      'action' => 'join',
+      'action' => 'read_value',
       'path' => $type_table . '.' . $type_column . '>cvterm.cvterm_id;cvterm.dbxref_id>dbxref.dbxref_id;dbxref.db_id>db.db_id',
       'chado_column' => 'name',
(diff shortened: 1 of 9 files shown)


laceysanderson:
Functional testing was done by @pdtouch and [described in this comment on the original issue](https://github.com/tripal/tripal/issues/1662#issuecomment-1758265395). As it worked well for him, automated testing passes and the approach behind this has been discussed already with Stephen, I am going to merge this. 

Thank you for the review @pdtouch!

PR #1664, opened 2023-10-16
Add dependency injection to the ChadoImporterBase

carolyncaron:
# Tripal 4 Core Dev Task

## Description

While developing a Tripal Importer, access to the database was needed via dependency injection. Thus, changes were made to the ChadoImporterBase class according to method 2 in Issue #1522. Some important notes:

- TripalImporterBase has not been changed except to rename the create() function since that name is needed by the ContainerFactoryPluginInterface. The old create() function has been renamed to createImportJob() and all references to it have been updated (hopefully!). Thus, TripalImporterBase is now _compatible_ to use dependency injection but **does not yet use it**.
- ChadoImporterBase function getChadoConnection() is still needed to set the database connection, the only change is that it no longer uses the static method (Drupal::service) to grab the connection.
- These changes are backwards compatible. 

## Testing?

I tested these changes by running the functional tests for the various Tripal importers. You can run them individually or run tests for all the importers at once using (assuming your docker container is named tripal-1522):
`docker exec --workdir=/var/www/drupal9/web/modules/contrib/tripal/ tripal-1522 phpunit --group TripalImporter`

If you have a custom Tripal importer, try it to see if the changes are indeed backwards compatible. I tested it on the importer I'm currently developing on, you can do so as well by:
- Cloning this repo: https://github.com/TripalCultivate/TripalCultivate-Germplasm.git 
- Switch to the branch `g2.2-germAccessionImporter`
- Run phpunit on `trpcultivate_germplasm/tests/src/Kernel/TripalImporter/GermplasmAccessionImporterTest.php` 

Laslty, you can try testing by setting up 2 different chado instances. When submitting an importer job, choose the 2nd instance and afterwards check that it imported into the right one.

### DIFF
--- a/tripal/src/Form/TripalImporterForm.php
+++ b/tripal/src/Form/TripalImporterForm.php
@@ -232,7 +232,7 @@ class TripalImporterForm implements FormInterface {
         return;
       }
 
-      $importer->create($run_args, $file_details);
+      $importer->createImportJob($run_args, $file_details);
       $importer->submitJob();
 
     }
--- a/tripal/src/TripalImporter/TripalImporterBase.php
+++ b/tripal/src/TripalImporter/TripalImporterBase.php
@@ -177,7 +177,7 @@ abstract class TripalImporterBase extends PluginBase implements TripalImporterIn
    * @return int
    *   Returns the import_id.
    */
-  public function create($run_args, $file_details = []) {
+  public function createImportJob($run_args, $file_details = []) {
 
     // global $user;
     $user = User::load(\Drupal::currentUser()->id());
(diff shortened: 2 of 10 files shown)


laceysanderson (approved):
✅ Code review looks good! 👍
     - includes important changes for DI for chadoimporterbase
     - renames the original create to createImportJob
     - consistently updates create => createImportJob everywhere that was used in core
 ✅ Functional review was good!
     - Checked the GFF3 and Fasta importer using the advanced "schema name" functionality to import into a clone chado database rather then the default one.
     - On this branch both importers worked beautifully and imported into the selected cloned chado as they should 👍 

Thanks @carolyncaron <3 This one is ready to merge! 🎉

PR #1665, opened 2023-10-16
Property Field Bug and reduce ChadoStorage complexity

### LINKED ISSUE #1398: Tripal 4 property field can't handle two properties at the same time

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR is currently in draft form as it is not complete. I believe I have reduced the complexity of the buildChadoRecord method (was 160 according to code climate when the threshold is 25) but was unable to check without making a PR...

So ignore this PR for now.

### DIFF
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -890,19 +890,7 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
           // stored in the Chado table indicated by this property.
           // ................................................................
           if ($action == 'store') {
-            $chado_column = $prop_storage_settings['chado_column'];
-            $value = $prop_value->getValue();
-            if (is_string($value)) {
-              $value = trim($value);
-            }
-            $records[$chado_table][$delta]['fields'][$chado_column] = $value;
-
-            // If this field should not allow an empty value that means this
-            // entire record should be removed on an update and not inserted.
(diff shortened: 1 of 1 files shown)


laceysanderson:
Apparently code climate won't tell me anyway 😒

PR #1667, opened 2023-10-23
Clean-up and Testing of TripalEntity

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

While working on the Property bug in ChadoStorage and boosting test coverage, I realized there was an unused class in TripalStorage, specifically, tripal/src/TripalStorage/TripalEntityBase.php was never used and most of the functionality had been moved into tripal/src/Entity/TripalEntity.php and evolved from there.

This PR removes `tripal/src/TripalStorage/TripalEntityBase.php` to reduce confusion and since it was useless. It also adds testing for TripalEntity as there were none which is part of why this wasn't caught.

Finally, it fixes two very minor things that came up during testing:
- Ensures that the default user set matches the current logged in user. Previously the wrong key was used for the user.
- Removes an unreachable block of code. According to the Drupal documentation, the id is unique and every entity will only have one id. The code block removed, checked to see if there were multiple ids set for a single entity object despite the variable holding the id being a string. I suspect it was added just to be safe since it took a Lot of digging to find the variable definition for the id.

## Testing?

Since the changes are very minor, this likely doesn't need any more manual testing than I've already done.

That said, if you do want to test manually (which is always appreciated!) then you would 
1. Create a new docker using this branch.
2. Create content for a couple of different entities while checking:

    - the fields you expect are there
    - it shows created by your user
    - there are no errors in the logs
    - the content creates as you expect.

### Automated testing

I just added a basic automated test for the TripalEntity class which creates an entity and checks the defaults are set properly. The goal was to get some coverage without spending a lot of time 🤷 and to start the process so it's easier to add to later.


dsenalik (approved):
Code review revealed one typo, fixed with https://github.com/tripal/tripal/pull/1667/commits/0e2d789ee14f64931f51474c5cf1b2bd74205732

I did do the manual testing, and things work as expected, no errors.
I did see one change, the "Authored by" field now works, it was broken before, thanks presumably to the fix on line 98
tripalentitycleanup

PR #1671, opened 2023-10-31
BUG FIX: ChadoAdditionalType field cannot handle multiple instances at once

### LINKED ISSUE #1669: Tripal 4 chado additional type field can't handle two types at the same time
### Tripal Version

4.x

### Drupal Version

10.1

### PostgreSQL Version

13.11

### PHP Version

8.2

### General Description

This issue is analogous to issue #1398 but in this case the field is the `ChadoAdditionalType` field "Chado Type Reference".

Example from the Array Design content type, enter this
20231028_arraydesign1

and when saved one of the values leaks through to the other field
20231028_arraydesign2

although both values are saved correctly in the `arraydesign` table
```
 arraydesign_id | manufacturer_id | platformtype_id | substratetype_id | protocol_id | dbxref_id |       name       | ...
----------------+-----------------+-----------------+------------------+-------------+-----------+------------------+ ...
              1 |               2 |            2996 |             2193 |             |           | Array Design One | ...
```

### Steps to reproduce

Create an array design with different values as illustrated above

### Error messages

_No response_

### Screenshots

_No response_

laceysanderson:
# Bug Fix

## Description

@dsenalik discovered (and demonstrated in the attached issue) that the ChadoAdditionalType struggled when there were two instances of it on the same TripalContent page. Specifically, on the Array Design content type, there is a Substrate and Platform Type which are both ChadoAdditionalType fields. This resulted in the term information for only one being shown in the formatter (see below screenshot) although at the database level the platformtype_id and substratetype_id were assigned correctly based on the user submission.

It turns out this was due to a bug in the read_value action for joins. Although the SQL generated was specific enough to keep the different joins separated, the code which retrieved the value after the query was run was not aware of the alias' added and could not accurate look up the value. As such it always defaulted to the first value added to the query as that name wasn't aliased.

This PR fixes the bug by ensuring all columns selected based on a join are aliased with the right table alias and developer specified column alias (e.g. jcvterm0_term_name) and that this alias is kept track of by field and property key to ensure accurate look-up and replacement.

## Testing?

Automated testing for this exact case has been added to the ChadoStorageActions_ReadValueTest test class in the `testReadValueActionJoinDouble` method. This tests the exact arraydesign case discussed above with a focus on the loadValues. We do not need to test the insertValues and UpdateValues because this bug is specific to the way the records array is built and values set based on the results which are the same between all three *Values methods.

### Manual Testing

1. Setup a testing environment on this branch.
2. Go to Tripal > Content > Add Tripal Content > Array Design (under Expression)
3. Fill out the form including specifying different terms for the Platform and Substrate Types. Take note of the terms and which you entered for each type.
4. Click save and ensure on the view page that both types have the appropriate term shown.

PR1669-ArrayDesign-Create
PR1669-ArrayDesign-View

### DIFF
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -91,6 +91,20 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
   protected $reverse_alias_mapping = [];
 
   /**
+   * A mapping of the field property to the alias used in the query.
+   * This is specific to properties with an action of read_value who
+   * specify a path.
+   *
+   * @var array
+   *  a nested array where mapping the new alias including the right table alias
+   *  to the original alias set in the property type. The structure is:
+   *   [field name]:
+   *     [property key]:
+   *       [original column alias]: [new alias]
+   */
+  protected $join_column_alias = [];
+
+  /**
    * Implements ContainerFactoryPluginInterface->create().
    *
    * Since we have implemented the ContainerFactoryPluginInterface this static function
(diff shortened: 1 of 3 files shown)


dsenalik (approved):
Wonderful 😍 both the fix and the added testing!
Code looks good (it's too bad chado storage is so complicated, but that's the way it is 🙈)
I tested with "DNA" and "RNA" and they saved correctly.
pr1671-1

PR #1674, opened 2023-11-01
Fixes to Tripal Importer Forms

### LINKED ISSUE #1672: OBO Importer form is currently broken
### Tripal Version

4.x

### Drupal Version

all supported versions

### PostgreSQL Version

all supported versions

### PHP Version

all supported versions

### General Description

It looks like some recent updates to the TripalImporterBase has broken some things in the OBO Importer. Specifically,

- There is a new fieldset for "New OBO File" which should not be there
- Creating a new OBO reference submits a job with a 0 obo_id
- Trying to run an existing OBO reference incorrectly shows validation errors for not filling in anything in the new obo reference section.
- Ajax appears to be broken

### Steps to reproduce

Go to Admon > Tripal > Data Loaders > OBO Vocabulary Loader.

### Error messages

_No response_

### Screenshots

Screenshot of OBO importer form

laceysanderson:
# Bug Fix

## Description

Specifically this PR fixes the following issues:

> There is a new fieldset for "New OBO File" which should not be there.

This was because the changes to annotation added defaults for missing values. The OBO importer did not set the `file_local` and this defaulted to TRUE; whereas apparently in the past an excluded annotation value always defaulted to FALSE. This was fixed by simply setting the annotation properly.

> Creating a new OBO reference submits a job with a 0 obo_id.

This was caused by a bug in the form submit for the TripalImporterForm. Specifically, any changes made during the importer formSubmit were not being carried through to the job arguments. Since the OBO importer sets the obo_id in it's own form submit after creating the new obo reference, this was not available in the job arguments.

> Trying to run an existing OBO reference incorrectly shows validation errors for not filling in anything in the new obo reference section.

This was actually due to broken AJAX although that was not immediately evident at first. The logic in the submit thought that the update button was being triggered even though it wasn't actually present visually due to failed AJAX.

> Ajax appears to be broken

It turns out the dependency injection PR for ChadoImporter caused this. Specifically, the AJAX callback was being specified using `$this` which serialized the entire importer object for an AJAX call just to be able to access the callback! This wasn't a problem before because the TripalImporter class didn't use dependency injection and thus was serializable... 

This was fixed by simply making the AJAX callbacks static (an easy change since they didn't use any initialized info in the class anyway) and then changing the callback specification to pass in the class name instead of the object. This is also more efficient.

### There is a new fieldset for "New OBO File" which should not be there.

Just go to the OBO importer form (Admin > Tripal > Data Loaders > OBO Vocab Importer) and ensure the importer looks like this:
image

### Creating a new OBO reference submits a job with a 0 obo_id

1. Go to the OBO importer form (Admin > Tripal > Data Loaders > OBO Vocab Importer)
2. Expand the "Add a New Ontology Reference" fieldset and fill in the form as follows:
     - New Vocabulary Name: required but you can put in anything here
     - Local File: /var/www/drupal9/web/modules/contrib/tripal/tripal_chado/files/tcontact.obo
3. Then click Import OBO File

You should see a success message at the top saying the vocab was completed, it should show up in the existing list and it should have submitted a job.

**Make sure to run the job! This should work now whereas before you got the following message:
```
2023-11-01 21:54:50
Tripal Job Launcher
Running as user 'drupaladmin'
-------------------
2023-11-01 21:54:50: Job ID 5.
2023-11-01 21:54:50: Calling: tripal_run_importer(9)
Running 'OBO Vocabulary Loader' importer
NOTE: Loading of this file is performed using a database transaction. If it fails or is terminated prematurely then all insertions and updates are rolled back and will not be found in the database
ERROR: Invalid OBO ID provided: '0'.
[site http://default] [TRIPAL] ERROR: Invalid OBO ID provided: '0'.
```
### Trying to run an existing OBO reference incorrectly shows validation errors for not filling in anything in the new obo reference section.

***This also confirms that AJAX is not broken any more.***

1. Go to the OBO importer form (Admin > Tripal > Data Loaders > OBO Vocab Importer)
2. Select an existing obo record from the drop down.
3. Confirm that there are no errors on the page and that the "Use a Saved Ontology Reference" fieldset expands with additional info specific to the selected ontology, plus the update and delete buttons show up as shown in the screenshot shared below.
4. Click Import OBO File and confirm that you get the message that the job was successfully submitted.

Screenshot 2023-11-01 at 4 20 13 PM

### DIFF
--- a/tripal/src/Form/TripalImporterForm.php
+++ b/tripal/src/Form/TripalImporterForm.php
@@ -170,6 +170,23 @@ class TripalImporterForm implements FormInterface {
     $importer = $importer_manager->createInstance($plugin_id);
     $importer_def = $importer_manager->getDefinitions()[$plugin_id];
 
+    // Now allow the loader to do its own submit if needed.
+    try {
+      $importer->formSubmit($form, $form_state);
+      // Ensure any modifications made by the importer are used.
+      $form_values = $form_state->getValues();
+      $run_args = $form_values;
+    }
+    catch (\Exception $e) {
+        \Drupal::messenger()->addMessage('Cannot submit import: ' . $e->getMessage(), 'error');
+    }
+
+    // If the importer wants to rebuild the form for some reason then let's
+    // not add a job.
+    if ($form_state->isRebuilding() == TRUE) {
+      return;
+    }
+
     // Remove the file_local and file_upload args. We'll add in a new
     // full file path and the fid instead.
     unset($run_args['file_local']);
@@ -218,23 +235,8 @@ class TripalImporterForm implements FormInterface {
       $file_details['file_remote'] = $file_remote;
     }
     try {
-      // Now allow the loader to do its own submit if needed.
-      $importer->formSubmit($form, $form_state);
-      // If the formSubmit made changes to the $form_state we need to update the
-      // $run_args info.
-      if ($run_args !== $form_values) {
-        $run_args = $form_values;
-      }
-
-      // If the importer wants to rebuild the form for some reason then let's
-      // not add a job.
-      if ($form_state->isRebuilding() == TRUE) {
-        return;
-      }
-
       $importer->createImportJob($run_args, $file_details);
       $importer->submitJob();
-
     }
     catch (\Exception $e) {
         \Drupal::messenger()->addMessage('Cannot submit import: ' . $e->getMessage(), 'error');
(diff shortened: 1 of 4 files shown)


dsenalik (approved):
Awesome 😍, thanks for figuring this out.
I tested with drupal 10.1, php 8.2, and performed the manual testing steps and results were as described, no errors. (I used the sequence ontology)

laceysanderson:
Thanks for the quick review @dsenalik! ❤️ 🎉

PR #1679, opened 2023-11-06
Do not display fopen error messages

dsenalik:
# Bug Fix
## Description
All this does is hides error messages from `fopen()` in the taxonomy importer (they are not exceptions). ~I think~ these are preventing the retry loop from working during automated testing.

## Testing?
The errors cannot be reproduced as they are intermittent. This small change should be safe, and maybe we just need time to see if this does it.

**Update**
I found a way to reliably test this. I can add code to create a very similar error.
Add this line to `tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php` after line 810
`$xfh = fopen("https://www.ncbi.nlm.nih.gov/ spaces should give 400 entrez/eutils/efetch.fcgi", "r"); // to give HTTP/1.1 400 Bad Request`
and run the importer tests (my container is named `1381`)
`docker exec --workdir=/var/www/drupal9/web/modules/contrib/tripal 1381 phpunit --group 'TaxonomyImporter' --debug`
```
There was 1 error:

1) Drupal\Tests\tripal_chado\Functional\TaxonomyImporterTest::testTaxonomyImporterSimpleTest
fopen(https://www.ncbi.nlm.nih.gov/ spaces should give 400 entrez/eutils/efetch.fcgi): Failed to open stream: HTTP request failed! HTTP/1.1 400 Bad Request
```

Now change this line by adding `@` and run the tests again
`$xfh = @fopen("https://www.ncbi.nlm.nih.gov/ spaces should give 400 entrez/eutils/efetch.fcgi", "r");`
`       ^`
...
`OK (2 tests, 37 assertions)`

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
+++ b/tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
@@ -551,6 +551,7 @@ class TaxonomyImporter extends ChadoImporterBase {
         if (!empty($api_key)) {
           $search_url .= "&api_key=" . $api_key;
         }
+$ffh = fopen('https://will.fail. .surely', "r");  // Try to trigger the failure but not affect loading
         $rfh = NULL;
         // Query NCBI. To accomodate occasional glitches, retry up to three times.
         $retries = 3;
--- a/tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
+++ b/tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
@@ -551,7 +551,6 @@ class TaxonomyImporter extends ChadoImporterBase {
         if (!empty($api_key)) {
           $search_url .= "&api_key=" . $api_key;
         }
-$ffh = fopen('https://will.fail. .surely', "r");  // Try to trigger the failure but not affect loading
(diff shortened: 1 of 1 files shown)


laceysanderson:
Nice! Yes, they are definitely stopping the retry loop in phpunit despite not being exception. PHPUnit monitors for messages sent to the error logs, catches them and re-throws them as an error. This behaviour caught me when testing and expecting TripalLogger->error() to output a message that I wanted to catch with ob_*

laceysanderson (approved):
I do worry a bit about hiding all the messages from fopen... but realistically this is a good fix for a really annoying problem. As such I'm going to approve 👍 Thank you!!!

Now onto more discussion/thinking out loud, I wonder if switching to the [new Drupal HttpClient](https://gorannikolovski.com/snippet/drupal-89-and-guzzle-http-client-library) like I did [in the OBO loader](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Plugin/TripalImporter/OBOImporter.php#L2876-L2946)? Apparently this uses fsockopen instead of fopen and might have more ability to catch problems accurately... It might be worth creating an issue of something to look into just in case there is time or need at a later date for the switch :-) 🤷‍♀️

PR #1690, opened 2023-11-15
Add a default label for the Authored by field

### LINKED ISSUE #1689: The "Authored by" field is missing a title
### Tripal Version

4.x

### General Description

We should change the default for the "Authored by" field.
This was broken and not showing content until Lacey fixed it in PR #1667 
It is displaying on all pages without a title which is inconsistent with everything else.
In `tripal/src/Entity/TripalEntity.php`
```
    $fields['uid'] = BaseFieldDefinition::create('entity_reference')
      ->setLabel(t('Authored by'))
      ->setDescription(t('The user ID of author of the Tripal Content entity.'))
      ...
      ->setDisplayOptions('view', array(
        'label' => 'hidden',
        'type' => 'author',
        'weight' => 0,
      ))
      ...
```
I propose to change this to `'label' => 'above',` to match defaults for other fields.

### Steps to reproduce

Create just about any content type. It looks like the screenshot below

### Screenshots

no_heading

dsenalik:
# Bug Fix

## Description
The "Authored by" field is displaying on all pages without a title which is inconsistent with everything else.
This PR just changes the default. Admins can still change the display settings as with every field.

## Testing?
Build a docker on this branch, and create any content type. Here I created a project.
The "Authored by" field should display a title as shown
### Before
no_heading

### After
issue1689

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -286,13 +286,13 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
 
     $fields['uid'] = BaseFieldDefinition::create('entity_reference')
       ->setLabel(t('Authored by'))
-      ->setDescription(t('The user ID of author of the Tripal Content entity.'))
+      ->setDescription(t('The user ID of the author of the Tripal Content entity.'))
       ->setRevisionable(TRUE)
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Always in favour of consistency! Thanks for the change @dsenalik ❤️

PR #1691, opened 2023-11-15
Remove all '#type' => 'markup' because it doesn't do anything

dsenalik:
# Bug Fix

## Description
Removes `'#type' => 'markup'` from everywhere.
This is because since Drupal 8 it is no longer a valid type, and was not needed in Drupal 7.
For a reference see https://www.drupal.org/node/2036237

## Testing?
An easy place to test this is on the FASTA importer, the file format description element is one of the places changed.
### Before
fasta-before

### After
fasta-after

They look the same to me 😀


laceysanderson (approved):
Confirmed and tested! 🙈  Thanks @dsenalik!

PR #1694, opened 2023-11-17
Add formatter settings to fields that need them

### LINKED ISSUE #1692: Add formatter settings to fields that need them
### Task Description

Now that I know how to add formatter settings to a field formatter, I would like to add to other fields starting with the boolean field. Currently it is hardcoded to display either 'true' or 'false'. This should be configurable for each instance of this field. You might want 'Yes' and 'No', etc.

### Branch Name

tv4g1-issue1692-field-formatter-settings

### PR #1694

dsenalik:
# Tripal 4 Core Dev Task

## Description
This PR adds field display settings to:
 * ChadoBooleanTypeFormatter - configure the value displayed for each of true and false, _e.g._ "Yes" and "No"
 * ChadoSequenceLengthFormatterDefault - configurable prefix and suffix, _e.g._ suffix of " b.p.", as well as thousand separator character _e.g._ ",", ".", " "
 * ChadoSequenceFormatterDefault - configurable line length and autosizing window up to container width, defaults to 50 as before
 * ChadoSequenceChecksumFormatterDefault - configure either upper case, lower case, or as stored in the database table.
 * `ChadoSequenceLengthFormatterCommas.php` is deleted as it is now no longer needed (sorry @srobb1 )

## Testing?
 * Create a gene record for testing, with a sequence at least 1,000 b.p. long
 * Go to the gene display settings page, observe that with this PR there are now settings cogs.
settings_page

 * For the "Sequence Length" field, enter a suffix value, " b.p." is a logical choice, and a thousand separator character such as "," or "." or " " (space)
integer_type_settings

 * For the "Sequence Residues" field, change wrap value to "100". Validation prevents a negative number, or a non-number. (Default is 50)
residues

 * For the "Sequence MD5 Checksum" field, select either upper case or lower case. (Default is "As stored")
md5
 
 * For the "Is Analysis" or "Is Obsolete" fields, enter new text for true and false. The validation here is they cannot be blank, and they cannot both be the same. (Default is "True" and "False")
boolean

 * Observe the gene page. For example:
gene_example_1


laceysanderson (approved):
Beautiful!

I love that I can choose any word for the true/false (i.e. obsolete/current for the is_obsolete field 😁 ).

The validation all worked as expected. Although I can enter 0 for wrap length in the residues without error, the page then shows the default wrap length.

Code review looks perfect.

dsenalik:
Thanks @laceysanderson 😍 I think that was intentional in that wrap length of zero was the same as not specifying one. One might argue that zero could mean "do not wrap", but that would not be a reasonable display option. Still, you can set the wrap length to 1000000 if you want, but don't expect nice results 💣. I had considered setting an upper limit, but that didn't happen.

PR #1697, opened 2023-11-22
Chado Linker Analysis field

dsenalik:
# Tripal 4 Core Dev Task - Implement all the fields - Update "Analysis"

## Description
This PR follows the theme of PR #1431 which implemented the first "Linker" field, for Contact. Linker fields use the chado "Linking" tables that connect two different contact types, for example "Project" links to "Contact" through the `project_contact` table. 
This PR will add linking ability to the existing ChadoAnalysisDefault field.

Because the linking mechanism for each table will need much of the same code, several form fields and functions that were in the contact field have been moved to the base class, and annotations are used to tell the base class what to do.

In `FieldType/ChadoAnalysisDefault.php` we add the annotation
`object_table = "analysis",`
which will provide a form selector and validation for the "linking method". This specifies the table that contains the links, and the column name e.g. "feature_analysis → analysis_id" - We must specify both, because in a few cases, the column name is not based on the table name. For example, for `arraydesign` the foreign key to contact is `manufacturer_id` . This select list can't be populated until the base table is selected, so an ajax callback handles that. All of this is in the parent class, so this one annotation does a lot.

Linking tables sometimes include additional columns, and often they have `not null` constraints, so linking field widgets must supply values. Here we just supply the "null" cvterm_id of 1 when type_id is needed, and the $delta value is used when a rank is needed. If there are existing values, they should pass through unharmed. Ranks don't have a unique constraint, either.
In `FieldWidget/ChadoAnalysisWidgetDefault.php` we include these placeholder values.
If anybody wants the ability to specify these in the widget, that can be a future project 🙈.

The base formatter class now supports an option for a "token string" to configure how a field value is displayed. A site admin may want to display more than one column from a table, such as was done with the contact field, you can display the contact name as well as the contact type.
In `FieldFormatter/ChadoAnalysisFormatterDefault.php` we have another annotation that configures which columns in the analysis table can be displayed. This triggers inclusion of the base class' display settings and token string form field and validation.
```
 *   valid_tokens = {
 *     "[name]",
 *     "[description]",
 *     "[program]",
 *     "[programversion]",
 *     "[algorithm]",
 *     "[sourcename]",
 *     "[sourceversion]",
 *     "[sourceuri]",
 *   },
```

### Linker tables for analysis in core Tripal
 * analysisfeature (has several extra columns!)
 * project_analysis

### Linker tables for contact in core Tripal
 * feature_contact
 * featuremap_contact
 * library_contact
 * nd_experiment_contact
 * project_contact

## Testing
1. Make a docker on this branch
2. Go to `/admin/structure/bio_data/import-collection` and import all of the type collections
3. Run the job `docker exec CONTAINER drush trp-run-jobs --username=drupaladmin`
4. Create two contacts
5. Create two analyses
6. Manage fields on the project content type `/admin/structure/bio_data/manage/project/fields` and click on `+ Create a new field`
7. Select "Chado Analysis" and give it a unique machine name
add-project_analysis-1
8. On the next screen select the base table "Project" - this should trigger the ajax callback
add-project_analysis-2
9. The second select will only have one value, since there is only one linking table. Set the **cardinality to unlimited** and continue.
add-project_analysis-3
10. Enter something for the help text and save the field.
11. Remove the existing contact field and re-add it if you would like to test that one too, since it has had some changes in this PR
12. Now let's try to change the display settings. On the project content type manage display `/admin/structure/bio_data/manage/project/display` and you should see cogs on both contact and analysis
settings-1
13. Change the Analysis display settings, click update, then save.
settings-2
14. Try to enter an invalid token string, or none at all, to check validation.
settings-3
15. Now create a new project, and include both analyses and contacts, and confirm display is as expected. Confirm they are stored in chado `select * from chado.project_analysis;` and `select * from chado.project_contact;`
16. Edit and delete a project and a contact (change them to `-- Select --`), save, and confirm that they are gone in chado too `select * from chado.project_analysis;` and `select * from chado.project_contact;`


dsenalik:
A note about commit https://github.com/tripal/tripal/pull/1697/commits/c7263899f6d5f64ec7e89e2df9bf931cce1abc04
This was to handle the case where we have the field being used in a "single hop" manner, but the base table column name does not match the linked table. This happens for the `arraydesign` table, where the column `manufacturer_id` is a foreign key for the `cvterm` table `cvterm_id`. The update fixes the form when selecting the linking method, and updates the base table lookup in `getBaseTables()` to remove the assumption that the column names are always the same, now we just look at foreign keys. You can now add the contact field through the GUI on the arraydesign content type.

spficklin (approved):

spficklin:
Nice job @dsenalik .  I ran all of the tests. It works great.  I reviewed the code and those look good and @laceysanderson approved of them as well in the codefest this morning.  

I approve for merging.  The only thing I noticed is that there is an Analysis field on the Analysis content type which is named 'project_analysis'. I think that field is supposed to go on the Project content type, not the Analysis content type.

dsenalik:
Nice catch @spficklin  I will fix that!

PR #1704, opened 2023-12-02
Update or add all remaining linking fields from issue #1381

dsenalik:
# Tripal 4 Core Dev Task

## Description
This pull request will upgrade or add all the remaining fields mentioned in https://github.com/tripal/tripal/issues/1381#issuecomment-1823789804

These fields will include "linker capabilities" so that the field can be used directly on a base table, or connected through a linking table.

~~This is dependent on PR #1697~~ (now merged) which adds Analysis and updates Contact and the base class.
~~This is dependent on PR #1710~~ (now merged) which provides the ChadoDbxrefDefault field.

~~This will include a fix for issue #1703, which can't be a separate PR because we need the fields first to test the fix! These are the changes in `tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php`~~ This was reverted and #1729 fixed this instead.

This is a draft until I finish the yaml additions

## Testing?
1. Build a docker on this branch
```
port=8080
git clone https://github.com/tripal/tripal.git testing-1381
cd testing-1381
git checkout tv4g1-issue1381-chado_linker_remaining_fields
docker build --tag=tripaldocker:testing-1381 --build-arg drupalversion="10.1.x-dev" --build-arg chadoschema='chado' --file tripaldocker/Dockerfile-php8.2-pgsql13 ./
docker run --publish=${port}:80 --name=1381 -tid --volume=$(pwd):/var/www/drupal/web/modules/contrib/tripal tripalproject/tripaldocker:latest
docker exec testing-1381 service postgresql start
```
2. Create some content for testing: Organism, Contact, Publication. Make sure to put unique content in each of the publication fields to test them all. For ELocation try something with a `https:` prefix to make sure it displays as a link

3. Create a Study, and specify both contact and publication when you create it. This tests ~~the fix for issue #1703~~ the current version of chado storage - there should  be no error.
4. Import all the other content types: Page Structure ->  + Import Type Collection -> Check all boxes and click "Import"
`drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
There should of course be no errors.
5. Create an Array Design, then create an Assay using that Array Design
6. ...

PR #1707, opened 2023-12-06
Make zero a valid integer for ChadoIntegerType field when it is required

### LINKED ISSUE #1706: Is zero a number? (ChadoIntegerType field)
### Tripal Version

4.x

### General Description

If you have a ChadoIntegerType field and make it required, you cannot enter zero as a value

A fix for this is needed for PR #1704 to put the `is_obsolete` field on DNA Library content type, which has a not null constraint. In the `library` table it is integer instead of boolean 😠

### Steps to reproduce

To test you need to add an `ChadoIntegerType` field to the DNA Library content type for the `is_obsolete` column, and make it required. Then create a new library and try to enter zero.

### Screenshots

obsolete_zero_1
obsolete_zero_2

dsenalik:
# Bug Fix

## Description
If you have a ChadoIntegerType field and make it required, you cannot enter zero as a value

A fix for this is needed for PR #1704 to put the `is_obsolete` field on DNA Library content type, which has a not null constraint. In the `library` table it is integer instead of boolean 😠

### Steps to reproduce

To test you need to add an `ChadoIntegerType` field to the DNA Library content type for the `is_obsolete` column, and make it required. Then create a new library and try to enter zero. With this PR it should allow it to save.

### Screenshots

obsolete_zero_1
obsolete_zero_2


laceysanderson (approved):
✅ Code review: Beautiful simple fix!
✅ Manual Test: works perfectly!

Issue #1703, opened 2023-12-02
Chado storage again - The woes of too many joins


dsenalik:
### Tripal Version

4.x

### Drupal Version

10.1

### PostgreSQL Version

13

### PHP Version

8.2

### General Description

Chado storage is failing if we have too many joins, they may get added
to the SQL out of order. My example is I am working on a new field for
pub, and putting it on the study content type. This already has a
contact field, and both contact and pub have a join to the cvterm table.
Here is the sql generated by chado storage:
```
SELECT
"ct"."pub_id" AS "pub_id",
"ct"."contact_id" AS "contact_id",
"ct"."dbxref_id" AS "dbxref_id", 
"ct"."description" AS "description",
"ct"."name" AS "name",
"jpub0"."title" AS "jpub0_pub_title", 
"jpub0"."volumetitle" AS "jpub0_pub_volumetitle",
"jpub0"."volume" AS "jpub0_pub_volume", 
"jpub0"."series_name" AS "jpub0_pub_series_name",
"jpub0"."issue" AS "jpub0_pub_issue", 
"jpub0"."pyear" AS "jpub0_pub_pyear", 
"jpub0"."pages" AS "jpub0_pub_pages", 
"jpub0"."miniref" AS "jpub0_pub_miniref",
"jpub0"."uniquename" AS "jpub0_pub_uniquename", 
"jpub0"."is_obsolete" AS "jpub0_pub_is_obsolete", 
"jpub0"."publisher" AS "jpub0_pub_publisher", 
"jpub0"."pubplace" AS "jpub0_pub_pubplace",
"jcvterm0"."name" AS "jcvterm0_pub_type", 
"jcvterm1"."name" AS "jcvterm1_contact_type", 
"jcontact0"."name" AS "jcontact0_contact_name", 
"jcontact0"."description" AS "jcontact0_contact_description"
FROM chado.study "ct"
LEFT OUTER JOIN chado.pub "jpub0" ON ct.pub_id=jpub0.pub_id 
LEFT OUTER JOIN chado.cvterm "jcvterm0" ON jpub0.type_id=jcvterm0.cvterm_id 
   LEFT OUTER JOIN chado.cvterm "jcvterm1" ON jcontact0.type_id=jcvterm1.cvterm_id 
   LEFT OUTER JOIN chado.contact "jcontact0" ON ct.contact_id=jcontact0.contact_id
WHERE "ct"."study_id" = 1;
```
[...]

PR #1710, opened 2023-12-12
Finally, the Database Cross Reference Field!

### LINKED ISSUE #1550: Tripal 4: Implement some of the missing core fields and add to content types

dsenalik:
# Tripal 4 Core Dev Task

~~### DEPENDANT ON PR #1729.~~
~~Do not merge until PR #1729 has been merged. This PR does contain everything you need to fully test it with the new changes now though~~ ❤️ 
This has been merged!

## Description
This PR adds a Database Cross Reference field, corresponding to `sbo__database_cross_reference` in Tripal 3 

### Single-hop testing
1. ~~Add the "Chado Database Cross Reference" field to the "Study" content type, cardinality 1~~ (I added this to the yaml so it will already be there)
1. Create a new study. You can use "1" for publication_id, the null publication, and "-- Unknown --" for the contact to populate these not-null fields.
1. Select any database, I selected "NCBItaxon". The ajax should trigger
dbxref_1

1. After the ajax is done, enter an accession, I used "79200"
1. Save
dbxref_2

1. Verify it is in the study database table
```
sitedb=> select * from study;
 study_id | contact_id | pub_id | dbxref_id |   name    | description 
----------+------------+--------+-----------+-----------+-------------
        1 |          1 |      1 |      3460 | Study One | D1
(1 row)
```
7. Verify that the link works
8. Edit and remove the accession and save. The cross reference should be gone now, and also gone from the study table (it will remain in the dbxref table)
```
sitedb=> select * from study;
 study_id | contact_id | pub_id | dbxref_id |   name    | description 
----------+------------+--------+-----------+-----------+-------------
        1 |          1 |      1 |           | Study One | D1
(1 row)
```
### Double-hop testing
1. ~~Add the "Chado Database Cross Reference" field to the "Organism" content type, cardinality unlimited~~ (I added this to the yaml so it will already be there)
3. Create a new organism.
4. Under "Database Cross Reference" click "add another item", we will add two
5. Select "NCBItaxon" again and type in a "7", autocomplete should suggest the value we added earlier.
6. For the second one, select "PMID" as the database and enter a "1" for accession, the autocomplete should suggest `PMID:16027110`
dbxref_3

7. Save
dbxref_4

8. Verify the links work
9. Delete one of the accessions, save, and verify that it is gone.

### Combined Single-hop and Double-hop testing
1. Test on any feature, stock, or biomaterial content type e.g. "Germplasm Accession"
20240108_bothtypes

### List of affected content types, test on any of these
These tables have a dbxref_id column and will have a "single-hop" instance of this field
`arraydesign`
`assay`
`phylotree`
`protocol`
`study`
`element` (not a default content type)

These tables have a linker table and will have a "double-hop" instance of this field
`analysis`
`biomaterial` *
`feature` *
`featuremap`
`library`
`organism`
`pub` (not added because pub is not yet in the yaml configuration)
`project`
`stock` *
`cell_line` (not a default content type)
`nd_experiment` (not a default content type)
`phylonode` (not a default content type)

* These three tables also have a `dbxref_id` column, ~~this is not supported with the current yaml configuration, the linker table is used instead. See comment https://github.com/tripal/tripal/pull/1710#issuecomment-1852687600~~ so there will be two instances of this field, one for each.


dsenalik:
A note, there are three cases where a base table contains a `dbxref_id` column, but there is also a linker table. In these cases, in the yaml I added the linker table version, because cardinality is unlimited there. In the case someone wants to use the base table column, they will have to delete the field and re-add it so that it references the base table.
These tables are:
biomaterial & biomaterial_dbxref
feature & feature_dbxref
stock & stock_dbxref

laceysanderson (commented):
Manual testing
✅ Study
☑️ Array Design -field worked but bug #1703 😒 
☑️ Assay -widget worked but couldn't save since missing arraydesign_id and operator_id
☑️ Phylotree -field worked but bug #1703 😒 
☑️ Protocol -field worked but bug #1703 😒 

Ok, we're going to have to hold off review of this PR until after #1719/#1703 are fixed :-( Boo!

The widget looks great though! Eventually I would like to see settings for admin to indicate when they don't want a term to be created if one doesn't exist. 

Also, we will want to support both the stock.dbxref_id and the stock_dbxref linker 🤔 Was the problem with the field not working with multiple instances on the same content type or was there a new problem in the YAML that I'm unaware of?

dsenalik:
1. I found a bug in the widget, if autocomplete is bypassed, then the DB limit was also, and any matching accession will be selected from any DB. Fixed with commit https://github.com/tripal/tripal/pull/1710/commits/46db52eeeef072333f3530bfd9cd3451f56c6ac1
2. If you have more than one dbxref field, then the ajax doesn't work due to duplicated form `'#edit-accession-' . $delta`. Fixed with commit https://github.com/tripal/tripal/pull/1710/commits/f5b24381482bc10a07e0bcbf1e811ec0c172cc32
3. We cannot have two instances of this field with the same cv term on a given content type. The normal term is `SBO:0000554` = `database cross reference`, I propose using `SBO:0000552` = `reference annotation` for the linker table version when both are present. This is the parent term. See commit https://github.com/tripal/tripal/pull/1710/commits/351c6c1c18ff04cdea043c7b2b0b073f83120afe
4. ~~This PR currently adds both fields to only the four `stock` content types with commit https://github.com/tripal/tripal/pull/1710/commits/351c6c1c18ff04cdea043c7b2b0b073f83120afe.~~ This PR adds the field everywhere it is appropriate. Example of appearance
different-terms-works

dsenalik:
From discussion with @laceysanderson we will be using the following controlled vocabulary terms:
 * For content based on a base table containing a `dbxref_id` column, we will use the same term as Tripal 3, [SBO:0000554](https://www.ebi.ac.uk/ols4/ontologies/sbo/classes/http%253A%252F%252Fbiomodels.net%252FSBO%252FSBO_0000554?lang=en) = `database cross reference`
 * For content using a linking table we will use the parent ontology term [SBO:0000552](https://www.ebi.ac.uk/ols4/ontologies/sbo/classes/http%253A%252F%252Fbiomodels.net%252FSBO%252FSBO_0000552?lang=en) = `reference annotation`

In this way, content types that have both types of database cross references can have two fields and support both (_i.e._ feature, stock, biomaterial)

Example on a "Gene" content type:
20240103_display

dsenalik:
~~I have copied the draft chado storage fix from PR #1704 to allow testing of this PR, these changes are in commit https://github.com/tripal/tripal/pull/1710/commits/c844d6ef819c7db1257c95828549dce2efed14c2~~

~~This may not be the proper fix, but it does allow testing.~~
No longer needed with merged PR #1729

dsenalik:
Some explanation of how I propose to add linking fields to the yaml when we are linking between content types in two different optional groups. A separate **Tripal Entity Type Collection** is created for each pairing, for example
20240104_linkingfields
These collections do not create any new content types, rather just the appropriate linking fields.

laceysanderson:
Updated this PR to use and include the changes in PR #1729 but have not gotten the chance to test it yet. Thanks ot the work @spficklin did with the new `_v2` branch and me removing all changes in this PR to chadostorage there were no merge conflicts.

## Now Dependant on PR #1729 being merged first!

dsenalik:
LOL I was doing the same thing, I will compare our changes...
```
git diff origin/tv4g1-issue1550-dbxref-field:tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php
diff --git a/tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php b/tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php
index 2c9f6ca6a..24493a524 100644
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoDbxrefTypeDefault.php
@@ -168,7 +168,7 @@ class ChadoDbxrefTypeDefault extends ChadoFieldItemBase {
         'action' => 'store_pkey',
         'drupal_store' => TRUE,
-        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $linker_pkey_col,
+        'path' => $linker_table . '.' . $linker_pkey_col,
       ]);
 
@@ -182,7 +182,7 @@ class ChadoDbxrefTypeDefault extends ChadoFieldItemBase {
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, self::$object_id, $linker_fkey_term, [
         'action' => 'store',
         'drupal_store' => TRUE,
-        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $linker_fkey_col,
+        'path' => $linker_table . '.' . $linker_fkey_col,
         'delete_if_empty' => TRUE,
         'empty_value' => 0,
       ]);
@@ -194,7 +194,7 @@ class ChadoDbxrefTypeDefault extends ChadoFieldItemBase {
         $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_' . $column, $term, [
           'action' => 'store',
           'drupal_store' => FALSE,
-          'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $column,
+          'path' => $linker_table . '.' . $column,
           'as' => 'linker_' . $column,
         ]);
       }

```

dsenalik:
Tested with current branch and all appears good.
But now noticing that the autocomplete does not work with a single zero `0` but does work with two `00`, _e.g._ for the `OBI` or `EFO` ontology accessions.
Fixed with commit https://github.com/tripal/tripal/pull/1710/commits/b2cf43d8dec36581d9c5ae99f608b6515a39e9f4

laceysanderson (commented):
Testing complete for the general content types and all looks good so far!
| Content Type | base.dbxref_id | base_dbxref | Create | Edit | Link |
|---|---|---|---|---|---|
| organism | n/a | ✅ | ✅ | ✅ | ✅ |
| analysis | n/a | ✅ | ✅ | ✅ | ✅ |
| contact | n/a | n/a |  |  |  |
| project | n/a | ✅ | ✅ | ✅ | ✅ |
| protocol | ✅ 🔘 | n/a | ✅ | ✅ | ✅ |
| publication | n/a | ✅ | ✅ | ✅ | ✅ |
| study | ✅ 🔘 | n/a | ✅ | ✅ | ✅ |

For protocol and study the primary database reference is correctly set as optional :-) In all cases I was able to create content with/without the field filled in, edit it, rearrange if there were multiple values supported and the rendering was perfect as well.

laceysanderson (approved):
For the remaining content types, I checked the edit form for every single one and confirmed the right type of dbxref field was added as expected ✅ 

I also created a QTL, BioSample and Breeding Cross to confirm that it works when version of these fields are attached and it worked great ✅ 

This field feels beautifully solid. Thanks @dsenalik! Ready To Merge!
Screenshot 2024-01-08 at 4 17 26 PM

dsenalik:
Wow! My first stamp of approval! 😍

PR #1721, opened 2024-01-02
Create a Github Workflow to Sync Drupal.org on every push to 4.x

### LINKED ISSUE #1720: Workflow to keep github and drupal.org in sync
### Task Description

When making a release, we need Drupal.org to be in sync with github. If they diverge too far it causes issues as drupal.org takes a long time to see the updates and sometimes needs intervention from the drupal.org staff.

This PR will create a github workflow that automatically pushes to drupal.org whenever there are updates to the 4.x branch.

### Branch Name

tv4g0-issue1720-workflowSyncDrupalOrg

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

When making a release, we need Drupal.org to be in sync with github. If they diverge too far it causes issues as drupal.org takes a long time to see the updates and sometimes needs intervention from the drupal.org staff.

This PR creates a github workflow that automatically pushes to drupal.org whenever there are updates to the 4.x branch. It can also be triggered manually for specific branches.

It uses my drupal login and a Github secret with my access token. By using a github secret it ensures that no one has access to my token. Furthermore, I use the git credentials helper for an extra level of security as it ensures my password is not encoded in the remote URL.

## Testing?
I have tested these exact commands locally and it successfully updated [Drupal.org GitLab](https://git.drupalcode.org/project/tripal/-/tree/4.x?ref_type=heads). 

Unfortunately we cannot test this github workflow directly until it is in the default branch... So a quick code review will be sufficient.

### DIFF
--- a/.github/workflows/MAIN-syncDrupalOrg
+++ b/.github/workflows/MAIN-syncDrupalOrg
@@ -1,9 +1,9 @@
 name: Push Recent Commits to Drupal.org
 on:
+  workflow_dispatch:
   push:
     branches:
       - 4.x
-      - tv4g0-issue1720-workflowSyncDrupalOrg
 
 jobs:
   push_to_drupal:
--- a/.github/workflows/MAIN-syncDrupalOrg
+++ b/.github/workflows/MAIN-syncDrupalOrg
@@ -1,0 +1,19 @@
+name: Push Recent Commits to Drupal.org
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
I see nothing wrong in the code review 🙈

laceysanderson:
Great! Thanks for the quick review <3 I'm going to merge and see if it performs as hoped 🙏

PR #1727, opened 2024-01-04
Cleanup Terms: move all chado_table focused local cvs into the main 'local' cv.

laceysanderson:
# Tripal 4 Core Dev Task

## Description

At some point in the history of Tripal, we created a number of chado table focused cvs (e.g. feature_property). All terms in these cvs had dbxref's in the 'local' db. This bloats the cv table a bit and gives more credibility to these terms then they deserve as they were created for convenience and to fill gaps in ontologies.

This PR moves all those terms into the `local` cv and db since that is exactly what they are.

## Testing?

This PR needed to update the chado data dump, so part of testing is ensuring that there is a ✅ beside the automated testing ;-p 

Additionally, the chado table cvs were used in the following, so a quick run of these functionalities would be prudent.

- Taxonomy Importer
- GFF3 Importer
- Genetic Map Unit field widget: test by creating a genetic map and choosing a unit.


dsenalik:
One fix needed for phylotree, if you want I could add this:
```
diff --git a/tripal_chado/src/api/tripal_chado.phylotree.api.php b/tripal_chado/src/api/tripal_chado.phylotree.api.php
index 825306d48..a1bfa2597 100644
--- a/tripal_chado/src/api/tripal_chado.phylotree.api.php
+++ b/tripal_chado/src/api/tripal_chado.phylotree.api.php
@@ -971,14 +971,14 @@ function chado_phylogeny_get_node_types_vocab($options, $schema_name = 'chado')
     $values = [
       'name' => $name,
       'cv_id' => [
-        'name' => 'tripal_phylogeny',
+        'name' => 'local',
       ],
     ];
     $cvterm = chado_generate_var('cvterm', $values, [], $schema_name);
     if (!$cvterm) {
       \Drupal::service('tripal.logger')->error(
         "Could not find the leaf vocabulary term: '%name'. It should " .
-        "already be present as part of the tripal_phylogeny vocabulary.",
+        "already be present as part of the local vocabulary.",
         ['%name' => $name], $options['message_opts']);
       return FALSE;
     }
```


Preparer test fixes
```
diff --git a/tripal_chado/tests/src/Functional/Task/ChadoPreparerTest.php b/tripal_chado/tests/src/Functional/Task/ChadoPreparerTest.php
index f74530685..5ed407cc8 100644
--- a/tripal_chado/tests/src/Functional/Task/ChadoPreparerTest.php
+++ b/tripal_chado/tests/src/Functional/Task/ChadoPreparerTest.php
@@ -100,8 +100,8 @@ class ChadoPreparerTest extends ChadoTestBrowserBase {
     // 3: IMPORT ONTOLOGIES.
     // --------------------------------
     $expected_counts_by_table = [
-      'cv' => 60,
-      'db' => 43,
+      'cv' => 31,
+      'db' => 41,
       'cvterm' => 3145,
       'dbxref' => 3454,
     ];
@@ -112,7 +112,7 @@ class ChadoPreparerTest extends ChadoTestBrowserBase {
     }
 
     // Check for some specific cv / db which should have been inserted.
-    $cv_found = $chado2check->query("SELECT 1 FROM {1:cv} WHERE name = 'feature_property'")->fetchField();
+    $cv_found = $chado2check->query("SELECT 1 FROM {1:cv} WHERE name = 'tripal_contact'")->fetchField();
     $this->assertEquals(1, $cv_found, 'Found feature_property CV');
     $db_found = $chado2check->query("SELECT 1 FROM {1:db} WHERE name = 'TAXRANK';")->fetchField();
     $this->assertEquals(1, $db_found, 'Found TAXRANK DB');
@@ -130,7 +130,7 @@ class ChadoPreparerTest extends ChadoTestBrowserBase {
     // 5: POPULATE DB2CV_MVIEW.
     // --------------------------------
     $table_name = 'db2cv_mview';
-    $expected_count = 41;
+    $expected_count = 28;
     $count = $chado2check->query("SELECT count(*) FROM {1:$table_name}")->fetchField();
     $this->assertGreaterThanOrEqual($expected_count, $count,
       "There was not the expected number of records in the $table_name table after preparing.");
```

dsenalik (changes requested):
This looks good now!
My only request, IF APPROPRIATE, is to revert most of https://github.com/tripal/tripal/pull/1727/commits/d6e8e2709467cd680b45efda11aa999915ce5d3f by removing all the new lines where we set the value to `1` because that is the default and so these lines don't do anything.

laceysanderson:
Thanks for the review @dsenalik!

I removed the lines with `setval(1, false)` as that is the default and the tests still pass so 🤞 and updated the setval for contact and pub to `setval(1,true)` as there is a single value and the true ensures it will be incremented on next insert.

This should be ready to go now :-)

dsenalik (approved):
* Automated testing has passed
 * There is something unrelated wrong with the taxonomy importer, it is asking for a gff3 file. See Issue #1733 . (Update-fixed by PR #1737) But ignoring that, it appeared to work. My fake organism showed up as a warning (as I would expect) ` [warning] WARNING: The following 1 organisms do not have an NCBI taxonomy ID, and have not been included in the tree: "Tripalus bogusii"`
 * Imported the `TripalusDatabasicaChr1Genes.gff3` file with only this unrelated warning, noting for later
```
 [notice] Step 14 of 27: Add child-parent relationships...                   
Percent complete: 0%. Memory: 35,015,288 bytes.
 [notice] Percent complete: 0%. Memory: 35,015,288 bytes.
 [warning] Undefined array key "AT1G01010" GFF3Importer.php:2280
 [warning] Trying to access array offset on value of type null GFF3Importer.php:2280
```
and then I could publish Gene and mRNA content.
 * No problems with genetic map, could add and then change units.

Issue #1716, opened 2023-12-18
Trying to publish 2+ content types in the same call to the Tripal job launcher fails


laceysanderson:
### Tripal Version

4.x

### Drupal Version

10.1.x-dev

### PostgreSQL Version

13

### PHP Version

8.2

### General Description

**This one is hard to describe generally so look at the steps to reproduce to get a better idea.**

I'm seeing confusing but repeatable errors when publishing content BUT ONLY if I get impatient and submit 2+ publishing jobs before running them.

## Single publish job per call to job launcher works
1. Run tripal jobs to see that the queue is empty.
2. Go to Admin > Tripal > Content > Publish and choose a random content type with unpublished data.
3. Immediately run the tripal job to complete the publishing and then confirm.
4. Repeat above steps exactly and confirm all general content types publish perfectly.

## When multiple publish jobs are in the queue, the second one fails.
**From a fresh install of Tripal or a fresh docker container**
1. Run tripal jobs to see that the queue is empty.
2. Go to Admin > Tripal > Content > Publish and choose a random content type with unpublished data.
3. WITHOUT RUNNING THE JOB LAUNCHER, repeat step 2 for another content type.
4. Now run the tripal jobs and confirm you see a big error and the second content type, regardless of what you chose, failed because it tried to publish data from fields for the first content type.

### Error messages

The following error message was obtained following the steps to reproduce **when multiple publish jobs are in the queue, the second one fails.**

Content types chosen: 1) contact, 2) project.

```
tripal-4x on  4.x via 🐳 desktop-linux via 🐘 v8.1.6 
❯ docker exec -it $containerName drush tripal:trp-run-jobs --username=drupaladmin

2023-12-18 18:21:31
Tripal Job Launcher
Running as user 'drupaladmin'
-------------------
2023-12-18 18:21:31: Job ID 3.
2023-12-18 18:21:31: Calling: tripal_publish(contact, chado_storage, Array)
Step  1 of 6: Find matching records... 
 [notice] Step  1 of 6: Find matching records... 
Step  2 of 6: Generate page titles...
 [notice] Step  2 of 6: Generate page titles...
Step  3 of 6: Find existing published entities...
 [notice] Step  3 of 6: Find existing published entities...
Step  4 of 6: Publishing 7 new entities...
 [notice] Step  4 of 6: Publishing 7 new entities...
Step  5 of 6: Find IDs of entities...
 [notice] Step  5 of 6: Find IDs of entities...
[...]

PR #1729, opened 2024-01-06
Updates to ChadoStorage -- path argument and cleanup.

### LINKED ISSUE #1719: Support Joins for Finding Values
### Task Description

Our current approach in fields for modeling the joins needed for inserting, updating and selecting all field values when the primary key of a base table record is known works well. However, this approach is insufficient for finding records when values in non-base tables are needed to find records in base tables. 

## Join Use Cases
Currently we support joins in fields using two different methods:
1.	CASE 1: The `store_link` action.   A join is specified using the `left_table`, `left_table_id`, `right_table` and `right_table_id` property storage settings.   The `ChadoStorage` plugin uses these values to create a “path” that is passed to the `ChadoStorage::addChadoRecordJoins()` function.
2.	CASE 2: The `read_value` action.  A join is specified using the `path` property storage settings. This path is passed directly to the `ChadoStorage::addChaodRecordJoins()` function.

## Review of how joins are currently handled.
Both CASE 1 and CASE 2 end up using the `ChadoStorage::addChadoRecordJoins()` to add joins. The code from that function that adds the join to the records object is as follows:

https://github.com/tripal/tripal/blob/6779fccbf67d58950b6186eb9c82f990d48d33bd/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php#L1714-L1727 

Note that aliases for tables in a join are given a numeric suffix based on the order in which a table is added.  If a table is only added once to a join then the suffix is always 0. But the second time, it will have a suffix of 1.  This can happen for tables such as `featureloc` because it has a `feature_id` and `srcfeature_id` column that both join to the `feature` table.  

For insert, update or selecting of a single record these aliases don’t matter because we also give an alias to the column we want via the `as` property in the `read_value` action.  So regardless of the table aliasing we can always pull out a unique column. 

Also, note that the “base” table in the joins is referred to as “ct” (an abbreviation for “chado table”, or “base table”).

## Challenges for finding values.
For finding values, we intend to find all base table records that match a set of criteria.  Theoretically, we should be able to find records using values set for any field property on the content type.  

### Publishing Use Case:
One use case for finding values, is in publishing. For example, the “Genome Assembly” content type requires that the `type_id` for an analysis base record be set in the `analysisprop` table.  So, to publish, we must first find all the analyses with an associated record in the analysisprop table.   If we look at the `ChadoAdditionalTypeTypeDefault` plugin class. We see that the join is specified via a property with a `store_link` action:
https://github.com/tripal/tripal/blob/6779fccbf67d58950b6186eb9c82f990d48d33bd/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php#L132-L136
The value of the `type_id` comes from a property with a `store` action.
https://github.com/tripal/tripal/blob/6779fccbf67d58950b6186eb9c82f990d48d33bd/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php#L145-L150 

#### Problem  1
First, the `link_id` field is missing the `left_table`, `left_table_id`, `right_table` and `right_table_id` properties. That’s okay, it just hasn’t been updated yet.  But let’s assume they are there to indicate how the table should join to the base table.  I’ve fixed this in my current working branch `tv4g2-issue1515-publish_chado_content`

#### Problem 2
The record array structure that ChadoStorage used for representing how to query cannot currently handle adding conditions on joined tables.  Here’s what it current looks like (in YAML format) for finding analyses:

```
analysis:
- conditions:
    analysis_id:
      value:
      - REPLACE_BASE_RECORD_ID
      - analysis
      operation: =
  fields:
    analysis_id: ~
    name: ~
  joins:
    analysisprop:
      analysis_id:
        "on":
          left_table: analysis
          left_col: analysis_id
          right_table: analysisprop
          right_col: analysis_id
          left_alias: analysis
          right_alias: analysisprop
        columns:
        - - analysis_id
          - link_id
          - genome_assembly_type
          - link_id
```
In order to filter (or find) base records that have conditions in joined tables we need to specify the table. In my current working branch this is how I changed the array:

```
analysis:
- conditions:
    analysis:
      analysis_id:
        value:
        - REPLACE_BASE_RECORD_ID
        - analysis
        operation: =
    analysisprop:
      type_id:
        value: "31"
        operation: =
  fields:
    analysis_id: ~
    name: ~
  joins:
    analysisprop:
      analysis_id:
        "on":
          left_table: analysis
          left_col: analysis_id
          right_table: analysisprop
          right_col: analysis_id
          left_alias: analysis
          right_alias: analysisprop
        columns:
        - - analysis_id
          - link_id
          - genome_assembly_type
          - link_id
```
Notice how the table name for the condition is now above the criteria.  This fix allows us to find analyses that are only “Genome Assembly” records and publish those!  It’s currently working using this approach.

#### Problem 3
The change to the records array structure to support conditions on non-base tables works because I also had to make a change in the `ChadoStorage::addChadoRecordJoins()` function.  In this use case, the `analysisprop` table is joined to the `analysis` table because the `link_id` property provides that join. The `type_id` property takes advantage of that join.  In the current `addChadoRecordJoins()` function (see code above) the join will create the alias `janalysisprop0` for the anaysisprop table.   Unfortunately, the `type_id` property doesn’t know what the alias is.   It would be unreasonable to expect the field developer to know this alias, and even less to know the order in which tables with the same name are joined. As long is there is only one instance of a table joined all is fine if the developer knows to use the fromat similar to `janalysisporp0` in a `chado_table_alias` property setting.  Since that did not seem reasonable, I changed the `ChadoStorage::addChadoRecordJoins()` to not add the `j` prefix nor the numeric suffix if the table was the first one.   The developer need not provide any additional settings. 

### Future Issues
In the process of adding fixes for the use case above I realized that we would continue to have problems when finding values because:

1.	Field properties rely on other properties to specify a join.  This is ambiguous.   For example, If we wanted to find a feature (via the `featureloc` table) with a given name (gene) located on another feature of a given name (e.g. chromosome) then we have to join twice on the feature table.  There is no way for the field properties to know which join of the feature table they should use to set a conditional value.  The `store_link` property could set an alias, but the feature `name` property doesn’t know which alias to use.  
2.	The `addChadoRecordJoins()` sets the table aliases.  This means it’s not possible for a field to know what the table alias is.  My solution for the publishing use case works as long as the table only occurs once in the joins.

### Possible Solution
Use a specific `path` attribute for all properties.  This would follow the same convention as the `path` setting does for the properties with a `read_value` action. One change is to add the column where the value comes from to the end of the path. Consider this for the `ChadoAdditionlaTypeTypeDefault` field

```php
if ($type_table != $base_table) {
      $type_table_def = $schema->getTableDef($type_table, ['format' => 'Drupal']);
      $type_pkey_col = $type_table_def['primary key'];
      $type_fkey_col = array_keys($type_table_def['foreign keys'][$base_table]['columns'])[0];
      $link_term = $mapping->getColumnTermId($type_table, $type_fkey_col);
      $value_term = $mapping->getColumnTermId($type_table, 'value');

      $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'prop_id', $record_id_term, [
        'action' => 'store_pkey',
        'drupal_store' => TRUE,
        'path' => $type_table . "." . $type_pkey_col,
        # 'chado_table' => $type_table,
        # 'chado_column' => $type_pkey_col,
      ]);
      $properties[] =  new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'link_id', $link_term, [
        'action' => 'store_link',
        'drupal_store' => TRUE,
        'path' => $base_table . '.' . $base_pkey_col . '>' . $type_table . '.' . $type_fkey_col,
        # 'left_table' => $base_table,
        # 'left_table_id' => $base_pkey_col,
        # 'right_table' => $type_table,
        # 'right_table_id' => $type_fkey_col,
      ]);
      $properties[] =  new ChadoTextStoragePropertyType($entity_type_id, self::$id, 'value', $value_term, [
        'action' => 'store',
        'path' => $type_table . '.value',
        # 'chado_table' => $type_table,
        # 'chado_column' => 'value',
      ]);
    }

    // We need to store the numeric cvterm ID for this field.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'type_id', $type_id_term, [
      'action' => 'store',
      'drupal_store' => TRUE,
      'path' => $base_table . '.' . $base_pkey_col . '>' . $type_table . '.' . $type_fkey_col . ';' . $type_column,,
      # 'chado_table' => $type_table,
      # 'chado_column' => $type_column,
      'empty_value' => 0
    ]);
    // This field needs the term name, idspace and accession for proper
    // display of the type.
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'term_name', $name_term, 128, [
      'action' => 'read_value',
      'path' => $type_table . '.' . $type_column . '>cvterm.cvterm_id;name',
      # 'chado_column' => 'name',
      'as' => 'term_name'
    ]);
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'id_space', $idspace_term, 128, [
      'action' => 'read_value',
      'path' => $type_table . '.' . $type_column . '>cvterm.cvterm_id;cvterm.dbxref_id>dbxref.dbxref_id;dbxref.db_id>db.db_id;name',
      #'chado_column' => 'name',
      'as' => 'idSpace'
    ]);
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'accession', $accession_term, 128, [
      'action' => 'read_value',
      'path' => $type_table. '.' . $type_column . '>cvterm.cvterm_id;cvterm.dbxref_id>dbxref.dbxref_id;accession',
      # 'chado_column' => 'accession',
      'as' => 'accession'
    ]);
```
Notice that the `type_id` property fully specifies where the value comes from... as well as all of the fields There is no ambiguity as to where it's value comes from.  Since the `type_id` and `link_id` paths are the same (other than the last column name), then it's clear that the vlaues from those two properties comes from the same join.

spficklin:
# Tripal 4 Core Dev Task

## Description

This PR fixes the changes with joining tables that were described in issue #1719. The fix is needed to make publishing work as we need better specification for joins in order to find records.   The solution proposed on that issue was to rework field properties so that they have a `path` element and a `table_alias_mapping` element that allows developers to fully specify table aliases on joins.  

In the process of adding in this functionality I took the time to fix multiple problems with the way ChadoStorage was handling keep track of how queries were to be handled.  So, this PR includes a major overhaul of the ChadoStorage and introduces a new ChadoRecords class. The ChadoRecords class keeps manages the "records" array that we were using before to keep track of how queries were handled. It provides better error checking, notification of errors, and a better structured array that is more intuitive and useful for debugging.  

@laceysanderson I did have to comment out two functional tests because I think they are buggy. I have makred them with a @todo and my initials (spf) so you can take a look and confirm or correct me.  

## Testing?

To test this PR, you will have to test use of every field. Here is the list of tests you can perform to do so.

1. Create two or more "analysis" content types.  
    1. Tests the `ChadoStorage::insertValues()` for single value fields
    2. Tests the `ChadoStraoge::loadValues()` for single value fields
    3. The following fields are tested:  , `ChadoStringTypeDefault` and `ChadoTextTypeDefault`.
2. Edit the analysis and change all of the values and save.
    1. Tests the `ChadoStorage::updateValues()` for single value fields 
4. Create at least two "organism" content types.
    1.  Repeats the insert and load tests of the analysis above, but you need an organism to create other conten types.
5. Create a two or more "contact" content types. We will use these later for tesing.
6. Create a "gene" and only fill all of the values.  Add as many contact as you want (be sure to add at least two). Add as many synonyms as you like.  Be sure to add at least two.  Be sure to add a sequence because we want to test that the md5checksum and length are calculated.  Be sure to check as least one of the checkboxes at the bottom for the form.
    1.  This tests the following additional fields:  `ChadoBooleanTypeDefault`, `ChadoAdditionalTypeTypeDefault`, `ChadoOrganismTypeDefault`, `ChadoSequenceChecksumTypeDefault`, `ChadoSequenceLengthTypeDefault`, `ChadoSynonymTypeDefault`
    2. Results should show that fields that have joins to another table work (e.g., the organism name).
    3. Results should show that fields with multiple tables in a join work (eg., the synonyms).
    4. Results should show that the md5checksum and sequence length are calculated.
    5. Results should show that fields with multiple cardinality (e.g. contacts and synonyms) store and display all of the values.
7.  Edit the field and change as many values as you like.  Be sure to change the order of the contacts and the synonym.  After saving all of your changes should be visible.
8. Add two new properties to the Gene content type via Admin > Structure > Tripoal Content Types.  Select "Manage Fields" for the "Gene" Content type and click the button to "Create a new field".  Select to create a new "Chado Property".  Create one property with the label of "Note" and another with the label "Comment". Select a controlled vocabulary term for each that is appropiate but different.  Make sure the base table for both is the `feature` table.
9. Return to the gene and edit it.  You should now see to new "Note" and "Comment" fields. Add as many values to both of those.  Be sure to add at least two to each of them. Save the entity.
    1. This tests these additional fields:  `ChadoLinkerPropertyTypeDefault`
    2. Results should show that you can add as many propery values as you like.  You should be able to reorder those while editing as well.
    3. Results should show that you can have multiple properties on the same content type, for the same table  and the do not intefere with one another.
11. Create an "array design" content type.    Save and then edit it at as you like.
      1.  Results should show that you can have two fields both with joins to the same table (e.g. cvterm), via the substrate type and platform type fields, without interferring with one another.
13. Next we will test publishing.  Currently, ChadoStorage supports publishing of two types of content types: those where all fields from a single table are for the same type (e.g. organism), and those where the type is set in a `type_id` column of the base table (e.g. feature, stock).   Go to the Tripal Content List page and delete all of your content types you made above.
14.  Once removed, click the "Publish Tripal Content" button and select "Organism" as the type to publish.  Run the Tripal job. It should restore the two or more organisms that you created previously.
       1. This tests the `ChadoStorage::findValues()` function.
       2. Results should show that you can publish values where the Chado table represents a single content type.
15. Do the same for the gene content type.
      1. Results should show that you can publish values where Chado table distinguishes between contnet  types using the `type_id` column in the base table.  
17.


spficklin:
A note:  this PR uses the branch `tv4g1-issue1719-usePathForAllActions-v2` because merging in 4.x on my first attempt was a disaster.  So the `tv4g1-issue1719-usePathForAllActions` is borked and should be ignored going forward.

spficklin:
Also, I think we should NOT merge any other PRs into the 4.x branch before this one.  The mishap where the `tv4g1-issue1719-usePathForAllActions` was accidentally merged into 4.x and then files were reverted back to restore it, created challenges with merging 4.x back into this branch. If we don't get this merged before anything else then I'm worried I'll have to repeat fixing the merge conflicts again. It wasn't fun to fix.

laceysanderson:
I've added some updates to the tests that Stephen mentioned having to comment out which fixes them. All the original tests are now being run and passing. Specifically,

- testStorePKeyAction: there is a data provider that tests table alias' for the store_pkey action. There was a mistake in the field definition that got introduced at an unknown point. Essentially, the base_table for the field needs to be different from the table the store_pkey is on (in this case cv vs. db) and no path between the two tables need to exist.
- testStoreLinkAction: there were a couple of bugs introduced into this test at some point that are now fixed. I've made the expected_values array more clear and more complete and set the project record_id for the update.

dsenalik (approved):
I merged this branch into my dbxref field #1710 and tested including the dbxref field.
It took a while but everything worked as expected, including dbxrefs.
A minor thing, when publishing, if you run two publish jobs at the same time, the second one fails. This is almost certainly the same bug as #1716 so will not affect the approval of this PR.

spficklin:
Hey @dsenalik .... just curious what do you mean it "took awhile"...

risharde:
Sorry for the delay, I have an issue which may not be related to your branch @spficklin 
I just did a reinstall of Tripal V4 under your branch, I created a chado instance and ran the prepare step
I don't see the Tripal Content Types appearing - could this be because I'm using Drupal 9.x ?
Confirming the branch is up-to-date on my side
Confirming that I ran drush cr to clear cache - did not work
image

dsenalik:
@spficklin just all those testing steps LOL

spficklin:
Hey @risharde .  The content type creation was moved out of the prepare step.  To add them, go to Admin > Structure > Tripal Content Types > Click the "import type collection" button.  On the page that appears select all of the content types and click the button. That will add a job.  Unfortunately it won't give you the command-line to run it but you should be able to use the last one you did and just update the job number to the next one.

dsenalik:
@risharde if you are not using the docker, you will have to first create the content types by going to "Page Structure" -> "+ Import type collection"

risharde:
@spficklin @dsenalik thanks guys, worked like a charm!

risharde:
1. Works perfectly! Encountered absolutely no issues - no errors in reports!
2. Works perfectly! No errors in reports as well!
3. Works perfectly! No errors in reports as well!
4. Works perfectly (as expected!) 
5. Works perfectly! (displayed correctly signifying joins worked), md5 checksums generated, length created, checkboxes store as True etc!
6. Works perfectly! (order changes were seen, other changes were seen)
7. Works perfectly!
8. Works great! (Reordering could be seen, edits happened!) Note: Saving was noticeably slower than other task tests so I assume many operations are happening in the background to make this feature work, just a note!
9. Works perfectly! (Creation and editing could be seen!)
10. Worked perfectly! (Items were deleted)
11. Worked perfectly! (This was quite fast! Organisms were added back to the UI!)
12. @spficklin Semi-works (content was recreated but multi-value fields were not fully populated example there was only 1 contact recreated, 1 comment recreated, 1 note recreated even though the previous content added contained 2 contacts, 3 comments and 3 notes.

This was a super exciting experience to see this part of Tripal working! @spficklin I'm not sure if you are able to test this on your side to confirm if #12 is something worth looking into?

dsenalik:
You paid better attention than I did @risharde - I confirm that a gene with two contacts, deleted, and published now only shows one contact. 
20240106_gene_contact

I'll "quick" build a 4.x docker and see if it is the same there...

dsenalik:
Results: Under 4.x the contact field is not yet supported.
20240106_4 x-publish

spficklin:
Hey @risharde.  I confirm that test number 12 has a problem.  I'll check it.
@dsenalik I can't reproduce that error. How did you generate it?

spficklin:
Okay, problems with test number 12 are fixed!  it was actually a deeper problem to fix.  The TripalPublish class didn't support fields with cardinality > 1 so when I asked you all to test it I had forgotten I hadn't fixed that and in my head I thought I had.  So, it's now fixed and that test should work.  

I also had to fix the Chado Property field because it wasn't setting the value for the `type_id` property which caused a problem with publishing.  It now sets it similar to the Additional Type field.   If you want to test it on the same Tripal instance you did your other testing then you have to reset the settings on the Chado Property field.  To do this first clear the cache, then go to Admin > Structure > Tripal Content Types. Select "manage fields" under the Gene contnet type. Then edit the "Note" and "Comment" fields and simply save them (no need to make changes).  This will cause the settings to update.

spficklin:
I should be clear.  This PR is ready again for testing test 12 again.

dsenalik:
A brand new docker.
Adding the first content, trying to add any of these four Organism, Contact, Protocol, Publication gives this error
`Warning: Undefined array key 1 in Drupal\tripal_chado\Plugin\Field\FieldWidget\ChadoAdditionalTypeWidgetDefault->formElement() (line 66 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldWidget/ChadoAdditionalTypeWidgetDefault.php). `

This happens because `$fixed_value` is specified as a boolean TRUE for some fields `$settings['fixed_value'] = TRUE;` which is stored as `1` actually, and for some reason this is getting evaluated in the additional type field, which expects a string like `DB:accession`. In the past, only the additional type field supported this fixed value. It looks like you added this to other fields for publishing. I'm not sure what the best fix is then.
I will try commenting these booleans out and build a new docker to see what happens...
```
    // If this field needs to set a fixed value, set this to TRUE.
    // It indicates to the publishing step to include this field.
    // If not set, then the publishing step may not be able to find matches
    // for this field based on the fixed value.
//    $settings['fixed_value'] = TRUE;
``

File: tripal/src/TripalStorage/TripalStorageBase.php [hunk 1]
```diff
@@ -207,23 +207,43 @@ public function getStoredValues() {
...
+   *   An array of property values.
+   * @param string $field_name
+   *   The name of the field to addd an item to.
+   */
+  protected function addEmptyValuesItem(&$values, $field_name) {
+    $num_items = count($values[$field_name]);
+    $values[$field_name][$num_items] = [];
+    foreach ($values[$field_name][0] as $key => $value) {
+      $values[$field_name][$num_items][$key] = [];
+      $values[$field_name][$num_items][$key]['value'] = clone $value['value'];
```
laceysanderson, line 243: This line worries me because if this function is called after the value has been set for delta 0 then it will get cloned into other delta. This could cause bleeding of values between delta.

Perhaps we just set the value to empty as well 🤔 
```suggestion
      $values[$field_name][$num_items][$key]['value'] = clone $value['value'];
      $values[$field_name][$num_items][$key]['value']->setValue(NULL);
```

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
```diff
@@ -147,60 +147,70 @@ public static function tripalTypes($field_definition) {
...
     // An intermediate linker table is used
     else {
       // Define the linker table that links the base table to the object table.
+      // (e.g., project_analysis.project_analysis_id)
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id', $record_id_term, [
         'action' => 'store_pkey',
         'drupal_store' => TRUE,
-        'chado_table' => $linker_table,
-        'chado_column' => $linker_pkey_col,
+        'path' => $linker_table . '.' . $linker_pkey_col,
```
laceysanderson, line 174: Both the change and the original should work but I think ideally we specify the full join here to ensure that these properties are evaluated as using the same join path. Plus the original doesn't provide any additional information than before `path` was added to all properties.

```suggestion
        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $linker_pkey_col,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
```diff
@@ -147,60 +147,70 @@ public static function tripalTypes($field_definition) {
...
       ]);
 
       // Define the link between the linker table and the object table.
+      // (e.g., project_analysis.analysis_id)
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, self::$object_id, $linker_fkey_term, [
         'action' => 'store',
         'drupal_store' => TRUE,
-        'chado_table' => $linker_table,
-        'chado_column' => $linker_fkey_col,
+        'path' => $linker_table . '.' . $linker_fkey_col,
```
laceysanderson, line 196:
```suggestion
        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' $linker_fkey_col,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
```diff
@@ -147,60 +147,70 @@ public static function tripalTypes($field_definition) {
...
+      // not implemented in the formatter. Typically these are type_id and rank,
+      // but are not present in all linker tables, so they are added only if
+      // present in the linker table.
       foreach ($extra_linker_columns as $column => $term) {
         $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_' . $column, $term, [
           'action' => 'store',
           'drupal_store' => FALSE,
-          'chado_table' => $linker_table,
-          'chado_column' => $column,
+          'path' => $linker_table . '.' . $column,
```
laceysanderson, line 211:
```suggestion
          'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $column,
```

dsenalik:
New docker with the boolean `$fixed_value`s commented out.
 * I also made a change in `ChadoAdditionalTypeWidgetDefault.php` to allow for no default value for `fixed_value`:
```
-    $fixed_value = $field_settings['fixed_value'];
+    $fixed_value = $field_settings['fixed_value'] ?? null;
```

 * Publishing a gene there is an undefined variable, and as a result the total number of field values is not updated at the end.
```
...
 [notice] Step  6 of 6: Add field items to published entities...
  Checking for published items for the field: field_gene_comment...
 [notice]   Checking for published items for the field: field_gene_comment...
  Publishing items 4 for field: field_gene_comment...
 [notice]   Publishing items 4 for field: field_gene_comment...
 [warning] Undefined variable $num_new_items TripalPublish.php:886
  Checking for published items for the field: field_gene_note...
 [notice]   Checking for published items for the field: field_gene_note...
  Publishing items 4 for field: field_gene_note...
 [notice]   Publishing items 4 for field: field_gene_note...
 [warning] Undefined variable $num_new_items TripalPublish.php:886
  Checking for published items for the field: gene_contact...
...
Published 1 new entities, and 0 field values.
 [notice] Published 1 new entities, and 0 field values.
```

 * When I published the gene, the property values "leaked through"
Is this because I removed the fixed_value = true?
Before
20240107_genebeforepublish

After deletion and publishing
20240107_geneafterpublish

 * But all of the other gene fields including both contacts did publish this time 🎉

tripal_chado/src/Plugin/Field/FieldType/ChadoContactTypeDefault.php
```diff
@@ -164,26 +166,29 @@ public static function tripalTypes($field_definition) {
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id', $record_id_term, [
         'action' => 'store_pkey',
         'drupal_store' => TRUE,
-        'chado_table' => $linker_table,
-        'chado_column' => $linker_pkey_col,
+        'path' => $linker_table . '.' . $linker_pkey_col,
```
laceysanderson, line 169:
```suggestion
        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $linker_pkey_col,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoContactTypeDefault.php
```diff
@@ -164,26 +166,29 @@ public static function tripalTypes($field_definition) {
...
+        //'right_table_id' => $linker_left_col,
       ]);
 
       // Define the link between the linker table and the object table.
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, self::$object_id, $linker_fkey_term, [
         'action' => 'store',
         'drupal_store' => TRUE,
-        'chado_table' => $linker_table,
-        'chado_column' => $linker_fkey_col,
+        'path' => $linker_table . '.' . $linker_fkey_col,
```
laceysanderson, line 189:
```suggestion
        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $linker_fkey_col,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoContactTypeDefault.php
```diff
@@ -195,8 +200,9 @@ public static function tripalTypes($field_definition) {
         $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_' . $column, $term, [
           'action' => 'store',
           'drupal_store' => FALSE,
-          'chado_table' => $linker_table,
-          'chado_column' => $column,
+          'path' => $linker_table . '.' . $column,
```
laceysanderson, line 203:
```suggestion
          'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col . ';' . $column,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoLinkerPropertyTypeDefault.php
```diff
@@ -80,52 +93,96 @@ public static function tripalTypes($field_definition) {
...
+        'table_alias_mapping' => [$table_alias => $prop_table],
+        //'chado_table' => $prop_table,
+        //'chado_table_alias' => $table_alias,
+        //'chado_column' => 'type_id'
       ]),
     ];
   }
 
+  /**
+   * We need to set the type_id property value to match the cvter,_id.
```
laceysanderson:
```suggestion
   * We need to set the type_id property value to match the cvterm_id.
```

tripal_chado/src/Plugin/Field/FieldType/ChadoSynonymTypeDefault.php
```diff
@@ -167,47 +168,53 @@ public static function tripalTypes($field_definition) {
     $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_pkey_id', $linker_synonym_id_term, [
       'action' => 'store_pkey',
       'drupal_store' => TRUE,
-      'chado_table' => $linker_table,
-      'chado_column' => $linker_table_pkey,
+      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_table_pkey,
```
laceysanderson, line 171:
```suggestion
      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_fkey_column . ';' . $linker_table_pkey,
```

tripal_chado/src/Plugin/Field/FieldType/ChadoSynonymTypeDefault.php
```diff
@@ -167,47 +168,53 @@ public static function tripalTypes($field_definition) {
...
+      //'right_table' => $linker_table,
+      //'right_table_id' => $linker_fkey_column,
     ]);
     // E.g. feature_synonym.synonym_id
     $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_synonym_fkey_id' , $linker_fkey_id_term, [
       'action' => 'store',
       'drupal_store' => TRUE,
-      'chado_table' => $linker_table,
-      'chado_column' => 'synonym_id',
+      'path' => $linker_table . '.synonym_id',
```
laceysanderson, line 189:
```suggestion
      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_fkey_column . ';synonym_id',
```

tripal_chado/src/Plugin/Field/FieldType/ChadoSynonymTypeDefault.php
```diff
@@ -167,47 +168,53 @@ public static function tripalTypes($field_definition) {
...
+      //'chado_table' => $linker_table,
+      //'chado_column' => 'synonym_id',
     ]);
     // E.g. feature_synonym.is_current
     $properties[] = new ChadoBoolStoragePropertyType($entity_type_id, self::$id, 'is_current', $linker_is_current_term, [
       'action' => 'store',
-      'chado_table' => $linker_table,
+      'path' => $linker_table . '.is_current',
```
laceysanderson, line 196:
```suggestion
      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_fkey_column . ';is_current',
```

tripal_chado/src/Plugin/Field/FieldType/ChadoSynonymTypeDefault.php
```diff
@@ -167,47 +168,53 @@ public static function tripalTypes($field_definition) {
...
       'drupal_store' => FALSE,
-      'chado_column' => 'is_current',
+      //'chado_column' => 'is_current',
       'empty_value' => TRUE
     ]);
     // E.g. feature_synonym.is_internal
     $properties[] = new ChadoBoolStoragePropertyType($entity_type_id, self::$id, 'is_internal', $linker_is_internal_term, [
       'action' => 'store',
-      'chado_table' => $linker_table,
+      'path' => $linker_table . '.is_internal',
```
laceysanderson, line 205:
```suggestion
      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_fkey_column . ';is_internal',
```

tripal_chado/src/Plugin/Field/FieldType/ChadoSynonymTypeDefault.php
```diff
@@ -167,47 +168,53 @@ public static function tripalTypes($field_definition) {
...
+      //'chado_table' => $linker_table,
       'drupal_store' => FALSE,
-      'chado_column' => 'is_internal',
+      //'chado_column' => 'is_internal',
       'empty_value' => FALSE
     ]);
     // E.g. feature_synonym.pub_id
     $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_pub_id' , $linker_pub_id_term, [
       'action' => 'store',
+      'path' => $linker_table . '.pub_id',
```
laceysanderson, line 214:
```suggestion
      'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_fkey_column . ';pub_id',
```

tripal_chado/src/Services/ChadoFieldDebugger.php
```diff
@@ -140,21 +142,18 @@ public function summarizeChadoStorage($chadostorage, $message) {
   /**
    * This will summarize the results of ChadoStorage::buildChadoRecords().
    *
-   * @param array $base_record_ids
-   *   A simple array where the key is the chado table name and the value
-   *   is the known value of the primary key for the specified record.
    * @param array $records
    *   This is a nested array describing the records which will need to be
```
laceysanderson:
```suggestion
   *   This is an instance of the TripalStorage ChadoRecords class which contains
       all the information of records to be inserted/modifiedin chado. 
```

laceysanderson (commented):
I combed through the code for all but the ChadoStorage and ChadoRecord class. This is super exciting 😍 Here are some things that jumped out at me.

laceysanderson:
@dsenalik can you check PR #1730? I think this fixes the discrepancy that you were mentioning for fixed_value. Also @spficklin do you agree that this is the right approach based on how you designed it?

tripal/src/TripalStorage/TripalStorageBase.php
spficklin: Yeah, I tried that originally but it wouldn't work. I forget why.  My recollection is that each property has checks and wouldn't allow a null in all cases.  I can try again.

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
spficklin: Yes, my original thought is that we should specify the entire join too but in the end it doesn't matter.  We can't guarantee that someone will use the full path to the base table and since we didn't specify that before and I wanted to make as little changes as possible to the fields so ChadoStorage supports it.  I think if we leave it and we ever break something having a mix will ensure we always support it either way.

spficklin:
Thanks @laceysanderson for your review. I committed some and added comments for you to review.

@dsenalik . I'm baffeled why you're getting those errors.  The code you have I"m not sure matches what I'm committing for some reason. I recognize those errors you're getting but they are things I've already fixed and I can't reproduce them in my code.  I can't tell if Lacey's comment just before this means she resolved the issue or not.  But if not can you check to make sure you have the most up-to-date code in the same branch?

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
spficklin: There is also an advantage in using the shorter path length by not including the base table... as it's more readable.

tripal/src/TripalStorage/TripalStorageBase.php
spficklin: I was wrong. This fix does work. I do remember having it this way at one point but it broke so I took out the `setValue()`.  But I must have confused myself on something.  It's working with it so I'm accepting.

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
spficklin: But I'll leave it up to you. I really don't care... I'm just nervous about breaking things and having to spend more time on this PR

laceysanderson:
> @dsenalik . I'm baffeled why you're getting those errors. The code you have I"m not sure matches what I'm committing for some reason. I recognize those errors you're getting but they are things I've already fixed and I can't reproduce them in my code. I can't tell if Lacey's comment just before this means she resolved the issue or not. But if not can you check to make sure you have the most up-to-date code in the same branch?

I did see the same errors that @dsenalik did on the most recent commit from this branch and a fresh docker. That's why I tried to solve them in the other PR.

Specifically, I also saw the following errros when adding any of these four Organism, Contact, Protocol, Publication:

> Warning: Undefined array key 1 in Drupal\tripal_chado\Plugin\Field\FieldWidget\ChadoAdditionalTypeWidgetDefault->formElement() (line 66 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldWidget/ChadoAdditionalTypeWidgetDefault.php).

spficklin:
> Specifically, I also saw the following errros when adding any of these four Organism, Contact, Protocol, Publication:

You see these when manually adding via the GUI?  Do you get a warning on the screen or this appears in the recent logs?

dsenalik:
You see this error directly on screen after trying to add these types via the GUI

spficklin:
I can confirm that I do not get these errors.....

dsenalik:
I wonder if it is a PHP version thing then? I always test on the latest available, 8.2
Edit: no, this code throws the error both in php 8.1 and 8.2
```
$fixed_value = "1";
list($idSpace, $accession) = explode(':', $fixed_value);
```

spficklin:
Hi @dsenalik I think I know how to fix it even thought I don't see the warning.  I'm on PHP 8.1 and  I usually get warnings about missing keys so I'm not sure why I'm not getting it in this case.  I just pushed a fix that I think will work. Can you try it?

dsenalik:
Well commit https://github.com/tripal/tripal/pull/1729/commits/99cfebb595b28b4f7accb1118136e336d8612c1c has fixed this error, I do not see it any more on a fresh docker.

spficklin:
I think this PR is ready for merge after @laceysanderson responds to the comments above.

risharde:
I just pulled in the latest changes, drush cr, created a new gene, deleted it, ran tripal publish and it didn't work for me - none of the fields appeared except Title
image

I'm going to do a full reinstall and test back

dsenalik:
In the interests of time I just pushed a single typo fix, hope it doesn't confuse anything https://github.com/tripal/tripal/pull/1729/commits/22d612e5921559b28735450df485b26962e00335

This removes the undefined variable when publishing from comment https://github.com/tripal/tripal/pull/1729#issuecomment-1880093532
Value is now updated: `Published 1 new entities, and 13 field values.`

dsenalik:
I just tested on my most recent docker, and created an organism, two contacts, and then a gene.
Publishing that appeared to work fine.

Then I added the two properties, Note, and Comment from step 8, and added two of each on a second new gene.
When I went to publish this one - woe is us!
```
...
2024-01-07 17:01:27
Tripal Job Launcher
Running as user 'drupaladmin'
-------------------
2024-01-07 17:01:27: Job ID 3.
2024-01-07 17:01:27: Calling: tripal_publish(gene, chado_storage, Array)
Step  1 of 6: Find matching records... 
 [notice] Step  1 of 6: Find matching records... 
Step  2 of 6: Generate page titles...
 [notice] Step  2 of 6: Generate page titles...
Step  3 of 6: Find existing published entities...
 [notice] Step  3 of 6: Find existing published entities...
Step  4 of 6: Publishing 1 new entities...
 [notice] Step  4 of 6: Publishing 1 new entities...
Step  5 of 6: Find IDs of entities...
 [notice] Step  5 of 6: Find IDs of entities...
Step  6 of 6: Add field items to published entities...
 [notice] Step  6 of 6: Add field items to published entities...
  Checking for published items for the field: field_gene_comment...
 [notice]   Checking for published items for the field: field_gene_comment...
  Publishing items 2 for field: field_gene_comment...
 [notice]   Publishing items 2 for field: field_gene_comment...
  Checking for published items for the field: field_gene_note...
 [notice]   Checking for published items for the field: field_gene_note...
  Publishing items 2 for field: field_gene_note...
 [notice]   Publishing items 2 for field: field_gene_note...
  Checking for published items for the field: gene_contact...
 [notice]   Checking for published items for the field: gene_contact...
  Publishing items 4 for field: gene_contact...
 [notice]   Publishing items 4 for field: gene_contact...
 [warning] Undefined array key 2 TripalPublish.php:740
 [warning] Undefined array key "" TripalPublish.php:741
ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "entity_id" of relation "tripal_entity__gene_contact" violates not-null constraint
DETAIL:  Failing row contains (gene, 0, null, 1, und, 0, 2, 3, 2, 2, null, null, null).: 
      INSERT INTO tripal_entity__gene_contact
        (bundle, deleted, entity_id, revision_id, langcode, delta, gene_contact_record_id, gene_contact_linker_id, gene_contact_link, gene_contact_contact_id) VALUES
(:bundle_3, :deleted_3, :entity_id_3, :revision_id_3, :langcode_3, :delta_3, :gene_contact_record_id_3, :gene_contact_linker_id_3, :gene_contact_link_3, :gene_contact_contact_id_3),
(:bundle_4, :deleted_4, :entity_id_4, :revision_id_4, :langcode_4, :delta_4, :gene_contact_record_id_4, :gene_contact_linker_id_4, :gene_contact_link_4, :gene_contact_contact_id_4); Array
(
    [:bundle_3] => gene
    [:deleted_3] => 0
    [:entity_id_3] => 
    [:revision_id_3] => 1
    [:langcode_3] => und
    [:delta_3] => 0
    [:gene_contact_record_id_3] => 2
    [:gene_contact_linker_id_3] => 3
    [:gene_contact_link_3] => 2
    [:gene_contact_contact_id_3] => 2
    [:bundle_4] => gene
    [:deleted_4] => 0
    [:entity_id_4] => 
    [:revision_id_4] => 1
    [:langcode_4] => und
    [:delta_4] => 1
    [:gene_contact_record_id_4] => 2
    [:gene_contact_linker_id_4] => 4
    [:gene_contact_link_4] => 2
    [:gene_contact_contact_id_4] => 3
)

 [error]  ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "entity_id" of relation "tripal_entity__gene_contact" violates not-null constraint
DETAIL:  Failing row contains (gene, 0, null, 1, und, 0, 2, 3, 2, 2, null, null, null).: 
      INSERT INTO tripal_entity__gene_contact
        (bundle, deleted, entity_id, revision_id, langcode, delta, gene_contact_record_id, gene_contact_linker_id, gene_contact_link, gene_contact_contact_id) VALUES
(:bundle_3, :deleted_3, :entity_id_3, :revision_id_3, :langcode_3, :delta_3, :gene_contact_record_id_3, :gene_contact_linker_id_3, :gene_contact_link_3, :gene_contact_contact_id_3),
(:bundle_4, :deleted_4, :entity_id_4, :revision_id_4, :langcode_4, :delta_4, :gene_contact_record_id_4, :gene_contact_linker_id_4, :gene_contact_link_4, :gene_contact_contact_id_4); Array
(
    [:bundle_3] => gene
    [:deleted_3] => 0
    [:entity_id_3] => 
    [:revision_id_3] => 1
    [:langcode_3] => und
    [:delta_3] => 0
    [:gene_contact_record_id_3] => 2
    [:gene_contact_linker_id_3] => 3
    [:gene_contact_link_3] => 2
    [:gene_contact_contact_id_3] => 2
    [:bundle_4] => gene
    [:deleted_4] => 0
    [:entity_id_4] => 
    [:revision_id_4] => 1
    [:langcode_4] => und
    [:delta_4] => 1
    [:gene_contact_record_id_4] => 2
    [:gene_contact_linker_id_4] => 4
    [:gene_contact_link_4] => 2
    [:gene_contact_contact_id_4] => 3
)
 
[site http://default] [TRIPAL] ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "entity_id" of relation "tripal_entity__gene_contact" violates not-null constraintDETAIL:  Failing row contains (gene, 0, null, 1, und, 0, 2, 3, 2, 2, null, null, null).:       INSERT INTO tripal_entity__gene_contact        (bundle, deleted, entity_id, revision_id, langcode, delta, gene_contact_record_id, gene_contact_linker_id, gene_contact_link, gene_contact_contact_id) VALUES(:bundle_3, :deleted_3, :entity_id_3, :revision_id_3, :langcode_3, :delta_3, :gene_contact_record_id_3, :gene_contact_linker_id_3, :gene_contact_link_3, :gene_contact_contact_id_3),(:bundle_4, :deleted_4, :entity_id_4, :revision_id_4, :langcode_4, :delta_4, :gene_contact_record_id_4, :gene_contact_linker_id_4, :gene_contact_link_4, :gene_contact_contact_id_4); Array(    [:bundle_3] => gene    [:deleted_3] => 0    [:entity_id_3] =>     [:revision_id_3] => 1    [:langcode_3] => und    [:delta_3] => 0    [:gene_contact_record_id_3] => 2    [:gene_contact_linker_id_3] => 3    [:gene_contact_link_3] => 2    [:gene_contact_contact_id_3] => 2    [:bundle_4] => gene    [:deleted_4] => 0    [:entity_id_4] =>     [:revision_id_4] => 1    [:langcode_4] => und    [:delta_4] => 1    [:gene_contact_record_id_4] => 2    [:gene_contact_linker_id_4] => 4    [:gene_contact_link_4] => 2    [:gene_contact_contact_id_4] => 3)
```

This is all that got published
20240107_second_gene

risharde:
@dsenalik pulled your last commit, reinstalled, created 2 contacts, 1 organism, created gene, deleted it and published it back - noticing this

image

so confirming it doesn't work on my side either

Edit:
Notes should have been 3 entries with notes 1, notes 2, notes 3
Comments should have been 3 entries with comments 1, comments 2, comments 3

dsenalik:
@risharde but you didn't get any errors when publishing 😕 ?

risharde:
@dsenalik No errors there, it created the Gene page
One other thing I notice is that the author is not attached - not sure if this is super important but just mentioning it
image

dsenalik:
I noticed that too, but I assume it is expected behavior, as a published page does not have an "author".

tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php
laceysanderson: True, I just always think of the fields in core as the "Best Practices" example so want to make sure we give consistent and clear examples 🤔 I lean towards using the full path here but also don't want to break things again 😬 How about we leave these full paths out of this PR. I can make a separate PR that makes them consistent and hopefully that can get in before release but at least by putting them in a separate PR, if something goes wrong it doesn't hold this up.

spficklin:
I just tested it on my end with all of the recent fixes and it seems to be working just fine :-/  I'll try a complete reinstall and see if that makes a difference.

dsenalik:
If it helps, this part of the publish log
```
 [notice]   Publishing items 4 for field: gene_contact...
 [warning] Undefined array key 2 TripalPublish.php:740
 [warning] Undefined array key "" TripalPublish.php:741
```
There should only have been **2** items for gene_contact, not sure why it tried to do **4**. That's how many properties there were though...

spficklin:
But you all didn't have these problems prior to our few tweaks, right?

dsenalik:
As I recall, and looking at previous comments https://github.com/tripal/tripal/pull/1729#pullrequestreview-1807523715, no this did not happen earlier.

I just now tried a new gene with three each of note and comment, and as I suspected, somehow 3+3 showed up under gene_contact:
```
 [notice]   Publishing items 6 for field: gene_contact...
 [warning] Undefined array key 4 TripalPublish.php:740
 [warning] Undefined array key "" TripalPublish.php:741
```
20240107_threeeach

spficklin:
I'm now seeing the same as @risharde .  Digging into it..... I really want to cry right now.

dsenalik:
I bet it is a simple fix 😟 but it might take hours to find...

spficklin:
Okay. I think I have it fixed. The problem was related to that pesky `fixed_value`.  We definintely need to redesign that as per the discussinon on PR #1730 .  But for now, I think I have it fixed so things will work.  

I would suggested starting from a fresh install.

risharde:
@spficklin It works!!! The order of the multiple values changed but I suspect this is normal behaviour. Great going!! 

image

dsenalik:
Just tested a new docker, gene content type, three different properties and three values for each property.
Publishing worked perfectly 😀
Also three contacts and two synonyms were published correctly.
@risharde ~~as far as order changing, I don't know if publish yet supports using chado ranks, such as from the `featureprop` table?~~
edit: I just did the experiment, changed the rank in `featureprop` before publishing, and the order was **correct** based on those ranks.
```
sitedb=> update featureprop set rank=-1 where value like '%3';
sitedb=> select * from featureprop order by featureprop_id;
 featureprop_id | feature_id | type_id |   value   | rank 
----------------+------------+---------+-----------+------
             15 |          1 |    3029 | Comment 1 |    0
             16 |          1 |    3029 | Comment 2 |    1
             17 |          1 |    3029 | Comment 3 |   -1
             18 |          1 |     727 | DNA1      |    0
             19 |          1 |     727 | DNA2      |    1
             20 |          1 |     727 | DNA3      |   -1
             21 |          1 |    3143 | Note 1    |    0
             22 |          1 |    3143 | Note 2    |    1
             23 |          1 |    3143 | Note 3    |   -1
(9 rows)
```
20240108_rank

Confirmed a second time with `update featureprop set rank=5 where value like '%1';` and published in 3,2,1 order as expected.

laceysanderson (approved):
I did one more test with the most recent commit.

I created multiple pieces of content of almost every time, edited them, deleted them and tried to trigger validation errors. Page creation seems very stable 😌 

With 3 reviewers saying this is working and a code review, this can be merged! 🎉 Thanks for all the work @spficklin and thanks for the weekend reviews @risharde and @dsenalik!

One big step closer to release!

PR #1730, opened 2024-01-07
Fix for the fixed_value discrepancy that Doug pointed out on PR #1729

laceysanderson:
There were discrepancies between how the fixed_value field setting was being used. Based on the new documentation and implementation provided by @spficklin it should be a boolean but in the past it actually held the fixed_value.

This PR makes it consistently a boolean and updates the ChadoAdditionalTypeWidgetDefault to generate the value based on the term if fixed_value is true.

Note: I also say that the boolean fields are set to fixed_value as well and something in my memory thinks I had a convo about this with @dsenalik?

## Testing

1. Create a new docker/install
2. Create an organism and confirm the intraspecific type field works as expected without errors.
3. Create a gene and confirm that the type field works here as well. Note: you should not be able to edit the type here and it should be disabled.

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -129,8 +129,8 @@ field.field.tripal_entity.*.*:
           label: 'Flag to Enable field debugging'
           nullable: true
         fixed_value:
-          type: string
-          label: "A fixed value for the field that cannot be changed."
+          type: boolean
+          label: "Indicates if this field has a fixed value for one or more of it's property types."
           nullable: true
 
 ## For TripalEntityTypeCollection functionality
@@ -281,8 +281,8 @@ tripal.tripalfield_collection.*:
                      label: "Term Accession"
                      nullable: false
                    fixed_value:
-                     type: string
-                     label: "A fixed value for the field that cannot be changed."
+                     type: boolean
+                     label: "Indicates if this field has a fixed value for one or more of it's property types."
                      nullable: true
                display:
                  type: mapping
--- a/tripal_chado/config/install/tripal.tripalfield_collection.genetic_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalfield_collection.genetic_chado.yml
@@ -288,7 +288,7 @@ fields:
         settings:
             termIdSpace: SO
             termAccession: "0000771"
-            fixed_value: SO:0000771
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -342,7 +342,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_analysis
-            fixed_value: local:is_analysis
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -369,7 +369,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_obsolete
-            fixed_value: local:is_obsolete
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -583,7 +583,7 @@ fields:
         settings:
             termIdSpace: SO
             termAccession: "0001060"
-            fixed_value: SO:0001060
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -637,7 +637,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_analysis
-            fixed_value: local:is_analysis
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -664,7 +664,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_obsolete
-            fixed_value: local:is_obsolete
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -878,7 +878,7 @@ fields:
         settings:
             termIdSpace: SO
             termAccession: "0001645"
-            fixed_value: SO:0001645
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -932,7 +932,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_analysis
-            fixed_value: local:is_analysis
+            fixed_value: TRUE
         display:
           view:
             default:
@@ -959,7 +959,7 @@ fields:
         settings:
             termIdSpace: local
             termAccession: is_obsolete
-            fixed_value: local:is_obsolete
+            fixed_value: TRUE
         display:
           view:
             default:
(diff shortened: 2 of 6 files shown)


dsenalik (approved):
Okay, knowing the fixed_value was supposed to be boolean helps.
I tested another new docker, all of the fixed_value related issues from this problem that I saw are fixed by this, your testing steps are successful.

I continued on and this does not fix the problem in PR #1729 where only one property is published, but that is not supposed to be part of this fix 🙈

(The convo we had was where I mentioned that at the time, fixed_value was only applicable to the additional type field.)
((By applicable, I mean only that field ever had code to do anything based on fixed_value.))

spficklin:
So the `fixed_value` setting is a bit weird because I don't think it considers all use cases for it. Take a look at this code in TripalPublish where it is used:

https://github.com/tripal/tripal/blob/3c1d4d8b7e6e62d41b8e14612b664ba0054b4e92/tripal/src/Services/TripalPublish.php#L422-L425

Note it just checks to see if it's present so it doesn't care if it's TRUE or FALSE or a string or integer.   It actually should not be expected to be a boolean or an integer. It can be anything and should be allowed to be anything.   Now that I look at it, I think this code in the `ChadoAdditionalTypeTypeDefault` field is not needed:

https://github.com/tripal/tripal/blob/3c1d4d8b7e6e62d41b8e14612b664ba0054b4e92/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php#L46-L54

This is because it's already set in the YAML

https://github.com/tripal/tripal/blob/3c1d4d8b7e6e62d41b8e14612b664ba0054b4e92/tripal_chado/config/install/tripal.tripalfield_collection.genomic_chado.yml#L178-L181

I'm a bit confused because it seems like the code in the plugin should override the YAML and the 'FALSE' value will break it.  I need to look deeper to see what's going on there. 

But continuing on..... TripalPublish is the only place (other than the `ChadoAdditionalTypeTypeDefault` field) where this value is used.  TripalPublish uses it to check to see if the field needs to "always" be added for publishing.  During publishing, we don't want to add all fields because it makes TripalPublish take an exorbitant amount of time as it has to query all fields... which is too slow.....

To make it go faster we only query for the fields that are required to find the fields that can uniquely identify an entity and its member field values.   Fields that have a fixed value that shouldn't ever change (e.g. the `ChadoAdditionalTypeTypeDefault` field, and the `ChadoLinkerPropertyTypeDefault ` field)  should always be included so that we find the correct properties.  Otherwise, we can't distinguish between gene or mRNA content types when trying to find them for publishing, and we can't distinguish between different properties.   So if the `fixed_value` is present within a field setting (no matter the value) then it gets added for querying.  

I think what is confusing is that the  `fixed_value` setting works nicely for the additional type field as the "main property" is `term_name`:

https://github.com/tripal/tripal/blob/3c1d4d8b7e6e62d41b8e14612b664ba0054b4e92/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php#L38-L40

This matches what is provided for the `fixed_value` in the YAML. The value is appropriate for the `term_name` property:

https://github.com/tripal/tripal/blob/3c1d4d8b7e6e62d41b8e14612b664ba0054b4e92/tripal_chado/config/install/tripal.tripalfield_collection.genomic_chado.yml#L178-L181

But in the case of the `ChadoLinkerPropertyTypeDefault` field, the main property for the property linker is not set so it defaults to the `value` property.  But we don't need a fixed value for the `value` property. We need it for the `type_id` property.  

I noticed the confusion when I was fixing the code for test 12 on PR #1729  but didn't feel like fixing it because I was exhausted, so to make the linker property work I just set a `fixed_value` setting of TRUE and let the `tripalValuesTemplate` take care of setting the value for the `type_id` field (which is also what happens in the `ChadoAdditionalTypeTypeDefault`  field.

Now that I think about it.... we have two different types of fixed values. We have those that map directly to a property (e.g. the `term_name` of the `ChadoAdditionalTypeTypeDefault` field and fixed values that are derived from other settings, which is the case for the `ChadoLinkerPropertyTypeDefault` (it uses the `termIdSpace` and `termAccession` settings to find the value for the fixed `type_id` property. 

I think we do need a `fixed_value` setting but I think we change the design for it. Here's a proposal.

- Change `fixed_value` to `fixed_values`
- We have two keys:  'absolute' and 'derived'.  T
- The value for the `absolute` key is an array that has key/value pairs where the keys are the property keys and the value is the value that the property should have fixed.
- The value for the `derived` key is simply an array of property keys that have a derived fixed value.  

So, for the `ChadoAdditionalTypeTypeDefault` it would look like this in the YAML:

```yaml
        settings:
            termIdSpace: SO
            termAccession: "0000704"
            fixed_values: 
                absolute:
                   term_name: SO:0000704
                derived:  
                   - type_id
```

The `ChadoLinkerPropertyTypeDefault` is not defined in the YAML because properties currently get added manually.  So the function would look like this:

```php
 public static function defaultFieldSettings() {
    $settings = parent::defaultFieldSettings();
    $settings['fixed_values']['derived'] = ['type_id']
    return $settings;
  }
```

spficklin:
Also, there are no checks in Tripal to make sure that fixed values are set. It's up to the field developer to make sure they get set. If they forget then Publishing will break.  So, this structure has the added benefit that it will allow us to check to make sure the fields that should have derived or absolute values set are actually set, and we can take the burden away from the developer for setting absolute fixed values.  They will still have to set the derived ones.

spficklin (changes requested):
I'm not sure these changes are the right way to go.  See my comment

laceysanderson:
> Note it just checks to see if it's present so it doesn't care if it's TRUE or FALSE or a string or integer. It actually should not be expected to be a boolean or an integer. It can be anything and should be allowed to be anything. 

The Drupal Schema API won't allow it to be anything... Hence the schema errors you are seeing in this PR. Based on what you've said above, I think we leave it as a string and leave things they way they are in the `ChadoAdditionalTypeTypeDefault`. For the `ChadoLinkerPropertyTypeDefault` I suggest we make the value `DERIVED`.

Maybe not ideal yet but mucks with less things right before a release and doesn't summon all the schema errors in the automated testing that I'm getting in this PR 😒

spficklin:
Oh if you can't have a type that varies in the YAML then that's a problem.   Sounds like we need to think more on it.

spficklin:
I'll make a new issue for this.

spficklin:
I created a new issue #1731 and I agree that we should leave things as they are for now, so I'm going to close this PR.

Issue #1509, opened 2023-05-09
Tripal 4 update YAML for various missing fields


dsenalik:
# BUG/ERROR report

### Issue description

Currently publication content type has no fields defined except for "Authored by".
Presumably these are missing from
`tripal_chado/config/install/tripal.tripalfield_collection.default_chado.yml`

Other fields are also missing from various content types

#### Steps to reproduce
`/bio_data/add/bio_data_6`
 
#### Error messages and screenshots
![add-publication](https://github.com/tripal/tripal/assets/8419404/ba4913d9-021d-4fa0-9f0a-6d0d592e0be5)

spficklin:
I think this issue is corrected with the last alpha release, but please reopen if I'm wrong.

PR #1732, opened 2024-01-08
Publishing content types and fields linked to a property type_id

spficklin:
# Tripal 4 Core Dev Task

## Description

This PR allows you to publish content types and their fields when the content type is tied to a `type_id` in a "prop" table rather than a `type_id` in the base table.  Different types of analyses are the primary example of this. 

## Testing?

1. Create one more more Analysis content types
2. Create a Genome Assembly content type
3. Delete all of these new entities.
4. Publish the Analysis content type. This should result in restoration of all of the analyses you created before and the ndeleted.
5. Publish the Genome Assembly type. You should only have one type left.
6. Since we don't have functionl tests for some of the issues that were discovered when testing PR #1729 It would be good to test a gene with contacts, and two different properties with multiple values each.  The delete it and publish that to make sure all still works.


spficklin:
@laceysanderson somehow I have some changes in this PR to publications. I'm not sure how they got in there! I don't remember working on that.  Can you check to see if that is just something from PRs being merged out of order with how branches were merging?

laceysanderson:
Hi @spficklin, it's not 😬 it looks like just small naming changes that aren't needed so maybe just remove them? When I work on https://github.com/tripal/tripal/issues/1509 I'll get everything with the publication content type nicely cleaned up.

laceysanderson:
The deprecation error is fixed in PR #1727.
I'm not sure why the chadoStorage findValues kernel test is failing in this branch though 🤔 It is passing in 4.x so must be related to the changes here

dsenalik:
@laceysanderson when you work on #1509 note that I have all the tripal 3 publication fields added in #1704 including properties

spficklin:
Okay there were actually two problems which I've now fixed.  But one problem was, I think, with testing.  @laceysanderson can you see where I commented out the line in the functional test.  IT sets a property type_id but the `findValues()` fails because it can't find a gene with a property in the `featureprop` table that has that type_id.   Previously, the `findValues()` didn't honor those settings.  It now  does.  So if you set a value in any field it will make sure that whatever records it finds matches those values.   Right now, it only supports an exact match. In the future, when we start integrating Drupal Views we need to get smarter about that.  But for publishing, an exact match is sufficient.

dsenalik (approved):
Excellent! I did all of the testing steps, including three properties on a gene and everything worked as expected.

(I note that a genome assembly publishes both as an analysis, and as a genome assembly. This is the exact same behavior as Tripal 3. A potential thing to do - don't publish the generic version if a specific version exists. Sounds hard, though)
20240109_analyses

spficklin:
Thanks @dsenalik  for the review!  Yeah, I agree with not publishing it twice.   It's something we can work on.   I just want to wait on @laceysanderson to take a look at that change I made to the functional test and if she's okay with things then I think we can merge this!

laceysanderson (approved):
I'm okay with that change 👍

PR #1735, opened 2024-01-09
Fix unquoted leading zeros in yaml

### LINKED ISSUE #1734: Field not being added.
### Tripal Version

4.x

### Drupal Version

10.x

### PostgreSQL Version

n/a

### PHP Version

8.1

### General Description

When importing content types I get the following error:

```
 [error]  ERROR: The term accession. "SBO:362", is not known in the Term Id Space for field, "phenotypic_marker_dbxref_ann". Check the "termIdSpace" and "termAccession" elements. 

```

I've hit on this before and it happens in the YAML when the accession for a term starts with zero and is not surrounded by quotes.  It gets interpreted as a number and the leading zeros get dropped off. 

### Steps to reproduce

Just install a new site and import the content collection

### Error messages

_No response_

### Screenshots

_No response_

dsenalik:
# Bug Fix

## Description
I'm embarassed I let these unquoted values through. I thought I had checked them all but obviously not.

You can't have unquoted things that look like numbers in yaml.

## Testing?
Just install a new site and import the genetic content collection. Or maybe code review is sufficient for this one.

### DIFF
--- a/tripal_chado/config/install/tripal.tripalfield_collection.genetic_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalfield_collection.genetic_chado.yml
@@ -127,7 +127,7 @@ fields:
                 linker_fkey_column: dbxref_id
         settings:
             termIdSpace: SBO
-            termAccession: 0000552
+            termAccession: "0000552"
         display:
           view:
             default:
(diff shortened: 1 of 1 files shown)


dsenalik:
This is such a safe simple fix, and you won't want to have to wait for tests tomorrow, I am just going to abuse my powers and merge this now.

PR #1736, opened 2024-01-10
Fix to improve loading speed of content type listing

spficklin:
# Bug Fix

## Description

This PR fixes the content listing page so that it loads much faster.  Previously if you published tens of thousands of entities the page would take a very long time to load or could time out.  

## Testing?

1. Use the GFF loader to import a bunch of genes
2. Publish the gene records
3. Go to the Tripal Content Listing page.  The page should load extremely quickly. You can use the page to move between pages also very quickly.

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -559,6 +559,21 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
   public static function postLoad(EntityStorageInterface $storage, array &$entities) {
     parent::postLoad($storage, $entities);
 
+
+    // IF we are doing a listing of content types there is no way in Drupal 10
+    // to specify which fields to load.  By deafult the SqlContentEntityStorage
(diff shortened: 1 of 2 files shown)


spficklin:
If you need a GFF3 file for loading try this one:  http://rice.uga.edu/pub/data/Eukaryotic_Projects/o_sativa/annotation_dbs/pseudomolecules/version_7.0/all.dir/all.gff3

It's for the _Oryza sativa_ v7.0 genome.  When you load it into Tripal be sure to set the landmark CV term as 'chromosome' or 'supercontig', or it won't know how to import the landmarks.

dsenalik (approved):
Wow that file has 55,986 genes! Took a while to load ...
The code changes are simpler than I would have thought, excellent!
With 2G memory, I was able to publish everything, and indeed content listing is now fast!
(As an aside, to increase the php memory limit in a docker, you edit `/usr/local/etc/php/conf.d/docker-php-memlimit.ini`)

PR #1742, opened 2024-01-10
GFF3 loader validation 

spficklin:
# Bug Fix

## Description

When loading a GFF3 file with the Tripal GFF3 loader, if you do not specify the landmark CV term then you get a validation error that says the following:

```
Error message
The Sequence Ontology (SO) term selected for the landmark type is not available in the database. Please check the spelling or select another
Because not all GFF3 files specify properly the sequence type for the landmarks to prevent mistakes in loading we should fix this problem by just requiring it. @laceysanderson agrees.
```
This simply fixes the problem by making the landmark form field required.  

## Testing?

Pull the PR and try to load a GFF3 file without specifying the landmark... it won't let you. bwahahaha

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
+++ b/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
@@ -317,7 +317,8 @@ class GFF3Importer extends ChadoImporterBase {
     $form['landmark_type'] = [
       '#title' => t('Default Landmark Type'),
       '#type' => 'textfield',
-      '#description' => t("Optional. Use this field to specify a Sequence Ontology type
+      '#required' => TRUE,
+      '#description' => t("Use this field to specify a Sequence Ontology type
        for the default landmark sequences in the GFF fie (e.g. 'chromosome'). This is only needed if
        the landmark features (first column of the GFF3 file) are not already in the database."),
       '#autocomplete_route_name' => 'tripal_chado.cvterm_autocomplete',


dsenalik (commented):
I feel like this is not the right fix, but correct me if I am wrong. As the message says, this is optional if the landmark is already in the database. Elsewhere in the code an exception is thrown if this is indeed not the case
`The landmark, %landmark, cannot be added because no landmark type was provided. Please redo the importer job and specify a landmark type.`

So the actual problem might be where we lookup the cvterm from the empty field in the form validation. Maybe just add `    if ($landmark_type) {` at line 503?

This would need some testing to confirm though, and that's not happening today...

dsenalik:
I'm confirming a problem here.
I loaded a fasta file specifying sequence type as "DNA" - this defines the landmarks
I then loaded the GFF3 file, with landmark entered as "RNA" (to simulate the fact that I have to enter something)
It them made new second landmarks of type RNA as you would expect.
If I load leaving the type blank, there is a bug in the code and it does not try to look up from existing landmarks.
` [error]  ERROR: Could not determine a type for landmark name: <em class="placeholder">DCARv3_MT</em>. There was no default landmark type to force either. `

It will be good to get this working, because sometimes we have a mix of types in a gff3 file - Chromosome, Scaffold, Contig, etc. so having the ability to leave the default blank is helpful.

A helpful past comment https://github.com/tripal/tripal/issues/1324#issuecomment-1430175719

And a screenshot of how it is in Tripal 3, you can specify only one or the other of these
20230216_landmarks
which didn't make it to Tripal 4

dsenalik:
These suggestions will be added to a new issue # TBD

PR #1743, opened 2024-01-10
Update or add most remaining linking fields from issue #1381

### LINKED ISSUE #1381: Tripal fields don't exist for all of the linker tables, e.g. link analysis and project, Tripal 4 version
### Description
This extends issue #1318, port the ```chado_linker__xxx``` fields to Tripal 4, and is related to #1414

**Update:** 
With Tripal 4, with a bit of extra code (not trivial), we can now support this in the existing fields, we won't need additional fields. See PR #1431 for the first example.

### LINKED ISSUE #1387: Validation error when leaving optional fields empty
When I create a gene through the UI, I get an error indicating that the sequence length is not of the correct primitive type when I leave it empty and try to save. The validation error is "This value should be of the correct primitive type. " with the "Sequence Length" field highlighted in red. Since this field is not required, I should be able to leave it empty without a validation error. 

## Steps to recreate

1. Spin up a new Tripal site.
2. Create an organism through the UI (i.e. Go to Tripal > Content > Organism)
3. Create a  Gene through the UI (i.e. Go to Tripal > Content > Gene) leaving the "Sequence Length" field at the bottom of the page empty.
4. Note the validation error that occurs preventing save.
5. Enter a number of any type in the "Sequence Length" field and save.
6. Note that if the sequence length is filled in then the gene is saved as expected.

dsenalik:
# New Feature

## Description
This PR will add all of the linking fields needed by core tripal, and updates a few existing ones to have linking capability.
For issue #1550 it will add the remaining missing fields from columns of the content type's base table to several content types, especially pub.
It will also add these linking fields to most content types through the yaml files in `tripal_chado/config/install/tripal.tripalfield_collection.xxx_chado.yml`

What is not done here is adding fields for the "complicated" content types. These are the cases where there are multiple content types derived from a single chado table. Four example, there are four content types from the `stock` table, they are all listed https://github.com/tripal/tripal_doc/issues/32#issuecomment-1837528225

Also, linking fields between two different entity groups are not yet added, for example stock (germplasm) <->feature (genetic or genomic)
Any table columns needing a timestamp are not here because the ChadoTimeDefault field does not yet exist (https://github.com/tripal/tripal/issues/1550#issuecomment-1623912019)

## Testing?
1. Build a docker on this branch
```
port=8080
git clone https://github.com/tripal/tripal.git testing-1381
cd testing-1381
git checkout tv4g1-issue1381-chado_linker_simple_fields
docker build --tag=tripaldocker:testing-1381 --build-arg drupalversion="10.1.x-dev" --build-arg chadoschema='chado' --file tripaldocker/Dockerfile-php8.2-pgsql13 ./
docker run --publish=${port}:80 --name=1381 -tid --volume=$(pwd):/var/www/drupal/web/modules/contrib/tripal tripalproject/tripaldocker:latest
docker exec testing-1381 service postgresql start
```
2. Go to "Page Structure" -> "+Import type collection"  -> Check all boxes and click "Import"
`drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
There should of course be no errors.
3. Create some content for testing, this order lets you proceed nicely through the 15 tables use by core Tripal content:
Organism - Enter something for all the fields, and as you go down this list you will be able to link to the earlier content you just made. 
4. Contact
5. Publication - Make sure to put unique content in each of the publication fields to test them all. For ELocation and URL try something with a `https:` prefix to make sure it displays as a link
20240110_ElocationURL

7. Protocol
8. ArrayDesign - leave at least one of the later integer fields empty. Having a non-required integer field left blank tests the fix for issue #1387.
9. Study - specify both contact and publication when you create it. This tests the current version of chado storage - there should  be no error.
10. Assay
11. Analysis
12. Phylogenetic Tree
13. Project
14. Genetic Map
15. Germplasm accession (stock table)
16. Biological Sample
17. Gene (feature table)
18. DNA Library - Here the "Is Obsolete" field is a integer, and it is required. (Don't ask me why, chado defines it that way) Test using `0` to make sure zero is an integer! (Tests #1707)
20240110_15tables

19. Test as many other content types as you can stand, filling out all of the fields to make sure they work.

### DIFF
--- a/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
@@ -58,8 +58,8 @@ content_types:
         help_text: Use the publication page for books, journal articles, or other citable work.
         category: General
         id: pub
-        title_format: "[publication_title]"
-        url_format: "pub/[TripalEntity__entity_id]"
+        title_format: "[pub_title]"
+        url_format: "publication/[TripalEntity__entity_id]"
         synonyms:
             - bio_data_6
 
--- a/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
@@ -58,8 +58,8 @@ content_types:
         help_text: Use the publication page for books, journal articles, or other citable work.
         category: General
         id: pub
-        title_format: "[pub_title]"
-        url_format: "publication/[TripalEntity__entity_id]"
+        title_format: "[publication_title]"
+        url_format: "pub/[TripalEntity__entity_id]"
         synonyms:
             - bio_data_6
 
--- a/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
+++ b/tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
@@ -58,7 +58,7 @@ content_types:
         help_text: Use the publication page for books, journal articles, or other citable work.
         category: General
         id: pub
-        title_format: "[publication_title]"
+        title_format: "[pub_title]"
         url_format: "pub/[TripalEntity__entity_id]"
         synonyms:
             - bio_data_6
--- a/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
@@ -48,24 +48,20 @@ class ChadoOrganismFormatterDefault extends ChadoFormatterBase {
     $list = [];
     $token_string = $this->getSetting('token_string');
 
-    // Implement the shortcut token "[scientific_name]". Rather than the overhead
-    // of the chado_get_organism_scientific_name() api call, replicate its behavior
-    // with an equivalent token string.
-    $token_string = preg_replace('/\[scientific_name\]/', $token_string,
-        '[genus] [species] [infratype_abbrev] [infraname]');
-
     foreach ($items as $delta => $item) {
       $values = [
         'genus' => $item->get('organism_genus')->getString(),
         'species' => $item->get('organism_species')->getString(),
         'infratype' => $item->get('organism_infraspecific_type')->getString(),
         'infraname' => $item->get('organism_infraspecific_name')->getString(),
+        'scientific_name' => $item->get('organism_scientific_name')->getString(),
         'abbreviation' => $item->get('organism_abbreviation')->getString(),
         'common_name' => $item->get('organism_common_name')->getString(),
         'comment' => $item->get('organism_comment')->getString(),
       ];
 
       // Special case handling of abbreviations for genus and infraspecific type
+      // These are not available to web services!
       $values['genus_abbrev'] = substr($values['genus'], 0, 1) . '.';
       $values['infratype_abbrev'] = chado_abbreviate_infraspecific_rank($values['infratype']);
 
--- a/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
@@ -2,9 +2,7 @@
 
 namespace Drupal\tripal_chado\Plugin\Field\FieldFormatter;
 
-use Drupal\tripal\TripalField\TripalFormatterBase;
 use Drupal\Core\Field\FieldItemListInterface;
-use Drupal\Core\Form\FormStateInterface;
 use Drupal\tripal_chado\TripalField\ChadoFormatterBase;
 
 /**
@@ -12,11 +10,23 @@ use Drupal\tripal_chado\TripalField\ChadoFormatterBase;
  *
  * @FieldFormatter(
  *   id = "chado_organism_formatter_default",
- *   label = @Translation("Chado Organism Reference Formatter"),
- *   description = @Translation("A chado organism reference formatter"),
+ *   label = @Translation("Chado organism formatter"),
+ *   description = @Translation("A chado organism formatter"),
  *   field_types = {
  *     "chado_organism_type_default"
- *   }
+ *   },
+ *   valid_tokens = {
+ *     "[genus]",
+ *     "[genus_abbrev]",
+ *     "[species]",
+ *     "[infratype]",
+ *     "[infratype_abbrev]",
+ *     "[infraname]",
+ *     "[scientific_name]",
+ *     "[abbreviation]",
+ *     "[common_name]",
+ *     "[comment]",
+ *   },
  * )
  */
 class ChadoOrganismFormatterDefault extends ChadoFormatterBase {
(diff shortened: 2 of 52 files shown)


tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
```diff
@@ -58,8 +58,8 @@ content_types:
         help_text: Use the publication page for books, journal articles, or other citable work.
         category: General
         id: pub
-        title_format: "[publication_title]"
-        url_format: "pub/[TripalEntity__entity_id]"
+        title_format: "[pub_title]"
+        url_format: "publication/[TripalEntity__entity_id]"
```
spficklin: Reverts this change back to `publication` instead of `pub`.  We're not sure how this keeps popping up in our PRs, but this matches the fix @laceysanderson made earlier.

```suggestion
        title_format: "[publication_title]"
        url_format: "pub/[TripalEntity__entity_id]"
```

spficklin:
Everything worked breautifully except for one issue.  For the Phylotree record a cross refrence has to be specified because the `phytotree.dbxref_id` column is required.  I get the following message if I leave it blank:
``` 
The item cannot be saved because the following values are missing. Dbxref_id.
```
Next on to more code review.

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
```diff
@@ -2,37 +2,101 @@
...
+
+      // Substitute values in token string to generate displayed string.
+      $displayed_string = $token_string;
+      foreach ($values as $key => $value) {
+        $displayed_string = preg_replace("/\[$key\]/", $value, $displayed_string);
+      }
+      $list[$delta] = [
+        '#markup' => $displayed_string,
+      ];
+    }
```
spficklin: I'm confused why this code is in the formatter?  We have code built into the field for specifying a token string used to build the label.   If the token string isn't smart enough to build the values shown for the species then perhaps we need to use the `function` action for the field and put this code there.  That way it doesn't matter if you are getting results on the page or if you are getting results via the web services (coming soon).  With this code the user on the web browser will see something different than what is shown in web services.  

I'm not suggestion we fix it here, I just wanted to ask about it and if it sounds like we should use the `function` action on the field to generate this then we can change it later... we just need to add an issue so we don't forget.

spficklin: Also, the `function` action for fields isn't yet written.....  So another reason to leave things as they are here but create an issue to fix this later if need be.

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
```diff
@@ -3,131 +3,259 @@
...
+      'action' => 'read_value',
+      'drupal_store' => FALSE,
+      'path' => $linker_table . '.' . $linker_fkey_column . '>' . $object_table . '.' . $object_pkey_col . ';genus',
+      'chado_table' => $object_table,
+      'chado_column' => 'genus',
+      'as' => 'organism_genus',
     ]);
-    $properties[] =  new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'label', $label_term, $label_len, [
-      'action' => 'replace',
-      'template' => "<i>[genus] [species]</i> [infraspecific_type] [infraspecific_name]",
```
spficklin, line 111: I'm just marking this as a topic of dicussion for the Codefest day 2.

spficklin (changes requested):

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoOrganismFormatterDefault.php
dsenalik: This started with the boolean field, since a site administrator may wish to configure how true and false are displayed, e.g. Yes vs. No, Present vs. Absent, etc.
Then it seemed a really simple but helpful way for site admins to configure how linked content is displayed. You may want a displayed value to be different than the title of the entity page itself. This is explained in the now merged PR #1697.
A use case for this could be for a linked analysis, we might want to include the type of analysis, so the token string could be `[name] ([type])`
The default title for analysis is to just display the name.
We can discuss further...

spficklin: @laceysanderson and I had a conversation abou this. I now understand why it is done the way it is.  We had one suggestion though which I'll put on the next comment.

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
spficklin: So, @laceysanderson and I had a conversation abou this as the codefest was wrapping up.  We think this property should come back but have a key of 'scientific_name' instead.

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
```diff
@@ -3,131 +3,259 @@
...
+  protected static $object_table = 'organism';
+  protected static $object_id = 'organism_id';
 
   /**
    * {@inheritdoc}
    */
   public static function mainPropertyName() {
-    return 'label';
+    // Overrides the default of 'value'
+    return 'species';
```
spficklin: We need to have a property that can provide the main value for this field and species isn't quite right.  In another comment I suggested we put back the `label` field but name it `scientific_name`.  Here we can use that property.

spficklin (changes requested):

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
dsenalik: This will be fixed by commit https://github.com/tripal/tripal/pull/1743/commits/720af13025158a14f6edc9d1a9f45d8458bcdaea

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
dsenalik: This is done with commit https://github.com/tripal/tripal/pull/1743/commits/720af13025158a14f6edc9d1a9f45d8458bcdaea

dsenalik:
Recent commits move the [scientific_name] token to the field type.
Example of using it in the formatter settings and the resulting output
20240111_scientific_name

tripal_chado/src/Plugin/Field/FieldType/ChadoOrganismTypeDefault.php
```diff
@@ -3,131 +3,268 @@
...
```
spficklin:
```suggestion
    return 'organism_scientific_name';
```

spficklin (changes requested):

tripal_chado/config/install/tripal.tripalentitytype_collection.general_chado.yml
dsenalik: @spficklin why did you suggest this? This has now broken the title for publications.
Under Available tokens it says `[pub_title]` instead of `[publication_title]`

spficklin: @laceysanderson would like to fix this after this gets merged so she can marke sure this doesn't keep happening.  So let's restore it back the way you had it.

dsenalik:
As of commit https://github.com/tripal/tripal/pull/1743/commits/76444bfbc40d0af4cd5a4915dc3cc55dbe7778fc the publication title is working 
20240111_pub

dsenalik:
That last commit https://github.com/tripal/tripal/pull/1743/commits/2fa442e54de6879bfe483af695926d30a362947e fixes cases where the foreign key is a different name - these are listed here https://github.com/tripal/tripal_doc/issues/32#issuecomment-1837532399
They require a tiny bit more in the yaml and support by the field.
You can test on the biosample type, both biosample provider (contact) and taxon_id (organism) have this odd column naming.

This, except for undiscovered bugs 🙈, should be ready!

spficklin (approved):

PR #1745, opened 2024-01-10
Fixes Usage of TripalJob

spficklin:
# Bug Fix

## Description

The interface for importing multiple collection types works but would never report to the user the command-line needed to run the job. It turns out that code was use the Drupal service for jobs, but everywhere else that we are using jobs, we uesd the old legacy API function.

This PR makes 3 fixes.

1. It fixes the TripalJob::create() function so it prints out the command-line instructions for running the job.
2. It updates the tripal_add_job() legacy API function to fully use the TripalJob service. 
3. It updates everywhere that we used the old `tripal_add_job` to use the TripalJob service.

## Testing?

On a fresh install you'll need to test everywhere that a `tripal_add_job` function got replaced.  This includes
1. Installation of Chado
2. Preparing Chado.
3. Clone Chado -- the running of the job shoudl work but cloning at the moment is broken.
4. Running any importer
5. Publising a Tripal Content
6. Populating a materialized view

All jobs should launch!

### DIFF
--- a/tripal/src/Form/ImportTripalEntityTypeCollection.php
+++ b/tripal/src/Form/ImportTripalEntityTypeCollection.php
@@ -77,7 +77,7 @@ final class ImportTripalEntityTypeCollection extends FormBase {
 
     // Now create the job.
     $job_submitted = \Drupal::service('tripal.job')->create([
-      'job_name' => "Import Tripal Entity Type Collection",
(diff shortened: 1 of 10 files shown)


dsenalik (approved):
Nice.
So far I tested everything except prepare chado - Jobs all look good and run.
(But now I am stuck on the gff3 loader for issue #1742)

PR #1748, opened 2024-01-11
Chado fix for box_range

spficklin:
# Bug Fix

## Description

There is a bug in the `box_range` function of Chado that doesn't cast a 0 value passed to `create_point` to a bigint.  This causes the an error when cloning a Chado schema or deleting a feature.   

image

This PR applies a change to the Chado SQL file in Tripal that includes a fix in the Chado repo here: https://github.com/GMOD/Chado/pull/105/files

## Testing?

1. Install a fresh version of Tripal using this branch. 
2. Install Chado and then try to clone it. Cloning should work just fine.  
3. Create an Organism content type, then a Gene content type using the Tripal interface.  Next use your favorite SQL tool to manually delete the gene from the feature table.  It should delete without a problem.

### DIFF
--- a/tripal_chado/chado_schema/chado-only-1.3.sql
+++ b/tripal_chado/chado_schema/chado-only-1.3.sql
@@ -11046,3 +11046,51 @@ ALTER TABLE ONLY treatment
 
 ALTER TABLE ONLY treatment
     ADD CONSTRAINT treatment_type_id_fkey FOREIGN KEY (type_id) REFERENCES cvterm(cvterm_id) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED;
+    
+---
+--- The following additions are part of the pull request on Chado:
+--- https://github.com/GMOD/Chado/pull/105 
+--- Once a new version of Chado is made these changes will be integrated
+--- but we need them now to prevent create_point errors when deleting features
+--- or cloning Chado.
+---
+
+CREATE OR REPLACE FUNCTION fill_cvtermpath(INTEGER) RETURNS INTEGER
+    LANGUAGE plpgsql
+    AS $_$
+DECLARE
+    cvid alias for $1;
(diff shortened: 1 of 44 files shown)


dsenalik:
There is a ton of changes that are just additional white space at the ends of lines.
Does anybody know how to filter those out?

Here it is: `git diff HEAD..4.x --ignore-space-change`

dsenalik:
1. I can't replicate any error cloning the schema on the current 4.x branch on a docker.
```
2024-01-12 10:55:12
Tripal Job Launcher
Running as user 'drupaladmin'
-------------------
2024-01-12 10:55:12: Job ID 1.
2024-01-12 10:55:12: Calling: tripal_chado_clone_schema(chadoschema, clone_4x_branch)
```
It appears to be cloned correctly, and I can add it to Tripal
2024-01-12_cloned

 2. I can confirm the error deleting a feature on the 4.x branch
```
DELETE FROM chadoschema.feature WHERE feature_id=1;
ERROR:  function create_point(integer, bigint) does not exist
LINE 1: SELECT box (create_point(0, $1), create_point($2,500000000))
                    ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  SELECT box (create_point(0, $1), create_point($2,500000000))
CONTEXT:  SQL function "boxrange" during inlining
SQL statement "DELETE FROM ONLY "chadoschema"."featureloc" WHERE $1 OPERATOR(pg_catalog.=) "feature_id""
```

 3. On this branch both work correctly, I could delete the feature
 ```
sitedb=> DELETE FROM chadoschema.feature WHERE feature_id=1;
DELETE 1
```

laceysanderson (approved):
Looks perfect 👍 This fix is going to be So Nice! I get it often when I try to dump/backup my database in Tripal 4.

PR #1755, opened 2024-01-13
rename chado linker property field

### LINKED ISSUE #1754: "Linker" is no longer a needed field distinction
### Task Description

Back in Tripal 3 we distinguished fields that used values in linked tables with a "Linker" nomenclature.
We no longer need this distinction in Tripal 4, a field can handle both base tables and tables connected through a linker table in the same field.

So for consistency with other fields, the task here is simple, rename the "Chado Property" field file names and field names. `ChadoLinkerPropertyTypeDefault.php` becomes `ChadoPropertyTypeDefault.php`, `chado_linker_property_type_default` becomes `chado_property_type_default`
and so on for the widget and formatter.

### Branch Name

tv4g1-issue1754-rename-chado-linker-property-field

dsenalik:
# Tripal 4 Core Dev Task

## Description
Back in Tripal 3 we distinguished fields that used values in linked tables with a "Linker" nomenclature.
We no longer need this distinction in Tripal 4, a field can handle both base tables and tables connected through a linker table in the same field.

So for consistency with other fields, the task here is simple, rename the "Chado Property" field file names and field names. `ChadoLinkerPropertyTypeDefault.php` becomes `ChadoPropertyTypeDefault.php`, `chado_linker_property_type_default` becomes `chado_property_type_default`
and so on for the widget and formatter.

Note that when manually adding a property field, the name is already just "Chado Property".
20240113_property

## Testing?
Build a new docker or site. There should be no error when the publication content type is created.
Enter a publication and save values for some of the property fields like Authors, URL and make sure they save and display as expected.


spficklin (approved):

spficklin:
I like this change.  I was able to use the field with this change just fine.  I approve of merging

laceysanderson (approved):
This was also checked quickly behind the scenes by myself and @Ferrisx4 -definitely ready to merge :-) Thanks all!

PR #1756, opened 2024-01-13
Updating release numbers to 4.0.0-alpha3

spficklin:
Simple change to the version numbers of our Tripal modules.


dsenalik (approved):
💯
edit: actually 99, because of the branch name 😄

Issue #1761, opened 2024-01-23
Organism name is incorrectly case sensitive for taxonomy loader


dsenalik:
### Tripal Version

4.alpha2

### Drupal Version

NA

### PostgreSQL Version

NA

### PHP Version

NA

### General Description

The taxonomy loader, when you import existing species
![20240123_existing_species](https://github.com/tripal/tripal/assets/8419404/c1d200f5-2ed8-4a7e-b217-27437b03406b)
is failing due to NCBI returning all lower case organism names.

### Steps to reproduce

1. Create a real organism, e.g. `Citrus sinensis`
2. Run the taxonomy importer, just with defaults, don't enter anything.
3. When the job is run you will see ` [warning] WARNING: Partial match "citrus sinensis" to query "Citrus sinensis", no taxid available`
This should have matched, ignoring case. NCBI returned all lower case.

dsenalik:
The same thing happens in Tripal 3. Probably a change in the NCBI API.

pdtouch:
Thanks.

PR #1763, opened 2024-01-23
Tripal 3: Taxonomy importer name comparison now case insensitive

dsenalik:
# Bug Fix

## Description
Fix for taxonomy importer when importing existing species
20240123_existing_species

## Testing?

 1.  Create a real organism that exists in the NCBI taxonomy, e.g. `Citrus sinensis` - Make sure the first letter of the genus is capitalized to test this fix. You only need to enter genus and species.
 2.  Run the taxonomy importer, just with defaults, don't enter anything.
 3.  Before this PR, when the job is run you will see  `[warning] WARNING: Partial match "citrus sinensis" to query "Citrus sinensis", no taxid available`
   With this PR applied, you should see something like `Import phylotree: Associated <em class="placeholder">Citrus sinensis</em> to organism_id: <em class="placeholder">1</em>`
 4. Verify that info from NCBI has been imported into the `organismprop` table, e.g.
```
select * from chado.organismprop;
 organismprop_id | organism_id | type_id |                                                              value
-----------------+-------------+---------+-----------------------------------------------------------------------------...
               1 |           1 |      88 | Plants and Fungi                                                            
               2 |           1 |      87 | Standard                                                                    
               3 |           1 |      86 | 1                                                                           
               4 |           1 |      85 | Standard                                                                    
               5 |           1 |      83 | cellular organisms; Eukaryota; Viridiplantae; Streptophyta; Streptophytin...
               6 |           1 |      84 | 1                                                                   
```

### DIFF
--- a/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
+++ b/tripal_chado/includes/TripalImporter/TaxonomyImporter.inc
@@ -629,7 +629,7 @@ class TaxonomyImporter extends TripalImporter {
           if (isset($xml->TranslationStack)) {
             $matched = (string) $xml->TranslationStack->TermSet->Term;
             $matched = preg_replace('/\[All Names\]/', '', $matched);
-            if ($matched == $sci_name) {
+            if (strcasecmp($matched, $sci_name) == 0) {
               $taxid = (string) $xml->IdList->Id;
             }
             else {


laceysanderson (approved):
Same change as already tested in PR #1761 but in Tripal 3. Since the code change is so small/clean and uses a core PHP function I feel good approving/merging this with code review only.

PR #1768, opened 2024-01-26
Add function to get default schema and associated tests

Ferrisx4:
# New Feature

## Description

Adds a function within the ChadoSchema class to get the current default Chado schema

## Testing?

If writing code within a class:

```php
$default_chado_schema = $this->chado_connection->schema()->getDefault();
```
and if you are not within a class, this function can be called through the associated service:
```php
$default_chado_schema = $chado->schema()->getDefault();
```

This change also adds a PHPUnit test to get the current default Chado schema, it should match the test Chado schema string, `_test_chado_0123456789abcdef`, for example.

---

Thanks @laceysanderson for the suggestion and the sample code.

### DIFF
--- a/tripal_chado/tests/src/Functional/Database/ChadoSchemaTest.php
+++ b/tripal_chado/tests/src/Functional/Database/ChadoSchemaTest.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Drupal\Tests\tripal_chado\Functional;
+
+/**
+ * Tests for Chado Schema implementation of Tripal DBX Connection.
+ * 
+ * @group Tripal
+ * @group Tripal TripalDBX
+ * @group Tripal TripalDBX Chado
+ * @group TripalDBX Chado
+ */
+class ChadoSchemaTest extends ChadoTestBrowserBase {
+  /**
+   * Test the ChadoSchema::getDefault() method.
+   * 
+   * We will test that the default chado in this test returns the 'testchado' string, which should be the default chado.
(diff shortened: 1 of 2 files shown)


tripal_chado/tests/src/Functional/Database/ChadoSchemaTest.php
```diff
@@ -0,0 +1,29 @@
...
+   * There is no way to programatically set the default chado, so we will not test changing the default and seeing if the default is still reported correctly afterwards.
+   */
+  public function testGetDefault() {
+    $this->createTestSchema(ChadoTestBrowserBase::INIT_CHADO_EMPTY);
+    $chado = \Drupal::service('tripal_chado.database');
+    $default_chado_schema = $chado->schema()->getDefault();
+
+    // Test if the reported Chado version matches the test chado format,
+    // e.g. _test_chado_h87g97hkln64vy76
+    $this->assertRegExp('/(\_test\_chado\_[\w]{16})\b/', $default_chado_schema, "The default Chado could not be reliably determined.");
```
laceysanderson: Your test looks good :-) I would just add one more assert to check that the test schema name and the default chado schema name are the same:
```suggestion
    $this->assertRegExp('/(\_test\_chado\_[\w]{16})\b/', $default_chado_schema, "The default Chado schema returned did not match the format we expected within the testing environment.");
    
    // Test if the default chado schema returned matches the
    // one we created.
    $this->assertEquals($this->testSchemaName, $default_chado_schema,
      "The default chado schema name we retrieved did not match the test chado schema we just created.");
```

PR #1769, opened 2024-01-27
Add PHP8.3 + Drupal 10.2 to automated testing and docker builds

### LINKED ISSUE #1760: Add PHP8.3 + Drupal 10.2 to automated testing
### Task Description

PHP 8.3 was released Nov. 23
https://www.php.net/releases/8.3/en.php
Drupal 10.2 was released Dec. 15
https://www.drupal.org/project/drupal/releases/10.2.0
Add these newer versions to automated testing and docker builds
Add these versions to automated testing

Note that Drupal 10.0 and 10.1 do not support PHP 8.3

### Branch Name

tv4g9-issue1760-php8.3

dsenalik:
# Tripal 4 Core Dev Task

## Description
PHP 8.3 was released Nov. 23
https://www.php.net/releases/8.3/en.php
Drupal 10.2 was released Dec. 15
https://www.drupal.org/project/drupal/releases/10.2.0
Add these newer versions to automated testing and docker builds
Note that Drupal 10.0 and 10.1 do not support PHP 8.3

## Deprecations and other problems to fix

 - [x] fixed with commit https://github.com/tripal/tripal/pull/1769/commits/6c6b4e966fe9a7f7d7e102a4d0cc587a5b858aa0
Calling get_class() without arguments is deprecated
https://php.watch/versions/8.3/get_class-get_parent_class-parameterless-deprecated
 ```
tripal_chado/src/TripalImporter/ChadoImporterBase.php:106:    $chado->useTripalDbxSchemaFor(get_class());
tripal_chado/src/Task/ChadoPreparer.php:139:    $chado->useTripalDbxSchemaFor(get_class());
```
 - [x] fixed with commit https://github.com/tripal/tripal/pull/1769/commits/dcec10a8fc61c3399253e19a15a3b6bf228d9cf1
Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "test62608753chado_installations" does not exist
 This is a search path issue, this table exists in the `public` schema, but for some reason in drupal 10.2 `public` is not included in the search path. Example dump of earlier working version:
```
$this->connectionOptions['init_commands'] = array(1) {
  ["search_path"]=>
  string(55) "SET search_path="_test_chado_1653741241664665","public""
}
```
This in turn is caused by `tripal/src/TripalDBX/TripalDbx.php` function `getDrupalSchemaName()`
because formerly, `$connection_options['driver']` would contain the value `pgsql`. Under drupal 10.2 it contains `Drupal\pgsql\Driver\Database\pgsql`

 - [x] Fixed with https://github.com/tripal/tripal/pull/1769/commits/e6b75392ef394e1b906e0e0240028aeba23b0cb0
TripalRoutePermissionsTest::testTripalAdminPages
The privileged user should be able to access this admin page: Data Loaders which should be at 'admin/tripal/loaders'.
`tripal/tripal.routing.yml` defines permissions for this as:
```
# Importer Listing.
tripal.data_loaders:
  path: 'admin/tripal/loaders'
  defaults:
    _controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'
    _title: 'Data Loaders'
  requirements:
    _permission: 'administer tripal+allow tripal import'
```
The user created in the test
`$userTripalAdmin = $this->drupalCreateUser(['administer tripal']);`

If you change the order, the following will also cause the same error
`admin/tripal/storage`
`admin/tripal/extension`
`admin/tripal/files`
and what they all have in common is the controller
`_controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'`
however, the path `'admin/tripal` uses this same controller and is accessible in testing.
Fails also for just `admin` path

~~It might or might not be this issue https://www.drupal.org/project/drupal/issues/3413508~~

I needed to add an additional module, and additional permissions to the admin user.
I also removed one path from testing, `admin/tripal/extension`. This is because under Drupal 10.2 we no longer have access to this because there are no extensions defined in the testing environment. I don't like leaving it out, but this is the simple solution. On a "real" site you see
20240207_noextensions

 - [x] Fixed with commit https://github.com/tripal/tripal/pull/1769/commits/a5a1f292d73904cd5cfacb422e922b815865e0ea
 Drupal\Tests\tripal_chado\Functional\GFF3ImporterTest::testGFFImporterSimpleTest
unserialize(): Extra data starting at offset 616 of 617 bytes
We just need to remove the trailing return character from our cached feature before unserializing, this was added when we stored it in the cache file.

## Testing
 1. Verify that automated testing now includes PHP8.3 and that all tests pass
 2. Verify that you can build a PHP8.3 docker. For example `docker build --tag=tripaldocker:testing-1760 --build-arg drupalversion="10.2.x-dev" --build-arg chadoschema='testchado' --file tripaldocker/Dockerfile-php8.3-pgsql13 ./`

### CODE REVIEW

**Comment by dsenalik** [comment]
I should have checked this first, but only Drupal ~10.2 (and 11.0🙈) supports PHP 8.3
https://www.drupal.org/docs/getting-started/system-requirements/php-requirements

**Comment by dsenalik** [comment]
Debugging notes: after much painful back-tracing of the cause, it appears to be a problem with the menu generation in the test environment, as stored in the `public.menu_tree` table.
For example, the route `tripal.data_loaders` in the test environment appears to have no child menus.
In the "real" environment, the `has_children` column is set to `1`, for example
```
select menu_name, mlid, id, parent, route_name, has_children from menu_tree where p1=6 and p2=89 and depth=3 and mlid=92;
 menu_name | mlid |         id          |    parent    |     route_name      | has_children 
-----------+------+---------------------+--------------+---------------------+--------------
 admin     |   92 | tripal.data_loaders | tripal.admin | tripal.data_loaders |            1
                                                                                          ^
```

but in the testing environment it is `0`. The next step is to determine why this is the case...

File: tripal/tests/src/Functional/TripalRoutePermissionsTest.php [hunk 1]
```diff
@@ -19,7 +19,7 @@ class TripalRoutePermissionsTest extends BrowserTestBase {
   // protected $htmlOutputEnabled = TRUE;
   protected $defaultTheme = 'stark';
 
-  protected static $modules = ['tripal', 'file', 'field_ui'];
+  protected static $modules = ['tripal', 'tripal_chado', 'file', 'field_ui'];
```
laceysanderson: This shows a circular dependency 🤔 Which routes specifically needed this added? Instead of adding tripal_chado here we should create a TripalChadoRoutePermissionsTest in the tripal_chado module that tests those specific routes and then remove them from here.

laceysanderson: I see in the PR description that the following were all causing inaccessibility issues: loaders, storage, files. Did none of these work without tripal_chado added?

laceysanderson (changes requested):
Changes all look good and make sense although I'm still having some concerns about the route test as described more in an in-line comment.

tripal/tests/src/Functional/TripalRoutePermissionsTest.php
dsenalik: Correct, without tripal_chado all of those failed.

dsenalik:
As Lacey noted, I created a circular dependency when I added the `tripal_chado` module to a `tripal` test. There were two routes that required the `tripal_chado` module loaded, so I moved these to the `tripal_chado` module for testing as of commit https://github.com/tripal/tripal/pull/1769/commits/1c5cc49f5a159a92f94d3394a59bf1319704818f
These are 
`'Data Loaders' => 'admin/tripal/loaders',`
`'Data Storage' => 'admin/tripal/storage',`
Since I needed somewhere to put the new test file in tripal_chado, and to be consistent, I moved the tripal route tests into a new subfolder `tripal/tests/src/Functional/Permissions` and created a similar `tripal_chado/tests/src/Functional/Permissions` subfolder in tripal_chado.

laceysanderson (approved):
Changes all look great ✅ 
✅ built the docker image and ran the container locally
✅ manually checked some random functionality such as creating/editing content, managing display, running Tripal jobs, checking various admin listings, editing content types and it all worked as expected.

I also upped the `latest` docker tag to point to D10.2 + PHP 8.3 :-) 

This is ready to merge (once the tests finish re-running)! 🎉 
Thank you @dsenalik!!! ❤️

dsenalik:
Thank you for fixing the latest tag! 😀

PR #1771, opened 2024-02-02
Tv4g2 issue1758 chado custom table functionality

Ferrisx4:
# Bug Fix

## Description

This change serves to provide/upgrade Chado Custom Table functionality and also introduces tests for this functionality.

## Testing?

This change updates the Chado Custom Table Manager service. These can be tested by calling them in code:

```php
// Get an instance of the manager:
$ct_service = \Drupal::service('tripal_chado.custom_tables');

// Create a table, set 'chado' as the schema:
$custom_table = $ct_service->create('test_custom_table', 'chado');

// Get a ChadoCustomTable object for the new table later:
$custom_table = $ct_service->loadById('test_custom_table', 'chado');

// Get an array of custom tables:
$custom_tables = $ct_service->getTables('chado');
```

The functions above can also be called without specifying the Chado schema. They will use the newly created Chado getDefault() function to determine which schema to use.


tripal_chado/src/Services/ChadoCustomTableManager.php
```diff
@@ -93,6 +93,10 @@ public function loadbyName(string $table_name, string $chado_schema = NULL) {
    */
   public function findByName(string $table_name, string $chado_schema = NULL) {
 
+    // Retrieve the default name of the Chado schema if it's not provided.
+    if ($chado_schema === NULL) {
+      $chado_schema = chado_get_schema_name('chado');
```
laceysanderson: We'll want to switch these calls over to use the new method added in PR #1768.

laceysanderson (changes requested):
We'll also want to add dependencies to this service. Specifically, the Drupal connection (`database`) and the chado connection (`tripal_chado.database`). This will then allow you to remove all instances of `\Drupal::service` and to pass these connection into the ChadoCustomTable object on creation.

laceysanderson:
Thanks for the update @Ferrisx4. It looks like the [tests are failing](https://github.com/tripal/tripal/actions/runs/7820326061/job/21334766515?pr=1771#step:3:2763) because the Materialized Views service extends the Custom table service. You see this here: https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Services/ChadoMviewsManager.php#L15. This means you will need to add dependency injection to the materialized view service as well and then pass the dependencies into the parent on the line I linked.

Ferrisx4:
Thanks @laceysanderson, I've actually made this change and am testing locally before pushing (I'll also pull in the merge of 4.x you just made).
🤞

Ferrisx4:
Hi @laceysanderson, I think we've got something here! Github is blocking because of the review but I think the requirements have been satisfied. 
On the other hand, I can add a few more failure state tests if we want as well.

laceysanderson:
Yes, it's waiting on me to approve the changes. I'll test it right now :-)

laceysanderson (approved):
Works good and code changes are perfect :-) Thanks @Ferrisx4

Ferrisx4:
Thanks, I will merge it then!

PR #1681, opened 2023-11-07
Drupal 9 End of Life

dsenalik:
# Tripal 4 Core Dev Task

## Description
This removes all of the Drupal 9 automated tests and configurations, as well as PHP 8.0 which was only used for Drupal 9.

## Testing?
Let's build a docker and see if it works. Looks good
pr1681

PR #1775, opened 2024-02-06
Update Dockerfile Drupal version to 10.1.0 for PHP8.2

exnehilo7:
# New Feature

## Description

Changed the PHP 8.2 Dockerfile Drupal's version to 10.1.0. Updated the site name and var/www/drupal subfolder to reflect the versions used.

A colleague needed to add extensions that required Drupal 10+.

## Testing?

From my limited knowledge of Docker containers and testing within, if the build succeeds and the webpage is accessible, that's a pass! (Was built successfully on a Red Hat Ubuntu 20.04.6 VM)

### DIFF
--- a/tripaldocker/Dockerfile-php8.2-pgsql13
+++ b/tripaldocker/Dockerfile-php8.2-pgsql13
@@ -1,6 +1,6 @@
 FROM php:8.2-apache-bullseye
 
-ARG drupalversion='10.1.0'
+ARG drupalversion='~10.1.0'
 ARG modules='devel devel_php'
 ARG chadoschema='chado'
 ARG installchado=TRUE
--- a/tripaldocker/Dockerfile-php8.2-pgsql13
+++ b/tripaldocker/Dockerfile-php8.2-pgsql13
@@ -178,7 +178,7 @@ RUN cd /var/www/drupal \
   --account-name=drupaladmin \
   --account-pass=some_admin_password \
   --site-mail="drupaladmin@localhost" \
-  --site-name="Tripal 4.x-dev on Drupal 10.1.0" \
+  --site-name="Tripal 4.x-dev on Drupal ${drupalversion}" \
   && service apache2 stop \
   && service postgresql stop
 
--- a/tripaldocker/Dockerfile-php8.2-pgsql13
+++ b/tripaldocker/Dockerfile-php8.2-pgsql13
@@ -213,6 +213,7 @@ RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
 ## Make global commands.
 RUN ln -s /var/www/drupal/vendor/phpunit/phpunit/phpunit /usr/local/bin/ \
   && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
+  && ln -s /var/www/drupal /var/www/drupal9
 
 ## Set the working directory to DRUPAL_ROOT
 WORKDIR /var/www/drupal/web
--- a/tripaldocker/Dockerfile-php8.2-pgsql13
+++ b/tripaldocker/Dockerfile-php8.2-pgsql13
@@ -213,7 +213,6 @@ RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
 ## Make global commands.
 RUN ln -s /var/www/drupal/vendor/phpunit/phpunit/phpunit /usr/local/bin/ \
   && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
-  && ln -s /var/www/drupal /var/www/drupal10
 
 ## Set the working directory to DRUPAL_ROOT
 WORKDIR /var/www/drupal/web
--- a/tripaldocker/Dockerfile-php8.2-pgsql13
+++ b/tripaldocker/Dockerfile-php8.2-pgsql13
@@ -1,13 +1,13 @@
 FROM php:8.2-apache-bullseye
 
-ARG drupalversion='9.4.x-dev'
+ARG drupalversion='10.1.0'
 ARG modules='devel devel_php'
 ARG chadoschema='chado'
 ARG installchado=TRUE
 
 # Label docker image
 LABEL drupal.version=${drupalversion}
-LABEL drupal.stability="development"
+LABEL drupal.stability="production"
 LABEL tripal.version="4.x-dev"
 LABEL tripal.stability="development"
 LABEL os.version="bullseye"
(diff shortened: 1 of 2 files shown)


tripaldocker/Dockerfile-php8.2-pgsql13
```diff
@@ -213,7 +213,7 @@ RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
 ## Make global commands.
 RUN ln -s /var/www/drupal/vendor/phpunit/phpunit/phpunit /usr/local/bin/ \
   && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
-  && ln -s /var/www/drupal /var/www/drupal9
+  && ln -s /var/www/drupal /var/www/drupal10
```
dsenalik:
```suggestion
  && ln -s /var/www/drupal /var/www/drupal9
```

dsenalik: I am pretty sure the drupal9 directory is a relic and no longer needed since PR #1681 so this line can just be deleted. A recursive grep does not find `drupal9` mentioned anywhere else

dsenalik (changes requested):
Good catch fixing that drupal 9 reference. I didn't notice it when I did PR #1681 
Everything else looks good except my one comment below

tripaldocker/Dockerfile-php8.2-pgsql13
exnehilo7: Thanks! Glad to have helped. 🙂

Sounds good. Looks like the files in the drupal and drupal10 folders within the test build are the same.

dsenalik: That is correct, but we do not need a drupal10 folder. You can remove the line 216

dsenalik: You can just click on the "Commit suggestion" button. 
Some year soon it will be drupal11 and we don't want to be tied to a specific version.

exnehilo7:
The **Test Coverage on Drupal 10.1.x-dev + PHP 8.1 / Test Coverage** check is failing on **Publish code coverage to Code Climate** at line 43:

```
Error: response from https://api.codeclimate.com/v1/test_reports.
HTTP 401: You are not authorized for this action
```

Sorry, I'm still pretty new to a repo's full capabilities. Is this check something someone with the appropriate rights has to kick off?

tripaldocker/Dockerfile-php8.2-pgsql13
exnehilo7: Ah, "Resolve conversation" doesn't include accepting changes 🤦‍♂️. Thanks!

tripaldocker/Dockerfile-php8.2-pgsql13
```diff
@@ -213,7 +213,6 @@ RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
 ## Make global commands.
 RUN ln -s /var/www/drupal/vendor/phpunit/phpunit/phpunit /usr/local/bin/ \
   && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
-  && ln -s /var/www/drupal /var/www/drupal9
```
laceysanderson: This is here for backwards compatibility with old tripaldockers. The main directory is now not specific to verions (i.e. `/var/www/drupal`) so this line just adds a symbolic link for backwards compatiblity with previous images. We would like to keep it in place for a bit longer.

tripaldocker/Dockerfile-php8.2-pgsql13
```diff
@@ -178,7 +178,7 @@ RUN cd /var/www/drupal \
   --account-name=drupaladmin \
   --account-pass=some_admin_password \
   --site-mail="drupaladmin@localhost" \
-  --site-name="Tripal 4 on Drupal 10 DEVELOPMENT" \
+  --site-name="Tripal 4.x-dev on Drupal 10.1.0" \
```
laceysanderson: Use the Drupal version arguement here so that it changes based on the version requested. This is because on line 3 you specify the default version of Drupal to use. However, when building the image you can pass in a build-arg to specify a different drupal version and we do not want a mismatch in the name :-) 
 
```suggestion
  --site-name="Tripal 4.x-dev on Drupal ${drupalversion}" \
```

tripaldocker/Dockerfile-php8.2-pgsql13
```diff
@@ -1,13 +1,13 @@
 FROM php:8.2-apache-bullseye
 
-ARG drupalversion='9.4.x-dev'
+ARG drupalversion='10.1.0'
```
laceysanderson: There are actually multiple versions of Drupal 10.1 and 10.1.0 specifically has a security advisory on it. Since this version is used by composer we can use a tilde here to specify, any version between 10.1.0 and up to but not including 10.2.0. This ensures we get the most recent stable release of Drupal 10.1

```suggestion
ARG drupalversion='~10.1.0'
```

laceysanderson (commented):
Thanks for the contribution! and good catch :-) I have a few suggested improvements with clarification. Let me know if any of my suggestions cause issues with your use case and we can chat further :-)

dsenalik:
Thanks @laceysanderson 🧡

dsenalik:
@exnehilo7 I made an issue for the testing error you observed #1776

tripaldocker/Dockerfile-php8.2-pgsql13
exnehilo7: Doh, thanks. Using 10.1.0 for our purposes I had rigid version in the brain. 😅

tripaldocker/Dockerfile-php8.2-pgsql13
exnehilo7: @laceysanderson Thanks for the suggestion and teaching me something new! That's really cool.

tripaldocker/Dockerfile-php8.2-pgsql13
```diff
@@ -213,7 +213,6 @@ RUN mv /app/tripaldocker/init_scripts/supervisord.conf /etc/supervisord.conf \
 ## Make global commands.
 RUN ln -s /var/www/drupal/vendor/phpunit/phpunit/phpunit /usr/local/bin/ \
   && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
```
dsenalik:
```suggestion
  && ln -s /var/www/drupal/vendor/drush/drush/drush /usr/local/bin/ \
  && ln -s /var/www/drupal /var/www/drupal9
```

dsenalik (changes requested):
Here's the reversion based on Lacey's suggestion

tripaldocker/Dockerfile-php8.2-pgsql13
exnehilo7: Sounds good, I'll toggle it back.

dsenalik (approved):

laceysanderson:
So apparently we missed one small thing in PR review. Please test PR #1777 locally to confirm that the symbolic link is working.

dsenalik:
edit: nevermind, I was answering the wrong question...

PR #1782, opened 2024-02-13
Allow fields to link to their corresponding entity

### LINKED ISSUE #1632: Tripal 4 fields should link to entities
### Task Description

In Tripal 3, a field will link to it's corresponding entity if it is a content type.
For example, on the gene content type, the organism field links to the corresponding entity:
gene-tripal-three

So far in Tripal 4 this is not yet implemented. For example
gene-tripal-four

The task before us is to implement these links for Tripal 4 fields.
We may want to create a new service to look up entity types based on the base table.

### Branch Name

tv4g1-issue1632-add-field-links-to-entities

### PR #1782

dsenalik:
# New Feature

## Description
None of the Tripal 4 fields currently are "clickable" like they were in Tripal 3.
This PR will eventually make this possible.  For example
2024-02-26_organism-with-link

## Tasks
 - 🙈 ~~Create a new service to generate the entity titles for either one or all records. For example~~
```
    $title_manager = \Drupal::service('tripal.tripal_entity.title');
    $titles = $title_manager->getEntityTitles($datastore, $bundle, $record_id);
```
 ~~This will be a chunk of the code from the publish service, so make it generic enough that the publish service can also use it.~~

 - 🙈 ~~Update the publish service to use the new `tripal.entity_title` service.
 To implement this, make it a child class of the class above.~~
```
use \Drupal\tripal\Services\TripalEntityTitle;
class TripalPublish extends TripalEntityTitle {`
```

 - Abandoned these first two tasks, initial implementation is very slow at ~1.8 seconds to return the url. Generating a title has a large amount of overhead for setup. This was okay for publish, but not for this. Implement an alternative method.

 - [x] Create a new service `tripal.tripal_entity.lookup` to generate either the entity ID or the local URL for a record.
~~This will internally call the `tripal.tripal_entity.title` service.~~

 - [x] Also in this service implement a way to look up a bundle id from a CV term. Such as, if the field links to a gene, we need to know that the bundle name is gene. We can get this from the CV term, which in the field formatter can be retrieved with
```
  $item_settings = $item->getDataDefinition()->getSettings();
  $termIdSpace = $item_settings['termIdSpace'];
  $termAccession = $item_settings['termAccession'];
```
 - [x] Create a function to look up not-null columns in a chado table `getNotNullColumns()`
This will be the only slow step, but we can cache the result with `\Drupal::cache()->set($cache_id, $cache_values);` We only need a single cached set of values for each base table that shows up as a content type, and we don't even need to worry about expiring it after some time because this information won't change.

The reason we want this information is that the chado table and column for the corresponding field is guaranteed to be populated for every entity, because of the non-null constraint.

 - [x] Implement the entity lookup as follows
Select one of the entity tables with a not-null constraint. This insures there is a row for every published entity. Since we know the chado record_id, we can then look for the corresponding entity id with a query similar to this
```
$required_fields = $this->getRequiredFields($bundle_id, $entity_type);
$required_field = array_key_first($required_fields);
$pkey_property_id = 'record_id';  // This assumes all of our fields use this id - so far this is true
$required_property = $required_field . '.' . $pkey_property_id;
$query = \Drupal::entityQuery($entity_type)
      ->condition('type', $bundle_id)
      ->condition($required_property, $record_id, '=')
      ->accessCheck(TRUE);
// The values of the array are always entity ids. The keys will be
// revision ids if the entity supports revision and entity ids if not.
$ids = $query->execute();
```

 - [x] A high level function to simplify code in the  field formatters
~~public function getFieldUrl($displayed_string, $record_id, $item_settings)~~
`$renderable_item = $lookup_manager->getRenderableItem($displayed_string, $values['entity_id']);`

 - [x] Entity lookup is now implemented in the tripalTypes function for each field using a function callback.
```
    // This property will store the Drupal entity ID of the linked chado
    // record, if one exists.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'entity_id', self::$drupal_entity_term, [
      'action' => 'function',
      'drupal_store' => TRUE,
      'namespace' => self::$chadostorage_namespace,
      'function' => self::$drupal_entity_callback,
      'fkey' => self::$object_id,
    ]);
```
The field should then have clickable links, for example
2024-02-13_issue1632

 - [x] Handle the cases where the chado base table does not exactly match the drupal table name derived from the content type name, e.g. the `arraydesign` table contact reference is in the table `tripal_entity__array_design_manufacturer` because the content type id is `array_design`
 This also happens for any of the feature table content types, analysis (genome_assembly), etc.

- [x] Figure out URL generation - ~~currently hardcoded because of multi-second time required.~~
The solution was this `$url_object = Url::fromRoute('entity.tripal_entity.canonical', ['tripal_entity' => $entity_id]);` which takes only a few milliseconds.

 - [x] Move the entity lookup to the tripalTypes field functions. The entity will now be looked up when creating content and the formatter will use this value to generate the link.

 - [x] Create tests for the new service

 - [x] Create an update function to update the field tables with the new entity id property

## Testing?
 1. Import all the content type groups
 2. Start adding content, and as you go link to as much as possible. This order works well for content types that have dependencies: Contact, Project, Publication, Protocol, Array Design, Assay, Analysis, Biological Sample, Germplasm Accession, Genetic Marker, Genetic Map, and whatever else seems good.
Note on the Array Design page that the "Manufacturer" is a foreign key to contact, so we should get a link displayed.
Note on the Assay page that the "Operator" is also a foreign key to contact, however the term for this one is NCIT:C48036 which is not a content type, so no link will be added. 🙈 Likewise for Biosource Provider on Biomaterial content type...
 3. Make sure this field works
```
    -   name: analysis_software
        content_type: analysis
        label: Program, Pipeline, Workflow or Method Name
            storage_plugin_settings:
                base_table: analysis
                base_column: program
```

 4. To test the update function, 
   4a. Build a new docker on the 4.x branch
   4b. Add a project and add a contact (without any links)
   4c. Make a link `sitedb=> insert into project_contact (project_id, contact_id) values (1,2);`
   4d. Switch branch `git checkout tv4g1-issue1632-add-field-links-to-entities`
   4e. Run updates `drush updatedb`
   4f. Publish "Project" and "Contact" content types
   4g. run the jobs
   4h. rebuild caches
   4i. Go to your project and contact and confirm that the field is a link to the other e.g.
   2024-04-17_project


dsenalik:
I am quite disillusioned with the massive overhead of generating titles (fine for publish of one bundle - but not for display of an entity), and the thing I wanted to cache can't, because it contains DB references blah blah - I think I can take a different approach, and with a bit of drupal cache it is now in the millisecond range after the first instance of a given bundle!
2024-02-16_newmethod

dsenalik:
**Progress report**
I have done a second redesign. It turns out the approach of using database columns with a not-null constraint was often failing because field names don't always match column names. Such as `analysis_software` maps to the `analysis.program` column.
A better way to do this is consult the fields directly and find a field that has `is_required` set to true.
The new function `getRequiredFields($bundle_id, $entity_type)` does this, and caches the result to improve performance. Though this method only takes about a tenth of a second to look stuff up (although all content types were not loaded).

A remaining issue of sorts is some fields have a "different" CV term for which there is no bundle. For example, the `assay_operator` field on the Assay content type has a CV Term `NCIT:C48036` (operator) instead of the term for Contact `NCIT:C47954`, so without a bundle, no link is made. I'm not doing anything about this in this PR.

dsenalik:
Another progress report. I have fixed the `function` property so that information can be retrieved by the callback function. The callback function itself is implemented in the ChadoStorage class so that all fields can use the same callback. So for every linking field, we now have this property added:
```
    // This property will store the Drupal entity ID of the linked chado
    // record, if one exists.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'entity_id', self::$drupal_entity_term, [
      'action' => 'function',
      'drupal_store' => TRUE,
      'namespace' => self::$chadostorage_namespace,
      'function' => self::$drupal_entity_callback,
      'fkey' => self::$object_id,
    ]);
```
One bit of information that needs to be passed through the `fkey` is the base table primary key column name, _e.g._ for the Analysis field it would be `analysis_id`. The field knows this, but it would be tricky and possibly error prone to recalculate it in the callback. The other parameters are static variables in the parent class, so this one block of code is the same in every linking field.

Then in the formatters we have essentially three new lines of code
`$lookup_manager = \Drupal::service('tripal.tripal_entity.lookup');`
and for each delta
`        'entity_id' => $item->get('entity_id')->getString(),`
and 
`      // Create a clickable link to the corresponding entity when one exists.`
`      $renderable_item = $lookup_manager->getRenderableItem($displayed_string, $values['entity_id']);`

And a workaround, if you have a property where `drupal_store` is set to TRUE, then you can't store anything falsey, so I couldn't store a zero to mark when there wasn't a Drupal entity. It makes the whole field disappear. I worked around this by storing a `-1` instead, because I didn't want to even think about changing chado storage any more right now. 😬

**To do:**
 -  I was trying to inject the lookup service into the `ChadoStorage` class - however the function callback is static, so I'm not sure if or how to do that.
 - Remove the last two debugging `dpm`s
 - Tests 🙈

dsenalik:
Another thing to fix. For "Gene" content type, the organism link looks like
2024-04-03_gene-organism-link
The code is
```
      $url_object = Url::fromRoute('entity.tripal_entity.canonical', ['tripal_entity' => $entity_id]);
      $renderable_item = [
        '#type' => 'link',
        '#url' => $url_object,
        '#title' => $displayed_string,
      ];
```
which does not support us having the `<i>` tag embedded in the string.

edit: Yay, finally an easy fix, commit https://github.com/tripal/tripal/pull/1782/commits/1b73bba26e9fe13c69396d2a98c84802dfb43dac

dsenalik:
This is ready for review, but it includes changes from PR #1841, so that one needs to be merged before this one can.
There is one small fix to publishing included, https://github.com/tripal/tripal/pull/1782/commits/40b1b23773254030d4a7ffce6c38db22a119ff91 , which fixes a bug for publishing that only shows up in the test environment, which means we weren't testing as well as we might have 😲.
I think I saw what was wrong for issue #1843 so if I fix that I will try to add to the publish tests on that issue.

tripal/src/Services/TripalEntityLookup.php
```diff
@@ -0,0 +1,270 @@
...
+   *   The bundle's CV Term namespace e.g. "NCIT"
+   * @param string $termAccession
+   *   The bundle's CV term accession e.g. "C47954"
+   * @param string $entity_type
+   *   The type of entity, only 'tripal_entity' is supported.
+   *
+   * @return string|null
+   *   The bundle id, or null if no match found.
+   */
+  protected function getBundleFromCvTerm($termIdSpace, $termAccession, $entity_type) {
```
laceysanderson: There is actually a much easier and more performant way to do this. Instead of getting information for all the bundles and then loading each one, you can actually use the Entity Query API to load only entities matching specific properties. In this case, this entire function can become:
```php
$bundles = \Drupal::entityTypeManager()
  ->getStorage('tripal_entity_type')
  ->loadByProperties(['termIdSpace' => $termIdSpace, 'termAccession' => $termAccession]);
if (sizeof($bundles) == 1) {
  // Array keyed by bundle id; therefore, we get the key of the only element.
  return key($bundles);
}
```

tripal/src/Services/TripalEntityLookup.php
```diff
@@ -0,0 +1,270 @@
...
+      if (($termIdSpace == $bundleIdSpace) and ($termAccession == $bundleAccession)) {
+        $bundle_id = $id;
+        break;
+      }
+    }
+    return $bundle_id;
+  }
+
+  /**
+   * Retrieve the base table for a given bundle given the bundle's CV term.
```
laceysanderson: Since this function looks up the base table given the bundle, it doesn't actually matter how the bundle was retrieved.

```suggestion
   * Retrieve the base table for a given bundle.
```

laceysanderson (commented):
More suggestions.

tripal/src/Services/TripalEntityLookup.php
```diff
@@ -0,0 +1,260 @@
...
+   * @return array
+   *   The id for the requested Drupal entity in the tripal_entity table.
+   *   Will be an empty array if there was no corresponding entity. This can happen
+   *   if the content is not published, or if this is not a content type.
+   *   If a given record was published as more than one content type, the
+   *   returned array may have more than one entity id. This would happen
+   *   in Tripal 3, for example if an analysis is published as both
+   *   "analysis", and as "Genome Assembly".
+   */
+  protected function getEntityIdFromRecordId($base_table, $record_id, $bundle_id, $entity_type) : array {
```
laceysanderson: I was worrying about the assuming the field database table names and fell down a rabbit hole 🙈 
Apparently we can use the entity query api to search for all entities with a particular field value. Furthermore, we can dig into the properties of that field!

In the following example, I am grabbing the entity_id of all tripal entities with a bundle of contact and a contact_name field whose record_id property has a value of 2.
```php
$query = \Drupal::entityQuery('tripal_entity')
  ->condition('type', 'contact')
  ->condition('contact_name.record_id', 2)
  ->accessCheck(TRUE);
$results = $query->execute();
```

This can be used to get rid of the need to know the base table and to remove the assumptions regarding database tables!
```php
$query = \Drupal::entityQuery('tripal_entity')
  ->condition('type', $bundle_id)
  ->condition($required_field . '.record_id', $record_id)
  ->accessCheck(TRUE);
$results = $query->execute();
```
This does still make the assumption that we named the primary key of the base table `record_id` 🙈

dsenalik:
Some notes about why testing is failing. It is actually detecting a problem in the testing environment. Six entities are correctly created in the testing environment table
```
sitedb=> select * from test48953458tripal_entity;
 id |   type   | uid |     title      | status |  created   |  changed   
----+----------+-----+----------------+--------+------------+------------
  1 | project  |     | Project No. 1  |      1 | 1713281067 | 1713281067
  2 | project  |     | Project No. 2  |      1 | 1713281067 | 1713281067
  3 | project  |     | Project No. 3  |      1 | 1713281067 | 1713281067
  4 | analysis |     | Analysis No. 1 |      1 | 1713281068 | 1713281068
  5 | analysis |     | Analysis No. 2 |      1 | 1713281068 | 1713281068
  6 | analysis |     | Analysis No. 3 |      1 | 1713281068 | 1713281068
(6 rows)
```
however the fields themselves are not created in the testing environment
```
sitedb=> select * from test48953458tripal_entity__project_name;
 bundle | deleted | entity_id | revision_id | langcode | delta | project_name_record_id | project_name_value 
--------+---------+-----------+-------------+----------+-------+------------------------+--------------------
(0 rows)
```
instead they show up in the "real" tables
```
sitedb=> select * from tripal_entity__project_name;
 bundle  | deleted | entity_id | revision_id | langcode | delta | project_name_record_id | project_name_value 
---------+---------+-----------+-------------+----------+-------+------------------------+--------------------
 project |       0 |         1 |           1 | und      |     0 |                      1 | 
 project |       0 |         2 |           1 | und      |     0 |                      2 | 
 project |       0 |         3 |           1 | und      |     0 |                      3 | 
(3 rows)
sitedb=> select * from tripal_entity__analysis_name;
  bundle  | deleted | entity_id | revision_id | langcode | delta | analysis_name_record_id | analysis_name_value 
----------+---------+-----------+-------------+----------+-------+-------------------------+---------------------
 analysis |       0 |         4 |           1 | und      |     0 |                       1 | 
 analysis |       0 |         5 |           1 | und      |     0 |                       2 | 
 analysis |       0 |         6 |           1 | und      |     0 |                       3 | 
(3 rows)
```
### This is issue #1853, fixed by PR #1854

tripal/src/Services/TripalEntityLookup.php
dsenalik: This assumption will be dealt with in issue #1850

dsenalik:
This PR breaks publishing on an existing site. I built a docker on the 4.x branch, then switched to this branch. I created two projects
`insert into chado.project (name) values ('Project A'), ('Project B');`
then published
`[site http://default] [TRIPAL] ERROR: SQLSTATE[42703]: Undefined column: 7 ERROR:  column "project_analysis_entity_id" of relation "tripal_entity__project_analysis" does not existLINE 3: ..._id, langcode, delta, project_analysis_record_id, project_an...                                                             ^:       INSERT INTO "tripal_entity__project_analysis"        (bundle, deleted, entity_id, revision_id, langcode, delta, project_analysis_record_id, project_analysis_entity_id, project_analysis_linker_id, project_analysis_link, project_analysis_analysis_id) VALUES(:bundle_1, :deleted_1, :entity_id_1, :revision_id_1, :langcode_1, :delta_1, :project_analysis_record_id_1, :project_analysis_entity_id_1, :project_analysis_linker_id_1, :project_analysis_link_1, :project_analysis_analysis_id_1),(:bundle_2, :deleted_2, :entity_id_2, :revision_id_2, :langcode_2, :delta_2, :project_analysis_record_id_2, :project_analysis_entity_id_2, :project_analysis_linker_id_2, :project_analysis_link_2, :project_analysis_analysis_id_2); Array(    [:bundle_1] => project    [:deleted_1] => 0    [:entity_id_1] => 1    [:revision_id_1] => 1    [:langcode_1] => und    [:delta_1] => 0    [:project_analysis_record_id_1] => 1    [:project_analysis_entity_id_1] =>     [:project_analysis_linker_id_1] =>     [:project_analysis_link_1] => 1    [:project_analysis_analysis_id_1] =>     [:bundle_2] => project    [:deleted_2] => 0    [:entity_id_2] => 2    [:revision_id_2] => 1    [:langcode_2] => und    [:delta_2] => 0    [:project_analysis_record_id_2] => 2    [:project_analysis_entity_id_2] =>     [:project_analysis_linker_id_2] =>     [:project_analysis_link_2] => 2    [:project_analysis_analysis_id_2] => )`

dsenalik:
Marking as draft until I add the update function

dsenalik:
Example of running the update on a 4.x docker using this branch
```
root@2b12a100082b:/var/www/drupal/web# drush updatedb
 -------------- ----------- --------------- ----------------------------- 
  Module         Update ID   Type            Description                  
 -------------- ----------- --------------- ----------------------------- 
  tripal_chado   10403       hook_update_n   10403 - Updates entity_id    
                                             properties added to linking  
                                             fields in PR 1782            
 -------------- ----------- --------------- ----------------------------- 


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > 

>  [notice] Update started: tripal_chado_update_10403
>  [notice] Update completed: tripal_chado_update_10403
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__organism_pub", field 
> "organism_pub"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__analysis_pub", field 
> "analysis_pub"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__analysis_project", field 
> "analysis_project"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__project_contact", field 
> "project_contact"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__project_analysis", field 
> "project_analysis"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__project_pub", field 
> "project_pub"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__study_contact", field 
> "study_contact"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__study_pub", field 
> "study_pub"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__contact_project", field 
> "contact_project"
> 
>  [notice] Message: Adding entity id to Drupal table "tripal_entity__protocol_pub", field 
> "protocol_pub"
> 
>  [notice] Message: Added 10 entity id properties
> 
```

And publish now works!

Testing procedure added at the top of this PR

laceysanderson (commented):
✅ Code looks good :-) 
✅ Love the test and it has good test coverage too!
Screenshot 2024-04-26 at 11 46 09 AM

Now onto manually testing it 🙈

laceysanderson (approved):
✅ Manual testing looks good too! 🎉 

image

Thanks for this @dsenalik!

PR #1700, opened 2023-11-28
Tripal 4 sequence coordinate field

pdtouch:
Field Sequence Coordinates

#
## Description : 

This is required for Sequence Coordinates field to run in Tripal 4. Since the YML file for the field has been updated, this field should be active for the Gene page.  The Tripal-3 version of this field can be viewed here: 
https://github.com/tripal/tripal/tree/7.x-3.x/tripal_chado/includes/TripalFields/data__sequence_coordinates

Two formatters are present one for linear view of the coordinates and other for a tabular view.

## Testing?

1. Build docker image with the code in respective directories:
```
tripal_chado/config/install/tripal.tripalfield_collection.genomic_chado.yml
tripal_chado/src/Plugin/Field/FieldType/ChadoSequenceCoordinatesDefault.php
tripal_chado/src/Plugin/Field/FieldFormatter/ChadoSequenceCoordinatesFormatterDefault.php
tripal_chado/src/Plugin/Field/FieldFormatter/ChadoSequenceCoordinatesFormatterTable.php
tripal_chado/src/Plugin/Field/FieldWidget/ChadoSequenceCoordinatesWidgetDefault.php

```
2. Build and run docker image as required and verify Tripal can be accessed from its URL.

3.  Access docker container using branch name : tv4g1-issue1414-data__sequence_coordinates
```
cntr_br_name=tv4g1-issue1414-data__sequence_coordinates
docker exec -it $cntr_br_name /bin/bash
```

4. Change max_filesize value in settings file : `/usr/local/etc/php/php.ini` to a high value say 12MB.
`service apache2 restart` and exit from container. (This step is not required to just publish gene imported via Tripal -> Data Loaders -> Chado GFF3 File Loader )

5.  Log into Drupal/Tripal GUI and create Citrus Organism Tripal Content 
Ref: https://tripal.readthedocs.io/en/latest/user_guide/example_genomics/organisms.html

6. Create Analysis content :
Ref: https://tripal.readthedocs.io/en/latest/user_guide/example_genomics/analyses.html

7.  Structure -> Tripal Content Types -> Import type collection -> Genomic Content Types for viewing Gene content.

8. Load Feature Data (gene and sequence information)
Ref: https://tripal.readthedocs.io/en/latest/user_guide/example_genomics/genomes_genes.html
(Tripal -> Data Loaders -> Chado GFF3 File Loader )

[...]

PR #1787, opened 2024-02-17
Fix for error on job view and argument listing

spficklin:
# Bug Fix

## Description

While trying to fix the issue with ChadoStorage found on PR #1700  I noticed that Tripal no longer prints output to the terminal window for some types of jobs (e.g. preparing Chado and importing of content types).  So, I went to the job for importing content types to view it I got an error.  This PR fixes that error and allows the job to be viewed.    I also noticed that the job arguments were not properly displayed.  Just a `print_r` of the array which is hard to read, so  I took a moment to fix it so that job arguments show up nicely.  

This PR does not fix the problem that the log is missing.  I'm not sure why it doesn't show up on the command-line anymore and now doesn't show up in the job log either :0/  

## Testing?

Before using this branch go to the Tripal Job log where you imported content types and try to view it. You'll get an error.  Then pull the branch and try again, you should be able to see the job and the arguments should be nicely formatted into a table.

### DIFF
--- a/tripal/src/Controller/TripalJobController.php
+++ b/tripal/src/Controller/TripalJobController.php
@@ -100,13 +100,6 @@ class TripalJobController extends ControllerBase{
   private function buildArrayTable($array, $name = '') {
     $markup = '';
 
-    // If the table only has one key then simplify this down for display
-    $keys = array_keys($array);
-    if (count($keys) == 1) {
-      $key = $keys[0];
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code looks good
✅ Functional test works as well :-) 

Thanks @spficklin ❤️

PR #1788, opened 2024-02-17
ChadoStorage fix for joins that loop back to the root table

spficklin:
# Bug Fix

## Description

This PR addresses the problem where a field (such as the sequence coordinate field of PR #1700) wants to loop back on the same table that was previously joined. Fields that want to  use the `featureloc` table of Chado will need to do this if they want to get the landmark sequence.  

The fix for this is not too complicated because the `ChadoRecords::getJoinAliases()` function already created aliases for every table in a join. It ensures that every table has an alias and that if a join is used repeatedly the same alias is used for the same table.   The problem is that those aliases were available for querying but they weren't used when setting up the records array.  This PR fixes that and solves the problem. 

## Testing?

There is no way to manually test this until we have the Coordiante field working.  However, I did add a functional test named `ChadoStorageActions_ReadValueTest::testReadValueActionJoinLoop`.  This test adds an organism, feature, landmark feature and featureloc records to the table. It creates a temporary field that uses the featureloc table.  The field is then used to read values from the `featureloc` table and check that those value are properly returned.  

The test function worked for me. I think if all tests pass (assuming I didn't break anything else) and the code looks good to @laceysanderson then it's good to be merged and can be fully tested with a fix to PR #1700 .

### DIFF
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -572,6 +572,8 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
           $context = [];
           $context['is_find'] = $is_find;
           $context['base_table'] = $base_table;
+          $context['root_table'] = $base_table;
+          $context['root_alias'] = $base_table;
           $context['operation'] = array_key_exists('operation', $info) ? $info['operation'] : '=';
           $context['field_name'] = $field_name;
           $context['property_key'] = $key;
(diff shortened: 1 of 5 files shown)


tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -589,17 +589,17 @@ protected function buildChadoRecords($values, bool $is_find = FALSE) {
...
-            // Add to the context.
-            $context['path_string'] = $prop_storage_settings['path'];
-            $context['path_array'] = $path_array;
-
             // We only add joins when the action is 'read_value' because
             // they guarantee a single value (meaning a 1:1 join). For
             // other joins there may be a many to one so we don't want to add
             // those joins off the base table.
             if ($action == 'read_value' and array_key_exists('join', $path_array)) {
               $this->handleJoins($path_array, $context);
```
spficklin, line 603: I changed the `handleJoins` function to pass the $path_array be reference. This way the changes to the table aliases get added to the `$context` value that is then used to build the records.

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -126,16 +126,6 @@ protected function initTable($elements) : bool {
...
-    if ($base_table == $chado_table) {
-      if ($base_table != $table_alias) {
-        throw new \Exception(t('ChadoRecords::initTable(). Invalid attempt to '
-            . 'initalize a table as the base table cannot have an alias. '
-            . 'The element tries to set a table alias, which is not supported: '
-            . '@elements',
-            ['@elements' => print_r($elements, TRUE)]));
-      }
-    }
-
```
spficklin: This code had to come out because if the base table is used again later for another join (such as with the `featureloc` table) then it will violate this check.

spficklin:
@laceysanderson all tests are passing.  This is ready for another quick review.

laceysanderson (approved):
✅ Code looks good
✅ Tests are all passing + new functionality test looks good
✅ Manually tested that ChadoStorage is still working as expected for other fields by creating an organism, importing other content types, creating a gene and a cultivar and edited all of them.

Thanks for the fix @spficklin! ❤️ 🎉

PR #1792, opened 2024-02-19
Default Layouts

spficklin:
# Tripal 4 Core Dev Task

## Description

**_Note_**: This module is a draft pending a discussion about functional testing, but it's ready if someone wants to test it and see if anything doesn't work or if changes are needed.

This PR adds a new module named `tripal_layout` that allows a module developer to provide a default layout for Tripal content types using a YAML file.  It reads the YAML file and organizes the fields accordingly.   This module requires the [field_group](https://www.drupal.org/project/field_group) and [field_group_table](https://www.drupal.org/project/field_group_table) modules.  

Here's how it works:

The module developer creates a new YAML file named `tripal_layout.tripal_layout_default_view.<group>.yml` where `<group>` is the name (arbitrary) for a given group of layouts. This file is placed in the `modulename/config/install` folder of the module. For example, the module provides two YAML files:  `tripal_layout.tripal_layout_default_view.general_view.yml` and `tripal_layout.tripal_layout_default_view.genomic_view.yml`.  The idea is that this module will provide the layouts for the general content types, and when we move out the other groups (e.g., genomic, expression, germplasm, etc) that those modules will provide layouts for their content types.  

Here's an example of the YAML file for the organism layout:

```yaml
id: 'general_view'
label: 'General Content Types'
description: 'Layouts for general content types (analysis, organism, etc.) based on Chado and added via the Tripal Chado module.'
layouts:
  - tripal_entity_type: "organism" 
    hidden:
      - uid
      - title     
    field_groups:
      details:
        organism_details:
          label: "Organism Details"
          weight: 0
          show_empty: false
          open: true
          children: 
            - "organism_details_table"
        organism_properties_details:
          label: "Properties"
          weight: 10
          show_empty: false
          open: false
          children: 
            - organism_properties_table
        organism_cross_refs_details:
          label: "Cross References"
          weight: 20
          show_empty: false
          open: false
          children: 
            - "organism_dbxref"
        organism_pubs_details:
          label: "Publications"
          weight: 30
          show_empty: false
          open: false
          children: 
            - "organism_pub"
      field_group_table:
        organism_details_table:
          label: "Details Table"
          weight: 0  
          show_empty: false
          children:
            - "organism_genus"
            - "organism_species"
            - "organism_infraspecific_type"
            - "organism_infraspecific_name"
            - "organism_abbreviation"
            - "organism_common_name"
            - "organism_comment"    
        organism_properties_table:
          label: "Properties Table"
          weight: 0  
          show_empty: false            
          children:
            - "type:chado_property_type_default"
```
The field_group module provides several types of field groups, but for now, this module only supports the "details" and the "field_group_table" groups. The "details" group behaves like the fieldset of Drupal 7.  The "field_group_table" puts each field on a single row.   As you can see from the YAML above,  you can set fields that should be hidden and you can create a set of field groups.   All of the "details"  groups are together and all of the "field_group_table" groups are together.  You'll notice you can set some properties for each field group type and you can set the children that should be displayed in that group.  The children are the field names or the field group names (field groups can be children of other field groups).    

Once the module is installed, the configuration in the YAML files are known to Drupal and a site developer can apply the theme by going to **admin** > **Structure** > **Tripal Content Types** and selecting "manage display" on the drop down for the content type.  You'll see a page that looks like this:

image

If you use the Bootstrap 5 theme and look at the organism page it looks like this:
image

On the page to manage the display of the fields the layout can be applied by clicking the button "Apply Tripal Layout". If the content type has a layout configuration in the YAML file then it will be applied. After applying the YAML shown above the field display page now looks like this:

image

You'll see that there are new "details" field groups added and "field_group_tables" with their corresponding children fields.

If you load the organism page and have the bootstrap 5 theme installed it looks like this:

image

Here are examples an analysis page and a gene page:

image
image

The site developer can change the display as they like. For example, I had an image field for the organism which is a field I added. Notice in the organism page screenshot above and in the field layout screenshot the orgnaism image is on it's own. It's not in a field group.  Also, the description is at the top.  I manually edited the layout, saved it and now the image is in the details group and the description is lower down:

image

On the field display page there is a "Reset Layout" button as well, so that if you want to start over, you can click that and the layout will go back to its original state.   You are free to add new fields and organize them as you like after the layout is applied. If you click the 'Apply Tripal Layout' button and you already have a layout set, it will reset it to what is in the configuration.  

What's missing:
- This layout is different from Tripal 3.  There is no sidebar.  We may want to consider if we want to add a sidebar or let site developers figure that one out on their own.  Perhaps we just need a block that can be put anywhere that contains  table of content for the page and not built into the content section like on Tripal 3.
- This currently only containts layouts for 3 content types:  organism, analysis and gene, we need to create one for all of the content types provided by Tripal
- There is no functional testing, I wasn't sure of the best way to do that and wanted to chat on it.
- This module currently only support two of the field group types: "details" and "field_group_table" but there are others such as "tabs", "accordian", etc.  I don't think we have to support everything for this layout module to be useful though. People can still manually make changes. These two are just used for "default" layouts.
- Using the chado field type (e.g. see in the YAML where "type:chado_property_type_default" is used in the properties table) is not yet supported. The idea is that if you specify a child with a "type:" prefix then all of the fields of that type will be considered children. That's useful for properties where there are an unknown number.
- 

Limitations
- We will be depenedent on the field_group (used on ~320K sites) and field_group_table modules (used on ~4K sites).  
- The field_group module is not compatible with the built-in Drupal Layouts module.  The two can be used on the same site. For example, you can use Drupal Layouts to create a nice home page, and you can not use Drupal Layouts on our Tripal content types and all is happy. If you try to apply a Drupal Layout to a tripal content type and you have field groups it gets a bit wonky and needs some further testing.
- I realized fields should be ordered on the page how they are listed in the YAML and that's not happening.  The weights for the children aren't being adjusted. This needs to be fixed.

## Testing?

I reocmmend using the [Bookstrap 5 theme](https://www.drupal.org/project/bootstrap5) (as that's the only Drupal 10 theme that actually looks half-way decent INMHO). Pull this PR and do the following:

1. Make sure you have an organism, analysis or gene page.
2. View the organism or analysis page without the Tripal layouts module installed
3. Install the `tripal_layout` module.
4. View the page and it still looks the same.
5. Go to admin > Structure > Tripal Content Types and click the "manage display" button beside the content type you want to apply a layout for.
6. Click the "Apply Tripal Layout" button.
7. Go to the page and refresh. It should now be reorganized according to the default layout. 
8. Return to the fields display page and move things around as you like, add a new field, etc.  You are free to customize the layout as you like.  
9. View the content page again to make sure all of your change are applied.  
10. Return to the fields display page and click the "Reset Layout" button.
11. View the content page  again and original default Drupal layout with just a listing of fields is restored.  
12. Now, go to a content type that is not supported (e.g. contact, mRNA, germplasm, etc) and click the button to manage the display. Click the 'Apply Tripal Layout' button. It should warn you that it could not find a layout and it won't make changes.  
13. Site developers may need to reload a configuration if they want to make changes.  Suppose you are building a layout YAML file and want to reload it without uninstalling/enabling your module.  You can reload the configuration by
     1. Go to admin > Tripal > Configurations and clicking the 'Tripal Default Configuration".  You'll see two configurations, one for the general and one for the genomic.  
     2. Click the delete button beside each one. They will go away.
     3. Using drush, clear the cache,
     4. Reload the configuration page and the two should come back.


laceysanderson:
I just tested this on Drupal 10.2.x + PHP 8.3 in the TripalDocker with Olivero theme and it worked great :-) 

Here is the organism page with the (1) default template applied, (2) description moved out of the table, (2) fields re-arranged.
organismPageWithDefaultTemplate

I also confirmed I get the expected message when trying to apply a template to a content type that doesn't have one. I tested the Protocol content type.

spficklin:
Wonderful!   Glad it worked!

I could use some direction for how to apply some functional testing.  Drupal already handles throwing errors if the YAML format isn't correct.  But to test this it's all about the display.   I've not done any visual-level GUI functional testing so any pointers would be nice.

spficklin:
Otherwise, if you think this is good to go we can merge but I thought I'd ask about testing....

laceysanderson:
For testing, I think we definitely want some kernel tests for the Controller and form so those will be really easy. 

See [TripalImporterFormBuildTest](https://github.com/tripal/tripal/blob/4.x/tripal/tests/src/Kernel/TripalImporter/TripalImporterFormBuildTest.php) and [TripalImporterFormSubmitTest](https://github.com/tripal/tripal/blob/4.x/tripal/tests/src/Kernel/TripalImporter/TripalImporterFormSubmitTest.php) for form examples.

The controller should be really simple although I don't have an example for that 🤔 which brings me to ask why you didn't use a service? That's what we did for the Tripal Entity Type and Tripal Field Collections.

As far as testing the actual page afterwards, that would be more involved although likely worth it. Those fall under functional tests which should only be those that test the UI directly. We don't have any good functional browser-based tests in core but there is this one in one of my extension modules: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/tests/src/Functional/ConfigRRulesTest.php. There is this one in Tripal core but it doesn't show the parts detecting content in the page: https://github.com/tripal/tripal/blob/4.x/tripal/tests/src/Functional/Permissions/TripalRoutePermissionsTest.php. It's actually really easy and powerful :-) Here's the full Drupal tutorial on it https://www.drupal.org/docs/develop/automated-testing/phpunit-in-drupal/phpunit-browser-test-tutorial.

I think we should add the kernel tests in this PR but the more involved functional tests could wait until later.

spficklin:
Okay, thanks for the suggestions. I'll work on adding those tests for the controller.

To your questions about why I didn't use a service. I did start that way, but I noticed that the link to 'Check for new fields' resolved to this function `TripalEntityUIController::tripalCheckForFields()` so I figured I would try to be consistent and go the route of using a Controller.  I can convert this into a Service if you think it's better that way.

laceysanderson:
Ahhh, I see! I don't know that either one is better than the other in this case so we might as well leave it as is :-) The services approach feels more Drupal to me for some reason but I don't really have anything specific to base that on so who knows 🤪 Thanks for sharing the reasoning!

spficklin:
After thinking a bit more on how to test, I really need to have an instance of the fields form in order to edit the $display object.  Perhaps there's a way to programmatically get that without the GUI, but I'm not sure how to do that.  After discussion today in the PMC it was decided to go ahead and merge in this update and I'll create a new new issue so we remember to do the testing.

PR #1795, opened 2024-02-22
Fixes for the new Drupal 10.2 subform

### LINKED ISSUE #1791: Edit Field Settings WSOD on Drupal 10.2
### Tripal Version

4.x-dev

### Drupal Version

10.2.x-dev

### PostgreSQL Version

13

### PHP Version

8.3

### General Description

In Drupal 10.2 the UI for adding custom fields has changed quite a lot. This seems to have caused some incompatibilities as when you try to edit a Tripal/Chado field, you are greeted with the following WSOD:
image

The same happens when you try to create a new tripal/chado field.

### Steps to reproduce

1. Start a fresh site on Drupal 10.2.x-dev and PHP 8.3 (or use Tripaldocker)
2. Go to Admin > Tripal > Page Structure > Organism > Manage Fields
3. Click on "Edit" for any Tripal/Chado field

### Error messages

```shell

RuntimeException: The subform and parent form must contain the #parents property, which must be an array. Try calling this method from a #process callback instead. in Drupal\Core\Form\SubformState->getParents() (line 80 of core/lib/Drupal/Core/Form/SubformState.php).

Drupal\Core\Form\SubformState->getValues() (Line: 26)
Drupal\Core\Form\SubformState->getValue(Array) (Line: 44)
Drupal\tripal_chado\TripalField\ChadoFieldItemBase->storageSettingsForm(Array, Object, ) (Line: 123)
Drupal\field_ui\Form\FieldStorageConfigEditForm->form(Array, Object) (Line: 107)
Drupal\Core\Entity\EntityForm->buildForm(Array, Object) (Line: 85)
Drupal\field_ui\Form\FieldStorageConfigEditForm->buildForm(Array, Object, Object) (Line: 199)
Drupal\field_ui\Form\FieldConfigEditForm->form(Array, Object) (Line: 107)
Drupal\Core\Entity\EntityForm->buildForm(Array, Object)
call_user_func_array(Array, Array) (Line: 536)
Drupal\Core\Form\FormBuilder->retrieveForm('field_config_edit_form', Object) (Line: 283)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 627)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 181)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 76)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 28)
Drupal\Core\StackMiddleware\ContentLength->handle(Object, 1, 1) (Line: 32)
Drupal\big_pipe\StackMiddleware\ContentLength->handle(Object, 1, 1) (Line: 106)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 36)
Drupal\Core\StackMiddleware\AjaxPageState->handle(Object, 1, 1) (Line: 51)
Drupal\Core\StackMiddleware\StackedHttpKernel->handle(Object, 1, 1) (Line: 704)
```

### Screenshots

_No response_

dsenalik:
# Bug Fix

## Description
To quote @laceysanderson : "In Drupal 10.2 the UI for adding custom fields has changed quite a lot.
This is happening because in Drupal 10.2 the field settings and field storage settings have each become subforms and then been merged into a single form page. This means that where we originally were given the full form state, now we are given a subformstate which only has the values specific to our subform in it."

Drupal reference: https://www.drupal.org/node/3386675

## Testing?
**Test on BOTH Drupal 10.1 and 10.2**:
 1. Test a base column form element by adding a Chado Text Field Type, e.g. to project
2024-02-25_test1
You might notice that under 10.2 you can also see the description, so I took the opportunity to make the distinction between Text and String clear (I always seem to forget which is which)
Select the base table "Project" and after the Ajax does its thing, the "Table Column" should be populated with the only choice, "Description"
Select a CV Term e.g. "DNA (SO:0000352)" but any term will do - this is a two-form process under 10.1, but now all on the same form in 10.2
Save and verify the field looks good.

 2. Test the property form by adding a Chado Property to e.g. pub
 Use whatever CV term you like

 3. Test linking table form element by adding a Chado Contact field to e.g. project.
 The CV term should not need to be entered for this one.

 4. Test the Chado Database Cross Reference field
 5. Remove the original "Description", "Contact", and "Dbxref" (Database reference annotation) fields, we are going to test the ones we just added.
 6. Create some content. First create a "contact"
 7. Create a project and use this contact, fill in a description, your property, and a dbxref value.
2024-02-25_project


dsenalik:
Ready for review. A few notes:
 * I changed the order of some of the `use` statements for tidiness, put all Drupal classes together, then Tripal classes.
 * I changed descriptions of the String and Text fields for my own benefit as I always forget which is which
 `A string field.` -> `A text field with a maximum length.`
 `A text field.` -> `A text field with no length limit.`
 I was partially inspired by the new Drupal 10.2 options when adding a Drupal text field
2024-02-drupal_text_field

 * To reduce lines of code in many of the fields, I made a function in the parent class `get_linker_table_and_column()` to parse out the linking method, now this can be a one-liner in the fields.
 * In `TripalFieldItemBase.php` you will notice in the `defaultStorageSettings()` this:
```
-  'termIdSpace' => '',
-  'termAccession' => '',
+  'termIdSpace' => ($child_class::defaultFieldSettings()['termIdSpace'] ?? ''),
+  'termAccession' => ($child_class::defaultFieldSettings()['termAccession'] ?? ''),
```
The reason for this is that the empty strings here were now overriding the values that fields set in their `defaultFieldSettings()` functions. However, you can't just remove these initializations here, the variable must be declared here in this function, otherwise Drupal removes anything we set it to later (i.e. it becomes an unsupported setting). So, this code in the parent class copies the default values from the child class and it all works great. For an existing similar case of copying see `tripal/src/Services/TripalFieldCollection.php` https://github.com/tripal/tripal/blob/3aa66198c3b015c9dcf60b66b7eb696ed6295132/tripal/src/Services/TripalFieldCollection.php#L158-L160

tripal/src/TripalField/TripalFieldItemBase.php
```diff
@@ -214,7 +225,8 @@ public function fieldSettingsForm(array $form, FormStateInterface $form_state) {
                 ['@term' => $termIdSpace . ':' . $termAccession]));
             $is_open = TRUE;
           }
-          $default_vocabulary_term = !$default_vocabulary_term ? $term->getName()  . ' (' . $term->getIdSpace() . ':' . $term->getAccession() . ')' : '';
+          // @to-do - logic might be wrong here and would reset default_vocabulary_term?
+          $default_vocabulary_term = !$default_vocabulary_term ? ($term->getName() . ' (' . $term->getIdSpace() . ':' . $term->getAccession() . ')') : '';
```
laceysanderson: Hmmm. yah. If the default_vocabulary_term is set then this line would set it to an empty string 🤔 Maybe we should do this instead:
```suggestion
          $default_vocabulary_term = !$default_vocabulary_term ? ($term->getName() . ' (' . $term->getIdSpace() . ':' . $term->getAccession() . ')') : $default_vocabulary_term;
```

laceysanderson (commented):
✅ Code looks good :-)
✅ all 3 types of fields work without error in Drupal 10.2.x
✅ all 3 types of fields work without error in Drupal 10.1.x

Specifically, 
- I tested adding a text field, contact linker and two property fields. 
- I then edited the fields after they had been created and saved again to make sure defaults held.
- I created and edited a piece of content with the new fields ensuring that there were no errors on the page or in the log.

I only noticed one thing amiss:
When you edit a newly created text field, the base table is set but the table column is not. Both are rightly disabled and saving the form as is does not break the field functioning 🤷‍♀️  This happens in both D10.2 and D10.1 on this branch.
Note: the linking method default is set correctly on edit, just not the base table column 🤷‍♀️ 
D10 1 x-fieldsettings-textfield
D10 2 x-subform-textfield

dsenalik:
Thanks for catching that! I think I have fixed it https://github.com/tripal/tripal/pull/1795/commits/f13ae0d9721eb9a14f18d0bef5da6b6ca1c67fb9, you can now see the table column - I tested on Array Design because it has several text columns to choose from.
2024-02-26_table_column

laceysanderson (commented):
Tested the updates and now the base column is not disabled for some reason? Also, it doesn't seem to be set. The following were on Drupal 10.2.x.

**Screenshot taken when editing the existing description field on the project content type.**
image

**Screenshot taken when editing the existing infraspecies field on the organism content type.**
image

I did confirm that I have the most recent commit 🤔 I also built a new docker image and container on this branch before testing.
```
docker build --tag=tripaldocker:tripal1795 --build-arg drupalversion=10.2.x-dev --file tripaldocker/Dockerfile-php8.3-pgsql13 ./
docker run --publish=80:80 -tid --name=tripal1795 --volume=/Users/las166/Docker/tripal-1795-subforms:/var/www/drupal/web/modules/contrib/tripal tripaldocker:tripal1795
docker exec tripal1795 service postgresql restart
```

dsenalik:
Okay, fixing this last thing was more complicated than anticipated, it required a few steps.
1. I forgot to disable when `$default_table_column` is known, so that is easy. But...
2. In Drupal 10.2, when the table column is selected, an ajax callback is fired which updates the form, so if I disable it, then the user's first selection gets "locked in" and can't be changed, even though at this point changing it is fine.
To fix this, I introduced a `$is_ajax` variable, and disable the field only if the default_table_column is defined and this is not an ajax callback (plus the other existing cases for disabling it).
(Non of this need to happen in Drupal <=10.1 because there is no ajax for this selector.)

dsenalik:
Issue to fix: Import all content types. On the `pub` content type, add a field to link to QTL (chado feature) with the QTL CV term.
Create an organism, then a QTL. Then try to create a pub with a QTL, I get the error 

Cannot load the entity. See the recent logs for more details or contact the site administrator if you cannot view the logs. 

PDOException: SQLSTATE[42703]: Undefined column: 7 ERROR: column rvepokjzsukdlubayfus.taxon_id does not exist LINE 7: ...JOIN testchado.organism "rfojyuhxwkjyubisibmz" ON rvepokjzsu... ^ in /var/www/drupal/web/core/lib/Drupal/Core/Database/StatementWrapperIterator.php:111 

The column taxon_id is only present in the biomaterial table.
This only happens on Drupal 10.2
Fixed with commit https://github.com/tripal/tripal/pull/1795/commits/5219f89f0c8a0c881932966f112239692ac8ceac

dsenalik:
**Another thing to fix**
Under Drupal 10.2
Create in this order 
2024-02-29_content
Then try to create "Assay" content:
```

Drupal\Core\Entity\EntityStorageException: SQLSTATE[25P01]: No active sql transaction: 7 ERROR: 
  SAVEPOINT can only be used in transaction blocks: SAVEPOINT mimic_implicit_commit; Array ( ) in
  Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of
  core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).

Drupal\Core\Database\StatementWrapperIterator->execute(Array, Array) (Line: 851)
Drupal\Core\Database\Connection->query('SAVEPOINT mimic_implicit_commit', Array, Array) (Line: 239)
Drupal\pgsql\Driver\Database\pgsql\Connection->query('SAVEPOINT mimic_implicit_commit') (Line: 473)
Drupal\Core\Database\Transaction\TransactionManagerBase->addClientSavepoint('mimic_implicit_commit') (Line: 247)
Drupal\Core\Database\Transaction\TransactionManagerBase->push('mimic_implicit_commit') (Line: 540)
Drupal\pgsql\Driver\Database\pgsql\Connection->startTransaction('mimic_implicit_commit') (Line: 413)
Drupal\pgsql\Driver\Database\pgsql\Connection->addSavepoint() (Line: 153)
Drupal\pgsql\Driver\Database\pgsql\Select->execute() (Line: 176)
Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace->loadIdSpace() (Line: 122)
Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace->recordExists() (Line: 169)
Drupal\tripal\TripalVocabTerms\PluginManagers\TripalCollectionPluginManager->loadCollection('SIO') (Line: 69)
Drupal\tripal\TripalStorage\StoragePropertyBase->__construct('tripal_entity', 'chado_array_design_type_default', 'record_id', 'SIO:000729') (Line: 35)
Drupal\tripal\TripalStorage\StoragePropertyTypeBase->__construct('tripal_entity', 'chado_array_design_type_default', 'record_id', 'SIO:000729', 'int', Array) (Line: 32)
Drupal\tripal\TripalStorage\IntStoragePropertyType->__construct('tripal_entity', 'chado_array_design_type_default', 'record_id', 'SIO:000729', Array) (Line: 157)
Drupal\tripal_chado\Plugin\Field\FieldType\ChadoArrayDesignTypeDefault::tripalTypes(Object) (Line: 346)
Drupal\tripal\TripalField\TripalFieldItemBase::schema(Object) (Line: 480)
Drupal\field\Entity\FieldStorageConfig->getSchema() (Line: 517)
etc.
```

* Update: Only fails if a biological sample is specified. If left blank, then no error.

https://www.drupal.org/node/3381002
**New wrapper for database transactions**
Introduced in version: 10.2.0


`core/lib/Drupal/Core/Database/Transaction/TransactionManagerBase.php`
```
  /**
   * Commits the entire transaction stack.
   *
   * @internal
   *   This method exists only to work around a bug caused by Drupal incorrectly
   *   relying on object destruction order to commit transactions. Xdebug 3.3.0
   *   changes the order of object destruction when the develop mode is enabled.
   */
  public function commitAll(): void {
```

That was all a distraction, the real problem can be seen in `postgresql-13-main.log`
```
2024-03-02 15:32:07.175 UTC [424] drupaladmin@sitedb ERROR:  insert or update on table "assay_biomaterial" violates foreign key constraint "assay_biomaterial_channel_id_fkey"
2024-03-02 15:32:07.175 UTC [424] drupaladmin@sitedb DETAIL:  Key (channel_id)=(1) is not present in table "channel".
2024-03-02 15:32:07.175 UTC [424] drupaladmin@sitedb STATEMENT:  COMMIT
2024-03-02 15:32:07.189 UTC [424] drupaladmin@sitedb ERROR:  SAVEPOINT can only be used in transaction blocks
2024-03-02 15:32:07.189 UTC [424] drupaladmin@sitedb STATEMENT:  SAVEPOINT mimic_implicit_commit
```
In this case we can't use the default of `1` for channel_id.

This is fixed with commit https://github.com/tripal/tripal/pull/1795/commits/51f6a453a96e4264e8b55180d94687a4cd03f48e

laceysanderson:
This is now greatly impacted by #1440 and likely closed once some of these changes move into that branch instead.

dsenalik:
This PR will be replaced by #1805 and I have transferred all of the relevant fixes from this PR to that one (I hope) so I am closing this PR.

PR #1796, opened 2024-02-23
Update README.md

spficklin:
Simple update to the README


laceysanderson (approved):
Looks good. Just removing the alpha 2 todo list since it's already out 🎉

PR #1801, opened 2024-02-24
Tripal Layout naming change to human readable format in Admin -> Extensions

risharde:
The simplest PR ever but just to keep it standard on the UI
This fix will change tripal_layout to 'Tripal Layout' on the admin extensions UI
image


laceysanderson (approved):
Good catch and good call --Consistency is King 😄

PR #1802, opened 2024-02-27
Update older tests to use ChadoTestTrait instead of a hardcoded `testchado` schema

### LINKED ISSUE #1377: Remove old use of testchado schema in favour of Tripal DBX schema creation
Remove old use of testchado schema in favour of Tripal DBX schema creation.

Currently a large number of the old tests assume there is a `testchado` named chado schema installed and fail otherwise. These need to be updated to use the new chado schema created for each test to further weed out any name assumptions and ensure tests are more independent from each other.

laceysanderson:
# Tripal 4 Core Dev Task

## Description

When we first started developing tests, we required chado be installed before testing in a schema named `testchado`. After Tripal DBX was submitted, we were able to quickly prepare a chado schema for testing in the same method that Drupal uses. However, we had not gotten the change to go back and update the old tests.

This PR updates these old tests to use a Tripal DBX generated chado schema and removes the need to have `testchado` installed before running tests. Additionally, a few other things were noticed while getting these tests to pass again and were fixed in this PR as well.

Specifically,
- The old chadoInstaller service was removed since this is now done via the ChadoInstaller BioDB Task.
- The legacy ChadoSchema class and chado_get_version() function were updated to use the new TripalDBX method defining a valid schema name. Note: the old regex excluded test schema names.
- Switch github actions to install chado in a schema named `teacup` so that we catch any other assumed `testchado` schema names in tests.
- Various tests using `testchado` were updated as follows:
    - `$this->connection` was set with a TripalDBX generated chado instance for testing
    - previous `$connection` was updated to `$drupal_connection` to make it clear that it was a connection to the drupal schema outside of Tripal DBX.
    - anywhere that `testchado` was passed into a function or query, we now pass in `$this->testSchemaName`
    - inheriting from `ChadoTestKernelBase` instead of `BrowserTestBase`.
    - added better coverage to the chado installer task test.
    - removed a skipped test related to the legacy ChadoSchema class as we don't intend to ever implement it.

NOTE: Although git reports a lot of change in `testChadoQuery()` the changes were quite minimal functionally. This test was quite confusing in that it uses the Drupal connection to check various chado tables. The changes try to make this more clear by using `$drupal_connection` and `$chado_testschema` (which is the same as `$this->testSchemaName` but cleaner to concatenate into a string). Additionally, the `{1:tablename}` notation needed to be used in a couple of places as Tripal DBX struggles to choose the right schema in the testing environment.

## Testing?

Since only testing and legacy code has been impacted this is only really testable via the automated testing and perhaps code review.


tripal_chado/src/api/ChadoSchema.php
```diff
@@ -113,15 +113,18 @@ public function __construct($version = NULL, $schema_name = NULL) {
...
-      return FALSE;
-    }
     else {
-      $this->schema_name = $schema_name;
+      $tripalDbxApi = \Drupal::service('tripal.dbx');
+      if ($tripalDbxApi->isInvalidSchemaName($schema_name, TRUE)) {
+        // Schema name must be a single word containing only lower case letters
+        // or numbers and cannot begin with a number.
+        $this->logger->error(
+          "Schema name must be a single alphanumeric word beginning with a number and all lowercase.");
```
dsenalik: If comment is correct, then error message is wrong
```suggestion
        // or numbers and cannot begin with a number.
        $this->logger->error(
          "Schema name must be a single alphanumeric word beginning with a letter and all lowercase.");
```

laceysanderson:
There were a few more instances of `testchado` but I think I have them all now. Should be ready to review again.

dsenalik (approved):
Okay, I looked at all the changes and everything looks good to me and I cannot find anything more to comment on, but I am still a testing noob so don't yell if I missed anything 🙄

laceysanderson:
Awesome -thanks for the review @dsenalik!

PR #1805, opened 2024-03-05
New base table third party setting + Drupal 10.2 fixes

### LINKED ISSUE #1440: Tripal 4 You can add a field to a content type that does not support it

### LINKED ISSUE #1791: Edit Field Settings WSOD on Drupal 10.2

### LINKED ISSUE #1818: Provide a way to set the base table for a newly defined content type
### Task Description

When you define a field through yaml, you can set its base table, e.g.
```
    -   label: Organism
        term: OBI:0100026
        id: organism
        settings:
            chado_base_table: organism
```
When you create a new content type through the GUI, there is no way to save an equivalent setting, so when new fields are added to this new content type, you will have to set the base table for every one. 

We just need a way to select a base table when creating the new content type.

### Branch Name

~~tv4g1-issue1818-select-base-table-for-new-content-type~~

dsenalik:
# New Feature + Bug Fix

## Description
Because they are so intertwined, this PR does several things:
1. Create a third party setting so that fields can finally know what their base table should be (Thanks to @spficklin for this genius discovery).
2. Introduces an `is_compatible()` function that fields will implement a way to know what fields can be added to a particular content type.
3. Fix the form for adding fields through the GUI to support the new Drupal 10.2 subform.
To quote @laceysanderson : "In Drupal 10.2 the UI for adding custom fields has changed quite a lot.
This is happening because in Drupal 10.2 the field settings and field storage settings have each become subforms and then been merged into a single form page. This means that where we originally were given the full form state, now we are given a subformstate which only has the values specific to our subform in it."
Drupal reference: https://www.drupal.org/node/3386675
This PR will replace #1795
4. Edit (spficklin):  Also adds the YAML, library entities and CSS for grouping Tripal and Tripal Chado fields in the new 10.2 Drupal field grid layout.
5. For a new content type defined through the GUI, saves the base table selected for the first Chado field added to that content type as the base table for all subsequent Chado fields (Issue #1818)

## Testing?
**Test on BOTH Drupal 10.1 and 10.2**:
 1. Test a base column form element by adding a Chado Text Field Type, e.g. to project
2024-02-25_test1
You might notice that under 10.2 you can also see the description, so I took the opportunity to make the distinction between Text and String clear (I always seem to forget which is which)
The new third party setting will allow the field to know that the base table is `project`, so the selector should be pre-populated and grayed out. The "Table Column" should be populated with the only valid choice, "Description"
2024-03-05_project-description
Select any CV Term - this is a two-form process under 10.1, but now all on the same form in 10.2
2024-03-05_example-term
Save and verify the field looks good.
Go back to view with Edit and make sure everything is still there

 2. Test the property form by adding a Chado Property to e.g. pub. The base table selector will again be grayed out.
 Use whatever CV term you like

 4. Test the linking table form element by adding a Chado Contact field to e.g. project.
 The CV term should not need to be entered for this one.

 3. Test the linking table form element by adding an Analysis field to e.g. project.
 In this case there is no default CV term so you will need to specify one.

 6. Test the Chado Database Cross Reference field
 7. Remove the original "Description", "Contact", and "Dbxref" (Database reference annotation) fields, we are going to test the ones we just added.
 8. Create some content. First create a "contact", and then an "analysis".
 9. Create a project and use this contact, fill in a description, your property, your analysis, and any dbxref value.
2024-03-06_project-test
10. Other fields to test: Chado Type Reference (ChadoAdditionalTypeDefault) (e.g. on organism, pub, contact, or protocol)
11. Chado Synonym (on any feature table derived content, e.g. gene)
12. Chado Unit (on featuremap types only, Genetic Map or Physical Map)
13. Try to add a field that is wrong for the content type, e.g. put the Chado Unit field on a publication. Or consult the big linker table https://github.com/tripal/tripal_doc/issues/32#issuecomment-1837532399 and find a blank cell, e.g. try to put the Feature field on a Study. (Sometimes there is an obscure linking table e.g. Contact on an Analysis through the quantification table, but that is not by default a content type)


laceysanderson:
To fix the schema incomplete failures you will want to add the third party setting to the schema. Since the content type is defined in the Tripal module and this is chado specific you need to use the tripal_chado_config_schema_info_alter hook in the tripal_chado.module file to alter the schema.

* For future reference, an example of the error was 
```
45) Drupal\Tests\tripal_chado\Functional\api\SchemaAPITest::testGetBaseTables with data set #0
   ('1.3', array('organism', 'feature', 'stock', 'project', 'analysis', 'phylotree'))
  Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for 
  tripal.tripalentitytype_collection.expression_chado with the following errors: 
  tripal.tripalentitytype_collection.expression_chado:content_types.0.settings.chado_base_table missing schema, 
  tripal.tripalentitytype_collection.expression_chado:content_types.1.settings.chado_base_table missing schema, 
  tripal.tripalentitytype_collection.expression_chado:content_types.2.settings.chado_base_table missing schema
```

dsenalik:
Current status of this PR
 * Annotations are fully converted to static variables. The annotations no longer exist.
 * Third party setting added to the schema so testing can pass (thanks @laceysanderson )
 * All relevant changes from PR #1791 have been incorporated.
 * All chado `isCompatible()` functions have been implemented.
 * One complaint from Code Climate about complexity, but I think creating a sub-function just for this, and only called from one place is overkill and I would favor leaving as-is.

To-Do:
 * Need to properly implement the `isCompatible` check - currently I just have dpm messages showing the return value of `isCompatible()` 
 Do we present the message in the storage settings form, or do we have a validate for the earlier step that can stop progressing to the storage settings form?
 * Do the four tripal fields need `isCompatible()` functions? Haven't looked at this yet.

spficklin:
Thanks @dsenalik for adding in all of those fixes and writing all of this `isCompatible()` functions.

I've implemented the first bullet in @dsenalik 's TODO list.  If you select a field that is not compatible it short circuits the form and takes you back to the field selection form  It also prints a message like this: 

image

In answer to the question in the second bullet.  I don't think those generic Tripal fields need an `isCompatible()` function. Their values are not stored in Chado and have no connection to Chado tables.

dsenalik:
Adding any of the four tripal fields to Project gives this error (Drupal 10.2):

Exception: Cannot create a StorageProperty object for entity type "tripal_entity", field type "tripal_boolean_type", key "value" without a properly formatted term: ":" in Drupal\tripal\TripalStorage\StoragePropertyBase->__construct() (line 83 of modules/contrib/tripal/tripal/src/TripalStorage/StoragePropertyBase.php). 

Fails also in Drupal 10.1 but no WSOD
2024-03-07_tripal-field

Note that this error predates this PR so maybe this can become a separate issue unless the fix is easy

spficklin:
Hey @dsenalik  I dug into the issue you raised and I see exactly what the problem is and it's definitely not related to these code changes. It's been a problem since we designed the field class and we didn't know it.   I have a partial fix started, but I hit on a problem that I think we need to discuss.   I think we ignore that problem for this PR and I'll add an issue so we can fix it in a different PR.

spficklin:
Added issue #1807

dsenalik:
I removed the draft tag as I feel this may be done. Maybe @laceysanderson will want to review Drupal 10.2 fixes yet again?

laceysanderson (commented):
Tested on Drupal 10.2, PHP 8.3 and all looks good ✅ 
Will try to look at it on a previous version of Drupal and do a code review later.

tripal/src/TripalField/TripalFieldItemBase.php
```diff
@@ -176,19 +179,20 @@ public function fieldSettingsForm(array $form, FormStateInterface $form_state) {
     ];
 
     $default_vocabulary_term = '';
-    $vocabulary_term = $form_state->getValue(['settings', 'field_term_fs', 'vocabulary_term']);
+    // For Drupal ~10.2 our values are now in the subform
```
laceysanderson, line 182: Perhaps we should getFormStateSettings() here as well now that we have that function?

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
```diff
@@ -215,11 +217,43 @@ public static function storageSettingsFormValidate(array $form, FormStateInterfa
     $chado = \Drupal::service('tripal_chado.database');
     $schema = $chado->schema();
     if ($schema->tableExists($prop_table)) {
-      $form_state->setValue(['settings', 'storage_plugin_settings', 'prop_table'], $prop_table);
+      $drupal_10_2 = $form_state->getValue(['field_storage']);
+      if ($drupal_10_2) {
+        $form_state->setValue(['field_storage', 'subform', 'settings', 'storage_plugin_settings', 'prop_table'], $prop_table);
```
laceysanderson, line 222: Again, why not use getFormStateSettings here?

laceysanderson (approved):
| Task                                              | D10.2/PHP8.3 | D10.1/PHP8.2 |
|---------------------------------------------------|--------------|--------------|
| Add Chado Text field to existing content type     | ✅            | ✅            |
| Add Chado Property field to existing content type | ✅            | ✅            |
| Add Chado Analysis field to existing content type | ✅            | ✅            |
| Unable to add unsupported field to content type   | ✅            | ✅            |
| Add Chado Text field to new content type          | ❌            | ❌            |
| Add Chado Analysis field to a new content type    | ❌            | ❌            |

*In all cases when adding a field, I also tested that settings were kept when editing that field again and that content could be created using these new fields.*

## Unable to add to new content types created through the UI

**I'm ok with this being fixed in a new PR since this one is already substantial. That said, a quick fix might just be to ensure all the isCompatibility methods return true if the base table is not set.**

As mentioned above, we now cannot add any Tripal/Chado based fields to a content type created through the UI 🙈 The error is the same in both D10.2 and D10.1. Steps to reproduce:  
1. Create a new content type through the UI (Admin > Tripal > Page Structure then "Add custom type")
2. Manage fields for this new content type (Admin > Tripal > Page Structure > NEWCONTENTTYPE > Manage Fields)
3. Click on "Create new field" and provide a name/machine name. Choose a chado-based field and you will get the error provided by the isCompatible method.
Screenshot 2024-03-15 at 2 39 51 PM

## Testing the update hook

✅ I confirmed this works on D10.2 based on the following process.

1. Run a container on the current branch using the latest tripal docker rather then building your own.
```
docker run --publish=80:80 -tid --name=tripal1805 --volume=`pwd`:/var/www/drupal/web/modules/contrib/tripal tripalproject/tripaldocker:latest
docker exec tripal1805 service postgresql restart
```
2. Go to your site and login.
3. Run the updates by going to http://localhost/update.php and click through to apply pending updates.
4. Try to add fields to an existing content type and confirm that the base table is pre-selected ✅

spficklin:
Thanks @laceysanderson for the testing!  One question about this statement:

> I'm ok with this being fixed in a new PR since this one is already substantial. That said, a quick fix might just be to ensure all the isCompatibility methods return true if the base table is not set.

I'm thinking if the base table isn't set that it should always return False. My thinking is, that if the chado table isn't specified the "Chado" based field will never be compatible.   Doug suggested that the `isCompatible()` in the base class alwasy return False and I suggested it should return True so that it doesn't confuse developers if they can't attach their fields while development and forget to implement that function.  But now that I think on it more. I'm thinking for the reason of content types not having Chado tables at all, then we should change that base function to return FALSE.

tripal/src/TripalField/TripalFieldItemBase.php
spficklin: The `getFormStateSettings()` looks like it works directly with the `settings` element of the subform, but the `vocabulary_term` form element isn't stored in the settings. It's used to help collect the term name and cvterm_id for the term and then in the validate function the setting actually get set, so I think we need to leave this as it is.

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
spficklin: I'll have to defer to @dsenalik to answer that question. I don't see why it couldn't but there's that check from drupal_10_2 which I don't understand and I suspect it may be tied to that?

laceysanderson:
> I'm ok with this being fixed in a new PR since this one is already substantial. That said, a quick fix might just be to ensure all the isCompatibility methods return true if the base table is not set.
>> I'm thinking if the base table isn't set that it should always return False. My thinking is, that if the chado table isn't specified the "Chado" based field will never be compatible.

When content types are created through the UI we currently don't have a way to set the base table for the content type. I think isCompatibility should return true if the content type doesn't have the base table set. This then allows the admin to fill out the field storage settings form and set the base table for the field there. When the field base table is set manually in this way, the chado-based fields are compatible :-)

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
dsenalik: It can't be used here because we are setting a value, not getting it 🙃
I had to add a similar thing in the ajax callbacks with commit https://github.com/tripal/tripal/pull/1805/commits/b0dfc9d43253284786d306a75c64a1318cb35d27
Maybe in another PR I could make a `setFormStateSettings()` - but there's enough in this PR

dsenalik:
I have implemented the "Quick Fix" with commit https://github.com/tripal/tripal/pull/1805/commits/01c081f543eedb41a9e481c809763592101faec4 and `isCompatible()` will now return true if the base table is not defined.
You can now add Chado fields to a new content type.

When I did this, I could proceed, and discovered that the ajax was broken under Drupal 10.2.
So that was fixed with commit https://github.com/tripal/tripal/pull/1805/commits/b0dfc9d43253284786d306a75c64a1318cb35d27

A few fields are still broken when adding to a newly defined content type, but as Lacey said, there is a lot in this PR already so I will create a new issue for this, issue #1817

And probably a separate new issue would be to provide a way to define the base table when creating a new content type, this will make a site admin's job easier as it doesn't have to be selected over and over for each new field added. Issue #1818

spficklin:
> When content types are created through the UI we currently don't have a way to set the base table for the content type. 

I think we should fix this.  If we don't require that a field's base table matches the one for the content type then you could get a bunch of fields referencing differnet tables, which will then cause problems if some one wants to "check for new fields".

spficklin:
Perhaps the solution is that if the contnet type doesn't have a base table set, then when the first Chado field is added then it will set the base table for the content type.  Then all others added later must match the base table of the content type.    

Also , I think our implementations of the `isCompatible()` function is getting too repetitive and we're forcing developers to always remember to code the check for a missing base table in the content type.  If they forget to code it then their fields won't work like we want.  I think that check should be handled outside of the field and the `isCompatible()` for each field should only be responsible for making sure the base table on the content type is compatible with the field (true if so, false if not).  Perhaps we only call `isCompatible()`  if a base table is present in the content type.

spficklin:
AS per additional conversation with @laceysanderson on this last comment (on Slack) i'm working on the fix.

spficklin:
Okay, I've added a fix that sets the base table on a content type automatically when the first Chado field is added.  

However, I'm really confused by the code that adjust the base table drop down.  @dsenalik I think you need to look at it to see what's wrong.    Here's the problem.

1.  I create a new content type manually through the GUI
2. I add a new "Chado Contact" field and set the base table to be "library".   This should then set the base table for the content type to be 'library'.
3. Add a new "Chado Property" field to the same content type.  But even though it knows the base table is `library` for this content the base table drop down is empty.

dsenalik:
On Drupal 10.2 I created a library content type. First field I used Chado String Type and the `uniquename` column. Then I added a Chado Contact Type and the base table was preselected and I could see the one possible linking method.

The fix I made was for the error
`Warning: Undefined variable $base_table in Drupal\tripal_chado\TripalField\ChadoFieldItemBase->storageSettingsForm() (line 120 of modules/contrib/tripal/tripal_chado/src/TripalField/ChadoFieldItemBase.php).`
It bothers me that there are so many variables for base table, probably a close analysis could remove at least one, but for now the fix in commit https://github.com/tripal/tripal/pull/1805/commits/9b202e89b2495f15c746d60b6526cb07111c2c9a works.

I will build a 10.1 docker to test there too

dsenalik:
You get this error under Drupal 10.1
`TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in tripal_chado_form_alter() (line 230 of modules/contrib/tripal/tripal_chado/tripal_chado.module).`

Your new hook unfortunately needs to know how to handle both drupal versions it seems. The fix is not as easy as I hoped. In `tripal_chado_form_alter()` the form ID is different depending on Drupal version.
```
  // Under Drupal 10.1 we want 'field_storage_config_edit_form', the storage form
  // Under Drupal 10.2 we want 'field_config_edit_form', storage is a subform within this form
  if ($form_id == 'field_config_edit_form') {
```
so I could do this
`  if (($form_id == 'field_config_edit_form') or ($form_id == 'field_storage_config_edit_form')) {`

but then there is the check for entity type, which we cannot do from `field_storage_config_edit_form` in Drupal 10.1 because that is back in the `field_config_edit_form`
```
    $entity_type = $form['#entity'];
    if ($entity_type->getEntityTypeId() == 'tripal_entity') {
```
If I just take out the entity type check it works under Drupal 10.1.

```
    $entity_type = $form['#entity'] ?? NULL;
    if (!$entity_type or ($entity_type->getEntityTypeId() == 'tripal_entity')) {
```
 But I don't understand whether we need that or not.
I'm sort of stuck here @spficklin HELP!

spficklin:
I think we don't need the check for tripal entity, because the check for the `base_table` in the storage settings should be sufficien.t

spficklin:
Thanks for doing all that cleanup work.  It's unfortunate that the form_id's change as well as the form structure :-/

dsenalik:
Thanks! If that's the case, I just pushed changes https://github.com/tripal/tripal/pull/1805/commits/a4bc925e537e68d3700bb5078ca42483394cc99f which should fix it for Drupal 10.1
I used null coalescing `??` ~~to reduce the number of `array_key_exists()`~~ 😼
Edit: I forgot I changed to using form state instead of form, as is done elsewhere.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -341,6 +345,50 @@ public static function storageSettingsFormValidateBaseTable(array $form, FormSta
...
+    // 'chado_base_table', and then we don't need to select it when manually adding the
+    // field to a content type.
+    $form_state_storage = $form_state->getStorage();
+    $bundle = $form_state_storage['bundle'];
+    /** @var \Drupal\Core\Entity\EntityTypeManager $entity_type_manager **/
+    $entity_type_manager = \Drupal::entityTypeManager();
+    /** @var \Drupal\tripal\Entity\TripalEntityType $entity_type **/
+    $entity_type = $entity_type_manager->getStorage('tripal_entity_type')->load($bundle);
+    $entity_type_chado_base_table = $entity_type->getThirdPartySetting('tripal', 'chado_base_table');
+    dpm($entity_type_chado_base_table);
```
laceysanderson: Errant debugging code.
```suggestion
```

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -341,6 +345,50 @@ public static function storageSettingsFormValidateBaseTable(array $form, FormSta
...
+
+    // If the entity type is missing the base table then use the one from
+    // this field and set it. This should only happen the first time a Chado
+    // field is added to the content type.  Once the base table is set for
+    // the entity type then all fields added in the future must use this
+    // base table.
+    if (!$entity_type_chado_base_table) {
+      $base_table = $settings['storage_plugin_settings']['base_table'];
+      dpm($base_table);
```
laceysanderson: Errant debugging code.
```suggestion
```

laceysanderson (commented):
Ok, I retested and everything looks great except for a small hiccup on D10.0 regarding unsupported fields (see more below).

| Task                                              | D10.2/PHP8.3 | D10.1 (Doug) | D10.0/PHP8.2 |
|---------------------------------------------------|--------------|--------------|--------------|
| Add Chado Text field to existing content type     | ✅            |              | ✅            |
| Add Chado Property field to existing content type | ✅            |              | ✅            |
| Add Chado Analysis field to existing content type | ✅            |              | ✅            |
| Unable to add unsupported field to content type   | ✅            | ❌            | ❌            |
| Add Chado Text field to new content type          | ✅            |              | ✅            |
| Add Chado Dbxref field to a new content type      | ✅            |              | ✅            |
| Unable to add unsupported field to content type   | ✅            | ❌            | ❌            |

## D10.0 ONLY, Unsupported fields bug

On Drupal 10.0 ONLY, if you try to add an unsupported field you will experience the following:
- You do get an error that the field is unsupported
- When you go back to the field listing though the unsupported field was added anyway and is in the listing 👿 
- When you try to edit the unsupported field, again you get the message about it being unsupported and cannot set any details
- If you try to create content  of that type, it will now fail as chado storage sees that unsupported field and adds it to the verification code 🙈 

For example, I added a sequence checksum field on the existing project content type. When creating a project I now get errors about not having an organism, type, etc 😭

dsenalik:
For Drupal ≤ 10.1 this is the point at which the field comes into existence, this is the form prior to where we check compatibility. Ideally, if the field is detected as not compatible in the next form, we would then have to also remove the field.

2024-03-18_nooooooo
```
 Schema |              Name              | Type  |    Owner    
--------+--------------------------------+-------+-------------
 public | tripal_entity__field_noooooooo | table | drupaladmin
```

dsenalik:
I just pushed a commit that adds a function to "clean up" when an incompatible field is added under Drupal ≤10.1 😀
` $this->removeIncompatibleField($bundle, $machine_name);`
It's not particularly complicated, so this may be the simplest fix.

laceysanderson (approved):
Works perfectly now! I just tested again on Drupal 10.2 and Drupal 10.1. I specifically tested one unsupported field and one supported field on an existing content type.

Thanks @spficklin and @dsenalik for all the work on this one!!! ❤️

PR #1806, opened 2024-03-07
Publication Importer - V3

risharde:
This is an updated version of the Publication Importer where the design has changed to be further organized.
The functionality remains the same on the user side of things but key database interactions have been moved from the Form class into the TripalPubLibraryManager class to deal with adding, getting, updating and delete Publication Library Queries.


tripal/src/TripalPubLibrary/PluginManagers/TripalPubLibraryManager.php
```diff
@@ -42,4 +42,42 @@ public function __construct(
...
+    ->execute();
+  }
+
+  public function deleteSearchQuery(int $query_id) {
+    $public = \Drupal::database();
+    $public->delete('tripal_pub_library_query')
+    ->condition('pub_library_query_id', $query_id, '=')
+    ->execute();
+  }
+
```
spficklin: @risharde can you add function header commens to these functions.

tripal/tripal.install
```diff
@@ -1010,21 +1010,21 @@ function tripal_uninstall_cloner() {
 }
 
 /**
- * Adds the tripal_pub_import table TEMPORARY
+ * Adds the tripal_pub_library_query table TEMPORARY
  */
 function tripal_update_8403() {
```
spficklin: Is this the correct number for the update hook?  Shouldn't it be something like 10401?

tripal_chado/src/Form/ChadoNewPubSearchQueryForm.php
```diff
@@ -31,11 +31,13 @@ public function buildForm(array $form, FormStateInterface $form_state, $pub_impo
...
-      $publication = $public->select('tripal_pub_import', 'tpi')
-        ->fields('tpi')
-        ->condition('pub_import_id', $pub_import_id, '=')
-        ->execute()
-        ->fetchObject();
+      // $publication = $public->select('tripal_pub_library_query', 'tpi')
+      //   ->fields('tpi')
+      //   ->condition('pub_library_query_id', $pub_import_id, '=')
+      //   ->execute()
+      //   ->fetchObject();
```
spficklin: Can this come out just to keep things clean?

tripal_chado/src/Form/ChadoNewPubSearchQueryForm.php
```diff
@@ -549,18 +553,20 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
         // If form_mode is 'edit', this is an update to the database
         else {
-          $public->update('tripal_pub_import')
-            ->fields($db_fields)
-            ->condition('pub_import_id', $user_input['pub_import_id'])
-            ->execute();
+          // $public->update('tripal_pub_library_query')
+          //   ->fields($db_fields)
+          //   ->condition('pub_import_id', $user_input['pub_import_id'])
+          //   ->execute();
```
spficklin: Same comment here.

spficklin (changes requested):

tripal/tripal.install
risharde: Thanks, I missed this!

tripal/src/TripalPubLibrary/PluginManagers/TripalPubLibraryManager.php
risharde: Done!

risharde:
@spficklin Thanks for the review and requested changes, they've been made!

dsenalik (approved):
My functional test showed no problems (Drupal 10.2, PHP 8.3)
Code review I saw nothing to note, except
I am wondering if we should inject `\Drupal::database();` as a dependency for the service, if it would be more efficient. I am learning, so not sure about this. I did a test, and I don't think in this case it would make any significant difference. So this is just a thought in case anybody wants to comment
```
$t1 = microtime(true);
$public = \Drupal::service('database');
$t2 = microtime(true);
$et = $t2 - $t1;
printf ("Elapsed time %0.8f \n", $et);
```
Elapsed time 0.00000596

laceysanderson:
It is very important to inject the database service. Injecting services is really important for keeping dependencies clear, it improves auto-loading of classes which can help with performance, it is very helpful and less buggy in automated tests and it is best practice in general. The only time we should use `\Drupal::service()` is if we are not in a class context and there are very few cases where that should be the case :-)

risharde:
Thanks guys, my apologies regarding this injection theory - so do I just replace the current service line I have with @dsenalik line? I googled about Drupal injection but didn't find something specific in explanation. If you can confirm, I can surely make the change, thanks much!!

laceysanderson:
Dependency injection is documented in issue #1522 where I have linked to multiple Drupal pages and also one of my PRs :-) 

Your Manager class is a service so you will use the service example with that file. You know you have the right example if it's showing you to add the dependency to the services.yml file ;-)

risharde:
I think I've implemented the database dependency injection correctly since everything seems to work and tests are passing.
I found this link useful for figuring out the name of the database service which was @database . I was wondering if there was a way to track back these service names but I wasn't entirely sure.
https://www.drupal.org/docs/drupal-apis/database-api/instantiating-a-database-connection-object

@laceysanderson @dsenalik @spficklin Could you please review and perhaps merge if everything works on your side? Thanks!

spficklin (approved):

laceysanderson (approved):
I did a code review on the updates to include dependency injection and that looks right 👍 ✅

PR #1833, opened 2024-03-26
Publications Importer V4

risharde:
This version of the Publications Importer adds new functions to the TripalPubLibraryInterface such as retrieve, parse and run.
All Tripal Pub Library plugins must now define these specific functions to operate correctly and as such the included TripalPubLibraryPubMed plugin now uses these functions. The run function is still to be fully implemented and this should happen likely in the next version (V5).


dsenalik (dismissed):
I didn't see anything bad in code review, just TODOs for the next version 😜
I did functional testing and created a variety of searches, they all appeared to work correctly, except...

Create a new search
select "Abstract" and enter "capillifolius" (field and values not important for this problem)
Click the test button, there are five hits.
Click the save button.
Go back to the list, the search query is not saved.

If instead you click the save button first without testing, it is saved.

And if this gets fixed, I have a request, could we change the text on this second button
2024-03-28_save-test-buttons
to say "Save and Test Search Query" so that it is clear to the user (me) that changes are also saved, not just tested.

laceysanderson (commented):
I took a look at the code and it doesn't raise any red flags :-)

Although I do see a `dpm` in the TripalPubLibrary (not in current changes). Dpms should never have been committed to the main branch 🙈 **Please comment out all dpms before this can be merged or replace them with drupal messages.**

I also agree with @dsenalik request that the buttons be more clear... Is it necessary to make the test button also save? This is not very intuitive to me either 🤔 

Also, how does the order you press the buttons unsave things? If clicking test does both save + test then why when you press save again does that cause it not to be saved?

risharde:
I think I found the issue with the clicking of the test button and then clicking the save button not saving on the first fire. 
But what I have not been able to recreate is the issue when you click on the test button and it saves automatically (that does not happen on my side) - so just in my defense here - this isn't supposed to happen @laceysanderson 

Oh and dpms, I really do try to not commit these, I do apologize if there are any that slipped by - will keep trying to keep these out

Let me push the latest changes and maybe you guys can retest it - hopefully it works correctly this time?

laceysanderson (approved):
I did a manual test on Drupal 10.2.x-dev, PHP 8.3 and PostgreSQL 16 and it works great for me :-)

Specifically, 
1. I went to `admin/tripal/loaders/publications/manage_publication_search_queries`
2. Clicked "New search query" and entered the same details as Doug mentioned.
3. Clicked test and confirmed the test worked. ✅ 
4. Checked the manage listing in another tab to confirm that the test button did not save the query. ✅ 
5. Back in the other tab where I tested the query, I now clicked save. 
6. Checked the listing tab to see that it did now save. ✅ 
7. In the query tab clicked test again to ensure it would test and it worked. ✅ 
8. Back in the listing I created a new query with my own search options, clicked test and then clicked back to the listing effectively leaving the form without saving. It did not save ✅ 
9. Went into the create new query form and entered my own, multi-field query and clicked save without testing first and confirmed it did save happily ✅ 
10. Editing my new query that I saved without testing, and now clicked test and confirmed that it can test on the edit form ✅ 
11. Made a change to the query and the name and clicked save and confirmed it did save ✅ 

Also, thank you for removing the dpms I mentioned!

dsenalik (approved):
Okay I did my tests again and everything works as expected.
I was just confused about the test button, somehow I thought it was supposed to save, but the way it is now is good, so the buttons are labelled correctly as-is.

I'll go ahead and merge this so you can work on V5 😀

PR #1835, opened 2024-03-28
GFF importer, fix loading gene sequence from FASTA section

dsenalik:
# Bug Fix

## Description
A gff3 file can provide sequence information in a `##FASTA` section.
Currently this is only working for landmark sequence, it is not working for a feature sequence like gene.

## Testing?
This simple gff3 file can be used for testing. Import with the gff3 importer, publish gene content type, and a sequence should be present for gene-YAL068C
[PAU8-withgeneseq.txt](https://github.com/tripal/tripal/files/14792698/PAU8-withgeneseq.txt)

2024-03-28_gene-with-sequence

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
+++ b/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
@@ -1432,7 +1432,7 @@ class GFF3Importer extends ChadoImporterBase {
       if (array_key_exists($uniquename, $this->features)) {
         $findex = $this->features[$uniquename]['findex'];
         $feature = $this->getCachedFeature($findex);
-        $feature_id = $feature['feature_id'];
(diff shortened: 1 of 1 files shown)


mdondrup:
I think importing of genes sequences separately from the GFF fasta file is a good work-around to populate the sequence field. however, it would require quite some pre-processing on behalf of the user. I was hoping that either the GFF importer or the UI itself would extract the gene sequence from alignment, ideally as it was the case in Tripal 2 (I cannot say how this worked in Tripal 3 because I never migrated). 
Example: https://licebase.org/feature/Tigriopus/kingsejongensis/gene/maker-scaffold693_size110418-snap-gene-0.25?pane=sequences

The sequence of the gene and related records is extracted from the coodinates imported in the GFF file.

laceysanderson (approved):
✅ Code looks clean, simple and intuitive 🤷‍♀️ 
✅ Functional test worked beautifully
image

Thanks for the fix @dsenalik!

PR #1839, opened 2024-04-01
Move helper functions out of Functional base and into Test Trait

### LINKED ISSUE #1838: Methods for creating content types + fields should be in a TestTrait
### Task Description

Currently there are a number of methods in [TripalTestBrowserBase](https://github.com/tripal/tripal/blob/4.x/tripal/tests/src/Functional/TripalTestBrowserBase.php#L38-L256) that should be moved into a test trait so that it can be available to both functional and kernel tests.

### Branch Name

tv4g9-issue1838-testTrait

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

This PR:
- Creates a TripalTestKernel Base and uses it throughout since we didn't yet have one
- Move helper functions focused on creating content types, tripal terms and fields into a test trait and out of the functional test base.
- Ensures both functional and kernel Tripal test bases use the test trait so that all Tripal tests have access to the helper functions.
- Move the chado test trait into the traits namespace where it belongs.
- Ensure the chado functional and kernel test bases extend the tripal test bases so that they have access to the same helper functions.

## Testing?

Ensure automated tests pass 🤷‍♀️


dsenalik (approved):
Awesome! I fixed the two typos I saw, which restarted the automated tests, but otherwise looks good. THANKS @laceysanderson 🙌

laceysanderson:
Thanks @dsenalik! I don't understand why the docker build is failing on a few versions 😕 I've tried building those versions exactly locally and they work fine! Plus it's got nothing to do with my change but is rather focused on apt-get!? Plus they passed on Github on this same commit https://github.com/tripal/tripal/actions/runs/8512495073

dsenalik:
The failed tests I saw happen on another PR recently, I think it is that pesky "internet glitch" that shows up every so often, something doesn't download correctly. We can just rerun the failed tests...

PR #1840, opened 2024-04-01
Modified composer.json php version to 8.1

pdtouch:
#


composer.json
```diff
@@ -22,6 +22,6 @@
         }
   ],
   "require": {
-    "php": "^8.0"
+    "php": "^8.1"
```
laceysanderson: Add drupal 10 requirement too
```suggestion
    "php": "^8.1",
    "drupal/core": "^10.1"
```

laceysanderson (approved):
Thanks @pdtouch! 🙏

laceysanderson:
Note: It is expected that the test coverage will not pass on forked PRs due to permissions.

PR #1842, opened 2024-04-02
Automated Field Testing: focused on the field creation forms in Drupal 10.2+

### LINKED ISSUE #1822: Automated Testing for fields
### Task Description

All the automated testing we currently have for fields is actually testing of ChadoStorage with specific properties set by each field.

We still need automated testing for fields themselves.

Specifically,
- testing of the forms that add a field to a content type.
- testing that the widget for a field is present and as expected when editing a content that field is present on.
- testing that the formatter returns the expected markup for a given piece of content

Additionally, it would be good to have very basic tests from the field perspective to ensure ChadoStorage is doing what we expect. Specifically,
- when we edit content, ensure that the widget is given the expected values in the expected format by chadoStorage
- when we create content by filling out the widget form, ensure that chado storage creates the content correctly
- when we load a content page, ensure that the formatter is provided the values as expected by chado storage.

### Branch Name

tv4g9-issue1822-fieldTesting

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

This PR tests the field create UI in Drupal 10.2+ for four simple chado fields. I started this is reference to #1805 which did not have automated testing to confirm that we don't break this annoying collection of subforms in the future 🤪 

This test walks through the create a field forms. It has to submit every time there would normally be ajax involved since ajax doesn't work in the test environment. Specifically, it creates a field attached to a feature content type:
 - chado integer using the feature.feature_id column
 - chado boolean using the feature.is_analysis column
 - chado string using the feature.name column
 - chado text using the feature.uniquename column

In all these cases it fills in the field label, machine name, selects the storage information, provides some help text and chooses a cvterm from the autocomplete.

## Testing?

Automated testing changes only so they way to test is by confirming they pass and doing a code review.

You can also add `print $this->getSession()->getPage()->getContent();` at any point in the test to look at the full output of the virtual page in the test environment.


tripal_chado/tests/src/Functional/Plugin/Fields/CreateEditFieldForms/BasicDataTypeChadoFieldTest.php
```diff
@@ -0,0 +1,267 @@
...
+  protected $connection;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() : void {
+    parent::setUp();
+
+    // Installs up the chado with all the items added via the prepare.
+    // NOTE: This done not prepare Drupal so none of the TripalTerms we need are available.
```
spficklin:
```
    // Installs chado with all the items added via the prepare.
    // NOTE: This does not prepare Drupal, so none of the TripalTerms we need are available.
```

tripal_chado/tests/src/Functional/Plugin/Fields/CreateEditFieldForms/BasicDataTypeChadoFieldTest.php
```diff
@@ -0,0 +1,267 @@
...
+    // ONLY do this test if we are in Drupal 10.2+ since we are assuming
+    // that the field storage + settings form are combined into a single page.
+    if (version_compare(\Drupal::VERSION, 10.2) < 0) {
+      $this->markTestSkipped('Test only applies to Drupal 10.2+');
+    }
+
+    // Pages to access.
+    $manage_fields_path = 'admin/structure/bio_data/manage/' . $this->type . '/fields';
+    $add_field_path = '/admin/structure/bio_data/manage/' . $this->type . '/fields/add-field';
+
```
spficklin, line 181: The `$manage_fields_path` does not have a leading `/` in the path but $add_field_path does.... Is that supposed to be that way?  They both get used in a `$this->drupalGet()` function call.   I suppose the test is working but it seemed inconsistent.

spficklin:
Code review looks good.  I don't have any issues with it, just a few questions (above).  I'm just waiting for tests to finish then I'll approve.

tripal_chado/tests/src/Functional/Plugin/Fields/CreateEditFieldForms/BasicDataTypeChadoFieldTest.php
laceysanderson: Good catch :-) Nope, there is no reason for the difference here so we might as well make them consistent.

tripal_chado/tests/src/Functional/Plugin/Fields/CreateEditFieldForms/BasicDataTypeChadoFieldTest.php
laceysanderson: Is this meant to be a suggested change?

spficklin: Yeah. I guess I did it wrong :-/ I couldn't remember.

laceysanderson:
Thanks for the review @spficklin!

PR #1844, opened 2024-04-09
Implements the ChadoBuddy framework + Cvterm/Dbxref buddies

### LINKED ISSUE #1780: Chado CVTerm service: do we need one or do we just always use Tripal Terms?
### Discussion or Question

This issue is to hold a discussion for whether we want to add a chado cvterm service to support creating, updating, deleting terms in chado. In Tripal 3 these were the following circled cv and db api functions:
cv api
db api

In Tripal 3 we had the above in addition to Tripal Terms which were associated with content types and fields.

In Tripal 4 we already have the Tripal Terms API implemented and it includes a chado storage backend to allow use of chado cvterms directly.

However, we currently use the legacy versions of the functions from the screenshot when creating terms not used for content types and fields --or at least that is what the USASK team is doing in extension modules. These legacy functions are very slow and unreliable so I would like to update them to the new D10 service model...

However, in codefest it came up that maybe we should just use Tripal terms in all cases and we do not need a chado cvterm/db service like we had in Tripal 3.

This issue is to hold discussion on this idea and design if we do go forward with a cvterm/db service.

*Note: I've marked this urgent because the legacy functions provided in Tripal 4 are quite buggy and causing us issues in our extension modules. The urgent part is to decide a course of action :-) I am happy to be the person to implement.*

laceysanderson:
# Tripal 4 Core Dev Task

## Description

In Tripal 3 we had a cv/cvterm/db/dbxref function-based API in addition to Tripal Terms (associated with content types and fields). In Tripal 4 we already have the Tripal Terms API implemented and it includes a chado storage backend to allow use of chado cvterms directly.

We currently use the legacy versions of the cv/cvterm/db/dbxref function-based API when creating terms not used for content types and fields. These legacy functions are very slow and unreliable so this PR updates them to the new D10 plugin model. It also creates a ChadoBuddy base class and a ChadoBuddyRecord class.

### Documentation
https://github.com/tripal/tripal_doc/pull/90

### Task List

- [x] Implement ChadoBuddy plugin type
- [x] Create testing outline
- [x] Test base class/plugin manager fully
- [x] Cvterm Buddy: outline
- [x] Cvterm Buddy: getCv()
- [x] Cvterm Buddy: getCvterm()
- [x] Cvterm Buddy: insertCv()
- [x] Cvterm Buddy: insertCvterm()
- [x] Cvterm Buddy: updateCv()
- [x] Cvterm Buddy: updateCvterm()
- [x] Cvterm Buddy: upsertCv() and upsertCvterm()
- [x] Cvterm Buddy: associateCvterm()
- [x] Dbxref Buddy: outline
- [x] Dbxref Buddy: getDb()
- [x] Dbxref Buddy: getDbxref()
- [x] Dbxref Buddy: getDbxrefUrl()
- [x] Dbxref Buddy: insertDb()
- [x] Dbxref Buddy: insertDbxref()
- [x] Dbxref Buddy: updateDb()
- [x] Dbxref Buddy: updateDbxref()
- [x] Dbxref Buddy: upsertDb() and upsertDbxref()
- [x] Dbxref Buddy: associateDbxref()

### Tasks added to support GFF3 importer
- [x] Cvterm Buddy: getCvtermSynonym()
- [x] Cvterm Buddy: insertCvtermSynonym()
- [x] Cvterm Buddy: updateCvtermSynonym()
- [x] Cvterm Buddy: upsertCvtermSynonym()
- [x] Add `case_insensitive` to options array

### Automated Testing

This API is being built with full automated testing. You can run these independantly using the ChadoBuddy group. For example, where you are using Tripaldocker and the container name is `tripal1780`:

```
docker exec --workdir=/var/www/drupal/web/modules/contrib/tripal tripal1780 phpunit --group ChadoBuddy
```


dsenalik:
A List of Questions so far
1. Do we want getters/setters for the `schema_name` and `base_table` variables in the ChadoBuddyRecord, or make them public.
2. Internally the Cvterm buddy is calling the Dbxref buddy, so right now it creates an instance, though I store it in a class variable - It does work to save it for testing, and it is quick, 0.000383 seconds so I don't think it is a performance issue. Otherwise, is there at least a way to get at the service from inside the service?
3. Should the buddies include a function like Tripal 3 to get terms for a select list, or should that go somewhere else? Currently we have `tripal_chado/src/api/tripal_chado.cv.api.php:349:function chado_get_cvterm_select_options($cv_id, $rel_type = FALSE, $schema_name = NULL)`
4. For performance reasons, I have some associative arrays to convert options passed to the buddies into chadotable.chadocolumn - this simplifies cases where the column name by itself would not be unique - like cv.name and cvterm.name

dsenalik:
Working on documentation in https://github.com/tripal/tripal_doc/issues/89

tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
```diff
@@ -43,4 +44,75 @@ public function description(): string {
...
+   * indicates that the alias 'db_name' corresponds to the
+   * column 'name' in the 'db' table.
+   *
+   * @param array $user_values
+   *   An associative array to be validated.
+   * @param array $valid_values
+   *   An associative array listing all valid keys, and the
+   *   values are un-aliased database table alias and table
+   *   column as described above.
+   * @param bool $filter
```
laceysanderson: This function is a bit confusing as it appears to do two things. I think it would be more clear if you change this $filter parameter to a string where it's value is the table name. If filter is provided then only valid columns from that table are returned and the key is "unaliases" but if that parameter is NULL then the user_values are returned as is but an exception is thrown for invalid columns.

tripal_chado/src/ChadoBuddy/ChadoBuddyRecord.php
```diff
@@ -20,9 +22,118 @@ class ChadoBuddyRecord {
...
+  /**
+   * An associative array where the keys are chado table column
+   * names and the values are chado table record values.
+   * @var array
+   */
+  protected array $values;
+
+
+  /**
+   * Sets the value of the base table.
```
laceysanderson, line 44: It might be good to be clear what you mean here by the base table. For example
> In this context the base table is the core table for this buddy. For example, the base table would be cvterm for the CvtermBuddy. A more complex example, is for the PropertyBuddy where the base table is feature for a property buddy adding properties to the feature table.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -16,15 +17,91 @@
...
 class ChadoCvtermBuddy extends ChadoBuddyPluginBase {
 
+  /**
+   * Keys are the column aliases, and values are the
+   * table aliases and columns for the dbxref buddy.
+   * @var array
+   *
+   */
+  protected array $db_mapping = [
+    'db_id' => 'db.db_id',
```
laceysanderson: Why do we need this mapping? Can we not just keep the keys as `db.name` rather then `db_name`?

Additionally, it would be good if the valid columns for each table are pulled from ChadoSchema not hard-coded as this would allow better support for multiple chado versions. If you are worried about speed you could cache the result from chado schema and/or add a method to the ChadoSchema class that grabs the valid columns directly from the YAML to avoid extracting the info from the array returned from getTableDefn

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -34,26 +111,73 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase {
    *   If the select values return multiple records, then we return an array
    *     of ChadoBuddyRecords describing the results.
    *   If there are no results then we return FALSE and if an error is
-   *     encountered then an exception will be thrown.
+   *     encountered then a ChadoBuddyException will be thrown.
```
laceysanderson: Note: there is actually a drupal standard for documenting exceptions thrown :-) https://www.drupal.org/docs/develop/standards/php/api-documentation-and-comment-standards#throws To use it you would leave this as is but then add the `@throws` tag as well to make it easier to see as a developer.

tripal_chado/tests/src/Kernel/Plugin/ChadoBuddy/ChadoCvtermBuddyTest.php
```diff
@@ -13,31 +13,92 @@
...
 
-  protected ChadoConnection $connection;
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Open connection to a test Chado
+    $this->connection = $this->getTestSchema(ChadoTestKernelBase::INIT_DUMMY);
```
laceysanderson: It would be good to add our own cvterm-focused test chado instance rather then re-using this non-standard one added by Valentin.

I would suggest adding a new fixture `cvterm_focused_schema.sql` with an SQL comment at the top indicating which tables it includes and that it includes test data. Make it clear these table definitions match Chado 1.3 and that this schema can be used when testing cvterm-related functionality.

Then to make this work within our automated testing you would:
1. Add a block like this to the chado test trait: https://github.com/tripal/tripal/blob/4.x/tripal_chado/tests/src/Traits/ChadoTestTrait.php#L319-L328
2. Add the same constant to the functional and kernel test bases like this: https://github.com/tripal/tripal/blob/4.x/tripal_chado/tests/src/Traits/ChadoTestTrait.php#L319-L328 and https://github.com/tripal/tripal/blob/4.x/tripal_chado/tests/src/Traits/ChadoTestTrait.php#L319-L328

A good constant might be `CHADO_CVTERM_SUBSET`. For linked tables you could keep feature or move to something simpler like contact or protocol which have a type_id but no other FKs.

laceysanderson:
Response to questions from above:

> 1. Do we want getters/setters for the `schema_name` and `base_table` variables in the ChadoBuddyRecord, or make them public.

I do thing getters/setters would be good for this :-)

> 2. Internally the Cvterm buddy is calling the Dbxref buddy, so right now it creates an instance, though I store it in a class variable - It does work to save it for testing, and it is quick, 0.000383 seconds so I don't think it is a performance issue. Otherwise, is there at least a way to get at the service from inside the service?

Ideally you would use dependency injection here to get the dbxrefBuddy service inside the cvtermBuddy service. You would use [Method 2 described here](https://github.com/tripal/tripal/issues/1522). 

The Base ChadoBuddy uses dependency injection by having the plugin manager pass in it's dependancies rather then needing to use the create() approach described in the method above.

For children buddies, there are two different approaches we could take:

A. Use the create/constructor approach and just handle the base dependencies independently in the child.

B. Add an annotation where each buddy can indicate dependant services including other buddies and then have the plugin manager pass them through too.

I think we want to go with A actually but will think a bit on this. Do you have thoughts?

> 3. Should the buddies include a function like Tripal 3 to get terms for a select list, or should that go somewhere else? Currently we have `tripal_chado/src/api/tripal_chado.cv.api.php:349:function chado_get_cvterm_select_options($cv_id, $rel_type = FALSE, $schema_name = NULL)`

This is up for discussion but I think it should be a controller similar to what we do for the autocompletes now.

>     4. For performance reasons, I have some associative arrays to convert options passed to the buddies into chadotable.chadocolumn - this simplifies cases where the column name by itself would not be unique - like cv.name and cvterm.name

I addressed this above but for completeness, I think we use the chadotable.chadocolumn notation directly for the keys rather then worrying about a mapping and we should get the list of valid columns per table from the chadoSchema class.

tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
dsenalik: This is now split into two separate functions, so it should be much clearer.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: Implemented as you suggested, and values cached. Adding a method to the ChadoSchema class can be a future project.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: Adding @throws tags. There is also a @todo tag option, using that for the cvterm url comment!

tripal_chado/tests/src/Kernel/Plugin/ChadoBuddy/ChadoCvtermBuddyTest.php
dsenalik: I am seeing about 4.7 seconds to create a test chado with
`$this->connection = $this->getTestSchema(ChadoTestKernelBase::PREPARE_TEST_CHADO);`
vs. 0.3 seconds for the `INIT_DUMMY` schema.

So, I am back to using `PREPARE_TEST_CHADO`, because the super-simplified `INIT_DUMMY` schema does not allow looking up table columns or primary key. Making a new "faster" test schema can be a future project.

tripal_chado/src/ChadoBuddy/ChadoBuddyRecord.php
dsenalik: This change has been added

dsenalik:
Note for another feature to add. The GFF importer looks up a cvterm including synonym, the buddy does not yet support this.
```
    $sel_cvterm_sql = "
      SELECT CVT.cvterm_id, CVT.name, CVTS.synonym
      FROM {1:cvterm} CVT
        LEFT JOIN {1:cvtermsynonym} CVTS on CVTS.cvterm_id = CVT.cvterm_id
      WHERE CVT.cv_id = :cv_id and
       (lower(CVT.name) = lower(:name) or lower(CVTS.synonym) = lower(:synonym))
    ";

    $result = $chado->query($sel_cvterm_sql, [
      ':cv_id' => $cv_id,
      ':name' => $type,
      ':synonym' => $type,
    ]);
    $cvterm_match = $result->fetchObject();

```

dsenalik:
Buddy documentation is close to being done, see https://github.com/tripal/tripal_doc/pull/90

tripal_chado/tests/src/Kernel/Plugin/ChadoBuddy/ChadoCvtermBuddyTest.php
dsenalik: This is moved to issue #1947 so I will resolve the conversation here

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -0,0 +1,1016 @@
...
+   *   Associative array of options.
+   *     - 'case_insensitive' - a single key, or an array of keys
+   *                            to query case insensitively.
+   *
+   * @return bool|array|ChadoBuddyRecord
+   *   If the select values return a single record then we return the
+   *     ChadoBuddyRecord describing the chado record.
+   *   If the select values return multiple records, then we return an array
+   *     of ChadoBuddyRecords describing the results.
+   *   If there are no results then we return FALSE.
```
laceysanderson: After discussion we think we're going to change this. It was written this way based on the old API functions but we've decided to change the return value to simplify code in the future even if it makes it a bit harder to upgrade to these functions.
```suggestion
   * @return array
   *   An array of ChadoBuddyRecord objects. More specifically,
   *   (1) if the select values return a single record then we return an 
   *     array containing a single ChadoBuddyRecord describing the record.
   *   (2) if the select values return multiple records, then we return an 
   *     array of ChadoBuddyRecords describing the results.
   *   (3) if there are no results then we return an empty array.
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -0,0 +1,1016 @@
...
+    }
+    return $new_record;
+  }
+
+  /**
+   * Add a record to a controlled vocabulary term linking table (e.g. feature_cvterm).
+   *
+   * @param string $base_table
+   *   The base table for which the cvterm should be associated. Thus to associate
+   *   a cvterm with a feature the basetable=feature and cvterm_id is added to the
```
laceysanderson:
```suggestion
   *   a cvterm with a feature the base_table=feature and cvterm_id is added to the
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -0,0 +1,1016 @@
...
+   *   feature_cvterm table.
+   * @param int $record_id
+   *   The primary key of the basetable to associate the cvterm with.
```
laceysanderson:
```suggestion
   *   The primary key of the base_table to associate the cvterm with.
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -0,0 +1,1016 @@
...
+    if (!$base_pkey_col) {
+      $schema = $this->connection->schema();
+      $base_table_def = $schema->getTableDef($base_table, ['format' => 'Drupal']);
+      $base_pkey_col = $base_table_def['primary key'];
+    }
+    $fields = [
+      'cvterm_id' => $cvterm->getValue('cvterm.cvterm_id'),
+      $base_pkey_col => $record_id,
+    ];
+    // Add in any of the other columns for the linking table.
```
laceysanderson, line 971: What about if the user chooses not to care about performance and doesn't add the keys needed for their particular linking table? Shouldn't we then look them up?

What about having an option `lookup_columns` that is TRUE by default unless other keys in addition to the pkey are supplied? So the default would be determined as follows:
```php
$lookup_columns = $options['lookup_columns'] ?? TRUE;
$possible_columns = ['pub_id','rank','is_not','cvterm_type_id'];
$columns_defined = in_array($possible_columns, $options);
if ($lookup_columns === TRUE && count($columns_defined) > 0) {
  $lookup_columns = FALSE;
}
```

laceysanderson (commented):
Code review excluding tests complete :-)
Only two main suggestions:
1. Change the return value for the getters 🙈 to always be an array.
2. Lookup the linker table columns if none are supplied in the options and lookup_columns is not FALSE.

Just running the code coverage locally in order to confirm the tests :-)

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
laceysanderson: Marked this unresolved since the code change to the docblock was committed but the change to the code itself has not yet been made and we don't want to lose this.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: Resolving as I have added this.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: Implemented in code, and so this can be resolved

laceysanderson (commented):
Changes look great @dsenalik :-) 

I added a few more tests and some more placeholders for tests for the base class. This is where we currently are coverage-wise:
- Base Tests only (I'd like to see these close to 100% for all):
  Screenshot 2024-08-20 at 8 06 33 PM
- Buddy Record (I'd like to 100% here):
  Screenshot 2024-08-20 at 8 08 09 PM
- ChadoCvtermBuddy: I'm pretty happy with these :-) Most are database exceptions which I don't think need covering.
  Screenshot 2024-08-20 at 8 12 14 PM
- ChadoDbxrefBuddy: I'm pretty happy with these :-) Most are database exceptions which I don't think need covering.
  Screenshot 2024-08-20 at 8 13 43 PM

So I think once the base tests are completed, this should be good to go! I'm moving on to other PR reviews but if you need help implementing the remaining tests -let me know :-)

dsenalik:
Base class tests at 100% at least!
2024-08-21-base

laceysanderson (commented):
✅ Code review looks good!
✅ Good automated test coverage 😍 
✅ Basic manual testing in php:cli works as expected

This is ready to merge!!! 🎉 Thanks for all the help @dsenalik! This looks awesome 😍

Issue #1815, opened 2024-03-15
Locked Mview handling


Ferrisx4:
### Task Description

As an addition to issue #1779, we need to implement functionality surrounding locking materialized views in Tripal. These include:
 - Setting the locked status in the GUI when creating an mview
 - Not allowing editing if the mview is locked
 - ~~Do not allow populating the mview if it is locked~~

### Branch Name

tv4g2-issue1815-locked-mview-handling

Ferrisx4:
![Screen Shot 2024-04-16 at 2 43 22 PM](https://github.com/tripal/tripal/assets/9593623/32c6d511-b559-499c-8aff-f6f618abc102)

For the second point in the initial issue description, I just switched the "Edit" button to say "Locked" instead, as seen in the screenshot, and clicking the button does nothing. @laceysanderson do you remember if we talked about other functionality instead of just locking it?

---

Also, somewhat unrelated, but for the filter area, do we want to somehow indicate that that is what it is? There's no label on the box and the button just says "Apply". This could be its own issue or it could be a simple fix involved in the changes within this issue.

Ferrisx4:
Alternatively, we can make the default action be "Populate" and only include Edit in the dropdown if the mview is not locked

laceysanderson:
I like the "Locked" on the button but I would change it so that when clicked they see the edit page but the form elements are disabled. This would allow them to at least see what the locked settings are ;-p since we do not have a view. 

I do wonder if we may want to also make the default to populate for all of them. This might be better for usability now that you mention it as that is the most common action done on this page :-)

Ferrisx4:
Ok thanks @laceysanderson, I am on board with these suggestions.

I also remembered this comment from #1779: https://github.com/tripal/tripal/issues/1779#issuecomment-1939299988 which discussed adding a blurb at the top describing mviews and where to edit them.

Would I be able to sneak this into this issue? I've also added the filter description as I mentioned in the second comment above, for your consideration
![Screen Shot 2024-04-26 at 8 54 42 AM](https://github.com/tripal/tripal/assets/9593623/41b43499-72eb-4a79-9013-db0169d52bb0)

laceysanderson:
Yup, feel free to sneak that into this PR :-) I like your added filter description as well 👍

Ferrisx4:
> Setting the locked status in the GUI when creating an mview

[...]

PR #1847, opened 2024-04-12
Tv4g2 issue1779 no default custom tables

Ferrisx4:
# Bug Fix

## Description

This reverts some changes to the ChadoMview class that originally implemented setLocked and isLocked functions. Upon further discussion, that functionality should remain solely in the parent class, ChadoCustomTable, as we want the locked status of a custom table to match that of its own materialized view.

This finishes up 1779, further functionality surrounding Chado Custom Tables and their mviews will be attached to issue #1815


laceysanderson (approved):
✅ Code changes look good 👍 
✅ Custom tables and mview lists are now populated as I expect.

Thanks @Ferrisx4

PR #1846, opened 2024-04-12
Catch adding Chado field to non-Chado content type

dsenalik:
# Bug Fix

## Description
This adds a simple check to catch when attempting to add a Chado field to a non-Chado content type. Instead of a WSOD as described in issue #1837, we will now display a simple error message.

## Testing?
  1. Create a new Drupal content type. Go to `/admin/structure/types/add`
  2. Enter a "Name" and then click "Save and manage fields"
  3. Click "+ Add a new field"
  4. Enter a "Label"
  5. Select Chado fields
  6. Pick any Chado field, e.g. analysis
  7. Click "Continue"
 A friendly error message should replace the former WSOD
2024-04-11_invalid_field_appearance

PR #1849, opened 2024-04-12
Update chado prepare sql for automated tests

### LINKED ISSUE #1848: Testing is currently failing
### Tripal Version

4.x

### General Description

Automated tests are failing on recent PRs. It seems that we have 40 `db` records instead of the expected 41.

The changes in the `db` table from an older docker to a new one is
| Before | After |
|--|--|
| WIKI | wikipedia |
| XX |  |

The `WIKI` db had one dbxref `https\://en.wikipedia.org/wiki/Allele_frequency`
The `XX` db had one dbxref `www.ensembl.org/info/genome/variation/predicted_data.html#consequences`
The `wikipedia` db has one dbxref `Allele_frequency`

I suspect an ontology may have been updated. We might need to regenerate `tripal_chado/tests/fixtures/fill_chado_test_prepare.sql` and change expected count to 40?

### Steps to reproduce

`docker exec --workdir=/var/www/drupal9/web/modules/contrib/tripal containername phpunit --testdox --verbose --filter testChadoPreparer`

dsenalik:
# Bug Fix

## Description
Apparently some ontologies have been updated, so automated tests are failing as we get different numbers of table row than previously. Affected so far: PR #1844, PR #1846,  PR #1847

## Testing?
A quick code review is always good, but automated testing should do the real testing.


laceysanderson (approved):
These changes look perfect -thanks @dsenalik

PR #1851, opened 2024-04-15
New Widgets for Property Fields

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR adds two new widgets to the core property fields:
 - chado_property_string_widget_default: provides a simple textbox for a cleaner and more compact widget
 - chado_property_select_widget_default: provides a configurable select box

These were added to improve the experience of data entry for curators and to allow admin to control the entries accepted for various metadata.

Note: The select widget configuration UI specifically does not support a key => value mapping as we want to ensure the value entered for metadata matches what is displayed.

## Testing?

1. Start a fresh site on this branch or drush cr on an existing site.
2. Create 1+ property fields on a content type of your choice. In this example we will create 2 fields on the organism content type: Category (Term: Group (NCIT:C43359)) only supports a single value; Breeding Techniques (Term: Methods (local:Methods)) which supports unlimited values.
3. Go to Manage form display and confirm these property type fields have 3 options for widget (Long Text, Short Text, Select Drop-down) and the default is "Long Text". **Do not change them yet** Screenshot 2024-04-15 at 3 13 00 PM
4. Go to the add content page to create an organism and confirm that the two property widgets match the old long text style. Screenshot 2024-04-15 at 3 07 04 PM
5. Now go back to manage form display for organism and change one or both properties to use the short text widget. When you edit or create an organism these fields now use a textfield. They do not limit the length of text entered though! Screenshot 2024-04-15 at 3 09 41 PM
6. Go back to manage form display for organism and change one or both properties to use the select drop-down widget. Right after selecting it you should see a configuration cog appear and the summary "There are no options yet configured."Screenshot 2024-04-15 at 3 28 54 PM
7. Click on the cog and configure the options. Then click update and ensure the summary statement has been updated to indicate the number of options you configured. Click save at the bottom of the page. Screenshot 2024-04-15 at 3 31 22 PM
8. Now check existing content:
  - If you edit an existing organism who had an entry typed in that matched an entry in the select list then it should be automatically chosen.  Screenshot 2024-04-15 at 3 33 35 PM
  - If you edit an existing organism who had an entry NOT in the select list then `- None -` will be chosen and a warning will be shown to the curator. Screenshot 2024-04-15 at 4 01 14 PM
9. Confirm that you can edit and save till your heart's content.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
@@ -122,7 +122,7 @@ class ChadoPropertySelectWidgetDefault extends ChadoWidgetBase {
    */
   public static function defaultSettings() {
     return [
-      'options' => [],
+      'options' => '',
     ] + parent::defaultSettings();
   }
 
@@ -148,9 +148,13 @@ class ChadoPropertySelectWidgetDefault extends ChadoWidgetBase {
     $summary = [];
 
     $raw_options = $this->getSetting('options');
-    $count = sizeof(explode("\n", $raw_options));
-
-    $summary[] = $this->t("There are $count options configured.");
+    if (!empty($raw_options)) {
+      $count = sizeof(explode("\n", $raw_options));
+      $summary[] = $this->t("There are $count options configured.");
+    }
+    else {
+      $summary[] = $this->t("There are no options yet options configured.");
+    }
 
     return $summary;
   }
--- a/tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
@@ -1,0 +1,157 @@
+<?php
+
+namespace Drupal\tripal_chado\Plugin\Field\FieldWidget;
+
+use Drupal\tripal\TripalField\TripalWidgetBase;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\tripal_chado\TripalField\ChadoWidgetBase;
+
+/**
+ * Plugin implementation of default Tripal linker property widget.
+ *
+ * @FieldWidget(
+ *   id = "chado_property_select_widget_default",
+ *   label = @Translation("Chado Property Select"),
+ *   description = @Translation("Provides a configurable drop-down widget for Chado Properties."),
+ *   field_types = {
+ *     "chado_property_type_default"
+ *   }
+ * )
+ */
+class ChadoPropertySelectWidgetDefault extends ChadoWidgetBase {
+
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+
+    // Get the field settings.
+    $field_definition = $items[$delta]->getFieldDefinition();
+    $field_settings = $field_definition->getSettings();
+
+    // Get the default values.
+    $item_vals = $items[$delta]->getValue();
+    $record_id = $item_vals['record_id'] ?? 0;
+    $prop_id = $item_vals['prop_id'] ?? 0;
+    $linker_id = $item_vals['linker_id'] ?? 0;
+    $default_value = $item_vals['value'] ?? NULL;
+    $term_id = NULL;
+    if ($field_settings['termIdSpace'] and $field_settings['termAccession']) {
+      $idSpace_manager = \Drupal::service('tripal.collection_plugin_manager.idspace');
+      $idSpace = $idSpace_manager->loadCollection($field_settings['termIdSpace']);
+
+      $term = $idSpace->getTerm($field_settings['termAccession']);
+      $term_id = $term->getInternalId();
+    }
+
+    // Process the options from the configuration.
+    $raw_options = $this->getSetting('options');
+    $options = [];
+    foreach( explode("\n", $raw_options) as $item) {
+      $item = trim($item);
+      $options[$item] = $item;
+    }
+
+    $elements = [];
+    $elements['record_id'] = [
+      '#type' => 'value',
+      '#default_value' => $record_id,
+    ];
+    $elements['prop_id'] = [
+      '#type' => 'value',
+      '#default_value' => $prop_id,
+    ];
+    $elements['linker_id'] = [
+      '#type' => 'value',
+      '#value' => $linker_id,
+    ];
+    $elements['type_id'] = [
+      '#type' => 'value',
+      '#value' => $term_id,
+    ];
+    $elements['value'] = $element + [
+      '#type' => 'select',
+      '#default_value' => $default_value,
+      '#empty_option' => '- None -',
+      '#options' => $options,
+      '#required' => FALSE,
+    ];
+    $elements['rank'] = [
+      '#type' => 'value',
+      '#value' => $delta,
+    ];
+    return $elements;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
+    $storage = \Drupal::entityTypeManager()->getStorage('chado_term_mapping');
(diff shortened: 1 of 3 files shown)


ruobinLiu:
I have tested the setting on a Tripal docker site.  The drop downs all showed as expected after config 🎉

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,168 @@
...
+  public function settingsSummary() {
+    $summary = [];
+
+    $raw_options = $this->getSetting('options');
+    if (!empty($raw_options)) {
+      $count = sizeof(explode("\n", $raw_options));
+      $summary[] = $this->t("There are $count options configured.");
+    }
+    else {
+      $summary[] = $this->t("There are no options yet options configured.");
```
dsenalik:
```suggestion
      $summary[] = $this->t("There are no options configured yet.");
```

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $element['options'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Widget Options'),
+      '#description' => "Enter the options you want to be available in the widget drop down with each option on it's own line.",
```
dsenalik:
```suggestion
      '#description' => $this->t("Enter the options you want to be available in the widget drop down with each option on its own line."),
```

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,168 @@
...
+    $options = [];
+    foreach( explode("\n", $raw_options) as $item) {
+      $item = trim($item);
+      $options[$item] = $item;
+    }
+
+    // Check that the default value is one of the available options.
+    if (!array_key_exists($default_value, $options)){
+      $field_label = $field_definition->label();
+      $this->messenger()->addWarning("The value saved for the '$field_label' does not match any of the available options. As such it has been set to none and the original value ('$default_value') will be overwritten when you save this page. Please contact your administrator to get this added as an option before making any changes to this page.");
```
dsenalik:
```suggestion
      $this->messenger()->addWarning($this->t("The value saved for the '@field_label' does not match any of the available options. As such it has been set to 'None' and the original value ('@default_value') will be overwritten when you save this page. Please contact your administrator to get this added as an option before making any changes to this page.",
          ['@field_label' => $field_label, '@default_value' => $default_value]));
```

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsSummary() {
+    $summary = [];
+
+    $raw_options = $this->getSetting('options');
+    if (!empty($raw_options)) {
+      $count = sizeof(explode("\n", $raw_options));
+      $summary[] = $this->t("There are $count options configured.");
```
dsenalik:
```suggestion
      $summary[] = $this->t("There are @count options configured.", ['@count' => $count]);
```

dsenalik (changes requested):
I started out testing the select before I tried the others. This might trigger something, I am seeing this for my two properties (this is as I am first creating the project)
2024-04-17_bug
but it seems to be saving my selections just fine.

Otherwise this is awesome, I made a few minor code suggestions 🙈

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,168 @@
...
+    // Process the options from the configuration.
+    $raw_options = $this->getSetting('options');
+    $options = [];
+    foreach( explode("\n", $raw_options) as $item) {
+      $item = trim($item);
+      $options[$item] = $item;
+    }
+
+    // Check that the default value is one of the available options.
+    if (!array_key_exists($default_value, $options)){
```
dsenalik:
```suggestion
    if (isset($default_value) and !array_key_exists($default_value, $options)){
```
Maybe the fix is this simple...I tested it and the error is gone. (I used isset() so that zero can be a valid option)

dsenalik:
One last note, if I set up the select list after data is present, and then convert to a select lacking one of the values, the warning appears as expected. 
But if I go ahead and save anyway, the same warning appears a second time.
2024-04-17_message2x
I don't think it should appear the second time. By then, the data is gone anyway.

laceysanderson:
Thanks for the review @ruobinLiu and the suggestions @dsenalik! I just confirmed that I can still reproduce this and that the simple fix was not quite enough 🙈 I will look into this. I agree that it should not show 2X, especially after the data is already gone.

> One last note, if I set up the select list after data is present, and then convert to a select lacking one of the values, the warning appears as expected. But if I go ahead and save anyway, the same warning appears a second time. I don't think it should appear the second time. By then, the data is gone anyway.

Screenshot 2024-04-18 at 9 57 03 AM

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,169 @@
...
+      $item = trim($item);
+      $options[$item] = $item;
+    }
+
+    // Check that the default value is one of the available options.
+    if (isset($default_value) and !array_key_exists($default_value, $options)){
+      $field_label = $field_definition->label();
+      $this->messenger()->addWarning($this->t("The value saved for the '@field_label' does not match any of the available options. As such it has been set to 'None' and the original value ('@default_value') will be overwritten when you save this page. Please contact your administrator to get this added as an option before making any changes to this page.",
+          ['@field_label' => $field_label, '@default_value' => $default_value]));
+    }
```
dsenalik:
```suggestion
    if (isset($default_value) and !array_key_exists($default_value, $options)) {
      $request_method = \Drupal::requestStack()->getCurrentRequest()->getMethod();
      // Do not display the warning when saving
      if ($request_method == 'GET') {
        $field_label = $field_definition->label();
        $this->messenger()->addWarning($this->t("The value saved for the '@field_label' does not match any of the available options. As such it has been set to 'None' and the original value ('@default_value') will be overwritten when you save this page. Please contact your administrator to get this added as an option before making any changes to this page.",
            ['@field_label' => $field_label, '@default_value' => $default_value]));
      }
    }
```
Would this be an appropriate way to display the message only once?

Another way may be
`if (!$form_state->getUserInput()) {`

laceysanderson: Oh that is an interesting approach I didn't think of. It should work though and I don't see any reason why it's not appropriate. I'm going to commit and test it :-)

tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertySelectWidgetDefault.php
```diff
@@ -0,0 +1,173 @@
...
+    foreach( explode("\n", $raw_options) as $item) {
+      $item = trim($item);
+      $options[$item] = $item;
+    }
+
+    // Check that the default value is one of the available options.
+    if (isset($default_value) and !array_key_exists($default_value, $options)) {
+      $request_method = \Drupal::requestStack()->getCurrentRequest()->getMethod();
+      // Do not display the warning when saving
+      if ($request_method == 'GET') {
```
dsenalik:
```suggestion
      // Do not display the warning when saving
      if (!$form_state->getUserInput()) {
```
Cleaner version without the extra service

laceysanderson:
Thanks for all the suggestions @dsenalik, I've confirmed that it now works locally. Can you revise your review?

dsenalik (approved):

PR #1855, opened 2024-04-20
Fixes validation missing values bug

### LINKED ISSUE #1826: Analysis validation (maybe others) don't show the offending values
### Tripal Version

4.x

### Drupal Version

10.3

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

The validation for unique constraint on the analysis table doesn't correctly report the offending values. Setting the topic to data storage because in `tripal_chado/src/TripalStorage/ChadoRecords.php` line 1458 there is this, and it might be the cause
```
                // @todo need to use the column alias when getting the value.
                $col_val = $record['values'][$col];
```

### Steps to reproduce

 - Create an analysis with algorithm "science" and version "1.0"
 - Create a second analysis with the same algorithm and software version and observe this

### Screenshots

2024-03-22_analysis_unique_constraint
Doesn't display the program or version :see_no_evil:

dsenalik:
# Bug Fix

## Description
This PR does three very simple things (plus a stray double space)
 1. change `array_key_exists` to `in_array` - this is all that was needed to display the missing values
 2. deletes the lines that `continue` if the column in the unique constraint **can be null**, and is null.
 By removing this, we now get feedback on **all** of the columns that make up the unique constraint.
 3. Adds a short explanation of the unique violation (commit https://github.com/tripal/tripal/pull/1855/commits/d16e43f5e07655b8b09b9e205520d1517118c9fb)
The error message for the described testing steps, where there are more than one columns making up the unique constraint, will now be
`The item cannot be saved as another already exists with the following values: Program: 'science', Programversion: '1.0', Sourcename: ''. The combination of these 3 values must be unique for every item.`

For a unique constraint based on a single column, for example contact, the message will appear as
`The item cannot be saved as another already exists with the following values: Name: 'Borg'. This value must be unique for every item.`

## Testing?
Described in issue #1826

### DIFF
--- a/tripal_chado/src/TripalStorage/ChadoRecords.php
+++ b/tripal_chado/src/TripalStorage/ChadoRecords.php
@@ -1449,7 +1449,7 @@ class ChadoRecords  {
           if (($record_id == 0) or ($record_id != $match->$pkey)) {
             // Documentation for how to create a violation is here
             // https://github.com/symfony/validator/blob/6.1/ConstraintViolation.php
-            $message = 'The item cannot be saved as another already exists with the following values. ';
+            $message = 'The item cannot be saved as another already exists with the following values: ';
             $params = [];
             foreach ($ukey_cols as $col) {
               $col = trim($col);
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Nice clean fix. Code looks good and manual testing passes beautifully:

- ✅ Create two analyses with the same program + program version, leave data source empty. Gives the error as expected.
- ✅ For the above test, fill in the data source for the second record and confirm it now saves without error.
- ✅ Create two analyses with the same program + program version + non-empty data source. Gives the error as expected.
- ✅ For the above test, change the data source for the second record and confirm it now saves without error.
- ✅ Test the same situations on the organism content type
Screenshot 2024-04-22 at 2 20 21 PM

Issue #1816, opened 2024-03-15
Missing CV Terms


dsenalik:
### Task Description

Some chado table columns are missing terms.
Example, the `chado.featureloc` table does not have terms defined for: `is_fmin_partial`, `is_fmax_partial`, `residue_info`, and `locgroup`.

To list all chado table columns without terms assigned, this bit of code can be used. Run in devel module php window, or use `drush php:cli` Not everything here may need a term, ones that will be updated in this issue are marked with 🟢. Tables without a schema definition marked with `*` are all materialized view related, so can be ignored.
```
// Get the chado schema
$chado = \Drupal::service('tripal_chado.database');
$chado_schema = $chado->schema();

// Get the existing term mapping
$storage = \Drupal::entityTypeManager()->getStorage('chado_term_mapping');
$mapping = $storage->load('core_mapping');
$tables = $mapping->getTableNames();

foreach ($tables as $table_name) {
  $table_schema_def = $chado_schema->getTableDef($table_name, ['format' => 'Drupal']);
  if (!array_key_exists('fields', $table_schema_def)) {
    print "* No schema definition for table \"$table_name\"\n";
  }
  else {
    foreach ($table_schema_def['fields'] as $field => $properties) {
      $term = $mapping->getColumnTermId($table_name, $field);
      if ((!$term) and !($properties['type'] == 'serial')) {
        print "- $table_name.$field\n";
      }
    }
  }
}
```
output:
```
* No schema definition for table "analysis_organism"  (this is a materialized view)
- 🟢analysisfeature.rawscore
- 🟢analysisfeature.normscore
- 🟢analysisfeature.significance
- 🟢analysisfeature.identity
- analysisfeatureprop.analysisfeature_id
- biomaterial_treatment.treatment_id
- biomaterial_treatment.unittype_id
- cell_line_cvtermprop.cell_line_cvterm_id
- cell_lineprop_pub.cell_lineprop_id
- control.tableinfo_id
- control.row_id
* No schema definition for table "cv_root_mview"
- cvterm.is_relationshiptype
- cvterm_dbxref.is_for_definition
- cvtermpath.pathdistance
* No schema definition for table "db2cv_mview"
- 🟢eimage.eimage_data
- 🟢eimage.eimage_type
- 🟢eimage.image_uri
- expression_cvterm.cvterm_type_id
- expression_cvtermprop.expression_cvterm_id
- feature_cvterm_dbxref.feature_cvterm_id
[...]

PR #1861, opened 2024-04-27
Fix the sequence coordinates field

### LINKED ISSUE #1859: Fix the sequence coordinates field
### Tripal Version

4.x

### Drupal Version

probably any

### PostgreSQL Version

probably any

### PHP Version

probably any

### Branch

tv4g1-issue1859-sequence-coordinates-field
PR #1861

### General Description

In this comment https://github.com/tripal/tripal/issues/1823#issuecomment-2015135481 I noted that there is a bug in the sequence coordinates field.
Part of the fix was to add some missing terms, this was done in PR #1816 
With those in place we can now fix this field.

### Steps to reproduce

The steps in issue #1824 provide one way to trigger this error, so not repeating here.

### Error messages

`PDOException: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "featureloc_c1" DETAIL: Key (feature_id, locgroup, rank)=(4, 0, 0) already exists. in /var/www/drupal/web/core/lib/Drupal/Core/Database/StatementWrapperIterator.php:111`

dsenalik:
# Bug Fix

## Description
The sequence coordinates field was preventing us from editing a gene.
The reason is that the "linking table" for this field is the `featureloc` table. It contains a foreign key to the `feature` table i.e. `feature_id`, it contains another key back to the feature table. This is because a feature has a location on another feature, to put it briefly.

But this so-called "linking table" also has numerous other columns, such as `fmin`, `fmax`, `strand`, etc.
For chado storage to work, we need all of these columns defined in the field properties, and passed through in the widget. 

Making this harder was that some of the columns did not have terms defined. This part was fixed in PR #1816

This PR adds the missing properties, and includes an update function to fix this on an existing site.

## Testing?

 1. Import the type collection "Genomic Content Types"
 2. Create an organism [Saccharomyces cerevisiae](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=559292)
 3. Create an analysis
 4. Import this example file
For default landmark type, I used `DNA (SO:0000352)`
 5. Run the job `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
 6. Publish the "Gene" content type
 7. Run the job `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
 8. View the published "PAU8" gene. Observe the sequence location as `NC_001133.9:1806..2169-`
 9. Edit the gene, don't make any changes, and save.
 10. Verify that there are no errors, and that the sequence location is unchanged `NC_001133.9:1806..2169-`

It would also be good to build a 4.x docker and repeat steps 1 through 5, then switch to this branch, and run `drush updatedb` to run the update function before continuing


laceysanderson (approved):
Followed testing instructions via a container built on this branch.

✅ Worked perfectly
image

Then repeated as described on 4.x before switching as described in order to test the update script.

✅ Update script runs without error
✅ Sequence coordinates on gene look exactly as they did in the first screenshot

PR #1862, opened 2024-04-27
Enable Wysiwyg toolbar for Tripal/Chado Text fields

### LINKED ISSUE #1852: Chado Fields using the textarea do not support Wysiwyg / Editor toolbar
### Task Description

There are a number of Chado fields that provide a textarea in the widget. However, none of these have text filtering enabled and thus curators do not see the editing toolbar on these fields... This is a usability concern -especially for long description fields- as it makes it very difficult for formatted text to be entered.

I've found a tutorial, [Adding a WYSIWYG to text fields in Drupal 8 and 9](https://www.fourkitchens.com/blog/development/wysiwyg-text-fields-drupal-8-9/), indicating how we can alter existing Drupal core fields to support formatted text after the fact. I believe I can use the information here to enable our fields to display the Wysiwyg toolbar and allow admin to choose the filter.

### Branch Name

tv4g1-issue1852-wysiwygForChadoFields

laceysanderson:
# Tripal 4 Core Dev Task

## Description

There are a number of Chado fields that provide a textarea in the widget. However, none of these have text filtering enabled and thus curators do not see the editing toolbar on these fields... This is a usability concern -especially for long description fields- as it makes it very difficult for formatted text to be entered.

This PR alters the Tripal Text Field Widget + Formatter to support a configurable filter format. Choosing a filter format that the Wysiwyg toolbar is configured for, enables the toolbar on that field :-)

## Testing?

1. Checkout this branch on an existing Tripal site or in a docker container.
2. Ensure that you can still create an organism or any other content type with text fields without first going to the page structure form display settings. You should already see the Wysiwyg toolbar as it will use the default "Basic HTML" format.
3. Go to Page Structure > Organism > Manage Form Display and configure the Description field widget to use a filter format of "Full HTML". This will allow HTML to entered and rendered in the field without filtering.
4. Now go back to your organism and edit it. Confirm you now see the Wysiwyg toolbar and that you can use it to include formatted text :-) Confirm you can save the content without error. Confirm that it is rendered as expected on the page.

Screenshot 2024-04-27 at 3 14 23 PM

### DIFF
--- a/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalTextTypeFormatter.php
+++ b/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalTextTypeFormatter.php
@@ -28,7 +28,10 @@ class DefaultTripalTextTypeFormatter extends TripalFormatterBase {
 
     foreach($items as $delta => $item) {
       $elements[$delta] = [
-        "#markup" => $item->get("value")->getString(),
+        '#type' => 'processed_text',
+        '#text' => $item->get('value'),
+        '#format' => 'full_html',
+        '#langcode' => $item->getLangcode(),
       ];
     }
 
--- a/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalTextTypeFormatter.php
+++ b/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalTextTypeFormatter.php
@@ -26,6 +26,9 @@ class DefaultTripalTextTypeFormatter extends TripalFormatterBase {
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
 
+    // Default filter format.
+    $filter_format = 'basic_html';
+
     // We need to get the format set in the widget settings
     // because they need to match.
     $entity_type = $this->fieldDefinition->get('entity_type');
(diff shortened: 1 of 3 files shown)


tripal/src/Plugin/Field/FieldFormatter/DefaultTripalTextTypeFormatter.php
```diff
@@ -26,9 +26,27 @@ class DefaultTripalTextTypeFormatter extends TripalFormatterBase {
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
 
+    // Default filter format.
+    $filter_format = 'basic_html';
+
+    // We need to get the format set in the widget settings
```
dsenalik:
```suggestion
    // We need to get the format which was set in the widget settings
```

dsenalik (approved):
Made a docker on this branch, php 8.3 Drupal 10.2
 - [x] Tested creating a project, saw the wysiwyg toolbar by default
2024-04-27_project-basic-html

 - [x] Tested creating an organism, also saw the toolbar by default
 - [x] Changed to FULL HTML and entered `<h2>` and `<i>` tags and they rendered correctly
2024-04-27_fullhtml

 - [x] Tested Restricted HTML. No toolbar is displayed and only some tags can be used. I tested h2 i strong em. The i tag is filtered out.

2024-04-restricted

 - [x] Testing the same with plain, no tags are interpreted (as expected)
2024-04-plain

PR #1865, opened 2024-04-28
Implement the fallback entity lookup

### LINKED ISSUE #1850: Fallback entity lookup
### Task Description

PR #1782 implements entity lookup based on the CV term for a content type.
This allows us to distinguish multiple content types based on a single Chado table, for example the `analysis` table can be an `Analysis`, or it can be a `Genome Assembly`.
If we have a situation where we have a term that does not have a corresponding content type, but there is a chado record in a table that has a content type under a different term, we want to implement a fallback lookup for this content type using the third party setting where the chado table is stored.
For example the "Assay" content type has an "Operator" field with the term NCIT:C48036 = Operator which maps to the `chado.contact` table. Since "Operator" is not a content type, we would want to default to the "Contact" content type. But the current lookup just gives up since there is no "Operator" content type.

So the new lookup flow would be to 
 1. Try to lookup the chado table from the term.
 2. If there is no content type for that term, then try to lookup the chado table from the third party settings.

### Branch Name

tv4g1-issue1850-entity-lookup-third-party-settings

### PR #1865

dsenalik:
# New Feature

## Description
There are four cases in core Tripal where links use a different name, currently entity links do not work for these.
The goofy fields that we need to test are listed at the end of this comment https://github.com/tripal/tripal_doc/issues/32#issuecomment-1837532399
```
arraydesign FOREIGN KEY (manufacturer_id) REFERENCES contact(contact_id)
assay FOREIGN KEY (operator_id) REFERENCES contact(contact_id)
biomaterial FOREIGN KEY (biosourceprovider_id) REFERENCES contact(contact_id)
biomaterial FOREIGN KEY (taxon_id) REFERENCES organism(organism_id)
```

We had originally intended to consult the third party settings, but for reasons of efficiency I decided to have the TripalTypes property pass through the table name. It already knows this, and thus avoids a chunk of code to get at the third party settings to look it up.

Because I changed the entity_id property, we need an update function. Again for efficiency, I recycled the existing update function, so as not to duplicate it, it will just be run again with a higher number.

## Testing?
 1. Create your docker on the 4.x branch so we can test the update function
 2. Change to this branch `git checkout tv4g1-issue1850-entity-lookup-third-party-settings`
 3. Run the update `docker updatedb` - verify no errors are reported
 4. Import the gene expression type collection
 5. Create an organism
 6. Create a contact
 7. Create an array design. You will need a contact that you created in step 6
 8. Verify that the "Manufacturer" is a link 
 9. Create an assay. Use the contact as the "Operator"
 10. Again verify that the "Operator" is a link
 11. Create a biomaterial.
 12. Verify that both biosource provider and Organism (taxon id) are links
 13. I have added a kernel test for the fallback method (manufacturer on arraydesign) so make sure tests pass 🙌

### DIFF
--- a/tripal/src/Services/TripalEntityLookup.php
+++ b/tripal/src/Services/TripalEntityLookup.php
@@ -64,13 +64,16 @@ class TripalEntityLookup {
    *   The bundle's CV Term namespace e.g. for gene "SO"
    * @param string $termAccession
    *   The bundle's CV term accession e.g. for gene "0000704"
+   * @param string $base_table
+   *   The Chado base table for the requested entity e.g. for gene "feature".
+   *   Only needed if term does not map to a content type.
    * @param string $entity_type
    *   The type of entity, only 'tripal_entity' is supported.
    *
    * @return int|null
    *   The Drupal entity ID, or null if no match found.
    */
-  public function getEntityId($record_id, $termIdSpace, $termAccession, $entity_type = 'tripal_entity') {
+  public function getEntityId($record_id, $termIdSpace, $termAccession, $base_table = NULL, $entity_type = 'tripal_entity') {
 
     // Catch invalid entity type
     if ($entity_type != 'tripal_entity') {
@@ -80,8 +83,17 @@ class TripalEntityLookup {
     // Perform the lookup steps
     $entity_id = NULL;
     $bundle_id = $this->getBundleFromCvTerm($termIdSpace, $termAccession);
+
+    // If the term does not have a content type, fallback is
+    // to use the base table's default content type.
+    if (!$bundle_id and $base_table) {
+      $bundle_id = $this->getDefaultBundle($base_table);
+    }
+
     if ($bundle_id) {
-      $base_table = $this->getBundleBaseTable($bundle_id, $entity_type);
+      if (!$base_table) {
+        $base_table = $this->getBundleBaseTable($bundle_id, $entity_type);
+      }
       if ($base_table) {
         $entity_ids = $this->getEntityIdFromRecordId($record_id, $bundle_id, $entity_type);
         if ($entity_ids) {
@@ -112,6 +124,27 @@ class TripalEntityLookup {
       ->getStorage('tripal_entity_type')
       ->loadByProperties(['termIdSpace' => $termIdSpace, 'termAccession' => $termAccession]);
     if (sizeof($bundles) == 1) {
+      $bundle_id = key($bundles);
+    }
+    return $bundle_id;
+  }
+
+  /**
+   * Retrieve the default Tripal bundle for a given base
+   * table. This table name is expected to match the bundle id.
+   *
+   * @param string $base_table
+   *   The table name e.g. "feature"
+   *
+   * @return string|null
+   *   The bundle id, or null if no match found.
+   */
+  protected function getDefaultBundle($base_table) {
+    $bundle_id = NULL;
+    $bundles = \Drupal::entityTypeManager()
+      ->getStorage('tripal_entity_type')
+      ->loadByProperties(['id' => $base_table]);
+    if (sizeof($bundles) >= 1) {
       $bundle_id = key($bundles);
     }
     return $bundle_id;
@@ -255,7 +288,7 @@ class TripalEntityLookup {
       // Cache the values, specifying expiration in 1 hour.
       \Drupal::cache()->set($cache_id, $field_list, \Drupal::time()->getRequestTime() + (3600));
     }
-    return $field_list[$bundle_id];
+    return $field_list[$bundle_id] ?? NULL;
   }
 
 }
--- a/tripal/src/Services/TripalEntityLookup.php
+++ b/tripal/src/Services/TripalEntityLookup.php
@@ -130,11 +130,19 @@ class TripalEntityLookup {
   }
 
   /**
-   * Retrieve the default Tripal bundle for a given base
-   * table. This table name is expected to match the bundle id.
+   * Retrieve the default Tripal bundle for a given base table.
+   *
+   * If there is more than one bundle for a given base
+   * table, this function will only return a value if there
+   * is a bundle with the same name as the base table.
+   * If all bundle names are different, then we return NULL.
+   * For example with the "feature" table, there is no
+   * generic "feature" bundle, so return NULL in this case.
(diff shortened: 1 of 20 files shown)


laceysanderson (commented):
Sorry for the delay on this 🙈 

I have a couple of comments/questions:
1. I like that you are setting the foreign key table in the field settings now! That is a great approach and means we need to use the fallback checking all entities less often. ❤️ 
2. For the fallback, I see you are only grabbing a single backup entity type, whereas we had discussed looping through all content types where that is a base table. I still think what we discussed is a better approach 🤔 but am wondering why you decided to just focus on one?
3. I see that you deleted a migration 😬 Can you explain why you did that? I'm worried about people who did apply that original migration 🤔  although I doubt there are that many...

dsenalik:
For item 2: Thanks @laceysanderson for calling me out on that function, what it really did was just confirm that a bundle existed with the same name as the base table, and as you noted that was not what we planned so was not correct.

Here is my debugging code for the correct way to retrieve all the bundles with a given base table
```
// Should retrieve bundle = 'array_design' given base table 'arraydesign'
$base_table = 'arraydesign';

$bundles = \Drupal::entityTypeManager()
  ->getStorage('tripal_entity_type')
  ->getQuery()
  ->condition('third_party_settings.tripal.chado_base_table', $base_table)
  ->execute();
var_dump($bundles);
```
and this returns
```
array(1) {
  ["array_design"]=>
  string(12) "array_design"
}
```
Now for the `feature` table for example, no bundle name will match the base table name, so I will be implementing this as follows:
1. If nothing is returned, return null

2. If exactly one bundle is returned, return that bundle (for example base table = "arraydesign") even if bundle name ("array_design") does not match base table

  3. If more than one bundle is returned (for example base table = "analysis") {
  4. If one of the bundle names matches the base table name, then return that
  } else {
  5. Return null, because there is no "Default" bundle (for example base table="feature")
  There is no current Chado table which will cause us to get to this case, fortunately,
  }

dsenalik:
For Item # 3, actually I did not remove the migration, I effectively renumbered it, with some additions. That is not very obvious looking at the github changes, but here is the diff. Since the properties may have been added with 10403, or may be added for the first time with 10405, it now handles both cases.
But the short version is that if you never ran 10403 then 10405 will do everything needed.
I did add some better comments to explain this with commit https://github.com/tripal/tripal/pull/1865/commits/17fdcc657d75373d99eb9b7aace98a378e6575ec
```
4c5
< function tripal_chado_update_10403() {
---
> function tripal_chado_update_10405() {
32a34
>     $n_updated = 0;
48c50,56
<               if (!$field_exists) {
---
>               if ($field_exists) {
>                 $messenger->addMessage(t("Updating entity id in Drupal table \"@table_name\", field \"@field_name\"",
>                     ['@table_name' => $table_name, '@field_name' => $field_name]));
>                 $schema->changeField($table_name, $columns[$new_property], $columns[$new_property], $field_schema['columns'][$new_property]);
>                 $n_updated++;
>               }
>               else {
52d59
<                 $manager->updateFieldStorageDefinition($field_storage_definition);
57a65
>         $manager->updateFieldStorageDefinition($field_storage_definition);
60c68,69
<     $messenger->addMessage(t("Added @n_added entity id properties", ['@n_added' => $n_added]));
---
>     $messenger->addMessage(t("Added @n_added entity id properties, updated @n_updated entity id properties",
>         ['@n_added' => $n_added, '@n_updated' => $n_updated]));
```

dsenalik:
Commit https://github.com/tripal/tripal/pull/1865/commits/2004d64d46ca2a2f2144dc0b0846e2bdef689951 fixes the following unit test error
`Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for tripal.content_type.contact with the following errors: tripal.content_type.contact:third_party_settings.tripal.chado_base_table missing schema`
and includes a new test of `setThirdPartySetting()` + `getThirdPartySetting()`
This is because with the correct implementation of `getDefaultBundle()` we need to have the third party setting for `chado_base_table` set, but it was not set by default in the testing environment, so that had to be added.

laceysanderson:
Regarding 
> If more than one bundle is returned (for example base table = "analysis") {
>    a) If one of the bundle names matches the base table name, then return that
>    b) Return null, because there is no "Default" bundle (for example base table="feature") There is no current Chado table which will cause us to get to this case, fortunately.

I'm confused by what you mean when you say no current chado table will get to the case where there is no "Default" bundle. Wouldn't we hit this case for any field on a feature-based content type that does not have the foreign key table set?

As an alternative, maybe you should rename the function to indicate that it will return all possible bundles. Then in getEntityId() you loop through the array of possible bundles, checking each one until you find it?

laceysanderson:
> Commit [2004d64](https://github.com/tripal/tripal/commit/2004d64d46ca2a2f2144dc0b0846e2bdef689951) fixes the following unit test error `Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for tripal.content_type.contact with the following errors: tripal.content_type.contact:third_party_settings.tripal.chado_base_table missing schema` and includes a new test of `setThirdPartySetting()` + `getThirdPartySetting()` This is because with the correct implementation of `getDefaultBundle()` we need to have the third party setting for `chado_base_table` set, but it was not set by default in the testing environment, so that had to be added.

For this addition, maybe we should move this into the test trait createContentTypeFromConfig() function so that all tests have access to the new third party setting and better represent the live environment?

dsenalik:
Thanks for keeping me honest @laceysanderson 
  > As an alternative, maybe you should rename the function to indicate that it will return all possible bundles. Then in getEntityId() you loop through the array of possible bundles, checking each one until you find it?

Implemented this with commit https://github.com/tripal/tripal/pull/1865/commits/8aa263af9599d87ec25157b425757a13910a79c5

  > maybe we should move this into the test trait createContentTypeFromConfig() function so that all tests have access to the new third party setting and better represent the live environment?

Another great suggestion, implemented this with commit https://github.com/tripal/tripal/pull/1865/commits/dd7012ac531040a57a2308494e59365b76fb5c17

laceysanderson (approved):
✅ Code changes look perfect :-)
✅ Manual testing worked great!

image
image
image
image

Thanks for the work @dsenalik!!! This is ready to merge! 😍 🥕

PR #1869, opened 2024-05-04
Checking for new fields

### LINKED ISSUE #1793: Checking for new fields
## Issue Description
In Tripal v3 we had the option to add new fields to a content type.  A site admin would go to the content type page, manage the fields and click the button to "Check for new fields".  Then Tripal would run through all of the tables of Chado and add any missing fields to the corresponding content types.   This was useful in Tripal v3 because of these criteria:

1. If a user added a custom table they needed fields for the content type that was mapped to the custom table. In Tripal v3, content types were directly mapped to base tables.
2. Properties in Tripal are unique fields--1 per property type.  So, these can't be added during installation because there's no knowledge about how many properties there may be.  

In Tripal v4 adding new fields dynamically is a bit problematic because 
1.  Content types are not typed to Chado tables, instead fields are.
2.  The content types and fields are now created via YAML files. Theres is no function that looks through Chado to find new tables and new properties and then knows what content type to add them to.  
3.  In Tripal 3 we published content types and added fields, the fields then pulled the appropriate data form Chado. In Tripal v4 we publish fields.  So, some solution for adding new properties has to include publishing of the field records.

## Proposed Solution for Adding Property Fields
Add to the `tripal.tripalentitytype_collection.*` configuration files an entry (per content type) simlar to the following:

```yaml
    -   name: organism_properties
        abstract: true
        content_type: organism
        label: Properties
        type: chado_property_type_default
        description: "An organism property"
        cardinality: -1
        required: false
        storage_settings:
            storage_plugin_id: chado_storage
            storage_plugin_settings:
                base_table: organism
                prop_table: organismprop
        settings:
            termIdSpace: "schema"
            termAccession: "property"
        display:
            view:
                default:
                    region: content
                    label: above
                    weight: 10
            form:
                default:
                    region: content
                    weight: 10
```
In the definition note the new `abstract` element.  If it is true that means this is not a property to be added to the content type when those content types are imported (this field will be ignored), but, it represents the default settings for any field of type `chado_property_type_default`. It also signals that this type of field is appropriate for the given content type.  Then when it comes time to publish the content type, the TripalPublish service will know that any fields with an `abstract` of `true` could have new fields to add to the given type.

Second, we add to the `TripalFieldItemBase` class a static function named `getNewInstances()` function that is empty.  Fields that need to create new instances of fields can do so if they implement this function.   The function should return an array structure identical to the YAML of the `tripal.tripalentitytype_collection.*` configuration, and include the new fields that should be added.  

Prior to publishing, the TripalPublish service will call the `getNewInstances()` function for all entries in the YAML that have an `abstract` set to `true` and then automatically add the new fields.  The TripalPublish service can then (without changes to the publishing part) publish the entries for the new fields.

This will allow us to get rid of the "Check for new fields" button on the fields page of the content type. If a user adds new properties about an organism, for example, then they can re-publish the organism content type and the new properties and values will appear.   

### Branch Name

tv4g1-issue1793-checking_for_new_fields

spficklin:
#

## Description

This PR supports adding new fields to content types.   The support for this has been added to the `TripalFieldCollection` class which is used to provide service for adding fields to content types.  Currently, this service is used to import the fields associated to content types in the YML files when a module is installed.   This PR adds a `discover()` function to the service.  The `discover()` function gets called in the `TripalEntityUIController` when someone clicks the button "Check for new fields" on the field management page of a content type.  To find new field instances, each field can now implement it's own `discover()` method too.  It should return an array equivalent to the one we use in the YML file for adding a field to a content type.  The `TripalFieldCollection::discover()` function will call the discover function of the fields.

**What's missing:**
- There is no form for allowing a site admin to select which fields to import. This was something we spoke about and wanted to do so that if a site admin deletes a field it won't keep coming back if they don't want it to.  I felt we could add that later but just allowing users to be able to add fields for now was necessary.
- This PR is missing some functional test, so it will remain a draft until that's ready.  
- This PR only adds  a `discover()` function to the Chado properties field. Other fields that need to discover new instances of themselves will need to add a `discover()` method.

## Testing?

1. Checkout this branch
2. Create an _Oryza sativa_ organism page.
3. Create an analysis page for a genome assembly titled "MSU Rice Genome Annotation Project Release 7"
4. Download Chromosome1 of the rice GFF3 file for the rice genome here:  http://rice.uga.edu/pub/data/Eukaryotic_Projects/o_sativa/annotation_dbs/pseudomolecules/version_7.0/chr01.dir/Chr1.gff3
5. Import the GFF using the Tripal GFF3 importers
6. Publish the gene pages for rice.
7. Go to the page for managing fields for the "Gene" content type and clikc the button to "Check for new fields". You should see a new "Note" and "Score" field added.
8. Remove one or both of the new fields.  Then click the "Check for new fields" button. They should be readded.


spficklin:
I'm thinking now that I've created this PR that I should remove the adding of the fields in the `discover()` function and just lets it discover. It can still do the validation of the fields and organize the fields into states (e.g. error, ready, invalid, etc.)... just not add them.  This way we can easily add a form that lets the site admin select the fields before adding them.   Even though I don't want to include that functionality in this PR, I need to fix it so we can do that.

laceysanderson:
How about for now you split the TripalFieldCollection::discover into two functions: 1) discover, 2) add field list returned from discover. Then we can call both in the TripalEntityUIController::tripalCheckForFields() for this PR. 

Having them split apart will make it easier for another PR to design the UI where you can pick which discovered fields to add since then it will just move the addDiscoveredFields call into the UI submit.

That way we can get this functionality in sooner! Plus, the ability to choose is a new feature that we didn't have in Tripal 3 so I think it's fine to make that wait for a later PR :-)

spficklin:
@laceysanderson I went ahead and added a form for selecting fields that should get added. The more I thought about it, it wasn't too much work.   I just updated the PR so that now when you click the "Check for new fields" button you get a form that provides a list of checkboxes for each new field. You can turn off fields you don't want to add.   

image


Also, I add added a `discover()` function to the `chado_organism_type_default` field. I got to thinking, if a user accidentally deletes a field that is added during install, then we need a new way to get that field back if they want it.  It works! You can now delete the organism field and you get it back if you check for new fields.  This means that we should probably add `discover()` functions to all our fields.  And this may mean we don't need those YML files for install.

spficklin:
Okay. This PR is ready for testing.  It now has functional tests as well.   I had to create a new test module to provide a field that could be used  in the testing functions.  @laceysanderson, @dsenalik and anyone else following this PR, this is ready for your review.

tripal/src/Form/TripalContentFieldsForm.php
```diff
@@ -0,0 +1,201 @@
...
+   * @see \Drupal\Core\Form\FormInterface::buildForm()
+   */
+  public function buildForm(array $form, FormStateInterface $form_state,
+      TripalEntityType $tripal_entity_type = NULL) {
+
+    /** @var \Drupal\tripal\Services\TripalFieldCollection $tripal_fields **/
+    $tripal_field_collection = \Drupal::service('tripal.tripalfield_collection');
+    $fields = $tripal_field_collection->discover($tripal_entity_type);
+
+    // Save the tripal entit type for use in the form submit.
```
dsenalik:
```suggestion
    // Save the tripal entity type for use in the form submit.
```

tripal/src/Form/TripalContentFieldsForm.php
```diff
@@ -0,0 +1,201 @@
...
+      '#title' => 'Instructions',
+    ];
+
+    $form['instructions']['instr_text'] = [
+      '#type' => 'markup',
+      '#markup' => t('The following tabs provide the list of fields that were '
+          . 'discovered. Fields that are new can be added to the content type '
+          . 'by selecting them and clicking the "Add Fields" button. Fields '
+          . 'that were discovered but already added are shown but cannot be '
+          . 'added again. For module developers, fields that do no pass '
```
dsenalik:
```suggestion
          . 'added again. For module developers, fields that do not pass '
```

tripal/src/TripalDBX/TripalDbxSchema.php
```diff
@@ -690,6 +690,38 @@ public function primaryKeyExists(
...
+   * @param string $right_table
+   */
+  public function foreignKeyExists(string $left_table, string $right_table) {
+    $left_def = $this->getForeignKeyDef($left_table, $right_table);
+    if ($left_def) {
+      return TRUE;
+    }
+    return FALSE;
+  }
+
```
dsenalik:
```suggestion
  /**
   * If a FK exists between two tables, return the left and right columns.
   *
   * @param string $left_table
   *   The name of the table the foreign key resides in. E.g. 'feature' for
   *   the feature.type_id => cvterm.cvterm_id foreign key.
   * @param string $right_table
   *   The name of the table the foreign key refers to. For the example
   *   above it would be cvterm.
   * @return array
   *    The foreign key definition
   */
  public function getForeignKeyDef(string $left_table, string $right_table) {
    $left_def = $this->getTableDef($left_table, ['format' => 'Drupal']);
    if (!array_key_exists($right_table, $left_def['foreign keys'])) {
      return NULL;
    }
    return $left_def['foreign keys'][$right_table];
  }

  /**
   * Indicates whether a FK exists between two tables.
   *
   * @param string $left_table
   *   The name of the table the foreign key resides in. E.g. 'feature' for
   *   the feature.type_id => cvterm.cvterm_id foreign key.
   * @param string $right_table
   *   The name of the table the foreign key refers to. For the example
   *   above it would be cvterm.
   * @return bool
   *   TRUE if a foreign key exists, FALSE if not.
   */
  public function foreignKeyExists(string $left_table, string $right_table) {
    $left_def = $this->getForeignKeyDef($left_table, $right_table);
    if ($left_def) {
      return TRUE;
    }
    return FALSE;
  }

```
Let me know if this suggestion is correct!!! ~~Does the first function return an array or something else?~~ Edit: I confirmed that it is an array
2024-05-22_fkdef

tripal/tests/modules/tripal_test/tripal_test.info.yml
```diff
@@ -0,0 +1,6 @@
+type: module
+name: Tripal Test Module
+core_version_requirement: ^8 || ^9 || ^10
```
dsenalik: We don't support 8 or 9 any more, should this be changed?

tripal_layout/tripal_layout.routing.yml
```diff
@@ -1,4 +1,4 @@
-# Adds a +Check for new fields link on the 'Tripal Content Types' page.
+# Adds a +Apply Defeault Layout link on the 'Tripal Content Field Display' page.
```
dsenalik:
```suggestion
# Adds a +Apply Default Layout link on the 'Tripal Content Field Display' page.
```

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
```diff
@@ -54,14 +54,8 @@ public static function tripalTypes($field_definition) {
...
     $entity_type_id = $field_definition->getTargetEntityTypeId();
     $settings = $field_definition->getSetting('storage_plugin_settings');
-    $base_table = $settings['base_table'];
-    $prop_table = $settings['prop_table'];
-
-    // If we don't have a base table then we're not ready to specify the
-    // properties for this field.
-    if (!$base_table) {
-      return;
-    }
```
dsenalik: You took this out. I will continue tomorrow, but I am suspicious of this...
Edit: I confirm that this should not be removed, if I add a new property field to gene through the GUI I get
`TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in array_keys() (line 67 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php).`

dsenalik (dismissed):
As you may see I found a few typos.
I think the comments for getForeignKeyDef() and foreignKeyExists() were not complete, but check if my suggestion is correct!
I did the described functional test
(I needed to increase upload limit size in `/usr/local/etc/php/php.ini` to load a file that big)
I only had one new field suggested, `Note` and this matches what I see in the `featureprop` table
```
sitedb=> select distinct F.type_id, T.name from featureprop F left join cvterm T on F.type_id=T.cvterm_id;
 type_id | name 
---------+------
    3179 | Note
(1 row)
```
But I could add and delete the field as expected

I note that the new field gets a weight of `10` - so just thinking out loud here, in Tripal 3 the new fields would show up at the end, I wonder if a weight of `99` would be more appropriate? 

And the suspicion I had is confirmed, if you try to add a new property field through the GUI you get
`TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in array_keys() (line 67 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php).`

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
```diff
@@ -54,14 +54,8 @@ public static function tripalTypes($field_definition) {
...
+    $base_table = array_key_exists('base_table', $settings) ? $settings['base_table'] : NULL;
+    $prop_table = array_key_exists('prop_table', $settings) ? $settings['prop_table'] : NULL;
 
     // Get the base table columns needed for this field.
```
dsenalik:
```suggestion
    // If we don't have a base table then we're not ready to specify the
    // properties for this field.
    if (!$base_table) {
      return;
    }

    // Get the base table columns needed for this field.
```
so this needs to go back in here

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
```diff
@@ -232,4 +226,84 @@ public function isCompatible(TripalEntityType $entity_type) : bool {
...
+    $results = $query->distinct()->execute()->fetchAll();
+
+    // Create a field entry for each property type.
+    foreach ($results as $recprop) {
+      $field_list[] = [
+        'name' => self::generateFieldName($bundle, $recprop->cvterm_name),
+        'content_type' => $bundle->getID(),
+        'label' => ucwords($recprop->cvterm_name),
+        'type' => self::$id,
+        'description' => 'A record property with the following definition: ' . $recprop->definition,
```
dsenalik: Here is one last comment for now. I tested importing a different GFF file and I get all of the properties, but in every case there is no definition following this text `A record property with the following definition: `

2024-05-22_dcar

Well that is because there is no definition. I wonder if the fallback should be to use the term when there is no definition? Or always include the namespace and term anyway?

tripal/src/Form/TripalContentFieldsForm.php
```diff
@@ -0,0 +1,201 @@
...
+    $messenger = \Drupal::messenger();
+
+    // Get submitted form elements.
+    $tripal_entity_type = $form_state->getValue('tripal_entity_type');
+    $fields = $form_state->getValue('discovered_fields');
+    $add_fields = $form_state->getValue('new_fields_choice');
+
+    /** @var \Drupal\tripal\Services\TripalFieldCollection $tripal_fields **/
+    $tripal_field_collection = \Drupal::service('tripal.tripalfield_collection');
+    $fields = $tripal_field_collection->discover($tripal_entity_type);
```
laceysanderson: This is the second time you defined the `$fields` variable in this function. I'm assuming we can remove the one on line 167? In which case we can also likely remove that form element from the build() as well.

tripal/tests/modules/tripal_test/tripal_test.info.yml
laceysanderson: Yes, it should
```suggestion
core_version_requirement: ^10
```

tripal/tests/src/Functional/Services/TripalFieldCollectionTest.php
```diff
@@ -15,6 +15,14 @@
  */
 class TripalFieldCollectionTest extends TripalTestBrowserBase {
 
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'tripal_test',
```
laceysanderson, line 23: Doesn't it also need to include tripal?
```suggestion
    'tripal',
    'tripal_test',
```

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
laceysanderson: Yah, I agree this description needs help. Maybe the pattern should be:
- if the definition is empty: "Chado Properties tagged with the term $term_name ($term_accession)."
- if the definition is available: "Chado Properties tagged with the term $term_name ($term_accession) which is defined as '$term_definition'."

tripal/src/Services/TripalFieldCollection.php
```diff
@@ -197,6 +198,63 @@ public function setFieldDefDefaults(array $field_def = []) : array {
     return $new_defs;
   }
 
+  /**
+   * Discovers new fields for a given entity type.
+   *
+   * Fields can be specified in one of 2 ways.  Using the installation method
+   * where they are specified by a YML file created by the module developer or
+   * by fields that have a discover() function implemented. This function
+   * supports adding new fields to the collection using the discover apprach.
```
laceysanderson:
```suggestion
   * supports adding new fields to the collection using the discover approach.
```

tripal/src/Services/TripalFieldCollection.php
```diff
@@ -197,6 +198,63 @@ public function setFieldDefDefaults(array $field_def = []) : array {
...
+
+    $all_field_defs = $field_type_manager->getDefinitions();
+    $entity_field_defs = $entity_field_manager->getFieldDefinitions('tripal_entity', $bundle_name);
+    foreach ($all_field_defs as $field_id => $field_def) {
+      $field_class = $field_def['class'];
+      if (is_subclass_of($field_class, 'Drupal\tripal\TripalField\TripalFieldItemBase')) {
+        $discovered = $field_class::discover($tripal_entity_type, $field_id, $all_field_defs);
+        foreach ($discovered as $discovered_field) {
+
+          // If the doscovered field already exists then mark it as existing.
```
laceysanderson:
```suggestion
          // If the discovered field already exists then mark it as existing.
```

tripal/src/TripalField/Interfaces/TripalFieldItemInterface.php
```diff
@@ -120,4 +122,32 @@ public function tripalSave($field_item, $field_name, $prop_types, $prop_values,
...
   public function tripalClear($field_item, $field_name, $prop_types, $prop_values, $entity);
+
+
+  /**
+   * Finds new instances of this field for a given content type.
+   *
+   * Fields are added to Tripal using the tripal.tripalfield_collection.*
+   * configuration file during installation of a module.  In some cases,
+   * however, not all possible instances of a field can be added to a content
+   * type at installation of the module.  This function can be called to disover
```
laceysanderson:
```suggestion
   * type at installation of the module.  This function can be called to discover
```

laceysanderson (changes requested):
Results of a code review. Most of it looks sound but it would be good to improve the description provided for property fields specifically. Otherwise it's mostly an unused variable and spelling fixes.

I like the addition of the foreign key specific functions to Tripal DBX Schema and the tests look good 😍 Also I just love this approach of the fields deciding when they should be discovered!

Perhaps I'll let you address the outstanding comments before doing a manual review of the functionality?

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
spficklin: I'm not sure why I took that out.  I recall I needed too but I can't reproduce why. Putting it back in fixes the error you mention but doesn't cause a problem elsewhere.  So, back it goes.

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
spficklin: Fixed in most recent commit.

tripal/src/Form/TripalContentFieldsForm.php
spficklin: Oops, yeah that second one is not needed. Good catch. I've removed it.

spficklin:
Thanks @dsenalik and @laceysanderson for the thorough review.  I've merged in all of your suggested changes and as a quick overview.  
1.  I restored the code that checks if the base table is set for the property field. I'm not sure why I took that out.
2.  I fixed the text for a field such that if there is on term "definition" it doen't say anything about the description. It just prints the term and accession.
3. I took out that extra $fields variable that Lacey noted.

As long as my changes didn't goof up the testing, I think I've addressed any comments you both raised.

tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
```diff
@@ -232,4 +232,84 @@ public function isCompatible(TripalEntityType $entity_type) : bool {
...
+    $results = $query->distinct()->execute()->fetchAll();
+
+    // Create a field entry for each property type.
+    foreach ($results as $recprop) {
+      $field_list[] = [
+        'name' => self::generateFieldName($bundle, $recprop->cvterm_name),
+        'content_type' => $bundle->getID(),
+        'label' => ucwords($recprop->cvterm_name),
+        'type' => self::$id,
+        'description' => $recprop->definition ? 'A record property with the following definition: ' . $recprop->definition : '',
```
dsenalik, line 280: I think that both Lacey and I want the term included. How about this instead?
```suggestion
      $description = 'A record property with the following term: "' 
          . $recprop->cv_name . ':' . $recprop->cvterm_name . '"';
      if ($recprop->definition) {
        $description .= ' and definition: "' . $recprop->definition .'"';
      }
      $field_list[] = [
        'name' => self::generateFieldName($bundle, $recprop->cvterm_name),
        'content_type' => $bundle->getID(),
        'label' => ucwords($recprop->cvterm_name),
        'type' => self::$id,
        'description' => $description,
```
This will display as
2024-05-28_termanddef

dsenalik: I am now realizing that the term name is in here twice. So maybe this is not quite right. If you prefer ignoring this suggestion it is fine with me.

laceysanderson: Ok, lets ignore this for now then and we can think about this on a per field basis as we go through and work on these discover functions.

laceysanderson (approved):
I did a manual test and it works perfect. As all comments have been addressed, this is ready to merge! Thank you @spficklin!!! 🎉

PR #1871, opened 2024-05-05
URL and name field filters default to plain text

### LINKED ISSUE #1864: URI fields should be plain text
### Task Description

I just noticed the URI field on the Protocol content type has a formatting toolbar now that PR #1862 is merged.
2024-04-28_uri
It really shouldn't 🙈

And the protocol title! I entered plain text in the name field, but the title is displaying like this
2024-04-28_protocol

The tasks are
 - [ ] To set all of the URI fields to "Plain text", and maybe others?
 - [ ] And for the title problem, I wonder if we could allow embedded markup here, otherwise we need "Name" to be plain text too, and any other field that might be used in a title.
 
OR
do we set the default everywhere to plain text, and just set HTML for fields we know will be appropriate?

### Branch Name

tv4g1-issue1864-some-fields-to-plain-text

### PR #1871

dsenalik:
# Bug Fix

## Description
This changes the default filter for `text` (i.e. unlimited length) fields from `basic_html` (the default) to `plain_text` for those fields that should not have formatting. Specifically, any URI field should not allow formatting, and fields likely to be used in a content type title, such as a `name` or `uniquename` field, should also default to plain_text.

Other minor changes:
  * There was a duplicated field for the "Protocol" content type ‼️, so I removed the "extra" one. https://github.com/tripal/tripal/pull/1871/commits/951958cc101df34f933f053d3a51103b4ac12eb4
  * The label in the schema for the edit form was "View" so I fixed that. https://github.com/tripal/tripal/pull/1871/commits/9be2006ca5e7f7dc0a0f1d45011903b9e480f6aa
  * @laceysanderson pointed out that there was also a duplicated field for the Array Batch Identifier, fixed with commit https://github.com/tripal/tripal/pull/1871/commits/f149c96433349f7126123703223997ba4782ce04

## Testing?
 * Build a new docker on this branch.
 * Create a protocol.
The uri field should not show a formatting toolbar, and once saved, the title should no longer have HTML paragraph tags.
 * Test array design, assay, biomaterial, and study to make sure titles do not show paragraph tags
 * Test analysis, genome assembly, and genome annotation to make sure URI fields do not have a formatting toolbar.

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -340,6 +340,14 @@ tripal.tripalfield_collection.*:
                            type: integer
                            label: 'Weight'
                            nullable: false
+                         settings:
+                           type: mapping
+                           label: 'Settings'
+                           mapping:
+                             filter_format:
+                             type: string
+                             label: 'Filter format for a filtered text field'
+                             nullable: true
 
 field.formatter.settings.*:
   type: mapping
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -340,6 +340,10 @@ tripal.tripalfield_collection.*:
                            type: integer
                            label: 'Weight'
                            nullable: false
+                         filter_format:
+                           type: string
+                           label: 'Filter format for a filtered text field'
+                           nullable: true
 
 field.formatter.settings.*:
   type: mapping
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -320,7 +320,7 @@ tripal.tripalfield_collection.*:
                            nullable: false
                    form:
                      type: sequence
-                     label: 'View'
+                     label: 'Edit'
                      sequence:
                        type: mapping
                        mapping:
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -340,10 +340,6 @@ tripal.tripalfield_collection.*:
                            type: integer
                            label: 'Weight'
                            nullable: false
-                         filter_format:
-                           type: string
-                           label: 'Filter format for a filtered text field'
-                           nullable: true
 
 field.formatter.settings.*:
   type: mapping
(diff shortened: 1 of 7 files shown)


laceysanderson (changes requested):
- ✅ Protocol
- ✅ Array Design
- ❌ Assay
- ✅ Biological Sample
- ✅ Study

I also noticed in Assay there are two Array Batch Identifiers and the second one takes priority. There is an error shown related to the TripalEntityLookup service. See screenshots:

| View | Edit |
|------|------|
| assayContentType | assayContentTypeEdit |

dsenalik:
I have hopefully resolved the three noted problems
 1. I removed the first of the two duplicated fields with commit https://github.com/tripal/tripal/pull/1871/commits/f149c96433349f7126123703223997ba4782ce04 - the second one is more logically placed and better matches Tripal 3 I think.
 2. The error happens if the array specifying required fields is cached by `getRequiredFields()`, then a new Type Collection is imported after caching. I adjusted the code to re-cache if the `bundle_id` is missing, which will cover this case. Commit https://github.com/tripal/tripal/pull/1871/commits/a63c800006a2ac581611ee5370f0ad81b4fed4cc

dsenalik:
3. I changed `assay_name` (used for Assay title) to plain text to resolve the last remaining title with &lt;p&gt;&lt;/p&gt; tags, commit https://github.com/tripal/tripal/pull/1871/commits/9cf6bfc86e174daebac7804dc47cdb0ee0e56817

laceysanderson (approved):
✅ The code updates look good
✅ I don't get the error anymore
✅ Assay name is no longer rich text

Thus I am marking it as approved :-) 

That said, I did notice we have one more duplicated field 🙈 Specifically Array Identifier (assay_arrayidentifier) and Unique Name (assay_unique_name) both indicate assay.arrayidentifier as their chado column. I only caught it because I filled in uniquename but not array identifier and it kept not saving my indicated uniquename.

dsenalik:
I wrote a program to scan all the yaml just to make sure there are no others - that seems to be the only case - thanks for finding it!

laceysanderson:
Great! This is ready to merge once the tests pass again from the recent merge!

Issue #1688, opened 2023-11-15
Support two copies of a double-hop field (low priority)


dsenalik:
### Task Description

You can have more than one instance of some fields on a content type page, for example, two different property fields. Right now, double-hop fields such as ChadoContact do not support having two copies. For example, if you put two copies of the contact field on a project, and manually add a different contact to each, the values get swapped. You can set them both the same, but if you then edit the second one there is no change, if you edit the first one, both change.
@laceysanderson says that we may be able to support this using a table alias specific to the field.

This is a low priority issue right now, because it is not common to have two identical copies of a field. In the case of the property field, each would have a different CV term. Contact has a fixed term.

### Branch Name

tv4g1-issue1688-support-two-copies-of-a-field

PR #1875, opened 2024-05-08
Chado Contact by Role field

laceysanderson:
# Tripal 4 Core Dev Task

## Description

It would be good to have a set of attribution-focused fields automatically added to project + analysis based content types to encourage standards. This addition will make it easier for Tripal sites to meet FAIR standards as it will guide them on how to provide proper attribution and licenses for their project/analysis pages which typically describe a dataset as a whole.

This PR provides a specialized contact field type that can use the new type_id column in the _contact linker tables that will be added by https://github.com/GMOD/Chado/pull/144. Specifically, the new field type acts the same way the property field works --allowing multiple contactByRole fields on the same content type where each handles a different *_contact.type_id. **This new field type uses the existing contact widget and formatter.**

Additionally, some changes were made to the existing contact widget to set better defaults for a type_id and rank if it is present.

**NOTE: As this field expects a type_id in the linker tables that is not present in Chado 1.31, the compatibility function has been updated to only allow this field to be added if the type_id update is present and provide context if it is not.**

## Testing?

1. Create a fresh docker on this branch (needed because terms are added).
2. Go to Tripal > Page Structure > Project > Manage Fields and attempt to add a Contact By Role Field.
     a. Enter a field name (i.e. "Funder")
     b. Choose "Chado Field"
     c. Choose "Chado Contact By Role"
     d. Click "Continue"
     e. Observe an error that says this field cannot be added.
3. Update you installation of chado to include the type_ids provided in https://github.com/GMOD/Chado/pull/144. This is easiest by using `drush sql:cli` and then copying in the following commands from the PR.
```sql
SET search_path TO chado;
/* https://github.com/GMOD/Chado/issues/140 */
/* Upgrade unique constraints with nullable columns if PostgreSQL version > 15 */
CREATE OR REPLACE PROCEDURE addUniqueLinkerConstraint(table_name varchar, constraint_name varchar, columns varchar[])
  LANGUAGE plpgsql
  AS $$
  DECLARE
    newer_than_15 boolean;
  BEGIN
    -- Determine the version of PostgreSQL
    SELECT CASE WHEN current_setting('server_version_num')::INT > 150000 THEN true ELSE false END AS supported INTO newer_than_15;

    -- IF the version is newer then we can use the new UNIQUE NULLS NOT DISTINCT
    -- which does not treat 2 records that are the same but include NULL as distict.
    IF newer_than_15 THEN
      EXECUTE format('ALTER TABLE %s ADD CONSTRAINT %s UNIQUE NULLS NOT DISTINCT (%s)', table_name, constraint_name, array_to_string(columns, ','));
    -- IF the version is <15 then we use the original UNIQUE style constraint
    ELSE
      EXECUTE format('ALTER TABLE %s ADD CONSTRAINT %s UNIQUE (%s)', table_name, constraint_name, array_to_string(columns, ','));
    END IF;
  END
$$;
/* Contact Linkers */
/* -- Feature */
ALTER TABLE feature_contact ADD COLUMN type_id bigint;
ALTER TABLE feature_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE feature_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
ALTER TABLE feature_contact DROP CONSTRAINT feature_contact_c1;
CALL addUniqueLinkerConstraint('feature_contact', 'feature_contact_c1', ARRAY['feature_id', 'contact_id', 'type_id', 'rank']);
/* -- Featuremap */
ALTER TABLE featuremap_contact ADD COLUMN type_id bigint;
ALTER TABLE featuremap_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE featuremap_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
ALTER TABLE featuremap_contact DROP CONSTRAINT featuremap_contact_c1;
CALL addUniqueLinkerConstraint('featuremap_contact', 'featuremap_contact_c1', ARRAY['featuremap_id', 'contact_id', 'type_id', 'rank']);
/* -- Library */
ALTER TABLE library_contact ADD COLUMN type_id bigint;
ALTER TABLE library_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE library_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
ALTER TABLE library_contact DROP CONSTRAINT library_contact_c1;
CALL addUniqueLinkerConstraint('library_contact', 'library_contact_c1', ARRAY['library_id', 'contact_id', 'type_id', 'rank']);
/* -- ND Experiment */
ALTER TABLE nd_experiment_contact ADD COLUMN type_id bigint;
ALTER TABLE nd_experiment_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE nd_experiment_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
CALL addUniqueLinkerConstraint('nd_experiment_contact', 'nd_experiment_contact_c1', ARRAY['nd_experiment_id', 'contact_id', 'type_id', 'rank']);
/* -- Project */
ALTER TABLE project_contact ADD COLUMN type_id bigint;
ALTER TABLE project_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE project_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
ALTER TABLE project_contact DROP CONSTRAINT project_contact_c1;
CALL addUniqueLinkerConstraint('project_contact', 'project_contact_c1', ARRAY['project_id', 'contact_id', 'type_id', 'rank']);
/* -- Pubauthor */
ALTER TABLE pubauthor_contact ADD COLUMN type_id bigint;
ALTER TABLE pubauthor_contact ADD COLUMN rank int DEFAULT 0;
ALTER TABLE pubauthor_contact ADD FOREIGN KEY (type_id) REFERENCES cvterm (cvterm_id) ON DELETE SET NULL;
ALTER TABLE pubauthor_contact DROP CONSTRAINT pubauthor_contact_c1;
CALL addUniqueLinkerConstraint('pubauthor_contact', 'pubauthor_contact_c1', ARRAY['pubauthor_id', 'contact_id', 'type_id', 'rank']);
```
4. Now go back and re-do step 2. This time you will not get the compatibility error and will be able to continue. Set the field to be unlimited and set the term to Funder (EFO:0001736). The field should add without error.
5. Create an additional contact by role field: name "Data Collector", term "AGRO:00000379".
6. Go to Tripal > Content > Add > Contact and create multiple contact pages. Set the contact type to "Organization", etc or leave blank.
7. Go to Tripal > Content > Add > Project and create a generic project. With the two fields above added you should see something like the following screenshot:
Firefox_Screenshot_2024-05-08T17-04-00 046Z
8. Confirm that you can enter multiple items each and that you can enter the same item in both. Also confirm that you can edit, remove and rearrange elements.
image

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
@@ -150,6 +150,7 @@ class ChadoContactByRoleTypeDefault extends ChadoFieldItemBase {
       'drupal_store' => TRUE,
       'namespace' => self::$chadostorage_namespace,
       'function' => self::$drupal_entity_callback,
+      'ftable' => self::$object_table,
       'fkey' => self::$object_id,
     ]);
 
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
@@ -280,12 +280,12 @@ class ChadoContactByRoleTypeDefault extends ChadoFieldItemBase {
 
     // Ensure that the linker table has a type_id.
     $has_type_id = FALSE;
+    $schemaObj = \Drupal::service('tripal_chado.database')->schema();
     foreach ($linker_tables as $item) {
 
       [$table_name, $contact_fkey] = $item;
 
       // Check there is a type_id field.
-      $schemaObj = \Drupal::service('tripal_chado.database')->schema();
       $has_type_id = $schemaObj->fieldExists($table_name, 'type_id');
 
       if (!$has_type_id) {
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
@@ -1,0 +1,277 @@
+<?php
+
+namespace Drupal\tripal_chado\Plugin\Field\FieldType;
+
+use Drupal\tripal_chado\TripalField\ChadoFieldItemBase;
+use Drupal\tripal_chado\TripalStorage\ChadoIntStoragePropertyType;
+use Drupal\tripal_chado\TripalStorage\ChadoVarCharStoragePropertyType;
+use Drupal\tripal\Entity\TripalEntityType;
+
+/**
+ * Plugin implementation of default Tripal contact field type.
+ *
+ * @FieldType(
+ *   id = "chado_contact_by_role_type_default",
+ *   category = "tripal_chado",
+ *   label = @Translation("Chado Contact By Role"),
+ *   description = @Translation("Supports linking contacts fullfilling a certain role to content types."),
+ *   default_widget = "chado_contact_widget_default",
+ *   default_formatter = "chado_contact_formatter_default",
+ * )
+ */
+class ChadoContactByRoleTypeDefault extends ChadoFieldItemBase {
+
+  public static $id = 'chado_contact_by_role_type_default';
+  protected static $object_table = 'contact';
+  protected static $object_id = 'contact_id';
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function mainPropertyName() {
+    // Overrides the default of 'value'
+    return 'contact_name';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultFieldSettings() {
+    $settings = parent::defaultFieldSettings();
+    // If this field needs to set a fixed value, set this to TRUE.
+    // It indicates to the publishing step to include this field.
+    // If not set, then the publishing step may not be able to find matches
+    // for this field based on the fixed value.
+    $settings['fixed_value'] = FALSE;
+    return $settings;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultStorageSettings() {
+    $storage_settings = parent::defaultStorageSettings();
+    $storage_settings['storage_plugin_settings']['base_table'] = '';
+    $storage_settings['storage_plugin_settings']['linking_method'] = '';
+    $storage_settings['storage_plugin_settings']['linker_table'] = '';
+    $storage_settings['storage_plugin_settings']['linker_fkey_column'] = '';
+    $storage_settings['storage_plugin_settings']['object_table'] = self::$object_table;
+    return $storage_settings;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function tripalTypes($field_definition) {
+
+    $entity_type_id = $field_definition->getTargetEntityTypeId();
+    $storage_settings = $field_definition->getSetting('storage_plugin_settings');
+    $base_table = $storage_settings['base_table'];
+
+    if ($base_table === NULL) {
+      return;
+    }
+
+    $terms = [
+      'record_id' => 'SIO:000729',
+    ];
+    $max_lengths = [];
+
+    $schemaObj = \Drupal::service('tripal_chado.database')->schema();
+    $mappingObj = \Drupal::entityTypeManager()->getStorage('chado_term_mapping')->load('core_mapping');
+
+    // Get the column, term and any other schema-related details.
+    // A) BASE TABLE
+    $base_schema_def = $schemaObj->getTableDef($base_table, ['format' => 'Drupal']);
+    //    - primary key
+    $base_pkey_col = $base_schema_def['primary key'];
+    $terms['base_pkey'] = $terms['record_id'];
+    // B) OBJECT TABLE (i.e. contact)
+    $object_table = self::$object_table;
+    $object_schema_def = $schemaObj->getTableDef($object_table, ['format' => 'Drupal']);
+    //    - primary key
+    $object_pkey_col = $object_schema_def['primary key'];
+    $terms['object_pkey'] = $terms['record_id'];
+    //    - name
+    $terms['name'] = $mappingObj->getColumnTermId($object_table, 'name');
+    $max_lengths['name'] = $object_schema_def['fields']['name']['size'];
+    //    - description
+    $terms['description'] = $mappingObj->getColumnTermId($object_table, 'description');
+    $max_lengths['description'] = $object_schema_def['fields']['description']['size'];
+    //    - contact type
+    $cvterm_schema_def = $schemaObj->getTableDef('cvterm', ['format' => 'Drupal']);
+    $terms['contact_type'] = $mappingObj->getColumnTermId('cvterm', 'name');
+    $max_lengths['contact_type'] = $cvterm_schema_def['fields']['name']['size'];
+    // C) LINKING TABLE.
+    [$linker_table, $linker_fkey_column] = self::get_linker_table_and_column($storage_settings, $base_table, $object_pkey_col);
+    $linker_schema_def = $schemaObj->getTableDef($linker_table, ['format' => 'Drupal']);
+    //    - primary key
+    $linker_pkey_col = $linker_schema_def['primary key'];
+    $terms['linker_pkey'] = $terms['record_id'];
+    //    - left table foreign key
+    $linker_left_col = array_keys($linker_schema_def['foreign keys'][$base_table]['columns'])[0];
+    $terms['linker_left'] = $mappingObj->getColumnTermId($linker_table, $linker_left_col);
+    //    - right table foreign key
+    $terms['linker_right'] = $mappingObj->getColumnTermId($linker_table, $linker_fkey_column);
+    //    - linking type
+    $terms['linker_type_id'] = $mappingObj->getColumnTermId($linker_table, 'type_id');
+    if (empty($terms['linker_type_id'])) {
+      $terms['linker_type_id'] = $terms['contact_type'];
+    }
+    //    - rank
+    $terms['linker_rank'] = $mappingObj->getColumnTermId($linker_table, 'rank');
+    if (empty($terms['linker_rank'])) {
+      $terms['linker_rank'] = 'OBCS:0000117';
+    }
+
+    // We need to create a table alias for our linker table in order to ensure
+    // contact links with other roles are not combined.
+    // The type used when creating the linking record will be the same as the
+    // type set for the field. As such, we grab that here and use it in our
+    // table alias.
+    $field_settings = $field_definition->getSettings();
+    $term = $field_settings['termIdSpace'] . ':' . $field_settings['termAccession'];
+    $table_alias = $linker_table . '_' . preg_replace( '/[^a-z0-9]+/', '', strtolower( $term ) );
+
+    // FINALLY, THE PROPERTIES!
+    $properties = [];
+
+    // Define the base table record id.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $terms['base_pkey'], [
+      'action' => 'store_id',
+      'drupal_store' => TRUE,
+      'path' => $base_table . '.' . $base_pkey_col,
+    ]);
+
+    // This property will store the Drupal entity ID of the linked chado
+    // record, if one exists.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'entity_id', self::$drupal_entity_term, [
+      'action' => 'function',
+      'drupal_store' => TRUE,
+      'namespace' => self::$chadostorage_namespace,
+      'function' => self::$drupal_entity_callback,
+      'fkey' => self::$object_id,
+    ]);
+
+    // Define the linker table that links the base table to the object table.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id', $terms['linker_pkey'], [
+      'action' => 'store_pkey',
+      'drupal_store' => TRUE,
+      'path' => $table_alias . '.' . $linker_pkey_col,
+      'table_alias_mapping' => [$table_alias => $linker_table],
+    ]);
+
+    // Define the link between the base table and the linker table.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'link', $terms['linker_left'], [
+      'action' => 'store_link',
+      'drupal_store' => TRUE,
+      'path' => $base_table . '.' . $base_pkey_col . '>' . $table_alias . '.' . $linker_left_col,
+      'table_alias_mapping' => [$table_alias => $linker_table],
+    ]);
+
+    // Define the link between the linker table and the object table.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, $linker_fkey_column, $terms['linker_right'], [
+      'action' => 'store',
+      'drupal_store' => TRUE,
+      'path' => $table_alias . '.' . $linker_fkey_column,
+      'table_alias_mapping' => [$table_alias => $linker_table],
+      'delete_if_empty' => TRUE,
+      'empty_value' => 0,
+    ]);
+
+    // Linker type_id
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_type_id', $terms['linker_type_id'], [
+      'action' => 'store',
+      'drupal_store' => FALSE,
+      'path' => $table_alias . '.type_id',
+      'table_alias_mapping' => [$table_alias => $linker_table],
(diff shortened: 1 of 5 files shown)


laceysanderson:
I'm marking this as ready for review since it does check if the type_id is present and if not, provides a helpful error (as the field is not compatible). We can't change the YAML to add this field by default since not all chado instances will have the type_ids in the linker tables yet.

tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // Ensure that the linker table has a type_id.
+    $has_type_id = FALSE;
+    foreach ($linker_tables as $item) {
+
+      [$table_name, $contact_fkey] = $item;
+
+      // Check there is a type_id field.
+      $schemaObj = \Drupal::service('tripal_chado.database')->schema();
```
dsenalik:
```suggestion
    $schemaObj = \Drupal::service('tripal_chado.database')->schema();
    foreach ($linker_tables as $item) {

      [$table_name, $contact_fkey] = $item;

      // Check there is a type_id field.
```
This is absolutely not likely to be significant, but for performance we would only need to retrieve $schemaObj once

dsenalik (approved):
I have one tiny fussy suggestion, but manual testing worked exactly as it should, so I approve.
However I am wondering about longer term, if chado linker tables do all get type_id and rank in the future, I would not want two versions of other fields.
Is there any basic reason why we could not alter the existing contact field to include the functionality of this new field?
I tried to add a second contact field with a different term, but it leaks through to the first one (as you probably know 🙄 issue #1688)

laceysanderson:
> However I am wondering about longer term, if chado linker tables do all get type_id and rank in the future, I would not want two versions of other fields.
Is there any basic reason why we could not alter the existing contact field to include the functionality of this new field?

@dsenalik I was really on the fence about this. I ended up with two fields because:
1. It was starting to feel really messy logic-wise, having both in the same field. Instead it felt more maintainable to have two simple fields. In hindsight, perhaps instead of the two fields being completely separate, I could have inherited from the original one --that would be even more maintainable 🤔 
2. It also just felt more intuitive to me from the UI side of things rather than messing with more configuration.
3. I originally thought we may need a different widget/formatter (thankfully I was able to avoid that!).
4. I was also a bit worried because the schema changes needed for this field are not yet reviewed in Chado and are not available to users by default. It felt easier to indicate the entire new field type was not supported rather then greying out configuration options.

All that said, it was also easier to keep it separate so thanks for keeping me honest and getting me to explain my reasons. I'm happy to make changes :-) 

Anyway, what are your thoughts? Would you like me to
A) use inheritance to reduce the amount of duplicated code there currently is.
B) merge the two field types and add additional configuration to specify whether you want it filtered by role or not.

dsenalik:
Thanks for the detailed explanation! I also wanted to investigate what you have done as it may relate to issue #1688 since that may be a simple fix for all fields.
I am conflicted on what I want myself. Ideally, yes one field to rule them all. If Chado gets those changes that would make it nicely backward compatible. But the fields are getting complicated already (I have been working on and off on the field documentation updates, the linking fields are already too complicated for a beginner lesson 🙄)
I also like the inheritance idea if we want to keep two fields. I don't know how hard that is to implement.
I should look at the differences better, so right now I have no clear opinion on which way to go 😐

laceysanderson:
> I also wanted to investigate what you have done as it may relate to issue https://github.com/tripal/tripal/issues/1688 since that may be a simple fix for all fields.

Regarding this, my change likely won't help the original contact field. 

Specifically, my field uses table alias' to ensure that each field has it's own join and own copies of the values. This works because they are non-overlapping sets: the table is effectively partitioned by type_id.

The main contact field cannot partition in this way as it needs to handle all contacts. Two general contact fields on the same content type effectively results in two fields touching the exact same records. We could use table alias' here to ensure each general contact field selects it's own copy of the records -this would suck performance-wise but would work... EXCEPT that field supports editing and allowing editing of the same set of values by two fields is understandably buggy as heck 🙈 

I think we would need a way to see that there are two general contact fields on the same content type and then switch it so only one field has editing and the rest are read only.

dsenalik:
Ah so my light bulb is showing a dim glow now. So the main field displays everything, your field displays only contacts of one specific type. I got this much, but the main field can't filter by cv term type from the linking table, which I wasn't considering, that's a separate cv term from the field's cv term.
If I wanted to display two different fields for analysis, one a "Genome Assembly" and a second field a "Genome Annotation" where both are analysis content types, it would be the same situation. The main analysis field would display everything. I want to look at your code to understand better soon...
edit: and separate from the contact table's cv term!

spficklin:
My review worked as expected!  I like the field in that you can set the type.   To jump in on the conversation about why not just update the existing field. I think it's okay to have mutiple fields for the same data type. I suspect our extension modules will do that too.  I agree with Lacey that more simple fields are better than trying to do too much in one field. They get really hard to debug and manage.  

Two things to think about:

1. This new field will be compatible with a future version of Chado (or a manually tweaked version), whereas the current field is compatible with the current version (and all future versions) of Chado.   
2.  We designed Tripal to come with 'default' Chado fields.  The current field is the default now and this new fields is not, but when we move to the upgraded version of Chado do we deprecate the old default and make this new one the default?

I'm wondering if we need to update our default fields and put the word "Default" in their labels?  That way someone will know that this new field provides additional functionality.   I think we need to iron out how Tripal will support Chado upgrades.  Do we still with major releases only or support any flyway migration change?  One thing we may want to consider is a `compatiblity` setting for our fields so that developers can specify if it's compatible with a lower version of Chado.   This new field would not be compatible with 1.3, but would with 1.4  

Just some thoughts.

spficklin:
All of the tests worked, but I noticed one minor detail.  When on the project page, the default contact field links the contacts to their contact page, but this new field does not, it just lists the name of the contact. Do we want this new field to link to the contact page?

There is not a corresponding way to add a contact to a project (and specify the contact type) on the contact page.  It seems we would need a project field that was contact aware to do this?  Are there plans for that?

Also, I noticed a publishing problem we need to deal with. Maybe we need to address it here.... not sure. But on the contact page you can specify a project. If you save it, the project shows up on the contact page.  However, if you go to the project page, the contact is not there.  This is a publishing problem.  We would have to publish the contact record for the project page.    It seems we need a publishing step on save of an entity, to publish new field values on linked entities.  It should be quick to run. I don't think it will slow the save down too much, but otherwise we get out of sync across our pages.

spficklin:
One more comment on this statement I made:

> I'm wondering if we need to update our default fields and put the word "Default" in their labels? That way someone will know that this new field provides additional functionality. I think we need to iron out how Tripal will support Chado upgrades.

Actually, the two fields won't do the same thing.  The current field lists all contacts and could be upgraded in Chado 1.4 to list the type.  This field only lists contacts of a given type.  Perhaps we should rename both fields?   Instead of "Chado Contact" and "Chado Contact by Role" what if we named them "Chado Contacts (any)" and "Chado Contacts (by role)"?  Then it's more clear they deal with the same data, but handle it in different ways.

laceysanderson:
> All of the tests worked, but I noticed one minor detail. When on the project page, the default contact field links the contacts to their contact page, but this new field does not, it just lists the name of the contact.

I believe this should now happen automatically since PR #1865 has been merged. Is this right @dsenalik or is there something specific I need to change?

> Actually, the two fields won't do the same thing. The current field lists all contacts and could be upgraded in Chado 1.4 to list the type. This field only lists contacts of a given type. Perhaps we should rename both fields? Instead of "Chado Contact" and "Chado Contact by Role" what if we named them "Chado Contacts (any)" and "Chado Contacts (by role)"? Then it's more clear they deal with the same data, but handle it in different ways.

Yes, I tried to make this clear in the field description but it looks like more improvement is needed. What about "Chado Contacts: All" and "Chado Contacts: Specific Role"? I'll also expand the description of both fields to make it more clear.

>  I think we need to iron out how Tripal will support Chado upgrades. Do we still with major releases only or support any flyway migration change? One thing we may want to consider is a compatiblity setting for our fields so that developers can specify if it's compatible with a lower version of Chado. This new field would not be compatible with 1.3, but would with 1.4

YES! I've started a discussion issue about how we handle Chado upgrade here: #1876. I personally like the idea of supporting the migrations directly using the point versions assigned to them.

I love the idea of adding a chado compatibility string to each field! I'll try to start a new issue discussing this -unless you want to do so first @spficklin?

laceysanderson:
> There is not a corresponding way to add a contact to a project (and specify the contact type) on the contact page. It seems we would need a project field that was contact aware to do this? Are there plans for that?

Yes, this is the first of many fields that we will want to take advantage of the new type in the linking tables. This is why @dsenalik originally brought up the concern with having two contact fields... because it will potentially duplicate all the linking fields 🙈

spficklin:
I'm good with your suggested changes and suggestions for new fields and issues

spficklin (approved):

dsenalik:
> > All of the tests worked, but I noticed one minor detail. When on the project page, the default contact field links the contacts to their contact page, but this new field does not, it just lists the name of the contact.
> 
> I believe this should now happen automatically since PR #1865 has been merged. Is this right @dsenalik or is there something specific I need to change?

Yup, it is not working, so I will check why not, and rebuild the docker just in case

2024-05-22_contact

Edit: Fixed with commit https://github.com/tripal/tripal/pull/1875/commits/56f2ee475978b183888632f670df913b4aa4f18c

tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
```diff
@@ -0,0 +1,302 @@
...
+    ]);
+
+    // This property will store the Drupal entity ID of the linked chado
+    // record, if one exists.
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'entity_id', self::$drupal_entity_term, [
+      'action' => 'function',
+      'drupal_store' => TRUE,
+      'namespace' => self::$chadostorage_namespace,
+      'function' => self::$drupal_entity_callback,
+      'fkey' => self::$object_id,
```
dsenalik:
```suggestion
      'function' => self::$drupal_entity_callback,
      'ftable' => self::$object_table,
      'fkey' => self::$object_id,
```
This one line is all that is missing to make the entity links work!

tripal_chado/src/Plugin/Field/FieldType/ChadoContactByRoleTypeDefault.php
```diff
@@ -0,0 +1,303 @@
...
+    ]);
+
+    // The type of contact
+    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'contact_type', $terms['contact_type'], $max_lengths['contact_type'], [
+      'action' => 'read_value',
+      'drupal_store' => FALSE,
+      'path' => $table_alias . '.' . $linker_fkey_column . '>' . $object_table . '.' . $object_pkey_col
+        . ';' . $object_table . '.type_id>cvterm.cvterm_id;name',
+      'table_alias_mapping' => [$table_alias => $linker_table],
+      'as' => 'contact_type',
```
dsenalik:
```suggestion
    $table_alias = $linker_table . '_' . preg_replace( '/[^a-z0-9]+/', '', strtolower( $term ) );
    $table_alias_mapping = [$table_alias => $linker_table];

    // FINALLY, THE PROPERTIES!
    $properties = [];

    // Define the base table record id.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $terms['base_pkey'], [
      'action' => 'store_id',
      'drupal_store' => TRUE,
      'path' => $base_table . '.' . $base_pkey_col,
    ]);

    // This property will store the Drupal entity ID of the linked chado
    // record, if one exists.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'entity_id', self::$drupal_entity_term, [
      'action' => 'function',
      'drupal_store' => TRUE,
      'namespace' => self::$chadostorage_namespace,
      'function' => self::$drupal_entity_callback,
      'ftable' => self::$object_table,
      'fkey' => self::$object_id,
    ]);

    // Define the linker table that links the base table to the object table.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id', $terms['linker_pkey'], [
      'action' => 'store_pkey',
      'drupal_store' => TRUE,
      'path' => $table_alias . '.' . $linker_pkey_col,
      'table_alias_mapping' => $table_alias_mapping,
    ]);

    // Define the link between the base table and the linker table.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'link', $terms['linker_left'], [
      'action' => 'store_link',
      'drupal_store' => TRUE,
      'path' => $base_table . '.' . $base_pkey_col . '>' . $table_alias . '.' . $linker_left_col,
      'table_alias_mapping' => $table_alias_mapping,
    ]);

    // Define the link between the linker table and the object table.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, $linker_fkey_column, $terms['linker_right'], [
      'action' => 'store',
      'drupal_store' => TRUE,
      'path' => $table_alias . '.' . $linker_fkey_column,
      'table_alias_mapping' => $table_alias_mapping,
      'delete_if_empty' => TRUE,
      'empty_value' => 0,
    ]);

    // Linker type_id
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_type_id', $terms['linker_type_id'], [
      'action' => 'store',
      'drupal_store' => FALSE,
      'path' => $table_alias . '.type_id',
      'table_alias_mapping' => $table_alias_mapping,
      'as' => 'linker_type_id',
    ]);

    // Linker Rank
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_rank', $terms['linker_rank'], [
      'action' => 'store',
      'drupal_store' => FALSE,
      'path' => $table_alias . '.rank',
      'table_alias_mapping' => $table_alias_mapping,
      'as' => 'linker_rank',
    ]);

    // The object table, the destination table of the linker table
    // The contact name
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'contact_name', $terms['name'], $max_lengths['name'], [
      'action' => 'read_value',
      'drupal_store' => FALSE,
      'path' => $table_alias . '.' . $linker_fkey_column . '>' . $object_table . '.' . $object_pkey_col . ';name',

      'table_alias_mapping' => $table_alias_mapping,
      'as' => 'contact_name',
    ]);

    // The contact description
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'contact_description', $terms['description'], $max_lengths['description'], [
      'action' => 'read_value',
      'drupal_store' => FALSE,
      'path' => $table_alias . '.' . $linker_fkey_column . '>' . $object_table . '.' . $object_pkey_col . ';description',
      'table_alias_mapping' => $table_alias_mapping,
      'as' => 'contact_description',
    ]);

    // The type of contact
    $properties[] = new ChadoVarCharStoragePropertyType($entity_type_id, self::$id, 'contact_type', $terms['contact_type'], $max_lengths['contact_type'], [
      'action' => 'read_value',
      'drupal_store' => FALSE,
      'path' => $table_alias . '.' . $linker_fkey_column . '>' . $object_table . '.' . $object_pkey_col
        . ';' . $object_table . '.type_id>cvterm.cvterm_id;name',
      'table_alias_mapping' => $table_alias_mapping,
      'as' => 'contact_type',
```
What do you think about this change? I feel it simplifies reading the code not having this same array "hardcoded" several times
I just added a new `$table_alias_mapping` variable

laceysanderson: I see :-) I'm actually about to make quite a big change here to inherit a lot of this from the original contact field to reduce code duplication and make it really clear what the difference between the two fields is :-) But I'll keep this change in mind while I do it!

laceysanderson:
Ok, update on this. I made the change suggested by @dsenalik  regarding the table mapping array and I make the label/desciption of the two fields more clear as suggested by @spficklin.

AddFieldListing

I did not however, change the contact by role field to inherit from the original contact field. I tried it multiple ways and it kept producing inconsistent and buggy behaviour so at this point I would advise against inheriting from existing fields. We could create a shared base abstract class if we felt that would make the field more readable but I found this did not improve the readability of these two fields.

As such, once tests pass I am going to merge this as the two small changes do not warrant a full new review test IMO. Note: I did do my own test on the changes.

PR #1878, opened 2024-05-20
Render HTML tags in page titles, e.g. organism scientifc name

dsenalik:
# New Feature

## Description
  - [x] ~~`TripalTokenParser()` should strip outer &lt;p&gt;&lt;/p&gt; tags - These are added if a field is configured to one of the HTML filters and you want to use that value in the title. Maybe we want to include italics in a gene name and show that in the title. The value will always be stored in the database table with wrapping &lt;p&gt;&lt;/p&gt; tags around it.~~ This is not needed now that XSS::filter is used
  - [x] Add test for the above change
  - [x] The entity list should support HTML tags. Once implemented, the organism content pages can appear like this
2024-05-20_tripalcontentlist

  - [x] Create a hook to preprocess the title field, and convert the string value into markup. This will allow any HTML tags to be rendered.
  - [x] Create a hook to preprocess the page title block, again convert any string value into markup.
  - [x] Use XSS::filter to sanitize everything 🪣🧽
  - [x] The default organism title template can now include italics `<em>[organism_genus] [organism_species]</em> [organism_infraspecific_type] <em>[organism_infraspecific_name]</em>`
  - [x] Create an admin form to allow configuring the allowed HTML tags
    An organism page should look like this now
2024-05-20_allworking

## Testing?
Because the organism default title template now has HTML tags for italics, testing is simple
 1. Make sure tests on this branch had no errors.
 1. Create a new organism
 1. Both versions of the title should show italics
 1. The entity listing should show the title with italics
 1. For good measure, create a different content type to make sure a "normal" title still works as before.
 1. Finally, create a publication with bold and italics in the title. See that it displays correctly.
 1. Go to Manage -> Tripal -> Tripal Entity Settings and remove all tags and save. Rebuild cache and confirm no bold or italics now displays.
 1. Test form validation. In that same form, enter something not a tag like `123` - only characters and spaces should be allowed. Put in too many spaces, they should reduce to a single space only between tags.
 
2024-05-20_pub

 I note that currently &lt;strong&gt; tags don't seem to render as bold in the page title block 🙈
 However this is a theme issue as the tags are there (Olivero 10.2.7-dev (default theme))
`<h1 class="title page-title"><strong>Tripal 4</strong> and other hazardous adventures <em>in silico</em></h1>` 
The entire &lt;h1&gt; element is already rendered as "bold", thus we don't see a difference.
2024-05-20_h1


laceysanderson (commented):
Each place you create a markup object (2X in .module and 1X in list builder) you should call [XSS::filter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Xss.php/function/Xss%3A%3Afilter/11.x) on the string first to ensure we are not introducing a security issue. Additionally, you can use this to ensure we only have tags we support.

For example,
```php
$sanitized_value = Drupal\Component\Utility\Xss::filter($value, ['em','strong','i']);
$variables['items'][$delta]['content']['#context']['value'] = Drupal\Core\Render\Markup::create($sanitized_value);
```

Manual testing shows that it works as expected but also points out that we may want to be more restrictive in the tags we allow. For example, you will see I was able to put a `<br />` tag in the title which has the same undesirable effect as `<p>` 😉 
image
image

dsenalik:
Thanks for the security fix @laceysanderson 
&lt;BR&gt; tag is filtered out
`<em>polA</em> encodes DNA polymerase I <strong>strong</strong> <i>italic</i> <u>underline</u> BR tag inside:[<br>]`
2024-05-28_xss_filter

Ideally I would have the array of allowed tags as a configurable variable rather than hardcoded, but do we have any configurable variables currently in Tripal? I can't seem to find any.
For my reference the code to set and look it up could be something like
```
$config_api = \Drupal::service('config.factory')->getEditable('variable_get_set.api');
$tripal_allowed_tags = ['em', 'i', 'strong', 'u'];
$config_api->set('tripal.allowed_tags', $tripal_allowed_tags);
$config_api->save();

$tripal_allowed_tags = $config_api->get('tripal.allowed_tags');
```

laceysanderson:
We do have some settings already!

- The defaults are set here: [tripal.settings.yml files.quota](https://github.com/tripal/tripal/blob/4.x/tripal/config/install/tripal.settings.yml#L12)
- The schema is here: [tripal.schema.yml files.quota](https://github.com/tripal/tripal/blob/4.x/tripal/config/schema/tripal.settings.schema.yml#L15-L16)
- Here is an example usage: [TripalAdminManageQuotaForm files.quota](https://github.com/tripal/tripal/blob/4.x/tripal/src/Form/TripalAdminManageQuotaForm.php#L39)

dsenalik:
There was already a form for settings! `./tripal/src/Form/TripalEntitySettingsForm.php` It was just not configured in any menu because it hadn't been used yet. So I used it!

tripal/src/Form/TripalEntitySettingsForm.php
```diff
@@ -21,35 +21,86 @@ class TripalEntitySettingsForm extends FormBase {
...
+   *   Form definition array.
    */
-  public function submitForm(array &$form, FormStateInterface $form_state) {
-    // Empty implementation of the abstract submit class.
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $settings = \Drupal::config('tripal.settings');
+
+    $form['tripal_entity_settings']['#markup'] = 'Settings form for Tripal Content entities.';
+
+    // Provide overall server consumption (and space remaining)
```
laceysanderson:
```suggestion
    // Define the HTML tags that tripal supports in Tripal Entity titles.
```

tripal/src/Services/TripalTokenParser.php
```diff
@@ -127,6 +127,13 @@ public function initParser(TripalEntityType $bundle, TripalEntity $entity = NULL
    *   The property values
    */
   public function addFieldValue($field_name, string $key, $value){
+    // Special case handling, fields set to one of the HTML formats
+    // will be stored wrapped in paragraph tags <p></p>.
+    // These are disruptive to the title, so remove them.
+    // Any other HTML markup is allowed, e.g. <em> <strong>
+    if ($value) {
+      $value = preg_replace('/^<p>(.*)<\/p>[\r\n]*$/', '$1', $value);
```
laceysanderson: Shouldn't this one also use the allowed tags and XSS approach?

dsenalik: This change can be reverted, the &lt;p&gt; tags will now get filtered out by XSS::sanitize later.

dsenalik: I'm always happy to get rid of "special case" handling code 🙃

laceysanderson (approved):
✅ Code changes look good
✅ Manual testing for organism works as expected (see screenshot)
✅ Manual testing for other content types using bold, italic and underline work as expected (see screenshots of projects)
✅ Settings for worked as expected and changes were reflected immediately in pages.

NOTE: The settings for is actually at Manage > Tripal > Configuration > Tripal Entity Settings AND Tripal > More > Configuration > Tripal Entity Settings.

Screenshot 2024-06-04 at 11 43 47 AM

Screenshot 2024-06-04 at 11 34 08 AM
Screenshot 2024-06-04 at 11 36 12 AM
Screenshot 2024-06-04 at 11 44 14 AM

Issue #1843, opened 2024-04-03
Publish is not publishing more than one item when cardinality is unlimited


dsenalik:
### Tripal Version

4.x-dev

### Drupal Version

10.2

### PostgreSQL Version

13.14

### PHP Version

8.3

### General Description

It appears that publish is not publishing more than one item when cardinality is unlimited. Only the first one gets published.

### Steps to reproduce

1-4. Create two contacts, and two projects. Do not put a contact on the project or vice versa.
5. Generate links with this SQL: `insert into project_contact (project_id, contact_id) values (1,2), (1,3), (2,2), (2,3);`
`INSERT 0 4`
6. Publish contact and run the job
7. Publish project and run the job ~~(don't run both together because of issue #1716)~~ (fixed with PR #1841)
8. Rebuild cache `drush cache:rebuild`
9. Observe that only one contact is present on each project, only one project present on each contact.

laceysanderson:
I jumped in on this to see if I can figure out what is going on. I collaborated on @dsenalik existing branch, `tv4g1-issue1843-fix-for-cardinality-gt-1`, by bringing it up to date and expanding the findFieldItems return value. Previously this array just recorded if there was a published field item for a given field/delta. Now, it also saved the required property type values. This allows us to better confirm that the field item has in fact been published previously.

I also added some debugging so that we can see what is being published and what is being skipped, field item wise. This shows that the issue is not actually in the find/insertFieldItem functions. Rather it seems that the $matches array only includes one entry for these unlimited fields 🙈.

For example when I setup my environment as explained in this issue and then publish the contact content type repeatedly 🤪 this is the output I see on subsequent tries:
```
2024-04-28 18:58:11: Job ID 25.
2024-04-28 18:58:11: Calling: tripal_publish(contact, chado_storage, Array)
Step  1 of 6: Find matching records... 
 [notice] Step  1 of 6: Find matching records... 
Step  2 of 6: Generate page titles...
 [notice] Step  2 of 6: Generate page titles...
Step  3 of 6: Find existing published entities...
 [notice] Step  3 of 6: Find existing published entities...
Step  4 of 6: Publishing 0 new entities...
 [notice] Step  4 of 6: Publishing 0 new entities...
Step  5 of 6: Find IDs of entities...
 [notice] Step  5 of 6: Find IDs of entities...
Step  6 of 6: Add field items to published entities...
 [notice] Step  6 of 6: Add field items to published entities...
  Checking for published items for the field: contact_description...
[...]

PR #1867, opened 2024-04-29
Publish is not publishing more than one linking record

laceysanderson:
# Bug Fix

## NOT COMPLETE!

Putting up as draft so that the tests run automatically to ensure I don't break anything ;-p

## Description

There are a number of fields with unlimited cardinality (e.g. properties and linking fields). Currently if you have multiple records added to chado that these fields would normally show, publish will only show the first one. Furthermore, I'm not sure it's even doing that correctly. For example, if you have multiple contacts linked to a project, when you publish the projects only the first contact link will be published.

## Testing?

1-4. Create two contacts, and two projects. Do not put a contact on the project or vice versa.
5. Generate links with this SQL: `insert into chado.project_contact (project_id, contact_id) values (1,2), (1,3), (2,2), (2,3);`
`INSERT 0 4`
6. Publish contact and project and run the job
8. Rebuild cache `drush cache:rebuild`

Currently you will observe that only one contact is present on each project, and only one project present on each contact. There should actually be two contacts on each project and two projects on each contact.

PR #1882, opened 2024-05-31
Tv4g1 issue1843 publishing cardinality

spficklin:
# Bug Fix

## Description

This PR solves the issue described in issue #1843 where field values derived from linked tables in Chado were only showing the first item and ignoring the others.  In actuality the problem was that the testing we did made it look like it was publishing the first item but in fact it was publishing the wrong thing but because record IDs were similar between the base record and linked record we didnt notice. For example.  If you add a project and link two contacts to it, then publishin the contacts on the project, it looked like it only added one, but it was really publishing a contact with the same record ID as the project not the contact. Since the first contact was what was desired it looked like it was working .

There were multiple issues that this PR solves in both ChadoStorage and TripalPublish:

1.  There was a bug introduced that prevented title tokens from being renamed. I fixed that with the help of @dsenalik and @laceysanderson 
2.  The `ChadoStorage::findValues()` function sets the field property values after it finds all of the fields.  One property that all fields have is the `record_id` which always points to the base record.  When we find values, we always include the fields in the values array passed in because we want to find any field items to publish.  This resulted in the `findValues` function setting the `record_id` for the first item for all fields regardless if any items were found.  This resulted in publishing of incomplete items. They didn't seem to show up on the pages, but they were in the Drupal tables. This was fixed by adding a new function named `ChadoStorage::isFieldValid()` that the `findValues()` uses to make sure a field has all of the required values before including it in what is returned.  I also add a `has_values` to the `ChadoRecords` array that is set if the `values` for the item are really set.  This allows the `isFieldValid()` to really be sure that an item has a set value or not (i.e. empty values are sometimes okay).   Also, the `findValues()` function was presetting the IDs and Links on linked tables and it should not have been doing that.  
3. The reason that we could publish all properties but not all contacts on a project was because the `ChadoRecords::findRecords()` function was not looking at all of the ancialliary tables.  There was a bug in that it exepcted that at least on of the columns in the linked table had to be used. This is the case in a property table where the `value` column is used, but in the case of a linking table, we don't need any other columns but the IDs.  So, once the problem in item 2 above was fixed, no `project_contact` records were being published at all.  I fixed this so the `findRecords()` function now will go through all linked tables.  This is fixed.
4. The `TripalPublish::insertFieldItems()` had a bug in that it wasn't properly cycling through the entities correctly. This is fixed.  I also fixed the way the number of published field items was reported.  

## Testing?

1. Load a brand new site.
2. Create an analysis
3. Create an organism
4. Create 2 or more projects
5. Create 2 or more contacts
6. Add a lot of records using the GFF loader and publish content from it.  This tests that normal publishing is still working
7. Link the contacts to projects with this SQL statement: `insert into project_contact (project_id, contact_id) values (1,2), (1,3), (2,2), (2,3);`
8. Now publish the `project`. It should add two contact items to each of the two project pages.


spficklin:
I have not yet added any testing, but this is working so I thought I'd open it up for functional testing and code review.

dsenalik:
I built a new docker and created two contacts and five projects (so that I could make sure id's did not overlap)
```
 project_id |                  name                  |                                        description                                                                      
------------+----------------------------------------+-----------------------------------------------------------------------------------------
          1 | Colossus - The Forbin Project          | <p>Do not turn it on</p>
          2 | The Tripal Project                     | <p>Cool Stuff</p>
          3 | Manhattan Project                      | <p>dangerous</p>
          4 | Advanced LIGO                          | <p>A large-scale physics experiment and observatory to measure gravitational waves</p>
          5 | Superconducting Super Collider Project | <p>Cancelled</p>
(5 rows)
```
`insert into project_contact (project_id, contact_id) values (4,2), (4,3), (5,2), (5,3);`
Then published project:
```
 [notice]   Publishing 7 items for field: project_analysis...
ERROR: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "tripal_entity__project_analysis____pkey"
DETAIL:  Key (entity_id, deleted, delta, langcode)=(6, 0, 0, und) already exists.: 
      INSERT INTO "tripal_entity__project_analysis"
```
etc. but this was using the wrong branch, let's try again

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -526,6 +535,55 @@ protected function setPropValues(&$values, ChadoRecords $records) {
     }
   }
 
+
+  /**
+   * Checks if a field has all necessary elements to be considered 'foudn'.
```
dsenalik:
```suggestion
   * Checks if a field has all necessary elements to be considered 'found'.
```

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -526,6 +535,55 @@ protected function setPropValues(&$values, ChadoRecords $records) {
...
+   *
+   * The ChadoRecords class will search for all records necessary to
+   * populate the values of the fields for a content type. This works well
+   * when all conditions are set for the insertValues() and loadValues().
+   * However, for the findVales() function there are often no criteria set
+   * and we want to find all linked records associated with a base record.
+   * All Chado fields will have a `record_id` property and the value of that
+   * comes from the base table.  This means that all fields will ahve at
```
dsenalik:
```suggestion
   * comes from the base table.  This means that all fields will have at
```

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -1257,7 +1320,7 @@ static public function drupalEntityIdLookupCallback($context) {
 
     // During publish, record_id may be null. In this particular case, return null.
     if (!$record_id) {
```
dsenalik:
```suggestion
    if (!$record_id) {
```
If this is no longer a special case, the special case comment can be dropped.

spficklin:
Thanks @dsenalik for the quick check!  Hmmm looks like something is still wrong.  I'll see what's up.  I'm also seeing a functional test fail.

dsenalik:
OOPS! I used the wrong branch, from PR #1867 🙈 - building a new docker
On the branch in this PR publishing is working as expected.

dsenalik:
There is something wrong with publishing an organism. I create it manually with
`insert into organism (genus, species) values ('Tripalus', 'ficklinus');`

Then when I publish
(debugging shows that this happens on the infraspecific_name column)
...
 [notice] Step  2 of 6: Generate page titles...
 [warning] Undefined array key 0 TripalPublish.php:479
 [warning] Trying to access array offset on null TripalPublish.php:479
 [warning] Trying to access array offset on null TripalPublish.php:479
 [error]  Error: Call to a member function getValue() on null in Drupal\tripal\Services\TripalPublish->getEntityTitles() (line 479 of /var/www/drupal/web/modules/contrib/tripal/tripal/src/Services/TripalPublish.php) #0 /var/www/drupal/web/modules/contrib/tripal/tripal/src/Services/TripalPublish.php(870): Drupal\tripal\Services\TripalPublish->getEntityTitles(Array)
#1 /var/www/drupal/web/modules/contrib/tripal/tripal/src/api/tripal.publish.api.php(35): Drupal\tripal\Services\TripalPublish->publish()
#2 [internal function]: tripal_publish('organism', 'chado_storage', Array, Object(Drupal\tripal\Services\TripalJob))
#3 /var/www/drupal/web/modules/contrib/tripal/tripal/src/Services/TripalJob.php(395): call_user_func_array('tripal_publish', Array)
... and so on...

tripal/src/Services/TripalPublish.php
```diff
@@ -477,9 +477,10 @@ protected function getEntityTitles($matches) {
           $field = $this->field_info[$field_name]['instance'];
           $main_prop = $field->mainPropertyName();
           $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
-          if ($value !== NULL) {
-            $entity_title = trim(preg_replace("/\[$field_name\]/", $value,  $entity_title));
+          if ($value === NULL) {
+            $value = '';
           }
```
dsenalik:
```suggestion
          $value = '';
          if (array_key_exists($delta, $match[$field_name])) {
            $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
          }
          if ($value === NULL) {
            $value = '';
          }
```
With this change I can publish an organism.
I got this incorrect, marking as resolved and I will re-do.

tripal/src/Services/TripalPublish.php
```diff
@@ -477,9 +477,10 @@ protected function getEntityTitles($matches) {
           $field = $this->field_info[$field_name]['instance'];
           $main_prop = $field->mainPropertyName();
           $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
```
dsenalik:
```suggestion
          $value = '';
          if (array_key_exists($delta, $match[$field_name])) {
            $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
          }
```
With this change I can manually publish an organism, and run the failing test locally without error.

dsenalik:
There is another problem 😢
Create an organism in chado
`insert into organism (genus, species, common_name, comment) values ('Tripalus', 'ficklinuus', 'Ficklins Tripal', 'A rare blue variant of the common tripal');`
Publish organism
`Published 1 new entities, and 6 field values.`
Rebuild cache to be safe
View it - looks okay.
2024-06-01_good

Enter a second organism
`insert into organism (genus, species, common_name, comment) values ('Tripalus', 'senalikus', 'Senaliks Tripal', 'A rare purple variant of the common tripal');`
Publish organism
`Published 1 new entities, and 6 field values.`
Rebuild cache to be safe
The page only has a title and no other fields.
2024-06-01_nofields

If you try to edit, everything is blank, including genus and species.
The drupal tables don't contain anything for this entity_id, only the previous one is there (explaining the blank edit form)

tripal/src/Services/TripalPublish.php
```diff
@@ -733,25 +741,35 @@ protected function insertFieldItems($field_name, $matches, $titles, $entities, $
...
         $j++;
         $total++;
 
         // No need to add items to those that are already published.
         if (array_key_exists($entity_id, $existing)) {
-          continue;
+          if (array_key_exists($delta, $existing[$entity_id])) {
+            continue;
+          }
         }
```
dsenalik: I can't do a suggestion here, but the unpublished fields happens because of the continue. If we happen to be on the LAST iteration, we never get to the code below to perform the sql query!
We need to change this to something like
```
if (!array_key_exists($entity_id, $existing) or !array_key_exists($delta, $existing[$entity_id])) {
  then do the sql code below up to and including $num_inserted++;
}
```

dsenalik (approved):
This review is based on including my last two suggestions, and testing with publishing two organisms one at a time as described in comment https://github.com/tripal/tripal/pull/1882#issuecomment-2143489314 now works correctly.

I tested elsewhere more thoroughly, everything appears to be working as intended. 
In addition to project-contact, I linked two publications to analysis and these published correctly.
GFF importing and publishing gene, mRNA worked also.

tripal/src/Services/TripalPublish.php
```diff
@@ -476,10 +476,18 @@ protected function getEntityTitles($matches) {
...
-          if ($value !== NULL) {
-            $entity_title = trim(preg_replace("/\[$field_name\]/", $value,  $entity_title));
+          $value = '';
+          if (array_key_exists($delta, $match[$field_name])) {
+            $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
           }
+          $value = '';
+          if (array_key_exists($delta, $match[$field_name])) {
+            $value = $match[$field_name][$delta][$main_prop]['value']->getValue();
+          }
```
dsenalik:
```suggestion
```
I'm not sure what happened here but the code was duplicated. This removes the second copy. I'll just go ahead and commit this...

tripal/src/Services/TripalPublish.php
spficklin: I've integrated this fix. Thanks for checking deep enough to spot this.

spficklin:
Thanks @dsenalik for the testing! I've integrated all of your suggestions.   I'm working on functional tests right now, so hopefully will have some of that soon.

laceysanderson (commented):
I did some manual testing according to the instructions and it worked beautifully for me on the current commit 😍 I'll do some more intense manual testing soon.

I also did a code review and all of the changes are very reasonable and don't raise any red flags for me. The "why these changes fix this issue" is still percolating a bit in my mind but I believe makes sense based on the code changes 🤪 🙈 

PS. I love the fix to the reporting of what was actually published even though it is smaller in importance then the main issue 🙃 
PSS. More on my mention of more manual testing: I'm going to try publishing some content types on a Tripal 4 KnowPulse clone which should really test a large magnitude of property types and different relationships as well as magnitude in general 🤪 😈

spficklin:
Thanks for your reviews @laceysanderson and @dsenalik . I have now completed the functional testing for publishing.  I put it in the `tripal_chado` module rather than in the `tripal` module because it would have been really hard to create mocks to test the issues we are having in Chado.   One nice thing about it is that we now have an example for how to write a functional test for a field.   In the process of writing those tests I discovered there were more bugs in the ChadoStorage and ChadoRecords classes.  For example, publishing was working but suppose you had 3 property values on an organism.  When you published it would show only 3 on the website but in actuality there were more  than 3 records in the Drupal field tables.  This is now fixed, but it took a few changes to those classes.  So, you might want to redo your manual testing just to make sure nothing else broke in the process.   As long as the functional tests pass then I think we're good to go on this PR (post your additional reviews).

laceysanderson:
Awesome! Sounds good @spficklin -I agree with putting the testing in tripal_chado since in this case it allows us to also confirm Tripal Publishing's interactions with ChadoStorage.

I've only taken a quick look at this point but am excited by the added functions for readibility! 

One big question though: I see you used the bootstrap theme in testing 😬 Why did you end up going that route rather then using stark as recommended by Tripal?

tripal/src/Services/TripalPublish.php
```diff
@@ -476,12 +476,20 @@ protected function getEntityTitles($matches) {
...
+          if ($value === NULL) {
+            $value = '';
+          }
+          $entity_title = trim(preg_replace("/\[$field_name\]/", $value,  $entity_title));
         }
       }
+      // Trim any training spaces and remove double spaces. Double spaces
+      // can occur if a token replacment has no value but there are spaces
+      // around it.
+      $entity_title = trim(preg_replace('/\s\s/', ' ', $entity_title));
```
dsenalik:
```suggestion
      // Trim any trailing spaces and remove double spaces. Double spaces
      // can occur if a token replacement has no value but there are spaces
      // around it.
      $entity_title = trim(preg_replace('/\s\s+/', ' ', $entity_title));
```
I am suggesting this regex change in the unlikely chance there are more than two spaces, it will still collapse to one.

tripal/src/TripalStorage/TripalStorageBase.php
```diff
@@ -228,20 +228,34 @@ protected function cloneValues($values) {
   }
 
   /**
-   * A helper function to add a new item for a field by cloning delta 0.
+   * Sets the values for a field to be empty.
+   *
+   * If the detla value doesn't exist in the values array then a new values
```
dsenalik:
```suggestion
   * If the delta value doesn't exist in the values array then a new values
```

tripal_chado/src/Plugin/Field/FieldType/ChadoContactTypeDefault.php
```diff
@@ -162,20 +162,23 @@ public static function tripalTypes($field_definition) {
...
       // Define the link between the base table and the linker table.
+      // E.g.,:  project.project_id>project_contact.project_id
       $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'link', $linker_left_term, [
         'action' => 'store_link',
         'drupal_store' => TRUE,
         'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col,
       ]);
 
       // Define the link between the linker table and the object table.
+      // E.g.,:  project_contact.contact_id
```
dsenalik:
```suggestion
      // E.g.:  project_contact.project_contact_id
      $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'linker_id', $record_id_term, [
        'action' => 'store_pkey',
        'drupal_store' => TRUE,
        'path' => $linker_table . '.' . $linker_pkey_col,
      ]);

      // Define the link between the base table and the linker table.
      // E.g.:  project.project_id>project_contact.project_id
      $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'link', $linker_left_term, [
        'action' => 'store_link',
        'drupal_store' => TRUE,
        'path' => $base_table . '.' . $base_pkey_col . '>' . $linker_table . '.' . $linker_left_col,
      ]);

      // Define the link between the linker table and the object table.
      // E.g.:  project_contact.contact_id
```
LOL I feel like there is too much punctuation 😁

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -314,52 +314,51 @@ public function findValues($values) {
...
             }
 
-            // Add any additional items to the values array that are needed.
-            $num_items = $match->getNumTableItems($base_table, $table_alias);
-            for ($i = 0; $i < $num_items - 1; $i++) {
-              $table_fields = $match->getTableFields($base_table, $table_alias);
-              foreach ($table_fields as $field_name) {
-                $this->addEmptyValuesItem($new_values, $field_name);
+            // Prepare the values array to receive all the new values. We'll
+            // get all the fields for this ancialiary table and then
```
dsenalik:
```suggestion
            // get all the fields for this ancillary table and then
```
Typo

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -526,6 +528,55 @@ protected function setPropValues(&$values, ChadoRecords $records) {
...
   }
 
+
+  /**
+   * Checks if a field has all necessary elements to be considered 'found'.
+   *
+   * The ChadoRecords class will search for all records necessary to
+   * populate the values of the fields for a content type. This works well
+   * when all conditions are set for the insertValues() and loadValues().
+   * However, for the findVales() function there are often no criteria set
```
dsenalik:
```suggestion
   * However, for the findValues() function there are often no criteria set
```

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -1708,24 +1713,25 @@ public function findRecords(string $base_table, string $table_alias) {
           ['@table' => $chado_table, '@record' => print_r($record, TRUE)]));
       }
 
+      // Iterate through the results and create a new record for each one..
```
dsenalik:
```suggestion
      // Iterate through the results and create a new record for each one.
```

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -1876,7 +1877,11 @@ public function deleteRecords(string $base_table, string $table_alias, bool $gra
   }
 
   /**
-   * Selects a single record from Chado.
+   * Selects the items for a given table in a record object.
+   *
+   * This function is used for the findValues() and loadValues() functions so
+   * it needs to be able to find multiple records from the base table and
+   * multiple items from an anciallyar table.
```
dsenalik:
```suggestion
   * multiple items from an ancillary table.
```
😂

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -1887,26 +1892,26 @@ public function deleteRecords(string $base_table, string $table_alias, bool $gra
...
+    $items_found = 0;
 
     // Make sure all IDs are up to date.
     $this->setLinks($base_table);
 
     // Get the Chado table for this given table alias.
     $chado_table = $this->getTableFromAlias($base_table, $table_alias);
 
-    // Iterate through each item of the table and perform an insert.
+    // Iterate through each item of the table and perform an select.
```
dsenalik:
```suggestion
    // Iterate through each item of the table and perform a select.
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+   *   The content type bundle name (e.g. 'organism').
+   * @param string $field_name
+   *   The name of the field that should be queried.
+   * @param int $num_expected
+   *   The number of items that are expected to be found when applying the
+   *   conditions specified in the $match argument.
+   * @param array $match
+   *   An array of key/value pairs where the keys are the column names of
+   *   field table in Drupal and the values are those to match in a select
+   *   condition.  All fields other than the `entity_id', 'bundle', 'delta'
```
dsenalik:
```suggestion
   *   condition.  All fields other than the 'entity_id', 'bundle', 'delta'
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+
+    // Because we added properties for the first organism we should set it's
+    // entity in those returned, but not the gorilla organism.
+    $this->assertTrue(array_values($entities)[0] == '<em>Oryza species</em> subspecies <em>Japonica</em>',
+      'The Oryza species subspecies Japonica organism should appear in the published list because it has new properties.');
+    $this->assertTrue(count(array_values($entities)) == 1,
+      'There should only be one published entity for a single organism with new properties.');
+
+    // Check that the property values got published.  The type_id should be
+    // NULL because that's not stored in DRupal.
```
dsenalik:
```suggestion
    // NULL because that's not stored in Drupal.
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+
+    // Check that only the exact number of properties were published.
+    $this->checkFieldItem('organism', 'field_note', 3, ['entity_id' => 1], []);
+    $this->checkFieldItem('organism', 'field_comment', 2, ['entity_id' => 1], []);
+
+    //
+    // Test publishing a field that uses a linker table.
+    //
+
+    // Create an and publish the contacts and the project.
```
dsenalik:
```suggestion
    // Create and publish the contacts and the project.
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+    $this->assertTrue(count(array_values($entities)) == 3,
+        'Failed to publish 3 contact entities.');
+
+    $publish->init('project', 'chado_storage');
+    $entities = $publish->publish();
+    $this->assertTrue(count(array_values($entities)) == 2,
+      'Failed to publish 2 project entities.');
+
+    // Make sure that the linked records are also published for each project.
+    // The chado_contact_type_default is the field we're testing got published.
```
dsenalik, line 572:
```suggestion
    // The chado_contact_type_default is the field we're testing got published. (Issue #1809)
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+      'contact_id' => $contact_id2,
+    ]);
+    $project_contact_id3 = $this->addChadoProjectContact($chado, [
+      'project_id' => $project_id2,
+      'contact_id' => $contact_id2,
+    ]);
+
+    // Now publish the projects and contacts. We check that 3 items are
+    // published because there is a null contact and currently there is
+    // nothing to prevent that contact from being published.
```
dsenalik:
```suggestion
    // nothing to prevent that contact from being published. (Issue #1809)
```

dsenalik (approved):
I have tested everything I had tried previously, and everything is working as expected.
I do have a number of typos that I spotted, but otherwise this is a super-awesome PR!

spficklin:
Thanks @dsenalik for the additional review! It looks like your mispelling fixes got merged in.  I just approved the one where you made a change to the regex.  

@laceysanderson in answer to this question:

> One big question though: I see you used the bootstrap theme in testing 😬 Why did you end up going that route rather then using stark as recommended by Tripal?

I don't know :-) Something I brought in accidentally and it has been removed.

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -1123,10 +1123,16 @@ public function getColumnValue(string $base_table, string $table_alias, int $del
       return NULL;
     }
 
-    return $items[$delta]['values'][$column_alias];
+    // If the values were set then return it, otherwise return NULL;
+    if ($items[$delta]['values'] == TRUE) {
```
laceysanderson: I'm confused by this line. Did you mean to check has_values here instead?
```suggestion
    if ($items[$delta]['has_values'] === TRUE) {
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+    //
+    $organism_id2 = $this->addChadoOrganism($chado, [
+      'genus' => 'Gorilla',
+      'species' => 'gorilla',
+      'abbreviation' => 'G. gorilla',
+      'comment' => 'Gorilla'
+    ]);
+    $entities = $publish->publish();
+    $this->assertTrue(array_values($entities)[0] == '<em>Gorilla gorilla</em> <em></em>',
+        'The title of Chado organism with missing tokens is incorrect after publishing: "' . array_values($entities)[0] . '" != "<em>Gorilla gorilla</em> <em></em>"');
```
laceysanderson: It's better to use assert equals for this comparison because it provides more information if there is a failure. This way the error message will include the comparison just like you tried to add it into the message. So in this exact case it's equivalent but the formatting is better and it's easier to make sure the values shown don't get out sync in the future.
```suggestion
    $this->assertEquals('<em>Gorilla gorilla</em> <em></em>', array_values($entities)[0],
        'The title of Chado organism with missing tokens is incorrect after publishing');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+
+
+    // Now publish the organism content type again.
+    $publish->init('organism', 'chado_storage');
+    $entities = $publish->publish();
+
+    // Because we added properties for the first organism we should set it's
+    // entity in those returned, but not the gorilla organism.
+    $this->assertTrue(array_values($entities)[0] == '<em>Oryza species</em> subspecies <em>Japonica</em>',
+      'The Oryza species subspecies Japonica organism should appear in the published list because it has new properties.');
```
laceysanderson:
```suggestion
    $this->assertEquals('<em>Oryza species</em> subspecies <em>Japonica</em>', array_values($entities)[0],
      'The Oryza species subspecies Japonica organism should appear in the published list because it has new properties.');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+    $this->assertTrue(count(array_values($entities)) == 1,
+      'There should only be one published entity for a single organism with new properties.');
```
laceysanderson: Using assertCount will show the actual count obtained if it fails whereas using assertTrue and doing the comparison yourself does not provide that information.
```suggestion
    $this->assertCount(1, array_values($entities),
      'There should only be one published entity for a single organism with new properties.');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+      'contact_id' => $contact_id2,
+    ]);
+
+    // Now publish the projects and contacts. We check that 3 items are
+    // published because there is a null contact and currently there is
+    // nothing to prevent that contact from being published. (Issue #1809)
+    $publish->init('contact', 'chado_storage');
+    $entities = $publish->publish();
+    $this->assertTrue(count(array_values($entities)) == 3,
+        'Failed to publish 3 contact entities.');
```
laceysanderson:
```suggestion
    $this->assertCount(3, $entities,
        'Failed to publish 3 contact entities.');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+
+    $publish->init('project', 'chado_storage');
+    $entities = $publish->publish();
+    $this->assertTrue(count(array_values($entities)) == 2,
+      'Failed to publish 2 project entities.');
```
laceysanderson:
```suggestion
    $this->assertCount(2, $entities,
      'Failed to publish 2 project entities.');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+      }
+      $select->condition($column_name, $val);
+    }
+    $select->orderBy('delta');
+    $result = $select->execute();
+    $records = $result->fetchAll();
+
+    $this->assertTrue(count($records) == $num_expected,
+        'The number of items expected for field "' . $field_name .'" with bundle "'
+        . $bundle . '" is not correct: ' . count($records) . ' != ' . $num_expected);
```
laceysanderson:
```suggestion
    $this->assertCount($num_expected, $records,
        'The number of items expected for field "' . $field_name .'" with bundle "'
        . $bundle . '" is not correct.');
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -0,0 +1,590 @@
...
+
+      // Make sure the expected values are present.
+      foreach ($check as $key => $val) {
+        $column_name = $key;
+        if (!in_array($key, $drupal_columns)) {
+          $column_name = $field_name . '_' . $key;
+        }
+        $this->assertTrue($record->$column_name == $val,
+          'The value for, "' . $column_name . '", is not correct: '
+            . $record->$column_name . ' (actual) != ' . $val . ' (expected)');
```
laceysanderson:
```suggestion
        $this->assertEquals($val, $record->$column_name,
          'The value for, "' . $column_name . '", is not correct what we expected.');
```

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -687,9 +696,9 @@ public function setLinks(string $base_table) {
 
           // if this column is an ID field and links to this base table then update the value.
           if (array_key_exists($column_alias, $record['link_columns'])) {
-            $base_table =  $record['link_columns'][$column_alias];
-            $record_id = $record_id;
-            $this->setColumnValue($base_table, $table_alias, $delta, $column_alias, $record_id);
+            $base_table = $record['link_columns'][$column_alias];
+            $this->records[$base_table]['tables'][$table_alias]['items'][$delta]['values'][$column_alias] = $record_id;
```
laceysanderson, line 700: It might be good to add a comment here why you didn't use setColumnValue so it doesn't get changed back in the future 🙈

laceysanderson (commented):
Code review suggestions. The suggestions in the test are improvements only. I may have found a bug in chadorecord changes though.

laceysanderson (approved):
Thanks for all the changes @spficklin, this looks good to merge to me!

laceysanderson:
Merging this since we have 2 reviews and passing tests! Thanks @spficklin for fixing this pesky one!!!

Issue #1824, opened 2024-03-22
Publish doesn't get dbxref quite right


dsenalik:
### Tripal Version

4.x

### Drupal Version

10.2

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

If you import a GFF3 file with a database cross reference, _e.g._
`ID=gene-YAL068C;Dbxref=GeneID:851229`
The cross reference is imported correctly
```
 dbxref_id | db_id | accession | version | description 
-----------+-------+-----------+---------+-------------
      3485 |    45 | 851229    |         | 
```
but it is not displayed on the gene content page.

A second cross reference manually added is displayed as expected:
![2024-03-22_gene-dbxref](https://github.com/tripal/tripal/assets/8419404/2bf0b3b9-cdb6-43a3-8519-f73afcbba31d)

In the `public.tripal_entity__gene_dbxref_ann` table some columns are empty that maybe shouldn't be
`gene_dbxref_ann_linker_id`
`gene_dbxref_ann_dbxref_id`
`gene_dbxref_ann_linker_is_current`
`gene_dbxref_ann_dbxref_db_id`

### Steps to reproduce

 1. Import the type collection "Genomic Content Types" - **And to avoid any unrelated testing issues, remove the sequence coordinates field from the gene content type** (To be dealt with in issue #1859)
 2. create an organism [Saccharomyces cerevisiae](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=559292)
 3. create an analysis
 4. Manually enter a gene with a Database Reference Annotation. I used `NCBITaxon:559292`
 5. View the gene and the link shows up. This is expected behavior so far.
![2024-04-26_err4](https://github.com/tripal/tripal/assets/8419404/8e9e3596-3598-487e-99ba-6727bf91b23d)
 6. Verify in the Drupal field table. Looks good so far, this happens to be entity 6 when I tested.
 ```
sitedb=> select * from tripal_entity__gene_dbxref_ann;
[...]

PR #1889, opened 2024-06-05
Fields move record_id_term to parent class

### LINKED ISSUE #1879: Stop hardcoding record_id_term
### Task Description

We use this line of code in all but three fields:
`$record_id_term = 'SIO:000729';`

We should move this to the parent class.

And those three fields that use a different term `ChadoSequenceCoordinatesDefault`, `ChadoSourceDataTypeDefault`, and `ChadoUnitTypeDefault` should use this also (that could arguably be a separate PR)

### Branch Name

tv4g1-issue1879-record_id_term

dsenalik:
# Tripal 4 Core Dev Task

## Description
We have a hardcoded `$record_id_term = 'SIO:000729';` in most fields.
This moves this to the parent `ChadoFieldItemBase` class.

Three fields weren't consistent so they now also use this term, they can get extra careful code review 😼
ChadoSourceDataTypeDefault
ChadoUnitTypeDefault
ChadoSequenceCoordinatesDefault

## Testing?
Automated testing will test some of this. But I suppose ideally we need to create at least one instance of every field type. Importing all of the field type collections should do this.

To test these three fields, create an organism and analysis.
The analysis will test the source data field.
Run the GFF3 importer. I used this file for testing from issue #1824
Publish gene
Check the sequence coordinates field, should be `NC_001133.9:1806..2169-`
Create a genetic map to test the unit field.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoSequenceCoordinatesDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoSequenceCoordinatesDefault.php
@@ -89,19 +89,19 @@ class ChadoSequenceCoordinatesDefault extends ChadoFieldItemBase {
     // Get property terms using Chado table columns they map to. Return the properties for this field.
     $properties = [];
 
-    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', $record_id_term, [
+    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', self::$record_id_term, [
       'action' => 'store_id',
       'drupal_store' => TRUE,
       'path' => 'feature.feature_id',
     ]);
 
-    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'featureloc_id', $record_id_term, [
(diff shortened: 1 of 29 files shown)


tripal_chado/src/Plugin/Field/FieldType/ChadoSequenceCoordinatesDefault.php
```diff
@@ -70,11 +70,11 @@ public static function tripalTypes($field_definition) {
...
     // Get the property terms by using the Chado table columns they map to.
     $storage = \Drupal::entityTypeManager()->getStorage('chado_term_mapping');
-    $mapping = $storage->load( 'core_mapping' );
+    $mapping = $storage->load('core_mapping');
 
-    $record_id_term = $mapping->getColumnTermId( 'feature', 'feature_id' );
-    $ft_uniqname_term = $mapping->getColumnTermId( 'feature', 'name' );
+    $ft_uniqname_term = $mapping->getColumnTermId('feature', 'name');
 
+    $feature_id_term = $mapping->getColumnTermId('featureloc', 'feature_id');
```
laceysanderson, line 77: Good catch! I agree with getting the term for the featureloc.feature_id in case we want to be more specific there :-)

laceysanderson (approved):
✅ Code changes look great
✅ Fields attach without error
✅ Can create content with no errors

PR #1892, opened 2024-06-10
Unpublish UI

spficklin:
# Tripal 4 Core Dev Task

## Description

In order to implement the `ChadoStorage::deleteValues()` function we need to differentiate between an "unpublish" and  "delete".   This PR simply adds that functionality to the UI. It doesn't' fully implement a delete. or the `deleteValues()` function.  I separted this as a separte PR to simplify the number of files that have to be checked in a full PR for this issue.

## Testing?

1.  Go to a Tripal entity page. You'll see both an "Unpublish" and "Delete" tab at the to of the page.  Click and test each one. For the delete you should see a message that the delete is currently not fully implemented and it works like an "unpublish".
2. Edit any Tripal entity page. You'll see a new "Unpublish" button (it's red on the default theme) at the bottom of the page.  The delete button is the default button in Drupal.  They should also both work. They use the same underlying functions, but use a "pop-up" to confirm unpublish/delete (default Drupal behavior).
3. Go the the Tripal content page listing. In the action bar on the right you'll now see an unpublish option. It uses the same pop-up as the entity edit page.


dsenalik:
I built a docker on this branch, and if I log in and then try to add any content type I get:

Drupal\Core\Entity\EntityMalformedException: The "tripal_entity" entity cannot have a URI as it does not have an ID in Drupal\Core\Entity\EntityBase->toUrl() (line 161 of core/lib/Drupal/Core/Entity/EntityBase.php).

tripal/src/Form/TripalEntityForm.php
```diff
@@ -57,4 +58,30 @@ public function save(array $form, FormStateInterface $form_state) {
...
   }
 
+  /**
+   *
+   * {@inheritDoc}
+   * @see \Drupal\Core\Entity\EntityForm::actions()
+   */
+  protected function actions(array $form, FormStateInterface $form_state) {
+    $actions = parent::actions($form, $form_state);
+    $route_info = $this->entity->toUrl('unpublish-form');
```
dsenalik: This line (68) is what is generating the error - apparently at this point `$this->entity->id()` is NULL.

So if I wrap the remainder of this function in
```
if ($this->entity->id()) {
  $route_info = ...
...
}
```
then I can add content. But I don't understand enough to know if this is something that would be correct to add

spficklin: Thanks @dsenalik for hitting on this problem!  When I overwrote the `actions()` function I failed to include a check if the entity was new (similar to the parent function that adds the "edit" and "delete" actions).  I've updated the code and tested a brand new installation and it's working now.  Please feel free to try again.

dsenalik (changes requested):
Testing with the latest changes works as expected.
I have a request, though. Save is right next to unpublish and delete.
2024-06-14_deletebutton
In Tripal 3 we had the "dangerous" buttons on the far right side.
2024-06-14_tripal3delete

I request similar placement. And do we want similarly "Update" and "Cancel" when editing?

Tripal 3 button placement was discussed in PR #1061

spficklin:
Thanks @dsenalik for the review.  Regarding the button floating on the right like it did in Tripal 3. Something like this is required in the form element:

```
  '#attributes' => ['style' => 'float: right'],
```
I specifially decided not to do that because it requires a hard coding of a style that floats the button on the right.  In Tripal 3 we did some forcing of style which didn't always play nice with different themes.   It should be easy enough to do, I was trying to avoid forcing any style.

dsenalik:
As an idea, what if we just add a "Cancel" button, and have it between the "Save" and "Unpublish" buttons, It can act as a less-harmful buffer, and then you don't have to mess with placement and messing with themes. But just an idea.
\* But would that show up in the dropdown menu too? That would be confusing...

spficklin:
@dsenalik I think a Cancel button is reasonable. I added one.

image

The 'Cancel' won't show up in the dropdown actions, it's only on the form.

dsenalik (approved):
Awesome!
Verified that the cancel button works as expected.

laceysanderson:
It agree that it would be best floated to the right but also that there is definitely a balance with the overriding theme 🙈 

With both in mind, I like this solution (the cancel button to provide some buffer) :-)

PR #1896, opened 2024-06-13
add formatting settings to ChadoStringType field

### LINKED ISSUE #1893: Allow the ChadoStringTypeFormatter to format

Yes

Yes

### Is this feature urgent?

No

### Description

Currently on an organism page, genus etc. appear thusly
2024-06-11_genusspecies

If the ChadoStringTypeFormatterDefault field had a setting for italics, then we could just set that for these taxon fields and display them in italic. This should be much cleaner than having the formatting stored in the database tables, and no complication with web services.

### Your specific use case

.

### Generally Applicable

Maybe gene symbols?

PR #1896

dsenalik:
# New Feature

## Description
This PR adds display settings to the ChadoStringTypeDefault field.
These are used to wrap the value in &lt;em&gt; tags so that on an organism page, the genus,
species, and infraspecific name can be displayed in italics.
2024-06-13_T _bogusii

I recycled the existing `field_prefix` and `field_suffix` settings, ~~so no schema additions are needed~~.

A point for discussion can be the "ease of use" - there is not a checkbox for italic, you have to go in and add the HTML tags directly. For example
2024-06-13_genus_settings
but the advantage is you have more flexibility, maybe you want a dollar sign prefix or something, or a &lt;span&gt; tag with some css, or you want some units after the value.

And note that I am not using xss::sanitize, I'm wondering about this, so I may need to add something. But I wonder if all fields should be sanitized, since they are getting who knows what from database tables. Maybe this needs some discussion...

## Testing?
 - Build a docker on this branch. This will test the default settings in the YAML.
 - Add an organism with infraspecific type. Verify that the appropriate fields are displayed in italics.
 - Test the settings form. Go to the display settings for organism species, and remove the prefix and suffix, and save.
 - Try adding &lt;strong&gt; tags to the organism common name just for fun.
 - cache rebuild maybe, I didn't need to.
 - Go back to the organism and the species should now not be in italics, common name should be bold.
2024-06-13_manual_edit_test_1
2024-06-13_manual edit test

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -318,6 +318,18 @@ tripal.tripalfield_collection.*:
                            type: integer
                            label: 'Weight'
                            nullable: false
+                         settings:
+                           type: mapping
+                           label: 'Settings'
+                           mapping:
+                             field_prefix:
+                               type: string
+                               label: 'Field prefix'
+                               nullable: true
+                             field_suffix:
+                               type: string
+                               label: 'Field suffix'
+                               nullable: true
                    form:
                      type: sequence
                      label: 'Edit'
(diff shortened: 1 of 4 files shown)


spficklin (approved):

spficklin:
I like this PR and tested it and it works great.  Thanks @dsenalik .  The code review looks good, I  approve merging.

PR #1907, opened 2024-06-19
Rename duplicated terms in tcontact.obo

### LINKED ISSUE #1906: Duplicate terms in tcontact.obo
### Tripal Version

any

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

The file `tripal_chado/files/tcontact.obo` defines some terms for describing entries in the `chado.contact` table.
Two of these terms are duplicated: "Organization" and "Department". I put the relevant parts of this file below in the "Error messages" section.
The OBO importer handles this by renaming one, so you will have
"Organization" and "Organization(TCONTACT:0000018)"
"Department" and "Department(TCONTACT:0000016)"

Not very intuitive when you go to add a contact that is one of these
2024-06-Organization

We have to keep both for backward compatibility, but I propose renaming them to no longer be identical, then Chado prepare won't need to add the goofy suffix.

TCONTACT:0000016 will become "Affiliated Department"
TCONTACT:0000018 will become "Affiliated Organization"
and, for consistency,
TCONTACT:0000017 will become "Affiliated Institution"

### Steps to reproduce

.

### Error messages

```
[Term]
id: TCONTACT:0000004
name: Organization
is_a: TCONTACT:0000001 ! Contact Type

[Term]
id: TCONTACT:0000009
name: Department
is_a: TCONTACT:0000001 ! Contact Type

[Term]
id: TCONTACT:0000015
name: Affiliation

[Term]
id: TCONTACT:0000016
name: Department
def: The department of an institution or organization.
relationship: part_of TCONTACT:0000015 ! Affiliation

[Term]
id: TCONTACT:0000017
name: Institution
relationship: part_of TCONTACT:0000015 ! Affiliation

[Term]
id: TCONTACT:0000018
name: Organization
def: A generic term for any organization.
relationship: part_of TCONTACT:0000015 ! Affiliation
```

dsenalik:
# Bug Fix

## Description
This renames the two duplicated terms to be unique, and renames a third term for consistency. See more details in linked issue #1906
Also, this file had DOS style line endings, so converted this. See commit https://github.com/tripal/tripal/pull/1907/commits/eb58f33de9e25cf4528f59afbf4e1976cebd03ed to more clearly see the actual changes.

## Testing?
Build a docker
Add a new contact
Enter "%Departmen" for the type, the autocomplete should offer
2024-06-19_newDepartment

Change to "%Organizatio"
2024-06-19_newOrganization


laceysanderson (commented):
I think it would also be good to also change/add a definition to make it clear that these are specifically related to the contact (i.e. properties). The other terms with the same name are intended to be the type of contact.

Affiliated Department: The department of the institution or organization that the contact is affiliated with.

Affiliated Institution: The employer (e.g. company, university, corporation) that the contact is affiliated with.

Affiliated Organization: The formal group of people the contact is affiliated with. Organizations differ from institutions by being more focused on people coming together for a specific purpose; where as, institutions usually refer to physical locations of a particular corporation.

spficklin:
Have we always been able to use the `%` character in an autocomplete? Does that get fed into the SQL query?  If so, that really makes me nervous.  I think that characters is benign but can you give it others and hack the query?

dsenalik:
That's a very good question @spficklin that I had not thought about.
You could do this in Tripal 3 also
2024-06-21_autocomplete

The good news is that the database abstraction layer will handle sanitization for a `condition()` method as long as the user input is not used to select the operation - 
[Writing secure code for Drupal / Use the database abstraction layer to avoid SQL injection attacks](https://www.drupal.org/docs/security-in-drupal/writing-secure-code-for-drupal#s-use-the-database-abstraction-layer-to-avoid-sql-injection-attacks)
There is an example there of how to prevent use of the wildcard character, but because it is highly useful for the autocomplete we would not want to do that.

Our code, where $name comes from the autocomplete input, is
```
    if (array_key_exists('exact', $options) and $options['exact'] === True) {
      $query1->condition('CVT.name', $name, '=');
    }
    else {
      $query1->condition('CVT.name', $name . '%', 'LIKE');
    }
```

laceysanderson:
I think based on the code that it is safe to continue allowing this for the autocompletes :-)

The updated definitions look perfect @dsenalik <3 
Do we also want an update hook to change these term names in existing chado instances?

dsenalik:
I did not think we would need an update hook, because if we do the Chado prepare step when migrating, it will change these terms.

laceysanderson (approved):
😮 So you are right -no need for an update hook for this one.

I finally found the time to test this. I created a docker on the 4.x branch and queried to confirm that I had the old version of the department term:

```
2956	tripal_contact	Department
2963	tripal_contact	Department(TCONTACT:0000016)
```

Then I switched to this branch and ran the prepare again. The prepare ran again without an error. Then I queried the terms again.

```
2956	tripal_contact	Department
2963	tripal_contact	Affiliated Department
```

PR #1912, opened 2024-06-20
ChadoStorage::deleteValues()

spficklin:
# Tripal 4 Core Dev Task

## Description

This PR completes the work for the `ChadoStorage::deleteValues()` function.  It supports removing data from Chado using the values of field properties and is currently designed to work when deleting an entity and the records in Chado need to be removed as well.   Simlar to the `findeValues()` it doesn't not work yet with any type of filtering (e.g. ranges, >=, etc.)

## Testing?

1. On a new site, create a contact entity
2. Create a project entity and associate the contact with it
3. If you want to compare unpublishing to deleting, feel free to unpublish the project. You should be able to publish it again and it will come back with as it was before with the contact linked to it.
4. Delete the project.
5. Clear the cache and reload the contact page. It should not have been deleted even though it was linked to the project.
6. Try to republish the project, it will not exist.
7. Run the following SQL statement to make sure the link in the `project_contact` table is gone: `select * from chado.project_contact`


dsenalik (changes requested):
I followed the described test procedure and it all worked. 🙌

However, I wanted to test further and encountered an error, which can be reproduced as follows:
1. create a contact
2. create a project - do not link the contact
3. edit the project and add the contact - it does not appear on the page, however it is in the linker table
```
sitedb=> select * from project_contact;
 project_contact_id | project_id | contact_id 
                  1 |          1 |          2
(1 row)
```
2024-06-22_nocontact

I am also seeing problems with some dbxrefs not appearing, might be related

UPDATE: This happens also on the 4.x branch 😬
UPDATE 2: ~~May have happened with PR #1892~~ If I checkout before that PR it works
`git checkout e8d74534681abe51c088c06e38e2e2599fa7c4e8 .` = works
`git checkout 75ab06d888c839ef34affc5e9cd19ca7e0f491a2 .` = works
`git checkout 96d15dd05a5341c06cb05bb7f83c5550fc166366 .` = does not work
So the guilty PR is #1882

spficklin:
I noticed this problem with the analysis field and posted an issue about it.  https://github.com/tripal/tripal/issues/1913 .  I didn't realize the contact was having a similar problem.  I think this problem must have been introduced before this PR.  This PR only updates ChadoStorage/ChadoRecords for deleting and didn't touch the code for adding values for fields.  

So, I think if this PR works for a delete then we can merge it and I can go back and try to figure out and post a separate PR for why we're seeing the behavior you describe.

dsenalik:
I finally found the bug. This is really a separate issue from anything in this PR, so see issue #1919  
Edit: I closed that issue because it duplicates #1913, and I was sure I had found a bug, but that code sections isn't even being called, so more investigation needed...

dsenalik (approved):
So whatever bugs I mention here are unrelated to this PR so changing status to Approved

laceysanderson (approved):
Code-wise this looks good to me :-) Thanks for the PR and testing @spficklin! 🎉 

This is ready to merge once the tests pass again (had to update as another PR was merged first).

PR #1914, opened 2024-06-20
Fix the "Remove" button on edit form on linker fields

### LINKED ISSUE #1913: Analysis field not working
### Tripal Version

4.x

### Drupal Version

10.x

### PostgreSQL Version

15.1

### PHP Version

8.1.2

### General Description

**Bug 1**. The analysis field does not create a record in the Drupal field table, but does create a record in Chado.  This creates a problem because it doesn't show up on the entity page (e.g. on a project page) after the user added it.  Then when the user tries to add it again because it didn't work the first time, there is an error that a unique constraint is violated (in the recent logs) and it can't update the entity.

### I have given Bug 2 its own issue: #1920 and copied all of the relevant comments there

~~**Bug 2.** When editing an entity, the "Remove" button will remove a linked entity (e.g. remove analysis from project) in the Drupal table, but does not delete it from the underlying chado linker table. If instead it is changed to `-- Select --` then it is removed from both the Drupal and Chado tables. This is because the "Remove" button removes the item from the form state so then Chado storage doesn't know it existed and doesn't know to delete it in Chado.~~

### Steps to reproduce Bug 1

1. Create an analysis entity
2. Create a project entity.
3. Edit the project entity and add the analysis to it.
4. The analysis field will not show up on the page.
5. Look in the Drupal field table, a record for it is missing.
6. Look in the Chado `project_analysis` table and there is a record for it.

### Error messages

```shell
PDOException: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "project_analysis_c1" DETAIL: Key (project_id, analysis_id)=(2, 1) already exists
```

### Screenshots

image

Bug 1 was somehow introduced by PR #1882
`git checkout e8d74534681abe51c088c06e38e2e2599fa7c4e8 .` = works
`git checkout 75ab06d888c839ef34affc5e9cd19ca7e0f491a2 .` = works
`git checkout 96d15dd05a5341c06cb05bb7f83c5550fc166366 .` = does not work

### I have given Bug 2 its own issue: #1920 and copied all of the relevant comments there

1. Create an analysis entity
2. Create a project entity and link the analysis when creating it.
3. Edit the project entity and click the "Remove" button next to the analysis, and save.
4. Check the `project_analysis` table,, the record is still there

dsenalik:
# Bug Fix

## Description

When you have a field with cardinality more than one, in the edit form there is a "Remove" button on the right.
The bug is that if you use the Remove button on an existing item, when you save, the underlying chado table is not updated, because this item was removed from the form state.

I fixed this by removing the primary key from the conditions in the `deleteRecords()` function when used on a linker table

This is submitted as draft for discussion of whether my approach is suitable

## Testing?
1. Create two analyses
2. Create a project and link one analysis when creating
3. Edit the project and add the second analysis
4. Edit the project and click the "Remove" button next to the second analysis, and save
5. Edit the project and add the same second analysis
Before, step 5 would crash and burn

### DIFF
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -213,7 +213,7 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
       foreach ($base_tables as $base_table) {
         $tables = $this->records->getAncillaryTables($base_table);
         foreach ($tables as $table_alias) {
-          $this->records->deleteRecords($base_table, $table_alias, TRUE);
+          $this->records->deleteRecords($base_table, $table_alias, TRUE, TRUE);
         }
       }
 
--- a/tripal_chado/src/TripalStorage/ChadoRecords.php
+++ b/tripal_chado/src/TripalStorage/ChadoRecords.php
@@ -1813,10 +1813,14 @@ class ChadoRecords  {
    * @param bool $graceful
    *   Set to TRUE to not throw an exception if valid conditions are not
    *   set. If TRUE then it skips the record rather than performs the delete.
+   * @param bool $no_pkey
+   *   Set to TRUE to remove the primary key from the conditions.
+   *   This is used to remove all linker table records linked to a
+   *   base table record.
    *
    * @throws \Exception
    */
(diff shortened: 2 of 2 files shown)


dsenalik:
This may not be a valid approach. If you click "remove" on all of the records, then none get removed, because the form state didn't have any fields.

I think a different approach is needed

Issue #1637, opened 2023-09-15
Tripal 4 - Organism Image


Ferrisx4:
### Discussion or Question

In investigating the viability of the Display Suite and Field Group modules (#1615), I've attempted to add an image field to the Organism content type. This is a default field that we will want to use, so I've added it to the `tripal.tripalfield_collection.default_chado.yml` file.

In its current form, it simply adds a generic Drupal image field to the Organism content type, with all the usual Drupal image controls such as min/max resolution, alt text, allowed image types, and file directory, etc.
It's functional, but it doesn't directly use Chado. Past discussions have involved storing the image data itself within Chado or using the eimage table (as well as a linker table that will need to be created.
See also https://github.com/GMOD/Chado/issues/31

For now, this at least gets us a workable image field on Organisms. I'm open to discussion before I make a PR for it, which is why I'm using this issue type.

Ferrisx4:
Changes in branch https://github.com/tripal/tripal/tree/tv4g2-issue1637-t4_organism_image

Ferrisx4:
The validate() function in the TripalFieldCollection class checks if the storage_settings array exists and is not empty, so for the time being, this check is being somewhat bypassed by passing `storage_plugin_id: drupal`. This may or may not be a valid storage plugin, but the functionality to test for valid plugin id is still a @todo in that validate() function.
It is also passing 
```
termIdSpace: data
termAccession: 2968
```
although I am unsure if these are really being picked up by Tripal at all.

laceysanderson:
Thanks to a recent PR, you can now set it to `drupal_sql_storage` which is a valid TripalStorage plugin so that will help you test-wise in the future :-) 

I am also unsure whether the term + storage plugin information would be used/stored in a Drupal field since it doesn't have Tripal-style storage properties. This would definitely be interesting to investigate

Ferrisx4:
Ok that's good news about `drupal_sql_storage`! 

As for including term information, I imagine if we want to eventually store the image information in the chado.eimage table, we'd want to fully Tripal-ify it, as all the other Chado fields have that kind of information.

laceysanderson:
Definitely the question for investigation is:
Is setting the storage backend as DrupalSQLStorage enough to let the term information be stored or do we need to have a Tripal Wrapper for the Drupal Image field?

And the wider discussion about do we want to start supporting the eimage tables in chado through this field in which case a Tripal/Chado wrapper field is needed anyway. I don't feel strongly one way or the other for this part.

[...]

PR #1915, opened 2024-06-20
Adding Organism Image Field

spficklin:
# Tripal 4 Core Dev Task

## Description

Adds an orgnanism image field to the organism content type.  This PR adds the field which uses the Drupal image field and also adds it to the default layout of the organism.  

## Testing?

- On a new tripal site:
   1.  Create an oraganism and add as many images as you like.  Upon saving, you should see  the images on the page.  
   2.  Go to `Tripal` > `Page Structure` and click on "manage display" under the "organism' content type.  Click the 'Apply Tripal Layout' button.   Now, return to the organism page and the images should be in a new details box named "Images"
- On an existing tripal site
   1.  Pull the code from this branch
   2. Go to `Tripal` > `Configuration` and click on 'Tripal Field Collection'. 
   3. Delete the collection named "Default Chado Fields"
   4. Clear the cache
   5. Re-import the field collection by going to `Tripal` > `Page Structure`  and click "Import Type Collection".  Check the box labeled "General Content Types (Chado)".  Run the job. You should now have an organism image field.  
   6. Create or edit an oraganism and add as many images as you like.  Upon saving, you should see  the images on the page.  
   2.  Go to `Tripal` > `Page Structure` and click on "manage display" under the "organism' content type.  Click the 'Apply Tripal


spficklin:
Well dang. It's not passing tests.  Fixing now...

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
```diff
@@ -235,6 +234,58 @@ fields:
...
+        description: One or more images for the organism.
+        cardinality: -1
+        required: false
+        storage_settings:
+            storage_plugin_id: drupal_sql_storage
+            storage_plugin_settings: 
+                base_table: organism
+        settings:
+            termIdSpace: IAO
+            termAccession: "0000101"
```
dsenalik, line 250: Just a comment, I have been thinking for a while about creating a field for the `chado.eimage` table. On CarrotOmics I have a Tripal 3 field to do this and use this very term for the field. Would it be a conflict to use the same term as this field uses? (I now see there is much discussion about this in issue #1637)

dsenalik (changes requested):
I tested as described. I could add and remove images as expected.
A note, by accident I added the same file twice and this was allowed. I think this is fine though, and it's probably a Drupal thing anyway.

I did note something else. On a new site after applying layout, the organism appears as
2024-06-23_newsite

When I did the existing site test, I also tried applying the layout, and in this case the organism field is not inside a group.
2024-06-23_oldsite
I suspect the tripal layout YAML needs to be reloaded, so I wonder if this PR needs an update hook?

And tests are not passing 😆

spficklin:
Oh, good point about needing an update hook.   In the testing instructions I gave I walk you through reloading the YAML but we can't expect a user to do that on an existing site. I'll add an update hook.

Regarding the term used for this field.  I think we need to talk more on it.  We will probably want to use this term for any image for any content type.  I'm not sure if that's a problem or not. It is an image.  Does the term need to be more specific (e.g. organism image).  I think we need some conversation about that.

laceysanderson:
This PR is causing schema issues and brought to light the issue described in #1999. Since this PR (1) has very little code, (2) has been open for 1/2 a year and (3) is now dependant on the linked issue which is not easy to fix, I think it should just be closed at this point. We can revisit adding an organism field to these collections at a later point if there is enough demand.

PR #1917, opened 2024-06-22
Update TripalJobs to support callables instead of only functions.

### LINKED ISSUE #1916: Tripal Jobs should support callables as callbacks.
### Task Description

Currently Tripal jobs only support providing a function for the callback. This is very restrictive and requires creating random functions to call services, etc. Instead, we should support use of [callables](https://www.php.net/manual/en/language.types.callable.php) which call allow us to instead add the job setup to a static function within the class.

### Branch Name

tv4g3-issue1916-callableCallback

laceysanderson:
# Tripal 4 Core Dev Task

## Description

Currently Tripal jobs only support providing a function for the callback. This is very restrictive and requires creating random functions to call services, etc. Instead, we should support use of [callables](https://www.php.net/manual/en/language.types.callable.php) which call allow us to create a static function in the class that runs the Tripaljob.

Specifically, this PR
- Updates TripalJob load and create to is is_callable() as a check rather then function_exists()
- Serializes the callback to allow the array notation for callables to be used.
- Adds automated testing for TripalJobs specific to the callback.
- Updates TripalPublish to use a static method to run the TripalJob allowing the removal of the tripal_publish api file entirely.

NOTE: This PR is backwards compatible since functions are also callable ;-p 
NOTE: Since the callback is only serialized if it is array, any jobs created before this PR can still be run/re-run without issue.

### Backwards Compatibility
1. Create a fresh docker on the 4.x branch
2. Submit a tripal job (e.g. import content type collections) but do not run it.
3. Switch to this branch inside the docker and then run the previously submitted job.
4. Import any data using a tripal importer as these ones still use the old-style function based callback.

### Others
1. Submit a publishing job and ensure there are no issues with the job execution side of things.

### Automated testing

This PR includes an automated test that checks the 3 valid types of callables:
 - string-based static method: `'\Drupal\Tests\tripal\Kernel\Services\TripalJob\FakeClasses\callableClassForTripalJobs::myCallbackMethod'`
 - array-based static method: `[callableClassForTripalJobs::class, 'myCallbackMethod']`
 - simple function: `myFunctionCallback`

For each callable, the following is tested:
 - A Tripal Job can be created using the service
 - A tripal job can be loaded after it was created
 - The name, module, user, job_id, callback, arguments and job object call all be retrieved from the loaded job.
 - The job can be run without error.
 - A log message printed within the job is printed when the job is run.


tripal/src/Controller/TripalJobController.php
```diff
@@ -226,6 +226,11 @@ public function tripalJobsView($id) {
     // Get the submitter info.
     $submitter = \Drupal\user\Entity\User::load($job->getUID());
 
+    // If the callback is an array then formatting can make it more readible.
```
dsenalik:
```suggestion
    // If the callback is an array then formatting can make it more readable.
```

tripal/tests/src/Kernel/Services/TripalJob/CallablesJobTest.php
```diff
@@ -0,0 +1,163 @@
...
+    $this->assertEquals($details['uid'], $job->getUID(),
+      "Unable to retrieve the user of the job we just created.");
+    $this->assertEquals($details['job_name'], $job->getJobName(),
+      "Unable to retrieve the name of the job we just created.");
+    $this->assertEquals($details['modulename'], $job->getModuleName(),
+      "Unable to retrieve the module name of the job we just created.");
+    $this->assertEquals($details['callback'], $job->getCallback(),
+      "Unable to retrieve the callback of the job we just created.");
+    $this->assertEquals($details['arguments'], $job->getArguments(),
+      "Unable to retrieve the arguements of the job we just created.");
```
dsenalik:
```suggestion
      "Unable to retrieve the arguments of the job we just created.");
```

tripal/tests/src/Kernel/Services/TripalJob/CallablesJobTest.php
```diff
@@ -0,0 +1,163 @@
...
+      "Unable to return a job object from the job we just loaded.");
+
+    // Now we try to run it!
+    // If something goes wrong then an exception will be thrown.
+    // We use output buffering so we can condirm the log message within the
+    // callback is printed.
+    ob_start();
+    $job->run();
+    $output = ob_get_clean();
+    $this->assertStringContainsString('We were able to successfully run the job.',$output,
```
dsenalik:
```suggestion
    $this->assertStringContainsString('We were able to successfully run the job.', $output,
```

dsenalik (approved):
Nice!
 - [x] Switch to this branch inside the docker and then run the previously submitted job. = run was successful `Calling: import_tripalentitytype_collection(Array)`
 - [x] Import any data using a tripal importer as these ones still use the old-style function based callback. = imported gff file `Calling: tripal_run_importer(5)`
 - [x] Imported genomic content types `Calling: import_tripalentitytype_collection(Array)`
 - [x] Published gene and mRNA `Calling: Drupal\tripal\Services\TripalPublish::runTripalJob(gene, chado_storage, Array)`

A few picky typo corrections from code review 😼

laceysanderson:
Thanks for the review @dsenalik! I've added documentation for this in this PR https://github.com/tripal/tripal_doc/pull/84 and with your review, I now feel like I can merge this 😍

PR #1918, opened 2024-06-23
Synonym field was broken on Drupal ≥ 10.2

### LINKED ISSUE #1817: A few fields have errors when added to a newly defined content type
### Tripal Version

4.x

### Drupal Version

10.x

### PostgreSQL Version

13

### PHP Version

8.x

### Branch

tv4g1-issue1817-fields-on-new-content-types

### General Description

If you create a new content type manually through the GUI and add certain fields, you have problems.

### Steps to reproduce

1. Create a new content type through the UI (Admin > Tripal > Page Structure then "Add custom type")
I chose to create a "Primer" content type, so the base table will be feature, and most fields can be added to this. I used `primer (SO:0000112)` as the term.
2. Manage fields for this new content type (Admin > Tripal > Page Structure > NEWCONTENTTYPE > Manage Fields)
Click on "Create new field" and provide a name/machine name. Four fields are required to be able to save a primer entity
2024-03-16_minimal-fields
3. Create a primer entity (you will need to create an organism first)
4. Now add the "Chado Synonym" field
5. Edit your existing primer, and enter a synonym and save.
`TypeError: Drupal\tripal\TripalStorage\TripalStorageBase::addTypes(): Argument #2 ($types) must be of type array, null given, called in /var/www/drupal/web/modules/contrib/tripal/tripal/src/Entity/TripalEntity.php on line 423 in Drupal\tripal\TripalStorage\TripalStorageBase->addTypes() (line 115 of modules/contrib/tripal/tripal/src/TripalStorage/TripalStorageBase.php). `

6. to be continued if other fields have errors

dsenalik:
# Bug Fix

## Description
Since Drupal 10.2, the ChadoSynonymDefault field could not be added to a content type through the GUI, due to the subform introduced with this Drupal version.

One additional change:
`$storage_settings['storage_plugin_settings']['base_table'] = '';`
is present in the parent class, so I went ahead and removed it from all of the other fields too.

## Testing?
 1. Import genomic type collection
 2. Edit fields on "gene" and remove the existing synonym field
 3. Re-add it through the GUI. There should be no nasty errors now.
 4. Create a gene with a synonym and save it.


laceysanderson (approved):
✅ Code looks good
✅ Manual testing works as expected. Added both unlimited and limited single value versions of the synonym field and confirmed that content could be created and updated in both cases.

PR #1923, opened 2024-06-30
Set default for chado storage selector when there is only one option

### LINKED ISSUE #1922: Set default for chado storage selector when there is only one option

Yes

Yes

### Is this feature urgent?

No

### Description

This is a quality of life issue. When publishing, one needs to select the storage backend. But if there is only one, why not have it already be selected by default.

### Your specific use case

Every time I publish something 😁

### Generally Applicable

.

2024-06-30_PublishQOL

dsenalik:
# New Feature

## Description
One simple change for publishing. If there is only one storage backend, which will be the case for the majority of users, then you no longer have to select it from the list (which only contained one option)

## Testing?
 1. You don't even need a new docker or anything, you can just checkout this branch (assuming you only have one storage backend)
 2. Go to Tripal -> Content -> "+ Publish Tripal Content"
 3. You should now see the first two selectors pre-populated with values. Before you had to select the storage backend.
2024-06-30_default-now-set

### DIFF
--- a/tripal/src/Form/TripalEntityPublishForm.php
+++ b/tripal/src/Form/TripalEntityPublishForm.php
@@ -63,6 +63,14 @@ class TripalEntityPublishForm extends FormBase {
         'wrapper' => 'storage-options'
       ],
     ];
+
+    // If there is only one datastore available, set it as the default.
+    if (count($datastores) == 1) {
+      $datastore = array_key_first($datastores);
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
YAY! 😍 Yes, this has been bugging me too 🤪 

✅ Code changes are perfect
✅ Manual testing results in the pre-selected form as now expected for publish!

PR #1926, opened 2024-07-15
Deprecations for tripal_chado.query in branch tv4g2-issue1341-chadoQu…

pdtouch:
…eryDeprecations

Code changes to tripal_chado/src/api/tripal_chado.query.api.php incorporating deprecations to deprecated functions.

Tripal 4 Core Dev Task

## Testing?

When a developer uses the deprecated function,  a suitable notification including guideline to use the current alternative function in its place.


laceysanderson (approved):
Looks good! Sorry it took so long to get to this @pdtouch! You are good to go with this format for the other functions.

PR #1929, opened 2024-07-16
Resolve discovered field name length problems

### LINKED ISSUE #1928: Check for new fields fails if property name is too long
### Tripal Version

4.x

### Drupal Version

probably any

### PostgreSQL Version

probably any

### PHP Version

probably any

### General Description

Check for new fields fails if property name is too long

### Steps to reproduce

1. Create an organism with a property
```
sitedb=> INSERT INTO chado.organism (genus, species) VALUES ('Tripalus', 'databasica');
INSERT 0 1
sitedb=> INSERT INTO chado.organismprop (organism_id, type_id, value) VALUES
  ((SELECT organism_id FROM chado.organism WHERE species='databasica'),
  (SELECT cvterm_id FROM chado.cvterm WHERE name='mitochondrial_genetic_code_name'), 'test');
INSERT 0 1
```
2. Go to Tripal -> Page Structure
3. For "Organism" select "Manage Fields"
4. +Check for new fields. You should see this
2024-07-16_longproperty
5. Click "Add Fields" button. This error appears
2024-07-16_longpropertyerror
6. Check the db logs at `/admin/reports/dblog`, the error is
`ERROR: Error adding field, "organism_mitochondrial_genetic_code_name", to "organism": Attempt to create a field storage with an name longer than 32 characters: organism_mitochondrial_genetic_code_name`

### Notes
The error is generated by Drupal by this line in `core/modules/field/src/Entity/FieldStorageConfig.php`:357:
`throw new FieldException('Attempt to create a field storage with an name longer than ' . static::NAME_MAX_LENGTH . ' characters: ' . $this->getName());`

In `tripal/src/TripalField/TripalFieldItemBase.php` the maximum length seems to be too long, we could use that variable rather than hardcoding, but we may also need to check for uniqueness if we truncate the field name, e.g. there are two similar properties "mitochondrial_genetic_code" and "mitochondrial_genetic_code_name"
```
  /**
   * A helper function to create an appropriate field name.
   *
   * This function can be used by a field's `discover()` method to generate
   * an appropriate field name. It removes will always include the bundle
   * type as the prefix, followed by an underscore, followed by extra text
   * provided. It ensures that only alphanumeric values are present in the
   * name and that it doesn't exceed 50 characters.
   *
   * @param \Drupal\tripal\Entity\TripalEntityType TripalEntityType $bundle
   *   The TripalEntityType object with information about the bundle.
   * @param string $extra
   *   Extra text to add to the field name after the bundle name.
   * @return string
   *   The generated field name.
   */
  public static function generateFieldName(TripalEntityType $bundle, string $extra) {
    $field_name = strtolower($bundle->getID() . '_' . $extra);
    $field_name = preg_replace('/[^\w]/', '_', $field_name);
    $field_name = substr($field_name, 0, 50);
    return $field_name;
  }
```

Code to lookup the maximum length
```
use Drupal\field\Entity\FieldStorageConfig;
$max = FieldStorageConfig::NAME_MAX_LENGTH;
```

This is the Tripal 3 code that insured uniqueness for truncated field names, from `tripal_chado/includes/tripal_chado.fields.inc `
```
      // The field name can only be 32 chars, but if our name is longer we need
      // to add some random chars to ensure we don't have naming conflicts
      // with other terms (e.g. mitochondrial_genetic_code and
      // mitochondrial_genetic_code_name)
      if (strlen($field_name) >= 32) {
        $field_name = substr($field_name, 0, 20) . '_' . $term->cvterm_id;
      }
 ```

### PR DESCRIPTION
**Pull request by dsenalik**
# Bug Fix

## Description
Check for new fields fails if property name is too long.

Problems:
A value of 50 is used for truncation, but the actual maximum is 32.
With simple truncation, two fields might therefore get identical names, not good.

Here I get the actual maximum length from Drupal rather than hardcoding it.
In Tripal 3 we appended the cvterm_id to truncated field names to guarantee uniqueness.
There a fixed length of 20 characters was used, then underscore and then cvterm_id.
Here, I use the maximum length that can fit in 32 characters so that field names will be as long as possible.

If a cvterm_id value is not readily available, the PHP `uniqid()` function is used to generate the suffix for the field name.

## Testing?
I added an automated test for the truncation and even threw in a unicode character for good measure. This revealed that multibyte characters were being replaced with multiple underscores, fixed that with a `u` suffix on the regex.

For manual testing:
1. Create an organism with a property
```
sitedb=> INSERT INTO chado.organism (genus, species) VALUES ('Tripalus', 'databasica');
INSERT 0 1
sitedb=> INSERT INTO chado.organismprop (organism_id, type_id, value) VALUES
  ((SELECT organism_id FROM chado.organism WHERE species='databasica'),
  (SELECT cvterm_id FROM chado.cvterm WHERE name='mitochondrial_genetic_code_name'), 'test');
INSERT 0 1
```
2. Go to Tripal -> Page Structure
3. For "Organism" select "Manage Fields"
4. +Check for new fields. You should see this
2024-07-16_fixed
5. Click "Add Fields" button. There should be no errors.
2024-07-16_fixed2


laceysanderson (commented):
Code-wise this looks good.
My only comment is that the generateFieldName() method has had a parameter added which is not backwards compatible. I suggest making it optional by providing a default in the method header. Then if the cvterm_id is not passed in, use `uniqid()` instead. I do think the cvterm_id is better and am happy to highly recommend they don't leave that paramter empty in the doc block but this way no one looks up the cvterm if they don't have it. 

It may also be good to post a note that the generateFieldName() method may not generate the same name repeatedly for the same field if the cvterm is not passed in.

laceysanderson (approved):
Looks good to me!
✅ Code review
✅ Automated Testing
✅ Manual test

PR #1931, opened 2024-07-18
Remove double-space collapse for publish

### LINKED ISSUE #1930: Entity title token replacement if >1 space causes publish duplication
### Tripal Version

4.x

### Drupal Version

probably any

### PostgreSQL Version

probably any

### PHP Version

probably any

### General Description

Entity title token replacement is very basic, and if a field in the title is blank, you might then get a case where there are two consecutive spaces in the title. Publish will collapse these
https://github.com/tripal/tripal/blob/e01ea4881d55a92a91a552c9a7b585651e4b0047/tripal/src/Services/TripalPublish.php#L525
But when creating an entity manually, they are not collapsed. Because the title now differs by a space character, publish will re-publish as a second entity. This was introduced as a problem when we put `<em>` tags in the organism title to display in italics. Before, `trim()` would have made the title the same in both cases. But this may crop up in other cases, so we need to have the logic the same in both places.

### Steps to reproduce

1. Create an organism without any infraspecific type e.g _Daucus carota_
2. Publish organisms
3. Run the publish job
` [notice] Published 1 new entities, and 7 field values.`
4. You will now see two apparently identical entities.
2024-07-18_duplication

5. If you look in the `tripal_entity` table, you will see that they differ by a space. For example
```
sitedb=> select * from tripal_entity;
 id |   type   | uid |                       title                       | status |  created   |  changed   
----+----------+-----+---------------------------------------------------+--------+------------+------------
  1 | organism |   1 | <em>Daucus carota</em>  <em></em>                 |      1 | 1721142054 | 1721142054
  2 | organism |     | <em>Daucus carota</em> <em></em>                  |      1 | 1721311387 | 1721311387
```

### Another way to reproduce
1. Edit the contact title format to be `[contact_name] [contact_type] [contact_description]`
2. Enter a contact and enter a name and description, but do not enter a contact type
3. Publish contact
4. Observe two copies of the contact (plus the null contact)

dsenalik:
# Bug Fix

## Description

Publish included a replacement of two or more spaces in the entity title with a single space, in the well-intentioned goal of making things tidy.
Unfortunately, this creates a slightly different title for some entities in some cases, when a replaced field is empty and there is another field on each side. When this happens, a duplicate entity is published, because a manually entered entity does not have this, it retains both spaces.

## Changes in this PR

1. ~~The `preg_replace()` in the publish has been removed so that the title format will match that of a manually entered entity, and the double spaces remain in the entity title. Double spaces are usually collapsed in HTML so that a title `abc   def` will appear as `abc def` anyway, so it won't be a visually distracting difference.~~ Now uses token parser
2. Update the TripalTokenParser service to be less reliant on an entity for token replacement in order to allow it to be used by the publish service.
3. Add a static function, `getEntityTitle()` to the TripalTokenParser to make it really straight forward to generate a title from field values in order to ensure both the entity create/edit and the publish generate the titles in the exact same way and with the same code.
4. Add a simple function `getFieldValues()` to the TripalEntity class to make it easy to get the values of all fields for generic use with the TripalTokenParser and for other future APIs.

NOTE:

- TripalEntity still uses the old method to get the URLs which does do some preprocessing before passing things off to TripalTokenParser.
- We need additional automated tests for this functionality.

## Testing

1. Create an organism without any infraspecific type e.g _Daucus carota_
2. Confirm the entity title generation works and you get the title "*Daucus carota*" displayed on the page. There should not be any un-replaced tokens for the infraspecific name/type and it should be italicized.
3. Publish organisms
4. Run the publish job. On 4.x it will say the following but on this branch no new entities will be published.
` [notice] Published 1 new entities, and 7 field values.`
5. On 4.x you will now see two apparently identical entities but on this branch there will still be only the original one.
2024-07-18_duplication

6. If you look in the `tripal_entity` table on 4.x, you will see that there are two organisms that differ by a space (see below). On this branch there will only be a single entity for *Daucus carota*.
```
sitedb=> select * from tripal_entity;
 id |   type   | uid |                       title                       | status |  created   |  changed   
----+----------+-----+---------------------------------------------------+--------+------------+------------
  1 | organism |   1 | <em>Daucus carota</em>  <em></em>                 |      1 | 1721142054 | 1721142054
  2 | organism |     | <em>Daucus carota</em> <em></em>                  |      1 | 1721311387 | 1721311387
```
7. Unpublish the *Daucus carota* organism through the UI to remove the entity but not the record in chado. 
8. Now re-run the publish job for organisms and confirm that it picks up the un-published *Daucus carota* and creates a page for it.
9. View the page and ensure all the fields show up and that the title is as expected (see above).


laceysanderson (commented):
This change is simple and works as advertised. 

That said, I'm thinking that it would be better if we could call the entity method for generating the title so that it is done the same way in publish and manual. That way we can make it more intelligent in the future without more of these popping up :-)

dsenalik:
This should now be ready for final review. I added tests for the two new functions, calling them directly.

laceysanderson (approved):
✅ Code looks good
✅ Automated testing looks good
✅ Perfect code coverage 😍 
✅ Followed the manual testing instructions and it worked exactly as described!

Ready for merge once the tests pass again after updating the branch 😍 🎉

PR #1937, opened 2024-07-25
Support Chado Flyway Migrations

### LINKED ISSUE #1876: How are we supporting Chado flyway migrations?
## Background

Chado has changed it's model of managing updates to the schema. It now uses an open-source database migration system called [Flyway](https://flywaydb.org/). 

Steps to apply migrations with Flyway
1. Install software
2. Configure to work with your database
3. Run `flyway baseline` the first time using flyway with your database so that it can detect the state of your system.
4. Run `flyway migrate` to apply the migrations to your database.

Each "migration" is a small SQL script that is applies to the schema. There are a number of migrations already merged in the chado repository that Tripal users currently don't have access to: https://github.com/GMOD/Chado/tree/1.4/chado/migrations. Eventually these migrations could be compiled into a version 1.4 but with the slow rate of change in Chado it would be nice to have access to these sooner!

## Discussion

1. Do we want to support these migrations in Tripal? or just wait until 1.4?
2. If yes above, do we want to write our own code to apply migrations or get Tripal users to apply them on their own using Flyway -perhaps by providing documentation on how to configure and use it.

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR adds support to Tripal to manage chado migrations ourselves so that we don't add Flyway as an external migration and increase complexity for users. Users will still install Chado 1.3 as they do now. Once Chado 1.3 is installed, they can go to the Chado Schema Manager and apply any outstanding migrations as shown in the following screenshots.

ChadoSchemaManager

ChadoSchemaManager-ApplyMigrations

## Tasks

- [x] Store the Chado migrations in a good place in Tripal
- [x] Use Yaml to provide nice metadata for each migration for use by Tripal
- [x] Add a button to the Chado Schema Manager form to manage migrations
- [x] Develop a framework to check for migrations to be applied and display them to the user
- [x] Use a tripal job to actually apply the migrations and then update the UI to show the user that they were applied.
- [x] Ensure that the chado.chadoprop table with the version number is updated to reflect the migration applied.

## Testing?

For testing, please focus your testing on the UI and migration application. Any issues with Tripal's functioning on higher versions are beyond the scope of this PR so please file and issue so they can be fixed in another PR :-)

1. In a TripalDocker based on this branch, go to Tripal > Data Storage > Chado > Chado Schemas. You should see something like the first screenshot above.
2. Click on "Apply Migrations" and you should see the second screenshot above.
3. Click "Apply Migrations" again and a Tripal job should be submitted.
4. When you run the Tripal job it should apply the migrations. Ensure the job runs without error and all the migrations are listed.
5. Now using check the chado schema to ensure that the migrations were actually applied. You can do this using `psql --user docker sitedb` where the password is the same as the username. Using `\d+ chado.TABLENAME` you can see the structure of the table. For example, `\d+ chado.project` should show the project now has a type_id and checking contact_project should show there is a type_id and rank.
6. Next, go back to the "Manage Chado Schemas" page described in step 1. The schema version should now match that of the latest migration (i.e. 1.3.3.013) and the updated column should be different from the created column.
Screenshot 2024-07-29 at 2 26 25 PM

7. Additionally, the "Apply Migrations" button should be replaced with a "View Migrations" button.
8. Click the "View Migrations" button and ensure they all say "Successful"
image


laceysanderson:
I believe this PR is now ready for review! A large number of incompatibilities with higher versions of Chado were fixed in PR #2076 and tests there imply that Tripal should work with higher versions of Chado now.

This PR adds an interface to allow people to upgrade their chado without the need for Flyway and for Tripal to keep track of these updates.

For testing, please focus your testing on the UI and migration application. Any issues with Tripal's functioning on higher versions are beyond the scope of this PR so please file and issue so they can be fixed in another PR :-)

spficklin (approved):

dsenalik (approved):
I also briefly went through the testing and everything worked as expected.

laceysanderson:
Thank you both for the reviews!!! 😍

PR #1943, opened 2024-08-02
Make sure the contact widget/formatter know they are compatible with both fields

laceysanderson:
# Bug Fix

## Description

Simple fix that adds the Contact by role field to the supported fields for the contact widget + formatter.

## Testing?

This is such a small change I don't know that its worth the effort to test manually as it's not straightforward. 

Essentially if you create contact by role fields via the YAML without this fix then you see the following on the form manage and display manage pages. These fields are automatically disabled because Drupal doesn't know what compatible widget/formatters there are and the drop down to select one is empty. This does not happen if you add the field through the UI.

Screenshot 2024-08-02 at 10 36 25 AM

With this PR, you can add them via the YAML and they are correctly enabled with the supported contact widget/formatter.


dsenalik (approved):
I agree this is a simple clear fix!

laceysanderson:
Thanks for the quick review @dsenalik!

PR #1944, opened 2024-08-07
GFF3 importer now uses buddies

### LINKED ISSUE #1925: Improve performance of GFF3 importer
### Task Description

The chado API functions for creating new DBs, terms, dbxref, or properties are painfully slow.
For example, in the Taxonomy importer, creating a new property is taking 7+ seconds to delete and 17+ seconds to add a property, about 25 seconds per property, and about 6 are added for each taxonomy id ⏳.

~~I am thinking to create some functions in the ChadoImporterBase class, because making direct chado queries only takes a handful of milliseconds. What does anybody think about this?~~

See below, we need a "buddy"!

### Branch Name

tv4g10-issue1925-gff-importer-use-buddy

### PR #1944

dsenalik:
# New Feature

## Description
Remove old slow buggy cvterm API functions in the GFF3 importer in favor of new Chado Buddy service.

## Testing?
1. Create an organism
2. Create an analysis
3. Load the gff3 file from PR #1824
For default landmark type, I used `chromosome (SO:0000340)`
4. Run the job `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`

Note how fast it goes as compared to the 4.x branch. Creating a single cvterm is in the tens of millisecond range at most, as compared to the tens of seconds range before!


Ferrisx4 (approved):
I tested this with the provided GFF and also a 100MB GFF sample file. The load of the provided GFF was almost instant compared to the current 4.x branch which took around a minute with each cvterm taking longer than the entire load with the new changes.

For the 100MB GFF, load time went from 8.5 minutes to under 5 minutes which is also a significant increase in speed.

Everything looks good and it works as advertised.

laceysanderson (approved):
Did a code review and it all looks great ✅ That in combination with Sean's manual testing means this can be merged! Thank you @dsenalik --that much of a speed increase is amazing 🤩

PR #1949, opened 2024-08-09
don't add to render array if no value

dsenalik:
# Bug Fix

## Description
Publishing an organism on an imported Tripal 3 chado dump results in this error on every organism page
Deprecated function: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in Drupal\filter\Element\ProcessedText::preRenderText() (line 101 of core/modules/filter/src/Element/ProcessedText.php).

This PR checks for NULL or empty strings and doesn't pass them on to a render array.
This may or may not be an appropriate fix., I'm just saving my work here.

### To see the error, build a docker on the 4.x branch and do the following
1. Get to your psql prompt and do this
    `INSERT INTO organism (genus, species) VALUES ('Tripalus', 'bogusii');`
2. Publish organism content type
3. View that organism entity
2024-08-10_4 x


dsenalik:
### A summary of what I found.
The null vs. empty string gets loaded somewhere in Drupal, not Tripal.
When `postLoad()` is called it has already happened, manually entered organism gets empty string, published organism gets NULL.
2024-08-10_setValue

### There are (at least) three possible ways to fix this.
1. This PR - check for NULL in the formatter. Easy, works, and did I mention easy?
2. Modify the current publish SQL to include the `non-required-types` columns and set an empty string for text fields. I almost got that to work, but you need to know which are text and which are not. For integers you should set zero. Not sure about boolean etc.
3. Use the Drupal entity and I assume field API function to publish. This would be the "correct" thing to do, but has a significant performance downside. I timed creating a single entity, and it is around 1 second. And I don't have any idea how long adding the fields would take. Imaging publishing all the genes for one genome!

laceysanderson (approved):
✅ Code looks good
✅ Manual testing using the instructions does fix the issue

As such, I am approving this...

However, 🙈 I did want to mention that this only fixes 4 field formatters but theoretically the same problem exists in all the chado formatters as well 🙈 

Ideally we could override `public function viewElements(FieldItemListInterface $items, $langcode)` in [TripalFormatterBase](https://github.com/tripal/tripal/blob/5c4617d32573f6797a8123d9e08d58aa7a321262/tripal/src/TripalField/TripalFormatterBase.php) and filter out delta with empty elements there. Then all the formatters just need to have a call to the parent at the top of the method. However, that method doesn't allow us to change the $items, only to return $elements.

Instead we could add a helper to the base class to do this. So something like this in TripalFormatterBase:
```php
  public function removeEmptyItems(FieldItemListInterface $items, $langcode) {

    // Get main property name from the field :thinking:

    // now for each item, get the main property and check if it's empty.
    foreach($items as $delta => $item) {
      if ($item->get($main_property_name)->isEmpty()) {
        unset($items[$delta]);
      }
    }
    return $items;
  }
```

This would require adding an `isEmpty()` method to all the property types just like we did in PR #1952 to get the default value 🙈. Then each formatter would call this at the beginning before processing the items to generate the markup.

dsenalik:
Notes for working on this

All of the formatters except the four below inherit `ChadoFormatterBase` which itself inherits `TripalFormatterBase`
```
$ grep 'class ' *  | grep -v ChadoFormatterBase
ChadoBooleanFormatterDefault.php:class ChadoBooleanFormatterDefault extends DefaultTripalBooleanTypeFormatter {
ChadoIntegerFormatterDefault.php:class ChadoIntegerFormatterDefault extends DefaultTripalIntegerTypeFormatter {
ChadoStringFormatterDefault.php:class ChadoStringFormatterDefault extends DefaultTripalStringTypeFormatter {
ChadoTextFormatterDefault.php:class ChadoTextFormatterDefault extends DefaultTripalTextTypeFormatter {
```

laceysanderson:
We have determined in cofest that this is not a priority. The other fields are not reproducibly showing the behaviour this was added for and additionally, the PR #1952 fixes this same problem from the publishing side so we don't need to fix it 2X when we have enough work on our plates :-) Thanks for the fixes and discussion in cofest!

Issue #1910, opened 2024-06-19
Allow publishing in batches


dsenalik:
### Task Description

I have over 30k publications to publish, and am running out of memory. My docker has 2G but it is not enough, and I don't know how much is enough.
It would be great if we could set a maximum number to publish, and then just publish multiple times.
Assuming of course that this would reduce the memory requirement.
```
2024-06-19 11:04:36: Calling: tripal_publish(pub, chado_storage, Array)
Step  1 of 6: Find matching records... 
 [notice] Step  1 of 6: Find matching records... 
PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes) in
  /var/www/drupal/web/modules/contrib/tripal/tripal/src/TripalStorage/TripalStorageBase.php on line 222
```

### Branch Name

tv4g1-issue1910-partial-publish

dsenalik:
### Idea
To reduce memory usage, we add the ability to pass to `ChadoStorage.php` `findValues()` a list of base table record id's, and then in that function we only query for tables in that list, e.g. `... WHERE contact_id IN (1,2,3...)`
We can have a new function `findUnpublishedRecordIds()` and it will perform a lookup similar to `drupalEntityIdLookupCallback()` to get entity ids for base table records, and use that to generate a list of unpublished records.

dsenalik:
@laceysanderson suggested a while back that `findValues()` or maybe a new function returns a simple array of record ids rather than the memory-consuming objects that it currently does. Then we can take chunks of record ids and publish in batches.

PR #1952, opened 2024-08-12
fix publish to store placeholders of proper type

### LINKED ISSUE #1956: Empty entity title fields are not getting removed when publishing
### Tripal Version

4.x

### Drupal Version

10.3

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

Entity title when entering an organism through GUI is correct, but for publish includes the blank infraspecific name field

### Steps to reproduce

`INSERT INTO organism (genus, species) VALUES ('Tripalus', 'publish');`
then publish organism.

### Error messages

_No response_

### Screenshots

2024-08-26_publish-vs-gui

dsenalik:
# Bug Fix

## Description
This PR updates the publish function to 
1. no longer store NULLs in the drupal field tables
2. no longer add rows to the drupal field table when the field has any property where `drupal_store` is set to TRUE but there is no value for that property in Chado. This happens for linking fields or property fields.

3. It also adds a `getDefaultValue()` to the property type classes that specifies the appropriate type of value to store as a placeholder in the drupal field tables.

Publish now stores these placeholder values in the field tables, which matches exactly what is stored when you enter an entity through the GUI.

## Testing?
1. Get to your psql prompt and do this to create some things to publish
  ```sql
  INSERT INTO chado.organism (genus, species) VALUES ('Tripalus', 'bogusii');
  INSERT INTO chado.organism (genus, species, infraspecific_name, comment) VALUES ('Tripalus', 'fakerelli', 'luminarus', 'Adding a comment here for testing purposes');
  INSERT INTO chado.contact (name) VALUES ('notype');
  INSERT INTO chado.contact (name, type_id) VALUES ('hastype', 123);
  INSERT INTO chado.pub (title, uniquename, type_id) VALUES ('On the origin of species', 'UID1', (SELECT cvterm_id FROM chado.cvterm WHERE name='Book'));
  INSERT INTO chado.pubprop (pub_id, type_id, value) VALUES (2, (SELECT cvterm_id FROM chado.cvterm WHERE name='Author'), 'Charles Darwin');
  ```
2. Publish organism content type -there should be two organisms created.
3. View both organism pages, there should be no error, and the title should be as expected as the issue described in #1956 is fixed here. You should be able to edit and save them without error. On 4.x the title is incorrectly published as
2024-09-11_infraname
4. Enter another organism manually, just genus and species, for comparison.
5. Look at the field tables and make sure there are **no NULLs stored** and that the published entity matches the GUI-entered entity. All organisms should have rows in all but two of these tables where cardinality is unlimited:
`tripal_entity__organism_dbxref` `tripal_entity__organism_pub`

```
\pset null '<NULL>'
\pset pager 0

select * from tripal_entity__organism_abbreviation;
select * from tripal_entity__organism_comment;
select * from tripal_entity__organism_common_name;
select * from tripal_entity__organism_genus;
select * from tripal_entity__organism_infraspecific_name;
select * from tripal_entity__organism_infraspecific_type;
select * from tripal_entity__organism_species;
select * from tripal_entity__organism_dbxref;
select * from tripal_entity__organism_pub;
  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_abbreviation_record_id | organism_abbreviation_value 
----------+---------+-----------+-------------+----------+-------+---------------------------------+-----------------------------
 organism |       0 |         1 |           1 | und      |     0 |                               1 | 
 organism |       0 |         2 |           2 | und      |     0 |                               2 | 
 organism |       0 |         3 |           3 | und      |     0 |                               3 | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_comment_record_id | organism_comment_value 
----------+---------+-----------+-------------+----------+-------+----------------------------+------------------------
 organism |       0 |         1 |           1 | und      |     0 |                          1 | 
 organism |       0 |         2 |           2 | und      |     0 |                          2 | 
 organism |       0 |         3 |           3 | und      |     0 |                          3 | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_common_name_record_id | organism_common_name_value 
----------+---------+-----------+-------------+----------+-------+--------------------------------+----------------------------
 organism |       0 |         1 |           1 | und      |     0 |                              1 | 
 organism |       0 |         2 |           2 | und      |     0 |                              2 | 
 organism |       0 |         3 |           3 | und      |     0 |                              3 | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_genus_record_id | organism_genus_value 
----------+---------+-----------+-------------+----------+-------+--------------------------+----------------------
 organism |       0 |         1 |           1 | und      |     0 |                        1 | 
 organism |       0 |         2 |           2 | und      |     0 |                        2 | 
 organism |       0 |         3 |           3 | und      |     0 |                        3 | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_infraspecific_name_record_id | organism_infraspecific_name_value 
----------+---------+-----------+-------------+----------+-------+---------------------------------------+-----------------------------------
 organism |       0 |         1 |           1 | und      |     0 |                                     1 | 
 organism |       0 |         2 |           2 | und      |     0 |                                     2 | 
 organism |       0 |         3 |           3 | und      |     0 |                                     3 | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_infraspecific_type_record_id | organism_infraspecific_type_type_id | organism_infraspecific_type_term_name | organism_infraspecific_type_id_space | organism_infraspecific_type_accession 
----------+---------+-----------+-------------+----------+-------+---------------------------------------+-------------------------------------+---------------------------------------+--------------------------------------+---------------------------------------
 organism |       0 |         1 |           1 | und      |     0 |                                     1 |                                   0 |                                       |                                      | 
 organism |       0 |         2 |           2 | und      |     0 |                                     2 |                                   0 |                                       |                                      | 
 organism |       0 |         3 |           3 | und      |     0 |                                     3 |                                   0 |                                       |                                      | 
(3 rows)

  bundle  | deleted | entity_id | revision_id | langcode | delta | organism_species_record_id | organism_species_value 
----------+---------+-----------+-------------+----------+-------+----------------------------+------------------------
 organism |       0 |         1 |           1 | und      |     0 |                          1 | 
 organism |       0 |         2 |           2 | und      |     0 |                          2 | 
 organism |       0 |         3 |           3 | und      |     0 |                          3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | organism_dbxref_record_id | organism_dbxref_linker_id | organism_dbxref_link | organism_dbxref_dbxref_id | organism_dbxref_dbxref_db_id | organism_dbxref_dbxref_accession | organism_dbxref_dbxref_version | organism_dbxref_dbxref_description | organism_dbxref_dbxref_db_name | organism_dbxref_dbxref_db_description | organism_dbxref_dbxref_db_urlprefix | organism_dbxref_dbxref_db_url 
--------+---------+-----------+-------------+----------+-------+---------------------------+---------------------------+----------------------+---------------------------+------------------------------+----------------------------------+--------------------------------+------------------------------------+--------------------------------+---------------------------------------+-------------------------------------+-------------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | organism_pub_record_id | organism_pub_entity_id | organism_pub_linker_id | organism_pub_link | organism_pub_pub_id | organism_pub_pub_title | organism_pub_pub_volumetitle | organism_pub_pub_volume | organism_pub_pub_series_name | organism_pub_pub_issue | organism_pub_pub_pyear | organism_pub_pub_pages | organism_pub_pub_miniref | organism_pub_pub_uniquename | organism_pub_pub_type | organism_pub_pub_is_obsolete | organism_pub_pub_publisher | organism_pub_pub_pubplace 
--------+---------+-----------+-------------+----------+-------+------------------------+------------------------+------------------------+-------------------+---------------------+------------------------+------------------------------+-------------------------+------------------------------+------------------------+------------------------+------------------------+--------------------------+-----------------------------+-----------------------+------------------------------+----------------------------+---------------------------
(0 rows)
```
6. To test an integer field. Publish contact content type. (The pre-existing "null" contact will get pubished too)
7. Enter a contact through the GUI for comparison.
8. Verify the drupal tables and compare published vs. GUI entities.
```sql
\pset null '<NULL>'
\pset pager 0

select * from tripal_entity__contact_description;
select * from tripal_entity__contact_name;
select * from tripal_entity__contact_project;
select * from tripal_entity__contact_type;
 bundle  | deleted | entity_id | revision_id | langcode | delta | contact_description_record_id | contact_description_value 
---------+---------+-----------+-------------+----------+-------+-------------------------------+---------------------------
 contact |       0 |         4 |           4 | und      |     0 |                             1 | 
 contact |       0 |         5 |           5 | und      |     0 |                             2 | 
 contact |       0 |         6 |           6 | und      |     0 |                             3 | 
 contact |       0 |         7 |           7 | und      |     0 |                             4 | 
(4 rows)

 bundle  | deleted | entity_id | revision_id | langcode | delta | contact_name_record_id | contact_name_value 
---------+---------+-----------+-------------+----------+-------+------------------------+--------------------
 contact |       0 |         4 |           4 | und      |     0 |                      1 | 
 contact |       0 |         5 |           5 | und      |     0 |                      2 | 
 contact |       0 |         6 |           6 | und      |     0 |                      3 | 
 contact |       0 |         7 |           7 | und      |     0 |                      4 | 
(4 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | contact_project_record_id | contact_project_entity_id | contact_project_linker_id | contact_project_link | contact_project_project_id | contact_project_project_name | contact_project_project_description 
--------+---------+-----------+-------------+----------+-------+---------------------------+---------------------------+---------------------------+----------------------+----------------------------+------------------------------+-------------------------------------
(0 rows)

 bundle  | deleted | entity_id | revision_id | langcode | delta | contact_type_record_id | contact_type_type_id | contact_type_term_name | contact_type_id_space | contact_type_accession 
---------+---------+-----------+-------------+----------+-------+------------------------+----------------------+------------------------+-----------------------+------------------------
 contact |       0 |         4 |           4 | und      |     0 |                      1 |                    0 |                        |                       | 
 contact |       0 |         5 |           5 | und      |     0 |                      2 |                    0 |                        |                       | 
 contact |       0 |         6 |           6 | und      |     0 |                      3 |                    0 |                        |                       | 
 contact |       0 |         7 |           7 | und      |     0 |                      4 |                    0 |                        |                       | 
(4 rows)
```
9. To test a boolean field and property fields, publish the Publication content type.
10. Manually enter a publication for comparison.
11. Verify. Note that the default for a boolean field is `f`.
```
\pset null '<NULL>'
\pset pager 0

select * from tripal_entity__pub_abstract;
select * from tripal_entity__pub_authors;
select * from tripal_entity__pub_citation;
select * from tripal_entity__pub_dbxref;
select * from tripal_entity__pub_doi;
select * from tripal_entity__pub_eissn;
select * from tripal_entity__pub_elocation;
select * from tripal_entity__pub_is_obsolete;
select * from tripal_entity__pub_issn;
select * from tripal_entity__pub_issue;
select * from tripal_entity__pub_journal_abbreviation;
select * from tripal_entity__pub_journal_country;
select * from tripal_entity__pub_language;
select * from tripal_entity__pub_language_abbr;
select * from tripal_entity__pub_miniref;
select * from tripal_entity__pub_pages;
select * from tripal_entity__pub_pii;
select * from tripal_entity__pub_publication_date;
select * from tripal_entity__pub_publication_model;
select * from tripal_entity__pub_publication_type;
select * from tripal_entity__pub_publisher;
select * from tripal_entity__pub_pubplace;
select * from tripal_entity__pub_pyear;
select * from tripal_entity__pub_series_name;
select * from tripal_entity__pub_title;
select * from tripal_entity__pub_type;
select * from tripal_entity__pub_uniquename;
select * from tripal_entity__pub_url;
select * from tripal_entity__pub_volume;
select * from tripal_entity__pub_volumetitle;
select * from tripal_entity__study_pub;
 bundle | deleted | entity_id | revision_id | langcode | delta | pub_abstract_record_id | pub_abstract_prop_id | pub_abstract_linker_id | pub_abstract_value | pub_abstract_rank | pub_abstract_type_id 
--------+---------+-----------+-------------+----------+-------+------------------------+----------------------+------------------------+--------------------+-------------------+----------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_authors_record_id | pub_authors_prop_id | pub_authors_linker_id | pub_authors_value | pub_authors_rank | pub_authors_type_id 
--------+---------+-----------+-------------+----------+-------+-----------------------+---------------------+-----------------------+-------------------+------------------+---------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_citation_record_id | pub_citation_prop_id | pub_citation_linker_id | pub_citation_value | pub_citation_rank | pub_citation_type_id 
--------+---------+-----------+-------------+----------+-------+------------------------+----------------------+------------------------+--------------------+-------------------+----------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_dbxref_record_id | pub_dbxref_linker_id | pub_dbxref_link | pub_dbxref_dbxref_id | pub_dbxref_linker_is_current | pub_dbxref_dbxref_db_id | pub_dbxref_dbxref_accession | pub_dbxref_dbxref_version | pub_dbxref_dbxref_description | pub_dbxref_dbxref_db_name | pub_dbxref_dbxref_db_description | pub_dbxref_dbxref_db_urlprefix | pub_dbxref_dbxref_db_url 
--------+---------+-----------+-------------+----------+-------+----------------------+----------------------+-----------------+----------------------+------------------------------+-------------------------+-----------------------------+---------------------------+-------------------------------+---------------------------+----------------------------------+--------------------------------+--------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_doi_record_id | pub_doi_prop_id | pub_doi_linker_id | pub_doi_value | pub_doi_rank | pub_doi_type_id 
--------+---------+-----------+-------------+----------+-------+-------------------+-----------------+-------------------+---------------+--------------+-----------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_eissn_record_id | pub_eissn_prop_id | pub_eissn_linker_id | pub_eissn_value | pub_eissn_rank | pub_eissn_type_id 
--------+---------+-----------+-------------+----------+-------+---------------------+-------------------+---------------------+-----------------+----------------+-------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_elocation_record_id | pub_elocation_prop_id | pub_elocation_linker_id | pub_elocation_value | pub_elocation_rank | pub_elocation_type_id 
--------+---------+-----------+-------------+----------+-------+-------------------------+-----------------------+-------------------------+---------------------+--------------------+-----------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_is_obsolete_record_id | pub_is_obsolete_value 
--------+---------+-----------+-------------+----------+-------+---------------------------+-----------------------
 pub    |       0 |         8 |           8 | und      |     0 |                         1 | f
 pub    |       0 |         9 |           9 | und      |     0 |                         2 | f
 pub    |       0 |        10 |          10 | und      |     0 |                         3 | f
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_issn_record_id | pub_issn_prop_id | pub_issn_linker_id | pub_issn_value | pub_issn_rank | pub_issn_type_id 
--------+---------+-----------+-------------+----------+-------+--------------------+------------------+--------------------+----------------+---------------+------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_issue_record_id | pub_issue_value 
--------+---------+-----------+-------------+----------+-------+---------------------+-----------------
 pub    |       0 |         8 |           8 | und      |     0 |                   1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                   2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                   3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_journal_abbreviation_record_id | pub_journal_abbreviation_prop_id | pub_journal_abbreviation_linker_id | pub_journal_abbreviation_value | pub_journal_abbreviation_rank | pub_journal_abbreviation_type_id 
--------+---------+-----------+-------------+----------+-------+------------------------------------+----------------------------------+------------------------------------+--------------------------------+-------------------------------+----------------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_journal_country_record_id | pub_journal_country_prop_id | pub_journal_country_linker_id | pub_journal_country_value | pub_journal_country_rank | pub_journal_country_type_id 
--------+---------+-----------+-------------+----------+-------+-------------------------------+-----------------------------+-------------------------------+---------------------------+--------------------------+-----------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_language_record_id | pub_language_prop_id | pub_language_linker_id | pub_language_value | pub_language_rank | pub_language_type_id 
--------+---------+-----------+-------------+----------+-------+------------------------+----------------------+------------------------+--------------------+-------------------+----------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_language_abbr_record_id | pub_language_abbr_prop_id | pub_language_abbr_linker_id | pub_language_abbr_value | pub_language_abbr_rank | pub_language_abbr_type_id 
--------+---------+-----------+-------------+----------+-------+-----------------------------+---------------------------+-----------------------------+-------------------------+------------------------+---------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_miniref_record_id | pub_miniref_value 
--------+---------+-----------+-------------+----------+-------+-----------------------+-------------------
 pub    |       0 |         8 |           8 | und      |     0 |                     1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                     2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                     3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_pages_record_id | pub_pages_value 
--------+---------+-----------+-------------+----------+-------+---------------------+-----------------
 pub    |       0 |         8 |           8 | und      |     0 |                   1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                   2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                   3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_pii_record_id | pub_pii_prop_id | pub_pii_linker_id | pub_pii_value | pub_pii_rank | pub_pii_type_id 
--------+---------+-----------+-------------+----------+-------+-------------------+-----------------+-------------------+---------------+--------------+-----------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_publication_date_record_id | pub_publication_date_prop_id | pub_publication_date_linker_id | pub_publication_date_value | pub_publication_date_rank | pub_publication_date_type_id 
--------+---------+-----------+-------------+----------+-------+--------------------------------+------------------------------+--------------------------------+----------------------------+---------------------------+------------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_publication_model_record_id | pub_publication_model_prop_id | pub_publication_model_linker_id | pub_publication_model_value | pub_publication_model_rank | pub_publication_model_type_id 
--------+---------+-----------+-------------+----------+-------+---------------------------------+-------------------------------+---------------------------------+-----------------------------+----------------------------+-------------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_publication_type_record_id | pub_publication_type_prop_id | pub_publication_type_linker_id | pub_publication_type_value | pub_publication_type_rank | pub_publication_type_type_id 
--------+---------+-----------+-------------+----------+-------+--------------------------------+------------------------------+--------------------------------+----------------------------+---------------------------+------------------------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_publisher_record_id | pub_publisher_value 
--------+---------+-----------+-------------+----------+-------+-------------------------+---------------------
 pub    |       0 |         8 |           8 | und      |     0 |                       1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                       2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                       3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_pubplace_record_id | pub_pubplace_value 
--------+---------+-----------+-------------+----------+-------+------------------------+--------------------
 pub    |       0 |         8 |           8 | und      |     0 |                      1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                      2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                      3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_pyear_record_id | pub_pyear_value 
--------+---------+-----------+-------------+----------+-------+---------------------+-----------------
 pub    |       0 |         8 |           8 | und      |     0 |                   1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                   2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                   3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_series_name_record_id | pub_series_name_value 
--------+---------+-----------+-------------+----------+-------+---------------------------+-----------------------
 pub    |       0 |         8 |           8 | und      |     0 |                         1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                         2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                         3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_title_record_id | pub_title_value 
--------+---------+-----------+-------------+----------+-------+---------------------+-----------------
 pub    |       0 |         8 |           8 | und      |     0 |                   1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                   2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                   3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_type_record_id | pub_type_type_id | pub_type_term_name | pub_type_id_space | pub_type_accession 
--------+---------+-----------+-------------+----------+-------+--------------------+------------------+--------------------+-------------------+--------------------
 pub    |       0 |         8 |           8 | und      |     0 |                  1 |                0 |                    |                   | 
 pub    |       0 |         9 |           9 | und      |     0 |                  2 |                0 |                    |                   | 
 pub    |       0 |        10 |          10 | und      |     0 |                  3 |                0 |                    |                   | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_uniquename_record_id | pub_uniquename_value 
--------+---------+-----------+-------------+----------+-------+--------------------------+----------------------
 pub    |       0 |         8 |           8 | und      |     0 |                        1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                        2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                        3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_url_record_id | pub_url_prop_id | pub_url_linker_id | pub_url_value | pub_url_rank | pub_url_type_id 
--------+---------+-----------+-------------+----------+-------+-------------------+-----------------+-------------------+---------------+--------------+-----------------
(0 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_volume_record_id | pub_volume_value 
--------+---------+-----------+-------------+----------+-------+----------------------+------------------
 pub    |       0 |         8 |           8 | und      |     0 |                    1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                    2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                    3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | pub_volumetitle_record_id | pub_volumetitle_value 
--------+---------+-----------+-------------+----------+-------+---------------------------+-----------------------
 pub    |       0 |         8 |           8 | und      |     0 |                         1 | 
 pub    |       0 |         9 |           9 | und      |     0 |                         2 | 
 pub    |       0 |        10 |          10 | und      |     0 |                         3 | 
(3 rows)

 bundle | deleted | entity_id | revision_id | langcode | delta | study_pub_record_id | study_pub_entity_id | study_pub_pub_id | study_pub_pub_title | study_pub_pub_volumetitle | study_pub_pub_volume | study_pub_pub_series_name | study_pub_pub_issue | study_pub_pub_pyear | study_pub_pub_pages | study_pub_pub_miniref | study_pub_pub_uniquename | study_pub_pub_type | study_pub_pub_is_obsolete | study_pub_pub_publisher | study_pub_pub_pubplace 
--------+---------+-----------+-------------+----------+-------+---------------------+---------------------+------------------+---------------------+---------------------------+----------------------+---------------------------+---------------------+---------------------+---------------------+-----------------------+--------------------------+--------------------+---------------------------+-------------------------+------------------------
(0 rows)
```
12. We don't have datetime or float fields yet so we can't test those yet 🙈
13. See this comment for how I tested that this does not break an existing site with published content. https://github.com/tripal/tripal/pull/1952#issuecomment-2288898016


dsenalik:
We're back to draft status because the new "default values" propagate through to the formatters. I should have anticipated this since the nulls getting through was the original problem.
2024-08-12_arraydesign

dsenalik:
This is because publish is creating rows in the field table when there is no value defined. In the GUI, if you leave one of these blank, then there will not be a row in the field table. Publish is creating a row, and the default value in that row  (the zero) displays because no value was defined in chado that could replace it.
The solution is to match what happens in the GUI, and not create a field table row when no value is defined in chado.

dsenalik:
### I wanted to test that this PR will not break an existing site.
On a 4.x docker I entered a contact and array design
```
insert into teacup.contact (name, type_id, description) values ('Borg', 42, 'You will be assimilated');
insert into teacup.arraydesign (name, manufacturer_id, platformtype_id, description) values ('AD1', 2, 42, 'Test one');
```
and published those.

Then I switched to this branch and added two more
```
insert into teacup.contact (name, type_id, description) values ('Dr. Peaks', 43, 'Science rules');
insert into teacup.arraydesign (name, manufacturer_id, platformtype_id, description) values ('AD2', 3, 43, 'Test two');
```
and again published.

We look at the drupal field table and see that the records entered while on the 4.x branch have NULL while the record from this branch has the correct empty value ''. There are two contacts on the 4.x branch because the null contact was published also.
```
sitedb=> \pset null '<null>'
sitedb=> select * from tripal_entity__contact_description;
 bundle  | deleted | entity_id | revision_id | langcode | delta | contact_description_record_id | contact_description_value 
---------+---------+-----------+-------------+----------+-------+-------------------------------+---------------------------
 contact |       0 |         1 |           1 | und      |     0 |                             1 | <null>
 contact |       0 |         2 |           1 | und      |     0 |                             2 | <null>
 contact |       0 |         4 |           1 | und      |     0 |                             3 | 
(3 rows)
```

The same for the arraydesign table:
```
sitedb=> select * from tripal_entity__array_design_description;
    bundle    | deleted | entity_id | revision_id | langcode | delta | array_design_description_record_id | array_design_description_value 
--------------+---------+-----------+-------------+----------+-------+------------------------------------+--------------------------------
 array_design |       0 |         3 |           1 | und      |     0 |                                  2 | <null>
 array_design |       0 |         5 |           1 | und      |     0 |                                  3 | 
(2 rows)
```

This next table shows that after this PR we don't add a row to the table if there is no underlying chado value. The value here was from the first array design on 4.x. On this branch a row was not inserted.
```
sitedb=> select * from tripal_entity__array_design_num_array_columns;
    bundle    | deleted | entity_id | revision_id | langcode | delta | array_design_num_array_columns_record_id | array_design_num_array_columns_value 
--------------+---------+-----------+-------------+----------+-------+------------------------------------------+--------------------------------------
 array_design |       0 |         3 |           1 | und      |     0 |                                        2 |                               <null>
(1 row)
```


### And Finally
This also fixes the bug where we have a bunch of field headers for fields that don't have a value! 
### If published on the 4.x branch
20240814_4 x

### If published on this branch

laceysanderson (changes requested):
So I tested the organism as suggested but received unexpected behaviour 🙈 
1. There was no record in tripal_entity__organism_comment for Tripalus bogusii. This makes sense if by "no longer add rows when the underlying Chado record is empty" in the description you meant "doesn't add a record to the drupal table for a field that doesn't have a value in the linked chado record" but doesn't match what you said to expect in the testing instructions.
2. When you go to the published page, the title replacement no longer works as expected 🙈 Specifically, the title is `Tripalus bogusii [organism_infraspecific_type] [organism_infraspecific_name]`. This is likely a side effect of the fact that the tripal_entity__organism_infraspecific_name table no longer has an record for the infraspecific name whereas before it did but it was empty.
  image

Note: Interestingly when I edit the organism and then save again without making changes, it fixes the title but does not add a record to the tripal_entity__organism_infraspecific_name or tripal_entity__organism_comment tables 🤔

Note: I did also enter an organism with a comment and can confirm that when the comment is present, the organism_comment_value column is now an empty string as expected --no null :-)

tripal/src/TripalStorage/Interfaces/TripalStorageInterface.php
```diff
@@ -105,7 +105,16 @@ public function getFieldDefinition(string $field_name);
   public function getStoredTypes();
 
   /**
-   * Returns a list of property values for stored types..
+   * Returns a list of property types that should not be stored.
+   * This is the complement function of getStoredTypes.
```
laceysanderson: Make it more clear that by unstored here we mean that the value is not stored in the Drupal field tables. The property type does still have a column in the drupal field tables which is why we need to know this.


```suggestion
   * Returns a list of property types whose value should not be stored in Drupal
   * field tables.
   *
   * This is the inverse of the getStoredTypes() method. It's needed because all
   * field property types have a column in the Drupal table. However, the "Stored"
   * ones save the value both in Drupal and in the storage backend and the
   * "unstored" ones save an empty value in Drupal that is populated on load by
   * the storage backend.
```

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -115,36 +115,61 @@ public function addFieldDefinition(string $field_name, object $field_definition)
    * {@inheritDoc}
    * @see \Drupal\tripal\TripalStorage\Interfaces\TripalStorageInterface::getStoredTypes()
    */
-  public function getStoredTypes() {
+  public function getStoredTypes(bool $non_required = FALSE) {
```
laceysanderson: This change to parameters should not be happening as it does not match the interface... Actually I'm not sure why PHP is not screaming at us for that!

```suggestion
  public function getStoredTypes() {
```

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -115,36 +115,61 @@ public function addFieldDefinition(string $field_name, object $field_definition)
...
+   * @see \Drupal\tripal\TripalStorage\Interfaces\TripalStorageInterface::getNonStoredTypes()
+   */
+  public function getNonStoredTypes(bool $non_required = FALSE) {
+    return $this->getStoredTypesFilter(FALSE);
+  }
+
+  /**
+   * Helper function for getStoredTypes() and getNonStoredTypes().
+   *
+   * @param book $required
```
laceysanderson:
```suggestion
   * @param bool $required
```

tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -115,36 +115,61 @@ public function addFieldDefinition(string $field_name, object $field_definition)
...
+  public function getStoredTypes(bool $non_required = FALSE) {
+    return $this->getStoredTypesFilter(TRUE);
+  }
+
+  /**
+   *
+   * {@inheritDoc}
+   * @see \Drupal\tripal\TripalStorage\Interfaces\TripalStorageInterface::getNonStoredTypes()
+   */
+  public function getNonStoredTypes(bool $non_required = FALSE) {
```
laceysanderson:
```suggestion
  public function getNonStoredTypes() {
```

tripal/src/Services/TripalPublish.php
```diff
@@ -485,6 +493,59 @@ protected function addFixedTypeValues(&$search_values) {
     }
   }
 
+  /**
+   * Adds to the search values array any remaining property values.
+   *
+   * @param array $seach_values
+   */
+  protected function addNonRequiredValues(&$search_values) {
```
laceysanderson, line 478: This method feels much more complicated then it needs to be :thinking: but it took me a couple of reads to see why. I think we just need to add a helper method that checks if a field is supported :-) That helper can then be used both here and in the addRequiredValues() method which reduces duplicated code and makes the logic in both these methods simpler.

With this change we check if the field is supported before looping through the properties and we don't need any checks for supportedness in the loop. We can also move defining $field_definition and $field_class outside the property loop.

Something like this:
```php
protected function checkFieldIsSupported(string $field_name) {

  // This property may be part of a field which has already been marked
  // as unsupported. If so then it won't be in the field_info and we
  // should skip it.
  if (!array_key_exists($field_name, $this->field_info)) {
    // Add it to the list of unsupported fields just in case
    // it wasn't added before...
    $this->unsupported_fields[$field_name] = $field_name;
    return FALSE;
  }

  return TRUE;
}
```

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
```diff
@@ -414,27 +414,27 @@ public function testChadoTripalPublishService() {
...
 
     $this->checkFieldItem('organism', 'organism_abbreviation', 1,
         ['record_id' => $organism_id2],
-        ['bundle' => 'organism', 'entity_id' => 2, 'value' => NULL]);
+        ['bundle' => 'organism', 'entity_id' => 2, 'value' => '']);
 
     $this->checkFieldItem('organism', 'organism_infraspecific_name', 0,
         ['record_id' => $organism_id2],
-        ['bundle' => 'organism', 'entity_id' => 2, 'value' => NULL]);
+        ['bundle' => 'organism', 'entity_id' => 2, 'value' => '']);
```
laceysanderson, line 428: Since the infraspecific name and type don't have a value in chado, shouldn't they not have an entry in the drupal field table? I suspect this will also fail when the token replacement issue gets fixed and if it doesn't then it probably should... shouldn't it?

tripal_chado/tests/src/Kernel/Services/TripalEntityLookupServiceTest.php
```diff
@@ -224,16 +225,17 @@ public function testTripalEntityLookupService() {
     // this uses the fallback entity lookup function getDefaultBundle(),
     // because the term for manufacturer_id is NOT a content type.
     // For the fallback lookup we need to also pass the base table.
+    // Note that we don't depend on the order that contacts are published in.
```
laceysanderson, line 228: Why has this PR changed this? Could we not depend on order before and we just got lucky?

laceysanderson (commented):
Ok code review completed :-)

Also, after a cofest discussion, I confirmed that the following issues from my manual test are not caused by this PR:

> 1. There was no record in tripal_entity__organism_comment for Tripalus bogusii.

This was just outdated testing instructions 🤪 They have been updated.

> 2. When you go to the published page, the title replacement no longer works as expected

This is being fixed in issue #1956 and was not caused by this PR. This PR does exacerbate the issue though as now both type and name are not replaced in the title whereas before this PR it was only name.

tripal_chado/tests/src/Kernel/Services/TripalEntityLookupServiceTest.php
dsenalik: I am pretty sure it was just that we got lucky 🙈

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
dsenalik: It is correct that they should not have an entry in the drupal field table.
Since we are passing `0` for $num_expected, this is testing that this is indeed the case 😀

laceysanderson: Ahhh! A comment to that effect would be really helpful!

tripal/src/Services/TripalPublish.php
dsenalik: Updated with commit https://github.com/tripal/tripal/pull/1952/commits/0eea7d250468a4350a29a1297b1bf4b6f121ed76

tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
dsenalik: Comment added with commit https://github.com/tripal/tripal/pull/1952/commits/99d2724377c24622ac1b7a858394fed13bf6d98c

tripal_chado/tests/src/Kernel/Services/TripalEntityLookupServiceTest.php
dsenalik: This is an interesting question perhaps - should publish order be predictable?
There is so much going on inside publish and in chado storage that I think it would be fair to never assume an order when testing if we publish more than one entity at a time. 
This is perhaps related to issue #1910 where I want to be able to publish in smaller batches. That may turn out to be one of two things: 1. easy; 2. insanely difficult - I am voting for number 2 😫

dsenalik:
Here is something I did not understand. For fields with unlimited cardinality, when an entity is created through the GUI, if there is no record, then the drupal field table does not have a row. That's how this PR works now.

HOWEVER, if the field is cardinality 1, then the drupal field table DOES have a row!
The symptom @Lacey found is that if you create this:
`INSERT INTO organism (genus, species) VALUES ('Tripalus', 'bugii');`
and publish, you cannot edit this and add a comment, common name, or abbreviation.

dsenalik:
Looking up cardinality is easy, but the problem that is happening in this PR is that, in the `insertFieldItems()` function, the `$matches` array no longer has records for fields that don't have a value in chado. This is because after adding the "not required" items to the initial search query, these fields are coming back with a NULL for their value. The whole match record then is removed by `isFieldValid()` and at this point
https://github.com/tripal/tripal/blob/3ad2f09fa281ec66f338defb53699f744b453ae4/tripal/src/Services/TripalPublish.php#L794-L795
there is nothing there for that field. 
In 4.x, there _was_ a record in the $match, although it was empty except for the record_id.

laceysanderson:
> no longer add rows to the drupal field table when the underlying Chado record has an empty value for that field

I am seeing quite a few bugs related to this in my production site 🤔 Perhaps we shouldn't do this. Rather if there is a chado record for that field then there is a record in drupal even if it is empty.

dsenalik:
I can see how this might make better sense. We would have to figure out how to handle integer fields, though, that have zero as an empty value, which caused the problem in this comment https://github.com/tripal/tripal/pull/1952#issuecomment-2285026682 - although we could back to storing NULLs now that the formatter won't break from those.

@laceysanderson did you mean your production site problems are on 4.x or testing on this branch? If we abandon this whole thing, well that's okay.

laceysanderson:
> @laceysanderson did you mean your production site problems are on 4.x or testing on this branch? If we abandon this whole thing, well that's okay.

Ah yes, I should have been more clear. I have tested this branch on a clone of my production site and I think there were still issues though at this point I'm having trouble determining which issues were on 4.x and which were still on this branch 🤪 I do not want to abandon this all together as I think many of the changes are great! I just think we should go back to always having a drupal record if there is a chado record.

> I can see how this might make better sense. We would have to figure out how to handle integer fields, though, that have zero as an empty value, which caused the problem in this comment https://github.com/tripal/tripal/pull/1952#issuecomment-2285026682 - although we could back to storing NULLs now that the formatter won't break from those.

Ah yes... I think we should leave a default of 0 for integer fields and create a new issue for this. I think I would be tempted to solve it with an option in the formatter for integer fields that lets you indicate if you want to hide entries with a value of 0 or not 🤔 but there may be a better option.

dsenalik:
> I just think we should go back to always having a drupal record if there is a chado record.

I just want to clarify this, as this is already the case. The difficulty is that if there is a drupal record when there is NOT a chado record (for multiple cardinality fields), that's when a value "bleeds through".

dsenalik:
With commit https://github.com/tripal/tripal/pull/1952/commits/6ade77fc5b6bc9a14ee08201d9d77ee4d7b77e30 I have made two significant changes
1. Lookup cardinality. Any field with cardinality 1 is _always_ added to the drupal field tables. Otherwise, it is only added if the chado value is not empty. This should exactly match the behavior when adding an entity through the gui.
2. Bypass the `isFieldValid()` check for publish only. This hopefully fixes what I described in comment https://github.com/tripal/tripal/pull/1952#issuecomment-2318619049 - but I have yet to test thoroughly.

I will now go back and update the testing procedure yet again as I test this ...

dsenalik:
Something to check, what about this NULL in a property field:
`select * from tripal_entity__pub_abstract;`
```sql
 bundle | deleted | entity_id | revision_id | langcode | delta | pub_abstract_record_id | pub_abstract_prop_id | pub_abstract_linker_id | pub_abstract_value | pub_abstract_rank | pub_abstract_type_id 
--------+---------+-----------+-------------+----------+-------+------------------------+----------------------+------------------------+--------------------+-------------------+----------------------
 pub    |       0 |         6 |           1 | und      |     0 |                      1 |               <NULL> |                      1 |                    |                 0 |                    0
(1 row)
```
and several others, authors, citation, doi...

If you enter a pub with gui and the field is blank, no row is created in the drupal field table for these, even though cardinality is 1 😖

With commit https://github.com/tripal/tripal/pull/1952/commits/6ade77fc5b6bc9a14ee08201d9d77ee4d7b77e30 I am checking for cardinality 1 fields where the main property is NULL as opposed to empty, and for these do not create a drupal field table entry.

dsenalik:
With commit https://github.com/tripal/tripal/pull/1952/commits/b2f0fbf92f5db25c48ba78230b9f0c95e0617131 I have abandoned using cardinality. Now, add to drupal tables except for the case where there is a property with `drupal_store` set to TRUE, but there is not a corresponding value. This magically handles everything and I am getting the exact same results as entities entered through the GUI except for `revision_id`.

"revision id for an unversioned entity type is the same as the entity id"

dsenalik:
I have also fixed `revision_id` to match how it works for entities entered through the GUI, and updated testing instructions, so my work here is done 🤞

laceysanderson (approved):
✅ Code changes look good
✅ Manually tested on a clone of my production site and it works beautifully with our project-based grant, study and experiment pages. It correctly published a complex set of multiple property types, contact relationships, and dbxref associations. It did not break or interact badly with existing content either 😌 
✅ Manually tested according to the instructions on a fresh Tripal docker and all looked as expected!

Thanks for all the hard work @dsenalik!!! This is ready to merge!!! 🎉

PR #1951, opened 2024-08-10
Improve Tripaldocker: support VSCode devcontainers + single dockerfile with args approach

laceysanderson:
# Tripal 4 Core Dev Task

## Description

It would be good to improve the tripaldocker in a number of ways. This PR is currently specific to the following:

 - [x] Have supervisord actually start postgresql on run of the docker container.
 - [x] Add a .devcontainer VSCode configuration file to help people setup development environments configured with Drupal standards and helpful add-ons.
 - [x] Explore using ARGs before the FROM clause to reduce our multiple dockerfiles into a single one for better devcontainer and windows support
 - [x] Print out a helpful message on run with commands, branding and login information.
 - [x] Create a base tripaldocker-drupal image that handles the os and drupal install so that the tripaldocker is much faster to build
 - [x] Update the build docker workflows including scheduled updates

### Checking docker build and supervisord setup

1. Clone this repo and checkout this branch.
2. Build the docker image using the single Dockerfile :-) Try this with different versions of php (i.e. 8.1, 8.2, 8.3), Drupal (i.e. 10.2.x-dev, 10.3.x-dev, 10.3.2), and PostgreSQL (i.e. 13, 14, 15, 16). This should complete without error.
```
docker build --tag=tripaldocker:localdocker \
            --build-arg phpversion="8.3" \
            --build-arg drupalversion="10.3.x-dev" \
            --build-arg postgresqlversion="16" \
            --build-arg chadoschema='teapot' ./
```
3. Once you have an image, create a container and confirm that it (a) has the versions expected based on the parameters to the image, (b) that both apache and postgresql start on their own without having to run the service command!
```
docker run --publish=80:80 --name=tripaldocker -tid tripaldocker:localdocker
```

### Test out using the configured devcontainer with VSCode

1. Clone this repo and checkout this branch.
2. Open the directory in VSCode.
3. Ensure you have the Microsoft Dev Containers extension by going to Extensions in VSCode and searching for `ms-vscode-remote.remote-containers`.
4. Then click the DevContainers status icon in the bottom left corner. It will look like the green double arrow square in the following screenshot. This will pop open the command palette where you want to choose "Reopen in Container". This will build a docker image and start a container!

Screenshot 2024-08-10 at 5 00 55 PM
[...]

PR #1954, opened 2024-08-19
Fix failing tests in the version grid 🙈 

laceysanderson:
# Tripal 4 Core Dev Task
## Description

In the recent PR #1951 I forgot to update the grid workflows not to use named dockerfiles. This PR fixes that. It also adds testing for 10.4 because, why not?

## Testing?

I added workflow dispatch to allow us to dispatch the grid workflows on this branch to confirm all works as it should.


dsenalik (approved):
Looks good!

PR #1953, opened 2024-08-16
Support Drupal 11.0.x + 11.1.x

dsenalik:
# New Feature

## Description
Changes to support Drupal 11.0.x-dev and 11.1.x-dev.

~~The docker has been updated to not attempt to install tripal_layout on Drupal 11 since we are waiting on https://www.drupal.org/project/field_group_table/issues/3430507. We are no longer trying to get Tripal Layout D11 compatible in this PR as that issue has been RTBC for months 😒 Instead we are focusing on the other Tripal Core modules so we don't get further behind 🙈~~

Field Group Table has finally been made D11 compliant! This PR thus intends to bring all Tripal modules to be compatible with D11 :-)

Status: All automated tests are passing, warnings have been dealt with and deprecation notices have been addressed. However, a thorough testing of a D11 site through the UI has not yet been done.
Basic testing for D11 was done in comment https://github.com/tripal/tripal/pull/1953#issuecomment-2479180184

## Testing?
1. Build a docker on this branch, e.g. 
```
docker build \
  --tag=tripaldocker:testing-1953 \
  --build-arg drupalversion="11.0.x-dev" \
  --build-arg postgresqlversion="16" \
  --build-arg chadoschema="chado" \
  --progress=plain \
  ./
git='where you have this checked out'
docker run \
  --publish=8080:80 -tid \
  --name=1898 \
  --volume=$git:/var/www/drupal9/web/modules/contrib/tripal \
  tripaldocker:testing-1953
```
2. Go to your site and start doing stuff and find out what is broken,
    e.g. it looks like autocomplete is not working for infraspecific type.

3. Make sure the javascript fix from commit https://github.com/tripal/tripal/pull/1953/commits/3ad8feb7897052fb5672f3bdf47331f6e54e5555 
  works on both Drupal 11.0 and 10.x, you just need to go to the GFF3 importer
  and upload a file, you don't need to actually run the importer.

## Tasks
- [x] **Fix autocomplete**
[...]

PR #1955, opened 2024-08-19
Build tripaldocker on all the versions so extensions can better test.

laceysanderson:
# Tripal 4 Core Dev Task

## Description

Build tripaldocker on all the versions so extensions can better test... we had previously only built on a small subset but with the expanded drupal base, we should also expand the tripaldocker as much as supported too so that extension packages can get on all the versions too.

## Testing?

Automated testing only. Just make sure the docker versions build without error.


dsenalik (approved):
Nice. And I see you are getting ready for Drupal 11 here, so just mentioning PR #1953 so that it is linked

PR #1921, opened 2024-06-27
Revert one change from PR #1882

dsenalik:
# Bug Fix

## Description
This reverts a single change that causes the issue described in issue #1913 - You can't edit an entity and add a linked record.

~~I am submitting this as draft until testing reveals that there are no unintended side effects.~~

## Testing?
 1. Create an analysis entity
 2. Create a project entity. Do not include the analysis yet.
 3. Edit the project entity and add the analysis to it.
 4. The analysis field will not show up on the page before this PR, it will show up on this branch.
 5. Look in the Drupal field table, a record for it is missing (the bug before this PR) - or the record is present (intended behavior, only on this branch)
```
sitedb=> select * from tripal_entity__project_analysis;
 bundle  | deleted | entity_id | revision_id | langcode | delta | project_analysis_record_id | project_analysis_entity_id | project_analysi...
---------+---------+-----------+-------------+----------+-------+----------------------------+----------------------------+----------------...
 project |       0 |         1 |           1 | und      |     0 |                          1 |                          4 |                ...
(1 row)
```
 6. Look in the Chado `project_analysis` table and there is a record for it (intended behavior, happens before and after this branch)
```
 sitedb=> select * from project_analysis;
 project_analysis_id | project_id | analysis_id | rank 
---------------------+------------+-------------+------
                   1 |          1 |           1 |    1
(1 row)
```
 7. Let's test publishing too
```
INSERT INTO chado.project (name, description) VALUES ('Published Project', 'For Testing');
INSERT INTO chado.contact (name, description) VALUES ('Published Contact', 'Also For Testing');
INSERT INTO chado.project_contact (project_id, contact_id) VALUES ((SELECT project_id FROM chado.project WHERE name='Published Project'), (SELECT contact_id FROM chado.contact WHERE name='Published Contact'));
```
Then publish both project and contact and make sure the linked entity is present on the page. For example
2024-06-30_PublishedProject

Issue #1932, opened 2024-07-22
Enforce that two different fields cannot have the same cv term


dsenalik:
### Task Description

We should probably make sure that we never have a situation where two fields on an entity share the same cv term.

We probably can add this to validation (`tripal/src/TripalField/TripalFieldItemBase.php`) when manually adding a field to a content type - hopefully this is easy, but you never know...

### Branch Name

tv4g1-issue1932-enforce-fields-unique-cvterm

### Idea
To look up all currently used CV terms for fields on a content type:
```
$bundle = 'project';
/** @var \Drupal\Core\Field\FieldTypePluginManager $field_type_manager **/
$field_type_manager = \Drupal::service('plugin.manager.field.field_type');
/** @var \Drupal\Core\Entity\EntityFieldManager $entity_field_manager **/
$entity_field_manager = \Drupal::service('entity_field.manager');

$all_field_defs = $field_type_manager->getDefinitions();
$entity_field_defs = $entity_field_manager->getFieldDefinitions('tripal_entity', $bundle);
foreach ($entity_field_defs as $id => $field_config) {
  if (is_object($field_config) and ($field_config instanceof Drupal\field\Entity\FieldConfig)) {
    $settings = $field_config->getSettings();
    $term_id_space = $settings['termIdSpace'] ?? '';
    $term_accession = $settings['termAccession'] ?? '';
    print "$id = $term_id_space:$term_accession\n";
  }
}
```
About 10 milliseconds

PR #1958, opened 2024-08-27
Fix the "Remove" button on field widgets

### LINKED ISSUE #1920: The "Remove" button on an entity edit does not remove a linked record from Chado
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

I am transferring Bug 2 from issue #1913 to here, because it is unrelated, and so should be a separate issue. I will transfer my comments here.

When editing an entity, the "Remove" button will remove a linked entity (e.g. remove analysis from project) in the Drupal table, but does not delete it from the underlying chado linker table. If instead it is changed to -- Select -- then it is removed from both the Drupal and Chado tables. This is because the "Remove" button removes the item from the form state so then Chado storage doesn't know it existed and doesn't know to delete it in Chado.

### Steps to reproduce

 1. Create a contact entity
 2. Create a project entity and link the contact when creating it.
 3. Edit the project entity and click the "Remove" button next to the contact.
2024-06-29_removebutton
 4. After the ajax has removed the contact, click the "Save" button. The Contact appears to be removed from the project.
 5. Check the `project_contact` table, the record is still there.

### Branch
`tv4g1-issue1920-fix-remove-button`

dsenalik:
# Bug Fix

## Description
When editing an entity, the "Remove" button will remove a linked entity (e.g. remove analysis from project) in the Drupal table, but does not delete it from the underlying chado linker table. If instead it is changed to -- Select -- then it is removed from both the Drupal and Chado tables. This is because the "Remove" button removes the item from the form state so then Chado storage doesn't know it existed and doesn't know to delete it in Chado.

To resolve this problem, I am storing a few values in the form state on the very first creation of the form,
and then in the `massageFormValues()` function, we can check for records that were there when the form was first built, but have been eliminated since. If that is the case, then put a reference in the `$values` array which will inform chado storage to delete the record in chado.

This problem will exist only for fields which support cardinality greater than one. Fields that need the fix:
 - [x] ChadoAnalysisWidgetDefault
 - [x] ChadoArrayDesignWidgetDefault
 - [x] ChadoAssayWidgetDefault
 - [x] ChadoBiomaterialWidgetDefault
 - [x] ChadoContactWidgetDefault
 - [x] ChadoFeatureMapWidgetDefault
 - [x] ChadoFeatureWidgetDefault
 - [x] ChadoOrganismWidgetDefault
 - [x] ChadoProjectWidgetDefault
 - [x] ChadoProtocolWidgetDefault
 - [x] ChadoPubWidgetDefault
 - [x] ChadoStockWidgetDefault
 - [x] ChadoStudyWidgetDefault

These will probably be different
 - [x] ChadoDbxrefWidgetDefault
 - [x] ChadoPropertySelectWidgetDefault
 - [x] ChadoPropertyStringWidgetDefault
 - [x] ChadoPropertyWidgetDefault
 - [x] ChadoSynonymWidgetDefault

Changes not needed because cardinality is always 1
 - ChadoAdditionalTypeWidgetDefault 

## Testing?
~~(We need PR #1921 merged into this branch to test)~~ merged
1. Build a docker on the 4.x branch (to test update hook)
2. Import Genetic and Genomic type collections and run the job `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
3. switch to this branch `git checkout tv4g1-issue1920-fix-remove-button`
4. Run update hook `drush updatedb`
```
>  [notice] Update started: tripal_chado_update_10406
>  [notice] Update completed: tripal_chado_update_10406
>  [notice] Message: Updating properties for Drupal table "tripal_entity__qtl_synonym", field  
> "qtl_synonym"
> 
>  [notice] Message: Updating properties for Drupal table  
> "tripal_entity__sequence_variant_synonym", field "sequence_variant_synonym"
> 
>  [notice] Message: Updating properties for Drupal table "tripal_entity__genetic_marker_synonym",  
> field "genetic_marker_synonym"
> 
>  [notice] Message: Updating properties for Drupal table  
> "tripal_entity__phenotypic_marker_synonym", field "phenotypic_marker_synonym"
> 
>  [notice] Message: Updating properties for Drupal table "tripal_entity__gene_synonym", field  
> "gene_synonym"
> 
>  [notice] Message: Updating properties for Drupal table "tripal_entity__mrna_synonym", field  
> "mrna_synonym"
> 
>  [notice] Message: Updating properties for Drupal table "tripal_entity__dna_library_synonym",  
> field "dna_library_synonym"
> 
>  [notice] Message: Updated 7 properties
> 
```
5. Create three contacts
6. Create an organism
7. Create a "Genetic Marker", and a) link all of the contacts, b) add some "Database Reference Annotations", c) add some synonyms, and save.

8. Edit the Genetic Marker you created. Click the "Remove" button for one of the contacts, select "-- Select --" for another, then save.

9. Contacts deleted using either of these methods should no longer appear on the entity.

10. The two contacts should also be removed from the `project_contact` table and only one should remain - This is what was **not** happening before.
```
sitedb=> SELECT * FROM chado.feature_contact;
 feature_contact_id | feature_id | contact_id 
                  7 |          1 |          4
(1 row)
```
11. Edit again and do the same for synonyms, click "Remove" on one, empty the synonym box for another and save. Check the `feature_synonym` table.
12. And likewise for the "Database Reference Annotations". Check the `feature_dbxref` table.
13. To test properties, import a gff3 file, for example [issue1824.gff3.txt](https://github.com/tripal/tripal/files/15109905/issue1824.gff3.txt)
14. Add new fields to "Gene"
15. publish Gene
16. Test removing some properties, you can examine with
`select * from featureprop FP left join cvterm T on FP.type_id=T.cvterm_id where FP.feature_id = (select feature_id from feature where name='PAU8');`


dsenalik:
Commit https://github.com/tripal/tripal/pull/1958/commits/dfa92b5379d8825d60376e3f1f5a1b5b579961f1 changes the internal identifier I was using from the field cv term to the field machine name. The latter is guaranteed to be unique, the former likely is, but issue #1932 is there because it is not currently enforced.

laceysanderson (approved):
✅ Code looks good. There is an assumption that there will be a record_id and a field_name element in the widget but I just documented it.
✅ Manual testing was successful. I tested the following situations:
  - deleted the last/only linker record
  - deleted by emptying the element
  - deleted by clicking the remove button
  - removed multiple entries in different fields in the same save
  - removed an entry with the remove button and then added one in the same page load
  - created multiple property fields with the same values and removed some values from each field but not the same ones to ensure there was no cross over. Checked where the value was the same and checked delta-wise as well.
  - checked on genetic marker and study
  - checked multiple different fields
  - confirmed after each save that the database underneath matched my expectations
  - confirmed nothing removed came back after publish

PR #1962, opened 2024-09-03
Fix dbxref "Add another item" bug

### LINKED ISSUE #1961: Dbxref "Add another item" bug
### Tripal Version

4.x

### Drupal Version

10.3, 10.2

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

The dbxref field widget includes an ajax callback because it contains two parts, a DB selector, and an accession box.
When you add a second item, the accession for the first one disappears.
A workaround is to click "Add another Item" required number of times first, then select DBs.

### Steps to reproduce

 1. Tripal -> content -> + Add tripal content
 2. Click on "Analysis"
 3. Scroll to the bottom to "Database Reference Annotations" and select any DB
 4. Enter any accession
2024-09-03_error1

 5. Now click "Add another item"
 
The accession you entered disappears and the box is disabled.
2024-09-03_error2

### What gives?
When the "Add another item" ajax triggers, there is nothing in the item values so this `$item_vals = $items[$delta]->getValue();` gets back an empty array. This then triggers the field to be disabled, because `$db_id` is not defined, which in turn removes the value.

dsenalik:
# Bug Fix

## Description
The dbxref field widget includes an ajax callback because it contains two parts, a DB selector, and an accession box.
The form element for the accession is disabled until a DB is selected. This ajax callback then enables the element only if `$db_id` has a value.

The "Add another item" button ajax callback (part of Drupal) doesn't have access to the item values,
`$item_vals = $items[$delta]->getValue();` returns an empty array during this ajax callback. This causes the existing accession fields to be set to disabled, since they think that `$db_id` is not yet set, and if they are disabled, any value that they previously contained disappears.

The resolution is to store the values of `$db_id` in the form state when we have it, and retrieve it when we don't.

## Testing?
Follow the steps described in issue #1961 
For the last step, the accession should now remain!


laceysanderson (changes requested):
## Observation 1
The autocomplete shows `NCIT:C164386` for Database Accession but when you edit after saving the Database Accession is reloaded as `C164386`.
Screenshot 2024-09-11 at 5 08 57 PM

## Observation 2
The "Add another item" ajax and the "Remove" button are not playing together nicely 😭 
1. Create an analysis and add multiple database reference annotations (I added 2) and click save.
2. Now edit the same page and add two more database reference annotations.
3. Then click the remove button beside one of the original 2 entries. It seems to work... so click SAVE.
  Screenshot 2024-09-11 at 5 08 57 PM
4. See errors 😢  although the correct dbxrefs are listed... if you check analysis_dbxref you will see all 4 entries are still listed...
  Screenshot 2024-09-11 at 5 09 46 PM
  Screenshot 2024-09-11 at 5 10 20 PM

dsenalik:
Observation 1 is fixed so that the database name is there on edit.
Observation 2 is also fixed, it stems from the fact that the "Remove" button changes the delta indices - so I needed to **not** save the `db_id` if the triggering element was the remove button, since that caused the values to propagate down to the next delta element.
You can test observation 2 more easily by just:
1. create two dbxrefs and save
2. edit, and click the remove button, and save
3. confirm there is no record in the chado table for the removed xref.

laceysanderson:
So... I cloned tripal, checked out this branch and built a docker and it appears that some of the recent merges may have broken this again 🙈 

More specifically, when I create a new analysis the ajax is once again clearing the accession field when I click the add more button. It is doing that again on edit as well.

dsenalik:
I don't know how I did that, as my logic was quite wrong.
It should be fixed now 🙏

laceysanderson (approved):
Works perfectly now! Thanks @dsenalik ❤️

Issue #1966, opened 2024-09-06
Tripal (not chado) fields can't be added


dsenalik:
### Tripal Version

4.x

### Drupal Version

10.2, 10.3

### PostgreSQL Version

13, 16

### PHP Version

8.3

### General Description

I can't add a tripal field 🙈

### Steps to reproduce

Tripal -> Page Structure -> Project -> Manage Fields
+Create a new field -> Tripal Fields -> Continue
Label enter anything
Check any of the four field types -> Continue
WSOD as shown below

### Error messages

The website encountered an unexpected error. Try again later.

Exception: Cannot create a StorageProperty object for entity type "tripal_entity", field type "tripal_integer_type", key "value" without a properly formatted term: ":" in Drupal\tripal\TripalStorage\StoragePropertyBase->__construct() (line 83 of modules/contrib/tripal/tripal/src/TripalStorage/StoragePropertyBase.php).

```
Drupal\tripal\TripalStorage\StoragePropertyTypeBase->__construct('tripal_entity', 'tripal_integer_type', 'value', ':', 'int', Array) (Line: 32)
Drupal\tripal\TripalStorage\IntStoragePropertyType->__construct('tripal_entity', 'tripal_integer_type', 'value', ':') (Line: 37)
Drupal\tripal\Plugin\Field\FieldType\TripalIntegerTypeItem::tripalTypes(Object) (Line: 401)
Drupal\tripal\TripalField\TripalFieldItemBase::propertyDefinitions(Object) (Line: 794)
Drupal\field\Entity\FieldStorageConfig->getPropertyDefinitions() (Line: 65)
Drupal\Core\Field\TypedData\FieldItemDataDefinition->getPropertyDefinitions() (Line: 64)
Drupal\Core\Field\FieldItemBase->__construct(Object, 0, Object) (Line: 47)
Drupal\Core\TypedData\TypedData::createInstance(Object, 0, Object) (Line: 99)
Drupal\Core\TypedData\TypedDataManager->createInstance('field_item:tripal_integer_type', Array) (Line: 108)
Drupal\Core\TypedData\TypedDataManager->create(Object, NULL, 0, Object) (Line: 208)
Drupal\Core\TypedData\TypedDataManager->getPropertyInstance(Object, 0, NULL) (Line: 100)
Drupal\Core\Field\FieldTypePluginManager->createFieldItem(Object, 0, NULL) (Line: 41)
[...]

PR #1967, opened 2024-09-07
Provide validation for already-used field terms

### LINKED ISSUE #1932: Enforce that two different fields cannot have the same cv term

dsenalik:
# New Feature

## Description
Currently there is no mechanism to prevent you from adding a field to a content type, and using a CV term that happens to already be in use by another field. This PR adds a check to the validation when adding a field.

## Testing?
~~We would like to include non-chado fields in the test, but currently issue #1966 prevents that. We will make a small change to allow us to test with a generic tripal field.~~ Fixed with PR #1958
~~1. edit `tripal/src/Plugin/Field/FieldType/TripalIntegerTypeItem.php` and after line number 34 add this line
    `if (!$termIdSpace) { return; }`
    i.e.~~
```
     $termIdSpace = $storage_settings['termIdSpace'];
     $termAccession = $storage_settings['termAccession'];
+    if (!$termIdSpace) { return; }
```
2. Go to Tripal -> Page Structure -> Contact -> Manage fields
3. +Create a new field
4. Tripal Fields -> Continue
5. For Label enter "tripal integer" -> select "Tripal Integer Field Type" -> Continue
6. For the Controlled Vocabulary Term, enter any term you like, I used 'Sequence length (data:1249)' -> Save Settings
7. Now we test! Click "+Create a new field"
8. Chado Fields -> Continue
9. Enter a Label, like "test" or something more interesting.
10. Select something we can add to contact, I used "Chado Integer Field Type" -> Continue
11. For "Table Column" I used `contact_id`
12. Enter a Controlled Vocabulary Term that is already in use, try the one you created at step 6
13. Observe a validation error.
issue1932-1
issue1932-2
14. Try the other existing terms on the contact content type, all of these should fail validation: 
```
Communication Contact (NCIT:C47954)
description (schema:description)
name (schema:name)
Project (NCIT:C47885)
```
15. Try any other term, e.g. `DNA (SO:0000352)`, validation should pass, and the field should be added successfully.
issue1932-3
16. To test the other validation that I fixed, add another field (I added another project field) and enter something not of the correct format as a term, e.g. `xxx`. The validation should not allow this and should correctly highlight the field in red. You can test this part on 4.x and see that the validation catches this, but does not highlight the field in red.
issue1932-4
issue1932-5

This is the appearance on 4.x:
issue1932-6


laceysanderson (changes requested):
I confirmed that this works when creating a new Tripal String field and when creating a new chado field. I ensured that it was specific to terms being used on the same content type and not terms on another content type. I checked the validation of an incorrectly entered term.

I did find one issue though... you now trigger a validation error when trying to edit any field. For example,
1. Go to manage fields for the organism content type.
2. Click edit beside the organism common name field.
3. Then click save (the error is triggered whether you make a change or not)
4. Observe the "term already being used" validation error... even though the field using the term is this exact field 🙈 
Screenshot 2024-09-21 at 3 35 15 PM

dsenalik:
Oops, sorry I didn't test that 😬! Should be fixed now.

Ferrisx4:
I just tested this as well, I have not encountered any issues. I also confirmed that the issue @laceysanderson when editing fields has also been fixed as of https://github.com/tripal/tripal/pull/1967/commits/78121d829efb974f2fab426d03d5e7d1379908b3

laceysanderson (approved):
Switching to approved based on a code review of the change in https://github.com/tripal/tripal/commit/78121d829efb974f2fab426d03d5e7d1379908b3 and @Ferrisx4 updated testing.

PR #1971, opened 2024-09-14
configurable mechanism for hiding fields when they have an "empty" value

### LINKED ISSUE #1970: Empty Values for Boolean + Integer
### Tripal Version

4.x-dev

### Drupal Version

_No response_

### PostgreSQL Version

_No response_

### PHP Version

_No response_

### General Description

The Tripal boolean field, if it is not set to TRUE, does not display on the entity page. We have a display setting (tested that too) to set what to display for each TRUE and FALSE. I feel setting the false value to an empty string should cause it to be hidden, but otherwise it should display 'false' or whatever.
issue-1966-works

For reference, the chado boolean field works is able to display both true and false and has the same settings to change the display.
The chado boolean field works like this
chadoboolean

For the settings form here, you currently can't set it to display an empty string
chadoboolean2

However, it might be better to allow support for hiding these fields in the case of an "empty value"... In https://github.com/tripal/tripal/pull/1952#issuecomment-2327320716 it came up that we need to think about what the "default" value of the correct type should be for boolean or integer fields that are required. This was a hard question because FALSE and 0 are also valid true values 🤔 Maybe configuration is how we let users control this?

### Steps to reproduce

1. Add a TripalBoolean field to any content type.
2. Create a page of that content type and select false for that field.

If it is set to false it will be removed from the page but if set to true then it will show up.

### Error messages

_No response_

### Screenshots

_No response_

dsenalik:
# New Feature

## Description
This PR provides a configurable mechanism for hiding fields when they have an "empty" value.
For text fields this would generally be an empty string, for integer fields it might be zero or -1,
for boolean fields it might be FALSE.
Being configurable, however, we can specify some other value, for example, there is a CV term `no_rank` that can be used for the infraspecific type on an organism, and we might want to hide that.

## Testing?
Note - because the change in commit https://github.com/tripal/tripal/pull/1971/commits/44d3dce11e1e6120560e95e524976541cc63a774 could have unintended consequences, I recommend putting a `dpm()` in `tripal/src/Entity/TripalEntity.php` line 566 so we know when the change is having an effect, do this after step 5 below
```
          // Keep track of elements that have no value.
          foreach ($prop_values as $i => $prop_value) {
            $prop_value_value = $prop_value->getValue();
if (!$prop_value_value and !is_null($prop_value_value)) {
  dpm($prop_value_value, "This value would have been removed in 4.x for field $field_name delta $delta");
}
            if (is_null($prop_value_value)) {
              // A given delta should only be present once here.
              if (!array_key_exists($field_name, $delta_remove) or !in_array($delta, $delta_remove[$field_name])) {
                $delta_remove[$field_name][] = $delta;
              }
              continue;
            }
          }
```
1. Build docker on `4.x` branch
2. Import type collection "Expression Content Types"
3. Enter a publication (tests text, string, and boolean (Is Obsolete) fields). Only enter required fields, leave others blank.
2024-09-13_pub1
4. Enter an array design which has lots of integer fields. Enter various values including zero, and leave some blank.
2024-09-13_ad1
5. Checkout this branch `git checkout tv4g1-issue1970-hide-field-configuration` and make the `dpm()` addition described above
6. Clear cache just to be safe `docker exec -it CONTAINERNAME /bin/bash -c "drush cache:rebuild"`
6. View the publication, it should appear the same as before by default.
7. Try changing some of the settings Tripal -> Page Structure -> Publication -> Manage display
for example 
- series name change to "Never Hide"
- change the "Is Obsolete" as shown here to not display if FALSE
2024-09-13_pub3
- Unique local identifier change to "Hide if empty or equal to a specific value" and for the value use what you entered.
2024-09-13_pub4
- Don't forget to hit save
8. Note that Series name now appears even though it has no value, and Unique local identifier is hidden.
2024-09-13_pub2
9. For array design, change the settings for an integer field containing a zero to not be displayed
These are the test values I used
1970_AD0
2024-09-13_ad2
10. Change the settings for an integer field containing a specific value to not be displayed
2024-09-13_ad3
11. View the array design to confirm they are no longer displayed
2024-09-13_ad4
12. Add a **Tripal** boolean field to some simple content type, for example contact.
13. Create a contact each with true and false
14. On 4.x the false value will not appear, because it is not stored in the drupal field table. On this branch it will be present.
15. Similarly test a tripal integer field and enter the value zero. On 4.x it will not be there, on this branch it will. This should be the only time that the dpm we added triggers
1970_dpm
1970_tripalfieldsfalse
16. Unpublish and republish to make sure that works correctly.
😬 Null integers are published as zero. This is not related to this PR but is a new issue #1972

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -392,6 +392,14 @@ field.formatter.settings.*:
       type: string
       label: 'Field suffix'
       nullable: true
+    hide_condition:
+      type: string
+      label: 'Hide condition'
+      nullable: true
+    hide_value:
+      type: string
+      label: 'Hide value'
+      nullable: true
     thousand_separator:
       type: string
       label: 'Thousand marker'
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -330,6 +330,14 @@ tripal.tripalfield_collection.*:
                                type: string
                                label: 'Field suffix'
                                nullable: true
+                             hide_condition:
+                               type: string
+                               label: 'Hide condition'
+                               nullable: true
+                             hide_value:
+                               type: string
+                               label: 'Hide value'
+                               nullable: true
                    form:
                      type: sequence
                      label: 'Edit'
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -561,8 +561,9 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
           $item->tripalLoad($item, $field_name, $prop_types, $prop_values, $this);
 
           // Keep track of elements that have no value.
-          foreach ($prop_values as $prop_value) {
-            if (!$prop_value->getValue()) {
+          foreach ($prop_values as $i => $prop_value) {
+            $prop_value_value = $prop_value->getValue();
+            if (is_null($prop_value_value)) {
               // A given delta should only be present once here.
               if (!array_key_exists($field_name, $delta_remove) or !in_array($delta, $delta_remove[$field_name])) {
                 $delta_remove[$field_name][] = $delta;
--- a/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalBooleanTypeFormatter.php
+++ b/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalBooleanTypeFormatter.php
@@ -27,6 +27,7 @@ class DefaultTripalBooleanTypeFormatter extends TripalFormatterBase {
     $settings = parent::defaultSettings();
     $settings['true_string'] = t('True');
     $settings['false_string'] = t('False');
+    $settings['hide_condition'] = '';
     return $settings;
   }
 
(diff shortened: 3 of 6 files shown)


laceysanderson (approved):
✅ Code changes makes sense + look sound

### Manual testing

✅ Hiding a false boolean (feature.is_analysis) value on gene worked as expected.
✅ Hiding a false boolean was specific to the field: configured is analysis to be hidden on false, made both is_analysis and is_obsolete false, only is_analysis was hidden.
✅ Hiding false boolean and re-write false boolean worked together as expected: made is_analysis false but also re-wrote false to be `Sally`; confirmed that is_analysis was still hidden and not re-written.
✅ Hiding test field worked as expected: set uniquename on gene to `meme` and configured the field to hide if empty or specific value = `meme` and observed the field was hidden. Choose hide if empty but still filled in meme for configuration and confirmed that the field was now not hidden as expected since it is not empty.
✅ Configured one text field to be hidden and confirmed only the one field was hidden and the setting did not leak to other fields.
✅ Integer: confirmed we can now enter `0` for an integer field and without config it will be displayed.
✅ Integer: confirmed we can config to hide empty but set value to -1 and a field with -1 is displayed (i.e. specific value is ignored for conditions not requiring it)
✅ Integer: confirmed that config to "hide 0" only hides 0.
✅ Integer: confirmed that config to hide specific value -1, hides -1 and not 0.
✅ Integer: confirmed settings are specific to field and do not leak to other fields on the same content type.
✅ String: configured organism.species to be hidden for specific value `databasica` and left in the `<em>` and `</em>` prefix and suffix + confirmed the field was still hidden (i.e. that the specific value check happened before the prefix/suffix were added)
✅ String: checked the default matches the previous behaviour (i.e. hide if empty)
✅ String: Configured to never hide publication series name and confirmed it showed up even when empty.

Notes: 
- previously, 0 and empty would have been hidden for integer fields but this is not what happens by default. I'm not against the change but if we wanted to be consistent then the default should be "hide 0" 🤷‍♀️  What happened in Tripal 3?
- previously false was hidden for boolean fields, wasn't it? If we wanted to be consistent that would be the default but I think I like the new never hide default better.

dsenalik:
Your point is a good one, it appears integer zero is hidden in Tripal 3.
I had been thinking of zero as a perfectly good integer, but thinking more on it, in most "biological" cases a zero is not a value we typically need to see. I will change the default there to be consistent with Tripal 3.

Testing the two booleans on a gene page in Tripal 3 shows that the false values are hidden.
I agree with you here that new new default of never hide may be better. I suspect the "hiding" of "falsey" values may have just come about in Tripal 3 without deep design discussions, and so this may be the time to change the default.

PR #1977, opened 2024-09-16
Publish update - no drupal records when no chado records

### LINKED ISSUE #1972: Null integer fields are published as zeroes
### Tripal Version

4.x

### Drupal Version

10.2

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

Sorry, I missed something in PR #1952 
Integer fields that are null are incorrectly published as zeroes.

### Steps to reproduce

1. Tripal -> Page Structure -> + Import type collection -> Expression
2. run the job
3. At a psql prompt: `INSERT INTO arraydesign (manufacturer_id, platformtype_id, name) VALUES (1, 10, 'Basic Array Design');`
4. Publish array design content type
5. run the job
6. Observe a bunch of zero values in integer fields that should not be there
1972_publish_integer_bug

### Error messages

_No response_

### Screenshots

_No response_

dsenalik:
# Bug Fix
## Description
As described in the linked issue #1972, integer fields with no value in chado are getting published as zero.

I had been blindly assuming that entities entered manually through the GUI were the "correct" example of how the Drupal field tables should look. While mostly correct, I have realized that the ChadoAdditionalType widget behaves differently, and does not remove empty values as the other widgets do. This meant that publish was leaving out this field, while GUI entered entities included a blank field (field title but no value). Realizing that this should be changed makes the changes here easier.

I have made several changes here:
1. I reverted changes to `findValues()`, so as before, any value deemed as "invalid" is removed. That means, if any required field is NULL then it is invalid. 
2. I updated the ChadoAdditionalType widget to, as other widgets do, remove from the `$values` array any record that is "empty". In this case that means if there is no accession.
3. This would break the title for organism, since the infraspecific type might be null. I changed the tripal token parser so that a token with no value is _always_ removed. This removes the "feedback" if you enter an invalid token, it is just silently removed. However, I think this would be better handed by validation back when you enter content title tokens, rather than later seeing you have fat thumbs and have to go back and fix, ~~so as of the posting of this PR this validation still needs to be added to `tripal/src/Form/TripalEntityTypeForm.php`~~
4. This token validation has been added

## Testing?
1. Redo the testing from PR #1952
2. Redo the steps from issue #1972

### DIFF
--- a/tripal/src/Form/TripalEntityTypeForm.php
+++ b/tripal/src/Form/TripalEntityTypeForm.php
@@ -304,7 +304,7 @@ class TripalEntityTypeForm extends EntityForm {
    * @param object $tripal_entity_type
    *
    * @return array
-   *   The list of valid tokens
+   *   The list of valid tokens for URLs, and the list of valid tokens for entity titles.
    */
   protected function getValidTokens($tripal_entity_type) {
     $url_tokens = $tripal_entity_type->getTokens();
@@ -316,6 +316,17 @@ class TripalEntityTypeForm extends EntityForm {
   }
 
   /**
+   * Validate that all tokens present in a passed string are valid.
+   * A token is anything enclosed in square brackets [xxx].
+   *
+   * @param string $format_string
+   *   The string to be validated containing any number of tokens.
+   * @param array $valid_tokens
+   *   A list of valid tokens. The token is the array key.
+   *
+   * @return string
+   *   An empty string if all tokens are valid, otherwise return
+   *   the first invalid token found.
    */
   protected function validateTokens($format_string, $valid_tokens) {
     $invalid_token = '';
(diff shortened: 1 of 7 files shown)


laceysanderson (commented):
Came across an issue that is fixed in 7df74231c441d4775e4ea9f1e93b0d38f316ed49.

Specifically, there was a bug in validateTokens() where it was only actually checking the first matched token and not all tokens matched. This allowed in invalid tokens as shown in the next screenshot where this form saved without error.
Screenshot 2024-10-05 at 4 21 42 PM

laceysanderson (approved):
✅ Code review looks good :-)
✅ Manually tested according to #1952 and confirmed that the infraspecific type is correctly published and removed in the title
Screenshot 2024-10-05 at 4 52 17 PM
✅ Testing according to the instructions in #1972 showed that the empty integer fields are published properly and no longer appear on the view page as `0` but rather are actually empty :-)
Screenshot 2024-10-05 at 4 52 09 PM

PR #1978, opened 2024-09-16
Update ChadoProperty Text widget to support filtered HTML

### LINKED ISSUE #1974: ChadoProperty field needs to be able to support formatted text
### Task Description

With PR #1862 we added support for the WYSIWYG toolbar for text fields, and now we need to do the same thing for property fields. Furthermore, it would be good for the filtered text widget for properties to use a textarea rather then a textfield.

This is needed because properties are often used to store metadata which can be more then a single value. For example, ActivatingCrops uses properties to store methodology and experimental design for research studies. This are often long text and sometimes even include links and tables.

Additionally, we can currently enter HTML into these fields and it will be rendered! This has some security implications and thus rendering of HTML should only happen when the appropriate filter is chosen.

### Branch Name

tv4g1-issue1974-filteredTextProperties

laceysanderson:
# Tripal 4 Core Dev Task

## Description

With PR #1862 we added support for the WYSIWYG toolbar for text fields, and now we need to do the same thing for property fields. Furthermore, it would be good to allow admin to configure the number of rows for the textarea.

This is needed because properties are often used to store metadata which can be more then a single value. For example, ActivatingCrops uses properties to store methodology and experimental design for research studies. This are often long text and sometimes even include links and tables.

Additionally, ensure that all the other property widgets return filtered text because on 4.x curators can enter HTML into the property field even if they do not have the appropriate permissions.

Specifically, this PR
 - adds a filter setting to the "Chado Property: Long Text" widget that defaults to basic.
 - ensures that the Chado Property formatter uses this setting.
 - since all Chado Property widgets share the same formatter, ensure that basic_html is used for all other widgets or if the "Chado Property: Long Text" was not configured.
 - ensures the property widgets respect the "Required" setting set in the field settings.
 - allows admin to choose the number of rows shown in the "Chado Property: Long Text" widget

### Manual

1. Add 4 different property fields to the organism content type. The terms do not matter. Set all to unlimited values.
2. Configure the form display and set two of these fields to use the long text widget and the other two to use the other widgets. Configure the long text widgets to use "Basic" for one and "Full" for the other. Add some options to the select list widget.
3. Create content. Enter the same block of HTML into both long text widgets. Ensure this includes an HTML. Also enter values in the other two widgets.
4. Save and confirm that the long text with "Basic" setting has stripped out the table/list formatting but the one using "Full" has not. We have two fields here to also confirm that the setting is on a per field basis.
5. Check that no errors were thrown for the other two property fields (ie. string and select widget) even though they did not have a filter format configured.

Other things to test:
- set some of the fields to required and others not and confirm the form respects this.
- change the format from "Full" to "Basic" and confirm the formatter/page strips tables without you needing to edit the page. Also check that when you do edit the page, the HTML is stripped.
- change the setting to indicate the number of rows for the "Long Text" widget and confirm the widget reflects this setting.
- enter an HTML table one-liner (e.g. `<table><thead><tr><th>H1</th><th>H2</th></tr></thead><tbody><tr><td>D1</td><td>D2</td></tr></tbody></table>`) as an option for the "Select" widget when configuring it and confirm that while you can select it, when the page is viewed, the HTML is stripped. You should be able to use basic html in a select option (i.e. bold or italics)

### Automated Testing

I started adding automated testing for the property field based on the model I used for Tripal fields lately. My progress is included here and, while not complete, it is fully functioning and better coverage then we had previously.

Includes
 - ChadoFieldTestTrait: this is the Chado-focused version of the TripalFieldTestTrait. It includes methods for creating field types + instances as well as an assert to confirm the values match what was expected. It also includes a method to make setting up the test environment for field kernel tests easier.
 - ChadoPropertyTypeSimpleCRUDTest: focuses on testing a single property field on a project-based content type with only one delta. It provides 44% test coverage of the ChadoPropertyTypeDefault field type class and tests creating a TripalEntity with a property field attached. Specifically, it checks that
   - the property field can be created programatically and attached to a tripal entity type.
   - a tripal entity can be created programatically with values for a property field supplied.
   - the property field values are set properly in the tripal entity class
   - tripal entity can retrieve the values successfully which checks basic interactions between the fields and the TripalStorage backend
   - that the entity can be saved without throwing any errors
   - that tripalTypes can be called without error and all the property type objects are created as expected.

Test coverage shows that this test is travelling the expected paths through the ChadoPropertyTypeDefault, TripalFieldItemBase, TripalEntity, and ChadoStorage for inserting field data.

Next steps are to check the field values are inserted into chado; however, this is not producing the expected results... Since I have already spent more time then I should have, I am moving on. These tests are included because they provide a good model for moving forward on testing fields.

### DIFF
--- a/tripal/config/schema/tripal.schema.yml
+++ b/tripal/config/schema/tripal.schema.yml
@@ -403,3 +403,6 @@ field.widget.settings.*:
     filter_format:
       type: string
       label: 'Filter format for a filtered text field'
+    num_rows:
+      type: integer
+      label: 'Number of Rows to display in the text widget'
--- a/tripal/tests/src/Traits/TripalTestTrait.php
+++ b/tripal/tests/src/Traits/TripalTestTrait.php
@@ -99,7 +99,7 @@ trait TripalTestTrait {
    * @return TripalTerm
    *   Returns the tripal term that was created.
    */
-  public function createTripalTerm($values, $idspace_plugin_id, $vocab_plugin_id) {
+  public function createTripalTerm(&$values, $idspace_plugin_id, $vocab_plugin_id) {
 
     // Setting the default values:
     $random = $this->getRandomGenerator();
@@ -122,6 +122,7 @@ trait TripalTestTrait {
       $vocabulary = $vmanager->createCollection($values['vocab_name'], $vocab_plugin_id);
       $this->assertInstanceOf(TripalVocabularyInterface::class, $vocabulary, "Unable to create the Vocabulary.");
     }
+    $this->assertInstanceOf(TripalVocabularyInterface::class, $vocabulary, "Unable to load the Vocabulary.");
 
     // Create the ID Space.
     $idsmanager = \Drupal::service('tripal.collection_plugin_manager.idspace');
@@ -131,13 +132,13 @@ trait TripalTestTrait {
       $this->assertInstanceOf(TripalIdSpaceInterface::class, $idSpace, "Unable to create the ID Space.");
       $idSpace->setDefaultVocabulary($vocabulary->getName());
     }
-
+    $this->assertInstanceOf(TripalIdSpaceInterface::class, $idSpace, "Unable to load the ID Space.");
 
     $term = $idSpace->getTerm($values['term']['accession']);
     if (!$term) {
       // Now create the term.
-      $values['term']['idSpace'] = $idSpace->getName();
-      $values['term']['vocabulary'] = $vocabulary->getName();
+      $values['term']['idSpace'] = $values['id_space_name'];
+      $values['term']['vocabulary'] = $values['vocab_name'];
       $term = new TripalTerm($values['term']);
       $this->assertInstanceOf(TripalTerm::class, $term, "Unable to create the term object.");
       // and save it to the ID Space.
--- a/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoPropertyFormatterDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoPropertyFormatterDefault.php
@@ -28,15 +28,34 @@ class ChadoPropertyFormatterDefault extends ChadoFormatterBase {
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
 
+    // Default filter format.
+    $filter_format = 'basic_html';
+
+    // We need to get the format which was set in the widget settings
+    // because they need to match.
+    // Note: the default filter is used when the widget does not support
+    // filter formats (i.e. String and Select widgets).
+    $entity_type = $this->fieldDefinition->get('entity_type');
+    $bundle = $this->fieldDefinition->get('bundle');
+    $field_name = $this->fieldDefinition->get('field_name');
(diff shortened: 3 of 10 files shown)


dsenalik:
Everything was working great, but then I tried this,
I made the first and third field "required"
The third is set to "select drop-down"
I get proper validation if I fail to select anything, but
(A) HTML in the select options is not interpreted, e.g. I get `<em>em</em>` instead of _em_
The first one is set to "Long text" @ 5 rows , basic HTML
If I leave the first one empty I get the validation message, but
(B) also errors
 
Error message

    DNA Property (value 1) field is required.
    Warning: Undefined array key "#parents" in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: foreach() argument must be of type array|object, null given in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: Undefined array key "#parents" in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: foreach() argument must be of type array|object, null given in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: Undefined array key "#parents" in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: foreach() argument must be of type array|object, null given in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: Undefined array key "#parents" in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

    Warning: foreach() argument must be of type array|object, null given in Drupal\Core\Form\FormState->getError() (line 1174 of core/lib/Drupal/Core/Form/FormState.php).

laceysanderson:
Good catch @dsenalik! This is ready to review again :-)

> (A) HTML in the select options is not interpreted, e.g. I get <em>em</em> instead of em

This is a Drupal issue and was an issue with the select widget before this PR 🙈 I have no idea how to fix it and am not sure it's worth the effort 🤔 but we can add another issue for it if you think people are likely to use HTML in their select lists?

> (B) If I leave the first one empty I get the validation message, but also errors

This is fixed in https://github.com/tripal/tripal/pull/1978/commits/d67ce721a704a01d3a8fad31aed35a2dcd6042f2. Apparently I was being too liberal with silencing the format and added classes to more then just the text_format element... Good catch! Your process now produces the validation message and no additional errors ✅ 

PS. I also fixed the breaking tests in https://github.com/tripal/tripal/pull/1978/commits/ad8a17c218d03ce55dd1ee0a3093ac4917af41ca. This was caused because I changed the method definition for createTripalTerm() to pass by reference and PHP screams if you define the array on the spot when calling the method as compared to defining a variable first 🙈

dsenalik (approved):
For [A] no I don't think it is worth even creating an issue at this point for html in select options
[B] Great! Working as expected now
And with the tests fixed all looks good to me
And nothing to report from code review

PR #1980, opened 2024-09-24
Fix the creation of duplicate aliases every time an entity is edited

### LINKED ISSUE #1973: Multiple URL alias' for Tripal Content pages
### Tripal Version

4.x-dev

### Drupal Version

10.4.x-dev

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

We are getting multiple versions of the same URL alias created everytime Tripal Content pages are edited and re-saved. This makes it difficult to maintain these alias' and will eventually cause performance issues.

### Steps to reproduce

1. Create a new Tripal Content page (i.e. Organism), details do not matter.
2. Go to Admin > Configuration > Search and metadata > URL aliases (admin/config/search/path) and search for the alias (i.e. organism/1). On a fresh site, this will be the only alias.
3. Now, edit the Tripal Content page you just created and save it. No changes to the data are needed here. You can even do this multiple times.
4. Finally, go back to the URL aliases page and confirm you now have an entry for each time you edited + the time you saved the page... all with the same alias 🤦 

### Error messages

```shell
There are no errors. This is done silently and would likely go unnoticed until someone tries to make their own alias.
```

### Screenshots

image

dsenalik:
# Bug Fix

## Description
Currently, every time an entity is edited, another copy of the path alias is created. This PR fixes that by first checking if an alias already exists.

## Testing?
1. Create an entity of any content type. Here I used "project".
2. Check the alias table (or go to `/admin/config/search/path`):
```
sitedb=> select * from path_alias;
 id | revision_id |                 uuid                 | langcode |    path     |   alias    | status 
----+-------------+--------------------------------------+----------+-------------+------------+--------
  1 |           1 | 997dfd5b-7ebf-4079-8d35-f7b60202c5a9 | und      | /bio_data/1 | /project/1 |      1
(1 row)
```
3. Edit the entity, you don't need to make any changes, and save it again.
4. Verify that there is still only one alias:
```
sitedb=> select * from path_alias;
 id | revision_id |                 uuid                 | langcode |    path     |   alias    | status 
----+-------------+--------------------------------------+----------+-------------+------------+--------
  1 |           1 | 997dfd5b-7ebf-4079-8d35-f7b60202c5a9 | und      | /bio_data/1 | /project/1 |      1
(1 row)
```

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -175,12 +175,20 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
     // Make sure the path alias is URL friendly.
     $path_alias = str_replace(['%2F', '+'], ['/', '-'], urlencode($path_alias));
 
-    // Now finally, set the alias.
-    $path = \Drupal::entityTypeManager()->getStorage('path_alias')->create([
-      'path' => $system_path,
-      'alias' => $path_alias,
-    ]);
-    $path->save();
+    // Check if this alias already exists.
+    $alias_exists = FALSE;
+    if (\Drupal::service('path_alias.repository')->lookupByAlias($path_alias, 'und')) {
+      $alias_exists = TRUE;
+    }
+
+    // If the alias does not exist, then create it.
+    if (!$alias_exists) {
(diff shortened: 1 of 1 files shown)


dsenalik:
I am closing this as this will be fixed by PR #1981 which will make the changes here obsolete

PR #1981, opened 2024-09-25
add path alias to tripal entity base class

### LINKED ISSUE #1963: TripalEntity needs to have the URL alias base field added to better support URL alias'
### Task Description

See Drupal issue [#3055349](https://www.drupal.org/project/drupal/issues/3055349).

This issue shows that in order to have the URL alias settings on the Tripal Content page, we need to add a base definition for the path field.

```php
$fields['path'] = BaseFieldDefinition::create('path')
  ->setLabel(t('URL alias'))
  ->setTranslatable(TRUE)
  ->setDisplayOptions('form', [
    'type' => 'path',
    'weight' => 100,
  ])
  ->setDisplayConfigurable('form', TRUE)
  ->setComputed(TRUE);
```

This is needed because the path module only adds it to node, comment and media entity types by default 😒 It adds a field like the following which is extremely helpful when creating Tripal content through the UI that you may want a custom alias for. For example, we create Research experiment pages (project-based content type) and will often set the URL alias by hand. This is needed because the titles for these are often long and/or not unique. Since we recommend people link to these in their publications as a source for methodology, we set a shorter, readable URL alias for them to use. Right now, I can't do this easily in Tripal 4...

Screenshot 2024-09-03 at 2 01 06 PM

### Branch Name

tv4g1-issue1963-addUrlAliasField

### LINKED ISSUE #1973: Multiple URL alias' for Tripal Content pages

dsenalik:
# New Feature

## Description
This adds a path alias field to tripal entity widgets, allowing entry of an alias when content is first created, or changing it when the entity is edited.
If an alias is not added, or if it is removed, then the default alias is generated (this default is the same as before this PR).
This PR also fixes the problem of multiple copies of the default alias being generated. Now exactly one alias should exist for every Tripal entity.

image

## Testing?
1. Create any content type, and do not specify an alias when creating it
2. Check the alias table (or go to `/admin/config/search/path`)
`select * from path_alias;`
3. Edit, the default alias should populate the alias box. Change it to a new value `/p2` and save
4. Check the alias table (or go to `/admin/config/search/path`)
`select * from path_alias;`
```
sitedb=> select * from path_alias;         
 id | revision_id |                 uuid                 | langcode |    path     |   alias    | status 
----+-------------+--------------------------------------+----------+-------------+------------+--------
  1 |           1 | ad583371-adb5-4392-9f8f-5648349f2bc5 | und      | /bio_data/2 | /p2        |      1
```
5. Edit and leave the alias box blank and save.
6. Check the alias table (or go to `/admin/config/search/path`)
`select * from path_alias;`
The alias should revert to the default alias.
2024-09-25_aliascardinality1
7. Edit the content type and set the URL alias pattern to include field tokens (e.g. `organism/[organism_genus]-[organism_species]`)
8. Now either edit an existing page or create a new one and leave the alias blank or delete it. Ensure that there are no errors and the tokens are filled in as expected.
7. Create a new entity and specify an alias when creating it.


laceysanderson:
I made this dependant on the EntityForm upgrades so that we can move it into the advanced sidebar. I started by trying to fix this tests --I think I have it but I didn't run the entire test suite locally so we'll see if I missed any that were failing.

laceysanderson:
Ok, automated tests are now passing, the URL alias is in the advanced sidebar and it works for the most part.

Main outstanding issue:
If the URL pattern includes field-related tokens, they don't seem to actually get filled in 😬 For example, if the URL alias patter for an organism is `organism/[organism_genus]-[organism_species]` then you would expect the default alias to be something like `organism/Tripalus-databasica` but right now it ends up being `organism/-` 🙈 I have to switch focus again so I'm hoping you can pick it up from here @dsenalik?

dsenalik:
Yes! I'm working on tokens now so sure!

dsenalik:
The token problem exists on the 4.x branch.
Since it is looking like the fix will be a bit complicated, I propose to make a new issue for it and not try to fix in this PR which is already fairly large, and the problem is not really related to this PR.
See Issue #2024
I think we should go ahead and merge this PR, whaddaya think @laceysanderson

dsenalik (commented):
I reviewed the changes that you added @laceysanderson and thanks for figuring this out 😍. I saw a bit of white space and typos that I have fixed. I approve your additions!

laceysanderson:
Oh, ok 👍 If the problem exists in 4.x as well then I agree that we can merge this one and do the fix in a separate branch/fresh PR for easier review and to not hold things back any longer then needed :-)

laceysanderson (approved):

PR #1982, opened 2024-09-25
Update Tripal Layout  to support configuring the Form Display

### LINKED ISSUE #1863: Expand Tripal Layout to support form display as well
### Task Description

Tripal Layout currently allows module developers to define a layout of fields and fieldsets for the page display. We would like to expand the same functionality to handle the form display.

## Usage Example

In issue #1827, @dsenalik suggested that putting the three fields that make up the analysis unique constraint in a fieldset with a descriptive title might help with the usability concerns mentioned in that issue.
2024-03-23_analysis-unique

### Branch Name

tv4g7-issue1863-tripalLayoutFormDisplay

laceysanderson:
# Tripal 4 Core Dev Task

## Description

Tripal Layout currently allows module developers to define a layout of fields and fieldsets for the page display. We would like to expand the same functionality to handle the form display. For specific use cases, see the linked issue.

This PR creates a new config entity (TripalLayoutDefaultForm) to handle form display layouts based on the original TripalLayoutDefaultView config entity which supports configuring the default page display.

**NOTE: This PR includes an unavoidable BREAKING CHANGE.** All layout yamls made before this PR will need to be updated to match the new format. All core yamls have been updated. This was needed because the previous yaml was not supportable by Drupal translation. This was only caught when automated testing was added.

### Tasks
- [x] Create TripalLayoutDefaultForm config entity based on TripalLayoutDefaultView
- [x] Create the YAML to organize the form for one content type (i.e. organism) and ensure the config schema is in place to support that.
- [x] Ensure that new TripalLayoutDefaultForm entities are created for each YAML file matching the file naming pattern just like what happens for the display.
- [x] Add "Apply Tripal Layout" and "Reset Layout" buttons to the form display page.
- [x] Implement the functionality within the controller for the "Reset Layout" form display button.
- [x] Implement the functionality within the controller for the "Apply Tripal Layout" form display button.
- [x] Cleanup the controller to reduce code duplication as much as possible.
- [x] Fix schema errors caught by automated testing
- [x] Ensure child fields are in the order they are listed in the YAML.
- [x] Support descriptions for field groups
- [x] Remove any empty field groups
- [x] A form template for analysis content type.
- [x] Some/Any automated testing

### Automated Testing

This PR adds the first automated testing for the Tripal layout module. This is the coverage it adds:

Screenshot 2024-09-28 at 8 16 02 PM

#### Testing still missing

 - controller: actually applies the layouts

### Manual Testing

On a fresh docker on this branch,
1. Go to Admin > Tripal > Configuration > Tripal Default Form Layouts (admin/tripal/config/tripal_layout_default_form) and confirm that there is an entry listed for "General Form".
2. Go to the Manage form display for the analysis content type: Tripal > Page Structure > Analysis > Manage Form Display (admin/structure/bio_data/manage/analysis/form-display) and confirm that there are "Apply Tripal Layout" and "Reset Layout" buttons added.
  image
3. Make some manual changes to the form display and save. Then click the reset button and confirm they are reset.
4. Click the "Apply Tripal Layout" button and confirm the form is separated into fieldsets as designed.
  image
5. Now redo steps 2 and 3 for the "Manage Display". You should expect to see the following.
  image
6. Create an analysis and confirm that the form shows this new form display layout and that no errors are encountered during save. This form will also be available on edit.
  image
7. After saving your analysis confirm the display matches the set template for "Manage Display". For example, like this.
  image

Other things to test:
- confirm you can still customize the form display after the Tripal Layout has been applied and see these customization in the add/edit content form for that content type.
- confirm you can reset the tripal layout back to the drupal one.
- check the organism form display as well.
- try to apply a template for any of the other content types and expect to be told they don't have one.
- also test display layouts since the schema needed to change.


dsenalik:
Testing is going well except for one issue I noticed. The reset layout is generating different weights every time it is pressed and it doesn't match the original weights.
For example, without doing anything else first except displaying weights, just clicking the + Reset layout button on organism does this.
2024-10-04_reset_layout

laceysanderson:
> Testing is going well except for one issue I noticed. The reset layout is generating different weights every time it is pressed and it doesn't match the original weights.

Good catch 🙈 Fixed in b4d3bca72. 
Apparently Drupal just keeps adding to the Highest weight and never resets it. Instead the module now applies weights for everything based on order unless it is specifies in the YAML. Also, it now resets all weights to 0 when resetting the display.

Note: I can't get it reset back to the weights loaded via Chado prepare since I don't have anywhere to save such things and it doesn't feel worth it performance-wise to make this dependant on pulling from the original field collection YAMLs.

dsenalik (changes requested):
I have one requested change, otherwise looks great.
- [x] ... confirm that there is an entry listed for "General Form".
- [x] ... confirm that there are "Apply Tripal Layout" and "Reset Layout" buttons added.
image
- [x] ... click the reset button and confirm they are reset
Most fields reset to zero. Title gets -4 and "authored by" gets 5
- [x] ... redo steps 2 and 3 for the "Manage Display"
Here the "authored by" gets zero.
- [x] Confirm display is as expected
20241006_analysis

- [x] Tested organism similarly, all good
- [ ] Tested contact (which has no layout yet). We get the following message which is fine, but 
20241006_contact
**the layout has been reset.**

I think this code could be reordered to eliminate the reset if there is no layout. In `tripal_layout/src/Controller/TripalEntityUILayoutController.php `
```
    // First reset the display.
    $this->resetLayout($tripal_entity_type, $display_context, $display);

    // Get the layout for this bundle.
    $layout = $this->getLayout($bundle, $display_context);
    if (!$layout) {
      return FALSE;
    }
```
I tested changing the order and it appears to now not reset if there is no layout.
- [x] confirm you can still customize the form display after the Tripal Layout has been applied and see these customization in the add/edit content form for that content type.
- [x] confirm you can reset the tripal layout back to the drupal one.
- [x] also test display layouts since the schema needed to change. - I think everything looks correct

laceysanderson:
Good point! Fixed exactly as you suggested in commit 575aadaea18ce66fdc72ef9f369ba020f28ca4d8.
image

dsenalik (approved):
💓

PR #1983, opened 2024-09-27
Publishing in batches

### LINKED ISSUE #1910: Allow publishing in batches

### LINKED ISSUE #1959: Publish new vs. update existing entities
### Task Description

Easy task. 1. Make three contacts. 2. Make a project, and include the three contacts. 3. Then publish contacts.
`Published 4 new entities, and 6 field values.`
The null contact is published "new", but the other three already existed as entities. Actually they were just updated.
The message could read
`Published 1 new entity, updated 3 existing entities, and added 6 field values.`

### Branch Name

~~tv4g1-issue1959-update-publish-message~~
This will be done in PR #1983

dsenalik:
# New Feature

## Description
Suppose you want to migrate your Tripal 3 site with several genomes, each having tens of thousands of "gene", "mRNA", etc. records. Right now, publish may fail when it runs out of memory.
This PR implements publishing in batches of at most 1000 records at a time.
We might want to put the actual batch size on the publish form, but for now it is hard-coded.

As far as log messages, to be more succinct, I removed this field message,
```
        $this->logger->notice("  Checking for published items for the field: $field_name...");
```
 and only display the other when at least one field value was published
```
        if ($num_inserted) {
          $this->logger->notice("  Published " . number_format($num_inserted) . " items for field: $field_name...");
        }
```

If only a few records, output will be the same as before, however if enough records exist to trigger batching, then a batch prefix is included. For example
```
Finding candidate record IDs...
Batch 1 of 2, Step 1 of 6: Find matching records...
Batch 1 of 2, Step 2 of 6: Generate page titles...
Batch 1 of 2, Step 3 of 6: Find existing published entities...
Batch 1 of 2, Step 4 of 6: Publishing 1,000 new entities...
Batch 1 of 2, Step 5 of 6: Find IDs of entities...
Batch 1 of 2, Step 6 of 6: Add field items to published entities...
  Published 1,000 items for field: organism_abbreviation...
  Published 1 items for field: organism_comment...
  Published 3 items for field: organism_common_name...
  Published 1,462 items for field: organism_dbxref...
  Published 1,000 items for field: organism_genus...
  Published 95 items for field: organism_infraspecific_name...
  Published 1,000 items for field: organism_infraspecific_type...
  Published 1,000 items for field: organism_species...
Batch 2 of 2, Step 1 of 6: Find matching records...
Batch 2 of 2, Step 2 of 6: Generate page titles...
Batch 2 of 2, Step 3 of 6: Find existing published entities...
Batch 2 of 2, Step 4 of 6: Publishing 919 new entities...
Batch 2 of 2, Step 5 of 6: Find IDs of entities...
Batch 2 of 2, Step 6 of 6: Add field items to published entities...
  Published 919 items for field: organism_abbreviation...
  Published 4 items for field: organism_comment...
  Published 1 items for field: organism_common_name...
  Published 1,096 items for field: organism_dbxref...
  Published 919 items for field: organism_genus...
  Published 68 items for field: organism_infraspecific_name...
  Published 916 items for field: organism_infraspecific_type...
  Published 4 items for field: organism_pub...
  Published 919 items for field: organism_species...
Publish completed. Published 1,919 new entities, checked 0 existing entities, and added 10,407 field values.
```

## Testing?
Either you migrate a site with thousands of records of a particular type, or for a fast test you can temporarily change the batch size in `tripal/src/Services/TripalPublish.php`
```
  /**
   * Specifies the maximum number of records to publish at one time.
   * This limits memory consumption if there are many thousands of
   * records, for example gene records in the feature table.
   * @todo We might want to add this as an option on the publish form.
   * 
   * @var integer $batch_size
   */
  private $batch_size = 1000;
```

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -15,8 +15,12 @@
     <include>
       <directory suffix=".php">tripal/src</directory>
       <file>tripal/tripal.module</file>
+      <directory suffix=".php">tripal_biodb/src</directory>
+      <file>tripal_biodb/tripal_biodb.module</file>
       <directory suffix=".php">tripal_chado/src</directory>
       <file>tripal_chado/tripal_chado.module</file>
+      <directory suffix=".php">tripal_layout/src</directory>
+      <file>tripal_layout/tripal_layout.module</file>
     </include>
     <exclude>
       <!-- Exclude Interface classes from coverage -->
--- a/pub.log
+++ b/pub.log
@@ -1,0 +1,2 @@
+nohup: ignoring input
+nice: ‘drush’: No such file or directory
--- a/pub.log
+++ b/pub.log
@@ -1,2 +1,0 @@
-nohup: ignoring input
-nice: ‘drush’: No such file or directory
--- a/tripal/src/Form/TripalEntityTypeForm.php
+++ b/tripal/src/Form/TripalEntityTypeForm.php
@@ -331,9 +331,9 @@ class TripalEntityTypeForm extends EntityForm {
   protected function validateTokens($format_string, $valid_tokens) {
     $invalid_token = '';
     preg_match_all('/(\[[^\]]+\])/', $format_string, $matches);
-    foreach ($matches as $match) {
-      if ($match and !array_key_exists($match[0], $valid_tokens)) {
-        $invalid_token = $match[0];
+    foreach ($matches[0] as $match) {
+      if ($match and !array_key_exists($match, $valid_tokens)) {
+        $invalid_token = $match;
         break;
       }
     }
--- a/tripal/src/Form/TripalEntityTypeForm.php
+++ b/tripal/src/Form/TripalEntityTypeForm.php
@@ -22,6 +22,7 @@ class TripalEntityTypeForm extends EntityForm {
 
     $tripal_entity_type = $this->entity;
     $tripal_entity_type->setDefaults();
+    list($url_tokens, $title_tokens) = $this->getValidTokens($tripal_entity_type);
 
     // We need to choose a term if this is a new content type.
     // The term cannot be changed later!
@@ -135,12 +136,6 @@ class TripalEntityTypeForm extends EntityForm {
       '#title' => t('Advanced Settings'),
     );
 
-    $tokens = $tripal_entity_type->getTokens();
-    $title_tokens = $tokens;
-    unset($title_tokens['[title]']);
-    unset($title_tokens['[TripalBundle__bundle_id]']);
-    unset($title_tokens['[TripalEntity__entity_id]']);
-
     // Page title options:
     $form['title_settings'] = [
       '#type' => 'details',
@@ -151,7 +146,7 @@ class TripalEntityTypeForm extends EntityForm {
     $form['title_settings']['msg'] = [
       '#type' => 'item',
       '#markup' => t('
-<p>The format below is used to determine the title displayed on content pages of the current type. This ensures all content of this type is consistent while still allowing you to indicate which data you want represented in the title (ie: which data would most identify your content).</p>
+<p>The format below is used to determine the title displayed on content pages of the current type. This ensures all content of this type is consistent while still allowing you to indicate which data you want represented in the title (i.e.: which data would most identify your content).</p>
 
 <p>Keep in mind that it might be confusing to users if more than one page has the same title. <strong>We recommend you choose a combination of tokens that will uniquely identify your content</strong>.</p>'),
     ];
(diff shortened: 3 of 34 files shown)


laceysanderson (commented):
I did manual testing on this by using the GFF3 importer on the Tripalus databasica test gff3.

Publishing mRNA with any size of batch worked great.
Publishing gene failed however being unable to find the base table and there was no check to make sure it was found...

Specifically, I created and organism and analysis, I ran the GFF3 importer with the file `/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/fixtures/gff3_loader/TripalusDatabasicaChr1Genes.gff3` and ran the job. I also imported the "Genomic" content type collection. Then I changed the batch size to 5 and published mRNA. It ran without error and created beautiful pages. I then tried to publish Genes and it failed on the command-line with an error regarding the base table. The failure for genes was not reproducible in another docker so I'm not sure what the difference was...

That said, code review also turns up that we are using the chado base table 3rd party setting directly which makes TripalPublish only work for chado 🙈 In PR #1991 we are looking at making a chado specific publish and creating a plugin infrastructure so each backend can have it's own publish but in the meantime, we are keeping Tripal Publish and don't want to break it for other data backends (if it works for them... this hasn't actually been tested).

laceysanderson (changes requested):
Upon further thought, I do think we need to be careful not to assume chado storage here... I think this small change takes out that assumption (couldn't use github suggestion since it covers lines of code you hadn't originally changed):

```diff
diff --git a/tripal/src/Services/TripalPublish.php b/tripal/src/Services/TripalPublish.php
index 4d608c401..1c6c734cf 100644
--- a/tripal/src/Services/TripalPublish.php
+++ b/tripal/src/Services/TripalPublish.php
@@ -215,7 +215,6 @@ class TripalPublish {
       throw new \Exception(t($error_msg, ['%bundle' => $bundle]));
     }
     $this->entity_type = $entity_type;
-    $this->base_table = $entity_type->getThirdPartySetting('tripal', 'chado_base_table');
 
     // Get the storage plugin used to publish.
     /** @var \Drupal\tripal\TripalStorage\PluginManager\TripalStorageManager $storage_manager **/
@@ -228,6 +227,21 @@ class TripalPublish {
 
     $this->setFieldInfo();
 
+    // We need a way to get all the record ids for a bundle.
+    // If this is the chado storage backend then we do this using the chado table.
+    if ($datastore == 'chado_storage') {
+      $this->base_table = $entity_type->getThirdPartySetting('tripal', 'chado_base_table');
+    }
+    // But if this is not chado storage then the backend needs to provide the base
+    // table for a bundle.
+    else {
+      $this->base_table = $this->storage->getBaseTable($bundle);
+    }
+    if (empty($this->base_table)) {
+      $error_msg = 'Could not find the base table for the %bundle entity type.';
+      throw new \Exception(t($error_msg, ['%bundle' => $bundle]));
+    }
+
     // Get the required field properties that will uniquely identify an entity.
     // We only need to search on those properties.
     $this->required_types = $this->storage->getStoredTypes();
```

Essentially, the above now checks in the init where you set the base table

1. if the backend is chado_storage then I use the same code you did
2. if not then I expect the backend to have a method to retrieve the base table based on the bundle
3. I also throw an exception now if we are not able to determine the base table.

Whereas before you just grabbed the third party setting and assumed it was present.

The rest of your code then remains untouched.

laceysanderson (approved):
✅ Code review now looks good 👍 There are faster ways to do the query restrict to batch that we can explore in the future but for now this is a huge improvement.
✅ Manual testing of importing both genes and mRNA when there is less then the amount in a batch + new records, greater then the amount and new records, and the same two combinations when there are more then the batch size. Worked great in all these cases and nice and fast too :-)

Thanks for all this work on publish @dsenalik! This is ready to merge!

PR #1988, opened 2024-10-04
Create xdebug_output directory for performance profiling results and debugging in general

laceysanderson:
# Tripal 4 Core Dev Task

## Description

In https://github.com/tripal/tripal_doc/pull/75 we discuss recent versions of TripalDocker creating the configured output directory for xdebug automatically... However, this didn't actually get done -until now. This PR simply creates the directory that Tripaldocker XDebug is configured to use. Thanks to git, a README is needed to get the directory committed and I acquiesced as it is good practice anyway 😜 

## Testing?

No testing needed but if you wanted, you can follow the instructions in the linked PR to test performance profiling and verify you no longer need to create the xdebug directory as mentioned here:

Screenshot 2024-10-04 at 3 27 03 PM

### DIFF
--- a/tripaldocker/xdebug_output/README.md
+++ b/tripaldocker/xdebug_output/README.md
@@ -1,0 +1,5 @@
+XDebug Output
+===============
+
+This directory is here as a place for XDebug to store various output files. XDebug in TripalDocker 
+is already configured to use this directory.


dsenalik (approved):
Just to be safe I built a docker and verified the directory was there. That was overkill I know.

PR #1990, opened 2024-10-05
Fix the brittle PubMed test

### LINKED ISSUE #1989: Fix brittle pub tests
### Task Description

There is one pub test that is failing quite a lot and then works again when re-run. This is happening quite often so this issue is a place to work on fixing this. 

My hope is that we can find a way to skip the test when pubmed is not available.

```
Drupal\Tests\tripal_chado\Functional\TripalPubLibraryTest::testTripalPubLibraryTestSimpleTest
This should have returned one pubmed record
Failed asserting that null is not equal to null.

/var/www/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:122
/var/www/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:56
/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Plugin/TripalPubLibraryTest.php:55
/var/www/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:729
```

### Branch Name

tv4g9-issue1989-pubmedBrittleTest

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

Skips the pubmed test if it is unable to retrieve results (i.e. returns NULL from retrieve) rather then failing.

## Testing?

Observe if the tests still fail.

### DIFF
--- a/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php
+++ b/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php
@@ -24,7 +24,7 @@ class TripalPubLibraryPubmed extends TripalPubLibraryBase {
     // is needed to receive and process the criteria data. See below.
   }
 
-  /** 
+  /**
    * Plugin specific form submit to add form values for example to criteria array
    * The criteria array eventually gets serialized and stored in the tripal_pub_import
    * database table. (This code gets called from ChadoNewPublicationForm)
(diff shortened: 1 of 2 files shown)


dsenalik (approved):
This one has to be just code review because of course the error won't strike now for a while. But I approve 100% - it will be nice not having these false failures on unrelated PRs.

laceysanderson:
Thanks @dsenalik! I did actually strike multiple times on this PR 🙈 As you can see in these screenshots there are two examples that show **S** for skipped and one that doesn't. This is the new fix and if it weren't made the two with skipped would have failed.
combo1-noSkipped
combo2-skipped
combo3-skipped

Issue #1984, opened 2024-09-27
Don't publish one chado record under multiple content types.


dsenalik:
### Task Description

Right now, as with Tripal 3, if you have a "generic" content type and a "specific" content type both using the same chado table, a record may be published twice.
For example, if you manually enter a "Genome Project" entity, and then publish the more generic "Project" content type, it will be published as a generic "Project" entity.
We don't have this problem with content types derived from the `feature` table, as in this case there is no generic "Feature" content type.

I am thinking that publish could get a list of all content types from a particular base table, get all those terms, and then it would have to check for a `type_id` column in the base table, or if that doesn't exist, look in the property table (as for example Project uses).

### Branch Name

tv4g1-issue1984-only-publish-once

Ferrisx4:
A discussion in Slack about this topic came across a few possible solutions to this. The examples use the analysis type and its subtypes, typically Genome Assembly and Genome Annotation, but also apply to other "generic" types that have "specific" subtypes such as Project and Genome Project mentioned above.

1. Create a generic type for "Analysis", assigned to the existing analysis bundle. This way all content from the analysis table belongs to a specific bundle. A proposed term to use to identify these "generic" analyses could be `Analysis (operation:2945)`. We'd need an update hook to address sites that do not have this value stored in `analysisprop` and add the Chado Type Reference field to the Analysis bundle. 
2. When publishing generic analyses, exclude analyses that have a specified subtype in the analysisprop table. This may be a little less straightforward because the type_id in the analysisprop table refers to the (and matches) the term itself. Because of this, we'd have to maintain a list of terms (`value`) or `term_id`s that designate the entity as a subtype and aren't just some other property e.g. `is_obsolete`.
3. Just for completeness of record from the conversation, an unlikely approach would be to, when publishing Analysis content, also publish all content of all subtypes and publish them under their respective types. This would then prompt the user when publishing Analysis content that *"This may also publish Genome Annotation and Genome Assembly content"*. This is challenging as @dsenalik pointed out because the table is determined from the bundle, and not the other way around.

Ferrisx4:
A 4th approach not mentioned would be to simply not publish content that is already published, however this would create a situation where the order of publishing matters. From the example above, a site would have to publish all specific analyses (Genome Assembly, Genome Annotation, etc.) first before publishing generic Analysis content.

dsenalik:
Here is a potential workaround I am testing that a site admin could apply to a site.
[...]

PR #1991, opened 2024-10-05
Publish duplicated titles, and multi-bundle chado tables

### LINKED ISSUE #1986: TripalPublish: finding unpublished records may need a redesign
### Task Description

This issue is meant to connect a number of issues and provide a place for discussion. We are seeing a number of issues pop-up now that we are trying to publish real live sites that imply a bit of a redesign may be needed.

Specifically,

 - #1984 
 - #1924 
 - #1910 (Draft PR #1983)
 
The current logic for publish depends on
1. getting the titles of existing entities
2. generating titles for records returned by Tripal/ChadoStorage->findValues()
3. finding the set of titles from 2 that is not in 1 and publishing them.

As mentioned in #1910, this has serious performance/memory needs when publishing large numbers of records and requires support of batches. However, both 1 and 2 are large enough to cause issues which causes the above logic to fall apart. Furthermore, #1984 and #1924 adds complexity to the ability to determine which chado records would belong to a content type in a performant way.

The current thought between myself and @dsenalik is to take this approach:
1. Find the unique ids of unpublished records via Tripal/ChadoStorage
2. Batch the set returned by 1 for the next step
3. Use Tripal/ChadoStorage->findValues restricted to these ids to get the full details of the records to be published.

In order to make as few changes as possible, #1983 currently focuses on finding a batch of ids but then still using the titles to determine what is published vs not. 

However, this will fail where there are duplicate titles or where there are too many existing entities. With this in mind it may be better to have a single function in Tripal/ChadoStorage that finds unpublished records. For ChadoStorage this may be easiest by using a join across chado + public schema.

Even with a `findUnpublishedRecordIds()` style method, we still need a generic-ish way to indicate how to filter down to the right set of records in a single chado table. The following scenarios at a minimum need to be supported:
1. content type contains all records from a single table
2. content type contains all records with a specific type_id from a single table
3. content type contains all records with a specific property type associated with that base table.

Additionally, #1984 indicates that both 1 and 2 can be the case for a single chado table 🙈 

This leads me to think maybe we want a way to indicate that a specific field is key to defining the bundle. For example, in many cases our ChadoAdditionalTypeTypeDefault is this field but in some cases other extension modules may need to define additional fields -especially for alternative data backends. Furthermore, some content types won't have any such fields and just contain all records from that base table.

dsenalik:
# Bug Fix

## Description
Chado Publish has been separated out from the core "publish" by creating a new `TripalBackendPublish` thanks to @laceysanderson . Different datastores can implement their own publish, but for now there is just a single `chado_storage` publish plugin implementation.

This PR resolves several things related to publishing:
1. Entities with duplicated titles - These no longer cause publish to fail. This can happen for the `pub` content type, because the title does not have a unique constraint.
2. Content types where multiple content types are derived from a single chado table - e.g. gene, mRNA are both based on the `chado.feature table`. - Two third party settings are added so that publish knows for a given bundle, which records to publish. New third party settings needed to implement this were added with PR #1992.
3. Add a flag to either publish only new entities, or to also republish existing entities. The latter is necessary if additional fields are added to a content type, or if the entity title format is changed. But in other cases it is much more efficient to skip existing published entities entirely.
4. Batch size is changeable on the publish form.
PR1991 2
5. A new drush command is available for publishing (thanks @laceysanderson) - for example:
 `drush tripal-chado:publish pub --republish --batch_size=2000`
6. Record IDs are sorted in PHP prior to generating batches. This is so we can use range conditions instead of the less efficient `IN` for the sql queries. The PHP sort has minimal overhead since it is a simple numeric array. Timing in my docker for sorting in PHP: `Elapsed time to sort 38511 records = 0.022515`

## Maybe not done in this PR but added later
1. Issue #2015 = Check the `pyear` column for the `pub` table and see why it is required on the form but not in the DB table. Step 2 of testing instructions is how we can bypass the issue for this PR.
2. Issue #2014 = Support publishing from another schema. This should be a different issue and PR because right now we only store an integer `record_id` in the drupal field tables, so there is no indication of which schema that record ID derives. Currently you cannot enter new content into a second chado schema through the UI.

## Testing?
1. Import all of the type collections
Tripal -> Page Structure -> +Import Type Collection
2. To bypass issue #2015, go to Tripal -> Page Structure -> Publication -> Manage Fields - and then for both "Title" and "Publication Year" uncheck the "Required Field" box and save.
3. Generate some content to publish with duplicated titles
`INSERT INTO chado.pub (title, uniquename, pyear, type_id) VALUES ('Duplicated pub title', 'Pub 1', '1901', 3107), ('Duplicated pub title', 'Pub 2', '1902', 3107), ('Duplicated pub title', 'Pub 3', '1903', 3107);`
(pyear is included because it is defined as required, although not at the db table level)
4. Tripal -> Content -> +Publish Tripal Content -> Publication
go ahead and set the batch size to `1` if you want to see the batch code in action
5. Run the job `drush trp-run-jobs --username=drupaladmin --root=/var/www/drupal/web`
```
2024-10-28 08:53:46
Tripal Job Launcher
Running as user 'drupaladmin'
-------------------
2024-10-28 08:53:46: Job ID 2.
2024-10-28 08:53:46: Calling: Drupal\tripal\TripalBackendPublish\PluginManager\TripalBackendPublishManager::runTripalJob(pub, chado_storage, Array)
 [notice] Initializing publish
 [notice] Finding all candidate records in the "pub" chado table
 [notice] Preparing to publish 4 records in 4 batches
...
 [notice] Publish completed. Published 4 new entities, and added 19 field values.
```
6. Verify that the "null" publication, and the three test publications have all been published
PR1991 1

### Test publishing a table with multiple bundles
7. Enter some projects into chado
`sitedb=> INSERT INTO chado.project (name) VALUES ('General Project 1'), ('General Project 2'), ('General Project 3'), ('Genome Project 1'), ('Genome Project 2'), ('Genome Project 3');`
`INSERT 0 6`
Add the property that makes three of them into genome projects
`sitedb=> INSERT INTO chado.projectprop (project_id, type_id, value) VALUES (4, (SELECT cvterm_id FROM chado.cvterm WHERE name='Genome Project' AND cv_id=(SELECT cv_id FROM chado.cv WHERE name='local')), 'Genome Project'), (5, (SELECT cvterm_id FROM chado.cvterm WHERE name='Genome Project' AND cv_id=(SELECT cv_id FROM chado.cv WHERE name='local')), 'Genome Project'), (6, (SELECT cvterm_id FROM chado.cvterm WHERE name='Genome Project' AND cv_id=(SELECT cv_id FROM chado.cv WHERE name='local')), 'Genome Project');`
`INSERT 0 3`
`sitedb=> select * from project P left join projectprop PP on P.project_id=PP.project_id;`
```
 project_id |       name        | description | projectprop_id | project_id | type_id |     value      | rank 
------------+-------------------+-------------+----------------+------------+---------+----------------+------
          4 | Genome Project 1  |             |              1 |          4 |      94 | Genome Project |    0
          5 | Genome Project 2  |             |              2 |          5 |      94 | Genome Project |    0
          6 | Genome Project 3  |             |              3 |          6 |      94 | Genome Project |    0
          2 | General Project 2 |             |                |            |         |                |     
          1 | General Project 1 |             |                |            |         |                |     
          3 | General Project 3 |             |                |            |         |                |     
(6 rows)
```
8. In this case we will use the drush command to publish
`drush tripal-chado:publish project`
` [notice] Publish completed. Published 6 new entities, and added 6 field values.`
Note that six are published - see issue #1984 - for this PR we will consider this "correct".
9. Publish the genome projects
`drush trp-chado-publish "genome_project"`
` [notice] Publish completed. Published 3 new entities, and added 6 field values.`
10. **Test republishing** Enter a contact through the UI named "Borg", and link it to the first two general projects.
11. Republish and verify that the contacts get added to the two projects
`drush tripal-chado:publish project --republish`
` [notice] Publish completed. Published 0 new entities, checked 6 existing entities, and added 2 field values.`
12. Rebuild cache before checking `drush cache:rebuild`
PR1991 3

### Try incorrect values - informative message expected
PR1991 4

### Import a tripal 3 chado dump and start publishing
See https://github.com/tripal/tripal/issues/1986#issuecomment-2397476913 for some steps needed before attempting this.
Instructions at https://tripaldoc.readthedocs.io/en/latest/upgrade_guide/site/migrating_chado.html

You can publish everything with these commands, and see how long each takes
```
time drush tripal-chado:publish analysis 
time drush tripal-chado:publish contact 
time drush tripal-chado:publish organism 
time drush tripal-chado:publish project 
time drush tripal-chado:publish protocol 
time drush tripal-chado:publish pub 
time drush tripal-chado:publish study 
time drush tripal-chado:publish breeding_cross 
time drush tripal-chado:publish germplasm 
time drush tripal-chado:publish germplasm_variety 
time drush tripal-chado:publish ril 
time drush tripal-chado:publish dna_library 
time drush tripal-chado:publish gene 
time drush tripal-chado:publish genome_annotation 
time drush tripal-chado:publish genome_assembly 
time drush tripal-chado:publish genome_project 
time drush tripal-chado:publish mrna 
time drush tripal-chado:publish phylotree 
time drush tripal-chado:publish physical_map 
time drush tripal-chado:publish genetic_map 
time drush tripal-chado:publish genetic_marker 
time drush tripal-chado:publish phenotypic_marker 
time drush tripal-chado:publish QTL 
time drush tripal-chado:publish sequence_variant 
time drush tripal-chado:publish array_design 
time drush tripal-chado:publish assay 
time drush tripal-chado:publish biosample 
```

You can test raising the batch size very high if you want to see the out of memory error on a docker container
`drush tripal-chado:publish pub --republish --batch-size=10000`
...
`PHP Fatal error:  Allowed memory size of 1077936128 bytes exhausted (tried to allocate 20480 bytes) in /var/www/drupal/web/core/lib/Drupal/Core/Database/StatementWrapperIterator.php on line 297`

`Fatal error: Allowed memory size of 1077936128 bytes exhausted (tried to allocate 20480 bytes) in /var/www/drupal/web/core/lib/Drupal/Core/Database/StatementWrapperIterator.php on line 297`
`PHP Fatal error:  Allowed memory size of 1077936128 bytes exhausted (tried to allocate 45056 bytes) in /var/www/drupal/vendor/composer/ClassLoader.php on line 576`

`Fatal error: Allowed memory size of 1077936128 bytes exhausted (tried to allocate 45056 bytes) in /var/www/drupal/vendor/composer/ClassLoader.php on line 576`


dsenalik:
You may notice that if you republish the `pub` bundle, the null publication gets published again every time.
This will be fixed by issue #2015 - I added step 2 to testing instructions to bypass this for now.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -411,6 +411,22 @@ public static function storageSettingsFormSubmitBaseTable(array $form, FormState
...
+    $type_fkey = $settings['storage_plugin_settings']['type_fkey'] ?? NULL;
+    if ($type_fkey) {
+      $type_table = $entity_type->getThirdPartySetting('tripal', 'chado_type_table');
+      if (!$type_table) {
+        list($type_table, $type_column) = explode(self::$table_column_delimiter, $type_fkey, 2);
+        $entity_type->setThirdPartySetting('tripal', 'chado_type_table', $type_table);
+        $entity_type->setThirdPartySetting('tripal', 'chado_type_column', $type_column);
+        $entity_type->save();
+      }
+    }
```
laceysanderson: This no longer needs to be here, right? We removed it in #1992 in favour of adding it in the type field itself.

laceysanderson (commented):
Running through the testing instructions produced the expected results 🎉 
✅ 3 publications with the same title are able to be published
✅ 6 project records are able to be published
✅ 3 Genome Project records can be published that were also published as projects
✅ 2 contacts were created through the UI and connected to projects that a) when published was run without the re-publish option, no additional entities were created and b) when publish was run with --republish then the contact fields were published on the projects correctly.
✅ The publish form calls the new publish plugin
✅ The publish drush command works for existing bundles
✅ Publish provides an exception when provided either a fake bundle or a fake data backend

✅ Code review looks good too. Only one possible artifact jumped out at me and no concerns.
✅ Code coverage also still looks good

I am testing this on my KnowPulse T4 clone right now and will get back to you when it completes (I had to restart it unfortunately due to a forced IT OS update).

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: D'oh 😲, yes, good catch, it should be removed. Plus I later renamed those to bundle_type_table and bundle_type_column

Fixed with https://github.com/tripal/tripal/pull/1991/commits/d7a1d848cc2b0c703ea17b625a549c87c7c4331d

laceysanderson (approved):
Publishing KnowPulse completed without error! 🎉 

Some of the timings include:

- 4m for 308 contacts (1,836 field values)
- 432m for 1,195,372 genetic markers (11,070,792 field values)
- 406m for 1,245,010. sequence variants (11,657,969 field values)

✅ I also confirmed a random selection of 10 pages and checked that I could use the Tripal content listing including going to the last page.

PR #1992, opened 2024-10-07
Add new third party settings needed by publish

### LINKED ISSUE #1994: Check for new fields fails on certain content types
### Tripal Version

4.x

### Drupal Version

10.3.7-dev 

### PostgreSQL Version

16.4 (Debian 16.4-1.pgdg110+2) 

### PHP Version

8.3.12

### General Description

Checking for new fields on some content types works fine but when doing the same for analysis content types, there is a fatal exception due to the array key 'foreign_key' not existing in the TripalDbxSchema::getForeignKeyDef() function.

### Steps to reproduce

Certain basic content types like organism and contact work OK, but ones that are a little more complicated, such as those with foreign keys related to other content types, such as analysis and project, fail.

To test, click the "Check for new fields" for organisms (admin/structure/bio_data/manage/organism/fields) and see that it works, and then do the same for analysis (admin/structure/bio_data/manage/analysis/fields) and see that a fatal exception occurs.

### Error messages

```shell
TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in array_key_exists() (line 707 of modules/contrib/tripal/tripal/src/TripalDBX/TripalDbxSchema.php).

Drupal\tripal\TripalDBX\TripalDbxSchema->getForeignKeyDef('analysis', 'organism') (Line: 726)
Drupal\tripal\TripalDBX\TripalDbxSchema->foreignKeyExists('analysis', 'organism') (Line: 308)
Drupal\tripal_chado\Plugin\Field\FieldType\ChadoOrganismTypeDefault::discover(Object, 'chado_organism_type_default', Array) (Line: 233)
```

### Screenshots

_No response_

dsenalik:
# New Feature

## Description
This PR adds two new third party settings intended for use by the publish service.
These two settings are used for publishing content types where more than one bundle is based on a single chado base table. For example, both the "gene" and "mRNA" content types are derived from the `feature` table.
The third party settings allow publish to know where to find the term that defines the bundle, whether in the base table, or in a property table.

## Testing?
This PR isolates the described changes from other changes in the publish PR #1991 to simplify review.

Testing involves:
1. Build a docker on 4.x branch
2. Import all of the type collections
3. Switch to this branch and test the update hook `drush updatedb`
4. Verify that the settings have now been added
```
$bundles = ['genetic_map', 'QTL', 'sequence_variant', 'genetic_marker', 'phenotypic_marker',
            'gene', 'mrna', 'physical_map', 'genome_assembly', 'genome_annotation', 'genome_project',
            'germplasm', 'breeding_cross', 'germplasm_variety', 'ril'];
foreach ($bundles as $bundle) {
  $entity_type_manager = \Drupal::entityTypeManager();
  $entity_type = $entity_type_manager->getStorage('tripal_entity_type')->load($bundle);
  $entity_type_chado_base_table = $entity_type->getThirdPartySetting('tripal', 'chado_base_table');
  $entity_type_bundle_type_table = $entity_type->getThirdPartySetting('tripal', 'bundle_type_table');
  $entity_type_bundle_type_column = $entity_type->getThirdPartySetting('tripal', 'bundle_type_column');
  print "bundle=\"$bundle\"\n";
  print "  base_table=\"$entity_type_chado_base_table\"\n";
  print "  type_table=\"$entity_type_bundle_type_table\"\n";
  print "  type_column=\"$entity_type_bundle_type_column\"\n";
}
```
expected output
```
bundle="genetic_map"
  base_table="featuremap"
  type_table="featuremapprop"
  type_column="type_id"
bundle="QTL"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="sequence_variant"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="genetic_marker"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="phenotypic_marker"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="gene"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="mrna"
  base_table="feature"
  type_table="feature"
  type_column="type_id"
bundle="physical_map"
  base_table="featuremap"
  type_table="featuremapprop"
  type_column="type_id"
bundle="genome_assembly"
  base_table="analysis"
  type_table="analysisprop"
  type_column="type_id"
bundle="genome_annotation"
  base_table="analysis"
  type_table="analysisprop"
  type_column="type_id"
bundle="genome_project"
  base_table="project"
  type_table="projectprop"
  type_column="type_id"
bundle="germplasm"
  base_table="stock"
  type_table="stock"
  type_column="type_id"
bundle="breeding_cross"
  base_table="stock"
  type_table="stock"
  type_column="type_id"
bundle="germplasm_variety"
  base_table="stock"
  type_table="stock"
  type_column="type_id"
bundle="ril"
  base_table="stock"
  type_table="stock"
  type_column="type_id"
```
5. And for good measure, a docker built on this branch should already include these settings.

### Test creating a new content type - adding type field manually
1. Creating a new content type. Tripal -> Page Structure -> "+ Add custom type"
2. I created a "Tripal Project" content type based on the project table.
3. Add a "Tripal Project Name" field using the "Chado String Field Type" field - this gets the base table populated.
4. Add a "Tripal Project Type" field using the "Chado Type Reference" field. Term should match the bundle term. You will want to check the new box "This term defines the bundle"
issue1924_gui-term-defines-bundle

### Test creating a new content type - adding type field manually
1-3. Repeat the steps 1-3 above, but then
4. Discover new fields
5. Edit the type field and you can then check the  "This term defines the bundle" box and save.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -307,7 +307,7 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
     // publish will be able to restrict by the term.
     $type_fkey = $settings['storage_plugin_settings']['type_fkey'] ?? NULL;
     if ($type_fkey) {
-      $type_table = $entity_type->getThirdPartySetting('tripal', 'chado_type_table');
+      $type_table = $entity_type->getThirdPartySetting('tripal', 'bundle_type_table');
       if (!$type_table) {
         $form_state_storage = $form_state->getStorage();
         $bundle = $form_state_storage['bundle'];
@@ -317,8 +317,8 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
         $entity_type = $entity_type_manager->getStorage('tripal_entity_type')->load($bundle);
 
         list($type_table, $type_column) = explode(self::$table_column_delimiter, $type_fkey, 2);
-        $entity_type->setThirdPartySetting('tripal', 'chado_type_table', $type_table);
-        $entity_type->setThirdPartySetting('tripal', 'chado_type_column', $type_column);
+        $entity_type->setThirdPartySetting('tripal', 'bundle_type_table', $type_table);
+        $entity_type->setThirdPartySetting('tripal', 'bundle_type_column', $type_column);
         $entity_type->save();
       }
     }
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -504,7 +504,7 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
         $prop_type_id = $prop_table_def['fields'][$prop_type_column] ?? NULL;
         if ($prop_type_id) {
           $type_table = $prop_table;
-          $type_column = $prop_type_columns;
+          $type_column = $prop_type_column;
         }
       }
     }
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -291,37 +291,74 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
   }
 
   /**
-   * {@inheritDoc}
+   * Saves the location of the bundle term in the entity for easy access by publish.
+   *
+   * @param string $bundle
+   *   The bundle identifier, e.g. "analysis" or "project"
+   * @param string $type_table
+   *   The table where the term is stored, usually the base table or else a property table.
+   * @param string $type_column
+   *   The name of the column where the term is stored. Usually this is "type_id".
    */
-  public static function storageSettingsFormSubmitBaseTable(array $form, FormStateInterface $form_state) {
-    parent::storageSettingsFormSubmitBaseTable($form, $form_state);
+  public static function setEntityBundleType(string $bundle, string $type_table, string $type_column) {
+    /** @var \Drupal\Core\Entity\EntityTypeManager $entity_type_manager **/
+    $entity_type_manager = \Drupal::entityTypeManager();
+    /** @var \Drupal\tripal\Entity\TripalEntityType $entity_type **/
+    $entity_type = $entity_type_manager->getStorage('tripal_entity_type')->load($bundle);
 
-    $settings = self::getFormStateSettings($form_state);
-    if (!array_key_exists('storage_plugin_settings', $settings)) {
-      return;
-    }
-    // This field is used to specify a subset of a chado table for the bundle,
-    // e.g. for feature table it could be 'gene'.
-    // The first time this field is entered, store third party settings in the
-    // entity for the table and column where this term will be stored, so that
-    // publish will be able to restrict by the term.
-    $type_fkey = $settings['storage_plugin_settings']['type_fkey'] ?? NULL;
-    if ($type_fkey) {
-      $type_table = $entity_type->getThirdPartySetting('tripal', 'bundle_type_table');
-      if (!$type_table) {
-        $form_state_storage = $form_state->getStorage();
-        $bundle = $form_state_storage['bundle'];
-        /** @var \Drupal\Core\Entity\EntityTypeManager $entity_type_manager **/
-        $entity_type_manager = \Drupal::entityTypeManager();
-        /** @var \Drupal\tripal\Entity\TripalEntityType $entity_type **/
-        $entity_type = $entity_type_manager->getStorage('tripal_entity_type')->load($bundle);
+    $entity_type->setThirdPartySetting('tripal', 'bundle_type_table', $type_table);
+    $entity_type->setThirdPartySetting('tripal', 'bundle_type_column', $type_column);
+    $entity_type->save();
+  }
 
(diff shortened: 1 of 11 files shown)


laceysanderson (commented):
I started a fresh docker on this branch then
✅ confirmed that the settings were set properly for all the existing general content types
✅ imported all the other content type collections and confirmed they're settings were correct as well.
❌ Tried to create a new custom content type to check it but got this validation error since type is in the default URL pattern but a new content type doesn't have any fields yet 🙈 This is not related to this PR but must have slipped in elsewhere.
image
❌ Add a type field and have the 3rd party settings added.
Your instructions may not have been clear enough for me to get this working right 😞 Since it is really finicky we should look at making the type field self discoverable so people who create custom bundles can use the check fields to have this set properly.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -411,6 +411,22 @@ public static function storageSettingsFormSubmitBaseTable(array $form, FormState
...
+    $type_fkey = $settings['storage_plugin_settings']['type_fkey'] ?? NULL;
+    if ($type_fkey) {
+      $type_table = $entity_type->getThirdPartySetting('tripal', 'chado_type_table');
+      if (!$type_table) {
+        list($type_table, $type_column) = explode(self::$table_column_delimiter, $type_fkey, 2);
+        $entity_type->setThirdPartySetting('tripal', 'chado_type_table', $type_table);
+        $entity_type->setThirdPartySetting('tripal', 'chado_type_column', $type_column);
+        $entity_type->save();
+      }
+    }
```
laceysanderson: I feel like this should be in the additional type field. Then it doesn't have to do this check. That field can just override this method, call the parent version first and then do this.

tripal_chado/tripal_chado.install
```diff
@@ -575,3 +575,44 @@ function tripal_chado_update_10406() {
...
+    'genetic_map' => 'featuremapprop',
+    'physical_map' => 'featuremapprop',
+    // project table
+    'genome_project' => 'projectprop',
+    // stock table
+    'germplasm' => 'stock',
+    'breeding_cross' => 'stock',
+    'germplasm_variety' => 'stock',
+    'ril' => 'stock',
+  ];
```
laceysanderson: What about Tripal 4 sites that have their own custom content types already? They will not get touched by this update hook.

laceysanderson (commented):
I worry there is no way for the 3rd party setting to be added for a content type created through YAML by an extension module who doesn't know to provide that info in the yaml file 🤔

dsenalik:
I forgot to mention that validation problem in my testing instructions 🙊 - I made an issue for it #1996
I will look at your other comments...

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: This has been moved

dsenalik: And the type field is now discoverable

dsenalik:
> I worry there is no way for the 3rd party setting to be added for a content type created through YAML by an extension module who doesn't know to provide that info in the yaml file 🤔

This can now be done by adding or editing the additional type field.

tripal_chado/tripal_chado.install
dsenalik: > What about Tripal 4 sites that have their own custom content types already? They will not get touched by this update hook.

The update hook is now generic - it uses the "fixed_type" settings.

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
```diff
@@ -394,4 +465,99 @@ public function isCompatible(TripalEntityType $entity_type) : bool {
...
+        'storage_plugin_id' => 'chado_storage',
+        'storage_plugin_settings' => [
+          'base_table' => $base_table,
+          'type_table' => $type_table,
+          'type_column' => $type_column,
+        ],
+      ],
+      'settings' => [
+        'termIdSpace' => $termIdSpace,
+        'termAccession' => $termAccession,
```
laceysanderson, line 541: The newly discovered field should have the fixed value set by default. This allows that checkbox to be checked by default and makes the type table/column be automatically detected when the type field is added by check fields rather then requiring an edit to the field.

I know there are cases where this field doesn't define the content type but that feels like the exception rather then the default 🤔 and you do specifically mention the fixed value in the field description you generate :wink'

```suggestion
        'termAccession' => $termAccession,
        'fixed_value' => $fixed_value,
```

laceysanderson (commented):
This does now work when adding a custom content type and I like that a simple edit will fix things if your YAML was missing information!

That said, it took me a bit of detective work to find the checkbox I needed to select since it was hidden in the controlled vocabulary term fieldset which is collapsed by default... that said this is an appropriate place to put it 🤷‍♀️ so I think I'm ok with that as long as the discovered field has this checked by default rather then requiring edit. Alternatively, we could add a drupal message letting people know they may want to select this for the type field specifically when it is edited?

tripal_chado/tripal_chado.install
laceysanderson: Confirmed resolved in new update hook.

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
dsenalik: I removed the default of checked with commit https://github.com/tripal/tripal/pull/1992/commits/0bdb0f8801a19702fb33a4d6ec9b857e085e2147
My thinking was that this was a safer default, however on further consideration I agree with your thinking that it is a sensible default, as those other use cases will be uncommon.

laceysanderson (approved):
Tested again and it looks good 👍

PR #1993, opened 2024-10-07
Add in schema for options all supported tripal_layout field group options.

laceysanderson:
# Bug Fix

## Description

I missed this in the original PR referenced 🙈 We need schema definitions for all options supported in the YAML but the original PR only included the main ones we had previously used and that are represented in the test yamls. This PR adds all the supported ones.

Note: this came about because we are using the Tripal Layout in our TripalCultivate extension module and it's test yamls have a description. As such all our tests are failing with a missing schema note about the description. 
Screenshot 2024-10-07 at 4 40 57 PM

## Testing?

This is a schema only change. It is much more difficult to test manually then it is worth. As such, code review and confirm it doesn't break existing tests.

### DIFF
--- a/tripal_layout/config/schema/tripal_layout.schema.yml
+++ b/tripal_layout/config/schema/tripal_layout.schema.yml
@@ -87,13 +87,13 @@ tripal_layout_field_group_item:
       label: "The label of this field group"
     label_visibility:
       type: string
-      label: "Indicates if the label should be visable."
+      label: "Indicates if the label should be visible."
     empty_label_behavior:
       type: string
       label: "Indicates how to handle an empty label"
     region:
       type: string
-      label: "Region to put the field group in (i.e. 'Content')"
+      label: "Region to put the field group in (e.g. 'Content')"
     weight:
       type: integer
       label: "The weight to give this field group.  Field groups will be displayed in order of their weights with larger weights being lower on the list."
@@ -111,7 +111,7 @@ tripal_layout_field_group_item:
       label: "Field group caption or text description."
     desc_visibility:
       type: string
-      label: "Indicates if the description should be visable."
+      label: "Indicates if the description should be visible."
     first_column:
       type: string
       label: "First Column Header; only applies to field group tables."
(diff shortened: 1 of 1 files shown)


tripal_layout/config/schema/tripal_layout.schema.yml
```diff
@@ -85,15 +85,51 @@ tripal_layout_field_group_item:
     label:
       type: string
       label: "The label of this field group"
+    label_visibility:
+      type: string
+      label: "Indicates if the label should be visable."
```
dsenalik:
```suggestion
      label: "Indicates if the label should be visible."
```

tripal_layout/config/schema/tripal_layout.schema.yml
```diff
@@ -85,15 +85,51 @@ tripal_layout_field_group_item:
...
+    empty_label_behavior:
+      type: string
+      label: "Indicates how to handle an empty label"
+    region:
+      type: string
+      label: "Region to put the field group in (i.e. 'Content')"
```
dsenalik: should this be e.g. instead of i.e.?
```suggestion
      label: "Region to put the field group in (e.g. 'Content')"
```

dsenalik (approved):
Minor typos but once tests finish I approve

tripal_layout/config/schema/tripal_layout.schema.yml
```diff
@@ -85,15 +85,51 @@ tripal_layout_field_group_item:
...
       label: "Set to true if this field group should be shown even if all of it's children are empty"
     open:
       type: boolean
       label: "Set to true if this field group should be open by default. False if closed"
+    description:
+      type: string
+      label: "Field group caption or text description."
+    desc_visibility:
+      type: string
+      label: "Indicates if the description should be visable."
```
laceysanderson: Same typo brought up above by Doug 🙈 
```suggestion
      label: "Indicates if the description should be visible."
```

tripal_layout/config/schema/tripal_layout.schema.yml
laceysanderson: I think both work but you are likely more correct so we'll go with that :-)

laceysanderson:
Thanks for the quick review @dsenalik!

PR #2002, opened 2024-10-14
Fix url template for custom content types

### LINKED ISSUE #1996: Invalid URL template when creating a new content type through the UI
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

If you create a new content type, the default URL template is
`[type]/[TripalEntity__entity_id]`
But `[type]` is not yet available, so you get a validation error
See comment https://github.com/tripal/tripal/pull/1992#pullrequestreview-2357779357 for a screenshot

### Steps to reproduce

1. Create a new content type and hit "save"

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
The current default URL template `[type]/[TripalEntity__entity_id]` is not valid when you create a new custom content type.
This PR does two things
1. Change the default URL template to `[TripalBundle__bundle_id]/[TripalEntity__entity_id]`, which will now work as expected.
2. Adds two more tokens `[TripalEntityType__term_namespace]` and `[TripalEntityType__term_accession]`

## Testing?
1. Create a new content type. I used "Primer" based on the "feature" table. However I would recommend a two-word type to test something with spaces, i.e. "Oligonucleotide Primer" would be great. For example
PR2002finishedprimer
2. You can go ahead and add the required fields for a content type based on the feature table: name, uniquename, organism, type_reference
3. Create the required organism
4. Create a primer entity and confirm that the url alias is, in this case, `oligonucleotide_primer/2`

### DIFF
--- a/tripal/src/Entity/TripalEntityType.php
+++ b/tripal/src/Entity/TripalEntityType.php
@@ -494,7 +494,7 @@ class TripalEntityType extends ConfigEntityBundleBase implements TripalEntityTyp
       return $this->url_format;
     }
     else {
-      return '[type]/[TripalEntity__entity_id]';
+      return 'bio_data/[TripalEntity__entity_id]';
     }
   }
 
--- a/tripal/src/Entity/TripalEntityType.php
+++ b/tripal/src/Entity/TripalEntityType.php
@@ -528,10 +528,10 @@ class TripalEntityType extends ConfigEntityBundleBase implements TripalEntityTyp
 
     // ID Tokens:
     if ($options['include id'] == TRUE) {
-      $token = '[TripalBundle__bundle_id]';
+      $token = '[TripalEntityType__entity_id]';
       $tokens[$token] = [
-        'label' => 'Bundle ID',
-        'description' => 'The unique identifier for this Tripal Content Type.',
+        'label' => 'Content Type/Bundle ID',
+        'description' => 'The machine name for this Tripal Content Type. By default this will be similar to the label you entered. For example, if you created a content type with the label "Genome Annoation" then it\'s machine name/id would be "genome_annotation".',
         'token' => $token,
         'field_name' => NULL,
         'required' => TRUE,
--- a/tripal/src/Entity/TripalEntityType.php
+++ b/tripal/src/Entity/TripalEntityType.php
@@ -494,7 +494,7 @@ class TripalEntityType extends ConfigEntityBundleBase implements TripalEntityTyp
       return $this->url_format;
     }
     else {
-      return '[TripalBundle__bundle_id]/[TripalEntity__entity_id]';
+      return '[TripalEntityType__term_label]/[TripalEntity__entity_id]';
     }
   }
 
@@ -540,7 +540,7 @@ class TripalEntityType extends ConfigEntityBundleBase implements TripalEntityTyp
       $token = '[TripalEntity__entity_id]';
       $tokens[$token] = [
         'label' => 'Content/Entity ID',
-        'description' => 'The unique identifier for an individual piece of Tripal Content.',
+        'description' => 'The unique identifier for an individual piece of Tripal Content. This will be unique for each Tripal Content page and is an integer.',
         'token' => $token,
         'field_name' => NULL,
         'required' => TRUE,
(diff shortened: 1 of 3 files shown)


laceysanderson (commented):
This works as described ✅ 

However, it is a bit disorienting for the URL scheme to change without telling the user. Especially as it's hidden in a vertical tabs so on editing you would only notice if you open it up.

Instead, perhaps we should just add some more tokens:
- [TripalEntityType__term_name]
- [TripalEntityType__term_accession]
- [TripalEntityType__machine_name]

Then the URL schema can be `[TripalEntityType__term_name]/[TripalEntity__entity_id]` right off the bat without any worry of it not existing yet and no need to change it.

This would be done by adding the tokens like this: https://github.com/tripal/tripal/blob/4.x/tripal/src/Entity/TripalEntityType.php#L550-L558

And then updating the parser to handle theme here: https://github.com/tripal/tripal/blob/4.x/tripal/src/Services/TripalTokenParser.php#L222-L234

dsenalik:
So it turns out that the token that we need for this already exists, it is the `[TripalBundle__bundle_id]` one.

Well, since I am in here anyway, I did add two new tokens `[TripalEntityType__term_namespace]` and `[TripalEntityType__term_accession]` but I did not implement the machine name one, mainly because I don't know what it is supposed to be 😐.
PR2002_url_format

laceysanderson (commented):
I hope you're ok with me making a few changes @dsenalik? If not, feel free to revert or change :-)

Specifically, I
- fixed an inconsistency between `[TripalBundle__bundle_id]` and `[TripalEntityType__label]` which both refer to TripalContent Type and thus should use the same naming convention. See commit b9ab7b116fbb82f0230a046daef7d2d44234fab1
- Expanded descriptions to make each token more clear. The `[TripalBundle__bundle_id]` token is actually the machine name I referred to. I'm guessing the difference was not clear because we often use the the term name as the content type label. See my screenshots below to see the difference.
- Added a `[TripalEntityType__term_label]` token to be the cvterm.name / term name that we intended to be included in the default token.
- Moved the new tokens added by this PR out of the 'include id' if block since they are not IDs and thus should be available for use in the title.

Now when I test creating a new content type I see the new default URL template (shown here on edit since I didn't save my create screenshot 🤦 )
image

And when I create a primer page, I see the token is replaced properly:
Screenshot 2024-10-22 at 11 34 57 AM

laceysanderson (changes requested):
I did notice another issue though 🙈 I don't think this is new to this PR but it was aggravated by it. Specifically, if you try to create a new content type right now, the title format will fail validation. Technically that is because it has now gotten too long but more problematically, it is because we have no fields yet and therefore the default title pattern just includes all non-field tokens which A) is too long and B) is not unique 🙈 🫢

dsenalik:
> * Moved the new tokens added by this PR out of the 'include id' if block since they are not IDs and thus should be available for use in the title.

The reason I did that, keep them in the `include id` test, was because otherwise they got added to the default title and it was a mess. As you have noted this causes problems.

laceysanderson:
> > Moved the new tokens added by this PR out of the 'include id' if block since they are not IDs and thus should be available for use in the title.
>
> The reason I did that, keep them in the include id test, was because otherwise they got added to the default title and it was a mess. As you have noted this causes problems.

Ahhhh... I see. But they should be available 🙈 It sounds like we need to create a new issue for the title format when there are no fields. For now, I can add them back into an if block with a todo to remove once the title format issue is resolved.

Sound good?

dsenalik:
That would be okay. Or alternatively we could remove the new tokens entirely for now, and add them back with a title fix.
This would make this into a 1-line PR for now

dsenalik:
Well, you worked on the descriptions, so let's keep them

laceysanderson:
> That would be okay. Or alternatively we could remove the new tokens entirely for now, and add them back with a title fix.
>
> This would make this into a 1-line PR for now

We need the term_label token. The original bundle_id token you used is the machine name not the term. That's why I fixed the descriptions ;-p

Ok, I will add the commit and make the new issue.

laceysanderson (approved):
This is now described in issue #2009 and the change is made allowing the title to continue quietly being a problem 🙈 

image

Can you give this a try @dsenalik to confirm it works on your end just to make sure? I approve the current state of the code.

dsenalik:
It is working as desired.
I tried pushing the boundaries, using terms with spaces to define the content type, such as "`INR motif`" and this works just fine
http://localhost:8096/INR_motif/4
Editing the title format before creating the entity gives expected result.
Failing to edit it gives every entity the same name as the bundle, again as expected.

laceysanderson:
Thanks for the confirmation @dsenalik and for all the work on this!

PR #2013, opened 2024-10-25
Inhibit additional type field discovery if field already exists

### LINKED ISSUE #2012: Additional type discovery should only suggest addition if it does not already exists
### Task Description

This is a reminder to improve the discovery function for the chado additional type field.
On a few content types this field is present, but the discovery offers to add a second instance of this field, e.g. organism, contact. It looks like this for contact:
issue2012 1

The current check `$existing = $this->checkDiscoveredTerm($discoveredIdSpace, $discoveredAccession, $entity_field_defs);` is based on the CV term for the field, so if the existing field uses a term other than default, it will be suggested for addition again.

### Branch Name

tv4g1-issue2012-additional_type_discovery

dsenalik:
# New Feature - or - Bug Fix depending on your interpretation ...

## Description
For the additional type field, this PR will no longer discover the field if an instance of the field already exists for the bundle.

The reason behind this is that this field is **usually** used to indicate the CV term for multiple bundles based on a single chado table. For example, "gene' and "mRNA" bundles from the "feature" table.

But in some cases the field is used to indicate a type. For organism, the indicated type is the infraspecific type (subspecies, cultivar, etc.), or for contact the type is the contact type (Person, Institute, etc.)

If we discover a second field, then we are effectively defining a new bundle type, and this is likely not the intention in these latter cases. And doing this may break publishing later.

So, this PR does the following: if the field already exists on the bundle, we shall choose to not automatically discover this field. This does not prevent it from being added through the GUI, however, if we really did want to separate the "organism" table into more than one bundle.

The only problem with the changes here, is that if discovery is "inhibited" as is done here, then the existing field does not show up in the "Existing fields" list, mainly because it uses a different CV term. This is a "discovery", not a "list what's there". Is that bad? Please discuss.

pr2013 1

## Testing?
1. Tripal -> Page Structure -> Contact -> Manage Fields
2. "+Check for New Fields"
3. No new fields will be suggested on this branch. On 4.x the "Type" field will be offered for addition. See Issue #2012 for screenshot.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -515,6 +515,15 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
       return $field_list;
     }
 
+    // If any instance of this field already exists on this content type,
+    // we will not suggest it for discovery. If there is a need for multiple
+    // instances, then they will need to be added manually through the GUI.
+    // Note that this will also block the existing field from showing up in
+    // the "Existing Fields" list, though.
+    if (array_key_exists(self::$id, $field_definitions)) {
+      return $field_list;
+    }
+
     // Create a field entry in the list
     $termIdSpace = $bundle->getTermIdSpace();
     $termAccession = $bundle->getTermAccession();
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -10,6 +10,7 @@ use Drupal\tripal\Entity\TripalEntityType;
 use Drupal\core\Form\FormStateInterface;
 use Drupal\Core\Ajax\AjaxResponse;
 use Drupal\Core\Ajax\ReplaceCommand;
+use Drupal\tripal\Services\TripalFieldCollection;
 
 /**
  * Plugin implementation of Tripal additional type field type.
@@ -469,7 +470,7 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
    * {@inheritDoc}
    * @see \Drupal\tripal\TripalField\Interfaces\TripalFieldItemInterface::discover()
    */
-  public static function discover(TripalEntityType $bundle, string $field_id, array $field_definitions) : array {
+  public static function discover(TripalEntityType $bundle, string $field_id, array $field_types, array $field_instances): array {
 
     /** @var \Drupal\tripal_chado\Database\ChadoConnection $chado **/
     $chado = \Drupal::service('tripal_chado.database');
@@ -515,12 +516,14 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
       return $field_list;
     }
 
-    // If any instance of this field already exists on this content type,
-    // we will not suggest it for discovery. If there is a need for multiple
-    // instances, then they will need to be added manually through the GUI.
-    // Note that this will also block the existing field from showing up in
-    // the "Existing Fields" list, though.
-    if (array_key_exists(self::$id, $field_definitions)) {
+    // Check for existing fields of this type.
+    if (array_key_exists(self::$id, $field_types)) {
+      // If yes, then add it to the field list.
+      foreach ($field_instances as $instance) {
+        if ($instance->getType() == self::$id) {
+          $field_list[] = TripalFieldCollection::getFieldArrayFromFieldInstance($instance);
+        }
+      }
       return $field_list;
     }
 
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -518,13 +518,15 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
 
     // Check for existing fields of this type.
     if (array_key_exists(self::$id, $field_types)) {
-      // If yes, then add it to the field list.
+      // If an existing field exists, then add it to the field list and return.
       foreach ($field_instances as $instance) {
         if ($instance->getType() == self::$id) {
           $field_list[] = TripalFieldCollection::getFieldArrayFromFieldInstance($instance);
         }
       }
-      return $field_list;
+      if ($field_list) {
+        return $field_list;
+      }
     }
 
     // Create a field entry in the list
--- a/tripal/src/Form/TripalContentFieldsForm.php
+++ b/tripal/src/Form/TripalContentFieldsForm.php
@@ -117,7 +117,7 @@ class TripalContentFieldsForm implements FormInterface {
       . 'discovered but which are already attached to this content type. They '
       . 'are shown here but cannot be added again.');
     if (empty($existing_fields)) {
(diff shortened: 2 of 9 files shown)


laceysanderson:
> The only problem with the changes here, is that if discovery is "inhibited" as is done here, then the existing field does not show up in the "Existing fields" list, mainly because it uses a different CV term. This is a "discovery", not a "list what's there". Is that bad? Please discuss.

This is interesting and non-intuitive 🤔 Especially that the infraspecific type field does not show up here by default 🤔 I will keep this in mind as I review and see if something twinkles in the back of my mind solution-wise 🙃

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
```diff
@@ -515,6 +515,15 @@ public static function discover(TripalEntityType $bundle, string $field_id, arra
...
 
+    // If any instance of this field already exists on this content type,
+    // we will not suggest it for discovery. If there is a need for multiple
+    // instances, then they will need to be added manually through the GUI.
+    // Note that this will also block the existing field from showing up in
+    // the "Existing Fields" list, though.
+    if (array_key_exists(self::$id, $field_definitions)) {
+      return $field_list;
+    }
+
```
laceysanderson: Perhaps instead of inhibiting detection here, we should check for existing instances of this field and return them in the discover if they are there. Then we can also grab the column information to determine if we have any type columns that do not yet have a field of this type.

something like
```
// Check for existing fields of this type.
if (array_key_exists(self::$id, $field_definitions)) {
  // If yes, then add it to the field list
  // @todo check if this is the right method to get the same type of array we create below.
  $field_list[] = $field_definitions[self::$id]->getSettings(); 

  // Now, check which chado columns the existing fields are managing
  // and save this info for latter.
}

// Get the table definition for this base table.
// Foreach column that is a cvterm foreign key, check if it has a field above
// if not, add the field as done below.
// @todo think about if we should use the term associated via semweb here rather then the bundle term 🤔 
```

laceysanderson (commented):
So I confirmed the approach here does what it was described to do.
Specifically, on this branch, I go to the organism > Manage Fields and click on Check for new fields. There is no type field in the new or existing sections.

However, I am thinking more and more that we want a solution that goes a bit further. While the field collection checks for existing fields based on the term, it does also give us all the field definitions. With that information, we can check for fields using the field type we define and return them in the field list. I believe this is how it was designed so that fields would show up in the existing.

Also, since we know that this field is chado-dependant we can use that information to confirm beyond the term associated with the field in order to suggest a field for all cvterm fkeys.

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
dsenalik: This is what we have for `$field_definitions[self::$id]` = can't call `getSettings()` on this
PR2013 1

we need an array suitable to pass to `addBundleField()`, as defined in the yaml

If we had the id `contact_type` (which we do not in the array above) we could use
```
use Drupal\field\Entity\FieldStorageConfig;
$field_storage = FieldStorageConfig::loadByName('tripal_entity', 'contact_type');
```
which gives some of the info needed


The parent `discover()` has access to `$entity_field_defs` which specify existing fields.

laceysanderson: Darn, it looks like we should have passed $entity_field_defs into the discover method for field types instead (or in addition to $all_field_defs). That would give us access to the the full field definitions for each field rather then this array form that you showed above :-(

If you are willing I would add $entity_field_defs as an additional parameter to discover here: https://github.com/tripal/tripal/blob/4.x/tripal/src/Services/TripalFieldCollection.php#L233 and then we will have access to the information we need.

This would also change the interface and the two other fields that implement the discover. For the interface, the code at https://github.com/tripal/tripal/blob/4.x/tripal/src/TripalField/Interfaces/TripalFieldItemInterface.php#L127-L152 would change to something like:

```php
  /**
   * Finds new instances of this field for a given content type.
   *
   * Fields are added to Tripal using the tripal.tripalfield_collection.*
   * configuration file during installation of a module.  In some cases,
   * however, not all possible instances of a field can be added to a content
   * type at installation of the module.  This function can be called to discover
   * if new instances of a field are appropriate for a given content type.  An
   * example of this is the `chado_property_type_default` field.  This function
   * should examine its storage backend and return a list of new fields
   * instances that could be added to the content type (i.e., bundle).
   *
   * @param \Drupal\tripal\Entity\TripalEntityType $bundle
   *   The entity type object for which new field instances should be found.
   *
   * @param string $field_id
   *   The id of the field.
   *
   * @param array $field_types
   *   An array where each item defines a field type that is attached to this 
   *   bundle. Each item is itself an array of the annotation defined at the top
   *   of that particular field type class.
   * 
   * @param array $field_instances
   *   An array of FieldConfig objects where each object defines a field attached
   *   to this content type. This differs from $field_types in that (1) there
   *   will be multiple entries if there a multiple instances of a single field
   *   type and (2) it includes the specific settings needed to create the field.
   *   As such, the FieldConfig object can be used to add existing fields to the
   *   field list returned by this method.
   *
   * @return array
   *   An associative array that follows the same structure as expected by 
   *   tripal.tripalfield_collection.* configuration.
   */
  public static function discover(TripalEntityType $bundle, string $field_id, array $field_types, array $field_instances) : array;
```

laceysanderson: These changes have been made in 756108ad76493a8339082c6bbc98c68cc6e952dc

laceysanderson:
Based on the discussion in review with @dsenalik, I've added a commit that allows us to check for an existing field. You can see it in action in this PR for the Additional Type and Organism fields as shown in the organism and gene content types in the screenshots below:
Screenshot 2024-11-02 at 4 37 23 PM
Screenshot 2024-11-02 at 4 38 42 PM

dsenalik:
Wonderful @laceysanderson - this works great!

Commit https://github.com/tripal/tripal/pull/2013/commits/4fd1444665389782ecc243c288a3485ec74ac592 adds counts to the discovered fields tabs, so that you will know if there is anything there without having to click on it. Now looks like this with Claro
PR2013 20241103 1

So I tested on a "Genome Project" and see the desired output
PR2013 20241103 2

But on a plain "Project" it does not offer to add a type field. Something to work on ...
PR2013 20241103 3

dsenalik:
Commit https://github.com/tripal/tripal/pull/2013/commits/23948bd54883dea3fb9b225e9ca5367188407195 fixes the lack of a discovered field on project - the array key may exist even if a field doesn't. Add a check to only return if a field was found. Now it looks like this
PR2013 20241103 4

dsenalik:
I imported a gff https://github.com/tripal/tripal/files/15109905/issue1824.gff3.txt and tried to check for fields on gene content type
`TypeError: Drupal\tripal\TripalVocabTerms\TripalTerm::setDefinition(): Argument #1 ($definition) must be of type string, null given, called in /var/www/drupal/web/modules/contrib/tripal/tripal/src/TripalVocabTerms/TripalTerm.php on line 95 in Drupal\tripal\TripalVocabTerms\TripalTerm->setDefinition() (line 215 of modules/contrib/tripal/tripal/src/TripalVocabTerms/TripalTerm.php). `

This happens because the `definition` for the terms created by the importer is NULL rather than an empty string.
We can fix this easily by converting NULLs to empty strings with commit https://github.com/tripal/tripal/pull/2013/commits/ce6c7cd0bba6e1d41356e11814a5c8a3203ea5b6 and then we get something like this
PR2013 20241103 5

laceysanderson (approved):

PR #2018, opened 2024-11-01
Fixes bug in discovery, cv should have been db for termIdSpace

### LINKED ISSUE #2017: Some discovered property fields are reported as invalid
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

The field discovery for properties will often find some invalid fields.
It is a bug where the CV name is being used to set the `termIdSpace` when it should be the DB name.

### Steps to reproduce

This is showing up when migrating a tripal 3 site, and then you try to discover new fields. Gene is a good one to test, because it may have a lot of property fields to discover.

### Error messages

```shell
.
```

### Screenshots

property_discovery 1

dsenalik:
# Bug Fix

## Description
This is a 1-line PR
This was a small bug in field discovery, termIdSpace should refer to the DB name, but the CV name was used instead.

## Testing?
Not sure of an easy set of steps, this showed up on trying to migrate Tripal 3 sites, and then discovering new property fields.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php
@@ -305,7 +305,7 @@ class ChadoPropertyTypeDefault extends ChadoFieldItemBase {
           ],
         ],
         'settings' => [
-          'termIdSpace' => $recprop->cv_name,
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Good catch @dsenalik 🙈 
This is a small, obvious enough change that code review is sufficient.
Thank you!

PR #1927, opened 2024-07-16
Taxonomy loader split

dsenalik:
# Tripal 4 Core Dev Task
## Description
This divides the two functions of the NCBI Taxonomy Importer into two separate parts
1. Importing from NCBI
2. Taxonomy tree generator

Automated tests for most importer forms were almost the same, these have all been combined into one test which loops over all of those forms.

~~I know this is **really** slow!~~ Large performance improvement after adding chado buddy - see PR #1925 which references PR #1844

## Testing Part 1 - Taxonomy Importer
 1. We will want to test the update hook, so build your test docker on the **4.x branch**
 2. Now checkout this branch `git checkout tv4g4-issue1799-taxonomy_loader_split`
 3. Rebuild cache so that the new importer will show up in the menu `drush cache:rebuild`
 4. Run the update hook by running `drush updatedb`
```
root@26a95fa2ac4f:/var/www/drupal/web# drush updatedb
  -------------- ----------- --------------- ----------------------------------- 
  Module         Update ID   Type            Description                        
 -------------- ----------- --------------- ----------------------------------- 
  tripal_chado   10408       hook_update_n   10408 - Updates missing mapping    
                                             of CV terms to some chado table    
                                             columns. PR 1927 This is the same  
                                             as former                          
                                             tripal_chado_update_10402, but     
                                             here we are also adding a new      
                                             cvterm local:lineageex             
 -------------- ----------- --------------- ----------------------------------- 


 ┌ Do you wish to run the specified pending updates? ───────────┐
 │ ● Yes / ○ No                                                 │
 └──────────────────────────────────────────────────────────────┘

>  [notice] Update started: tripal_chado_update_10408
>  [notice] Update completed: tripal_chado_update_10408
>  [notice] Message: Reloaded Chado content terms
> 
>  [notice] Message: Reloaded Chado term mapping
> 
 [success] Finished performing updates.
```
 5. Verify new 'lineageex' term is now present
```
sitedb=> SELECT * FROM chado.cvterm WHERE name='lineageex';
 cvterm_id | cv_id |   name    |               definition                | dbxref_id | is_obsolete | is_relationshiptype 
-----------+-------+-----------+-----------------------------------------+-----------+-------------+---------------------
      3189 |     2 | lineageex | Extended taxonomic lineage used by NCBI |      3496 |           0 |                   0
(1 row)
```
 6. Manually create an organism ~~with subspecies, this is to avoid the bug fixed in PR #1931 so until that is merged you need to have a subspecies.~~
[...]

PR #2027, opened 2024-11-07
re-add updated importer prepareEnvironment

dsenalik:
# Bug Fix

## Description
While the importer was waiting for review, some changes were made in PR #1981 to the importer tests.
To remove duplication, all of the importer form tests were merged into one file in PR #1927 but the change from PR #1981 was not added and so was reverted.
This PR puts that specific change back in.
I labelled urgent because this is causing test failures!

## Testing?
This is a change to automated testing so verify those work, and code review.


laceysanderson (approved):
Approved since this does fix the errors that it says it does. However, we have now found a bigger error described in #2028 which is still causing tests to fail.

laceysanderson (approved):

PR #2030, opened 2024-11-09
Smaller PR to get in testing/docker changes supporting D11

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR is focusing only on the updates needed to the Docker setup and workflows to test on Drupal 10.5.x, 11.0.x and 11.1.x. This will allow us to see our compatibility sooner. 

Furthermore, the is a dependency between the drupal docker and the tripaldocker that is currently making testing on PR #1953 fail because the scheduled docker build are removing the changes needed for tripaldocker on that PR to be able to build 🙈

Updates to the README can be viewed here: https://github.com/tripal/tripal/tree/tv4g0-1953-dockerUpdatesOnly?tab=readme-ov-file
image

## Testing?

Check that you can build some of these versions locally and confirm the README is rendering properly.

NOTE: in order to build the tripaldocker 11+ versions you will need to make local changes to the info.yml files.


dsenalik (approved):
There are a lot of version numbers to get right, I checked pretty closely, but they all look perfect!

PR #2035, opened 2024-11-17
Tripal 3 migration of existing entity ID (bio_data) values

### LINKED ISSUE #2034: Tripal 3 migration of existing entity ID (bio_data) values
### Task Description

After migrating a Tripal 3 site to Tripal 4, we may want to have the same entity ID numbers as before, if records are referenced by any external source.

This issue is a place to work on this, a script to run on the existing Tripal 3 site to generate a data file of existing chado to entity mapping, and a drush command to run on the migrated Tripal 4 site to either generate URL aliases, or to actually change the entity ID values.

### Branch Name

tv4g0-issue2034-migrate-tripal3-entity-ids

### PR #2035

dsenalik:
# Tripal 4 Core Dev Task 

## Description
After migrating a Tripal 3 site to Tripal 4, we may want to have the same entity ID numbers as before, if records are referenced by any external source.

This PR provides 
 - [x] A script to run on the existing Tripal 3 site to generate a data file of existing chado to entity mapping
 
 ␡ ~~A drush command to run on the migrated Tripal 4 site to update the URL aliases~~ (moved to new branch `tv4g0-issue2034-migrate-tripal3-url-aliases`)
 - [x] Ability to pass in the data file to publish, so that the entities are originally published with the Tripal 3 entity ID values. This way, the `/bio_data/` paths can also be migrated

Perhaps useful for people with existing sites might be a bulk unpublish, so that content could be published again with different ID values. This will be a separate PR #2073

## Testing?
1. Copy the script from `./tripal_chado/migration/export_tripal3_entity_mapping.php` to the root of your tripal 3 site (or wherever you want)
Rather than putting this script in the Tripal 3 repo, I am including it here in Tripal 4 just to make it easier for people, they won't have to update Tripal 3 just for this one script.
2. On the Tripal 3 site run this command
`drush php:script export_chado_entity_mapping.php tripal3_entity_mapping.tsv`
3. Copy the generated file to the root of your Tripal 4 site (or wherever you want)
4. On your Tripal 4 site reserve a block of entity IDs e.g. on Tripal 3 site `SELECT NEXTVAL('tripal_entity_id_seq');` then on Tripal 4 site `ALTER SEQUENCE tripal_entity_id_seq RESTART 1234567;`
## URL alias update test
This is removed from this PR, transferred to branch `tv4g0-issue2034-migrate-tripal3-url-aliases` because of performance issues, and the publish option is (I think) better.
5. ~~On the Tripal 4 site run this command
`drush tripal-chado:migrate-url-aliases tripal3_entity_mapping.tsv`~~
## Publish test
(You can reserve a block of entity IDs, or if you have no content published yet it will be handled automatically)
6. Migrate a chado dump from the Tripal 3 site https://tripaldoc.readthedocs.io/en/latest/upgrade_guide/site/migrating_chado.html
7. Publish a content type by command line, _e.g._
`drush tripal-chado:publish organism --migration-file=tripal3_entity_mapping.tsv`
8. Publish a content type through the UI. You will need to uncollapse the "Tripal 3 Migration Options" fieldset at the bottom.

9. If you upload the file, it will look like this

10. The file should then be available for a subsequent publish


Ferrisx4 (approved):
Tested this both with the drush command and the GUI interface, with and without the lenient options. bio_data IDs were successfully migrated and matched up to the T3 site in each case.

Issue #1891, opened 2024-06-06
HTML formatting in field values


spficklin:
### Task Description

Recently, a PR changed the `title_format` of an organism entity to have HTML tags in the title. This allows the organism to have an italicized genus and species.    The problem I see with this is that it puts formatting on a value and assumes that the visualization of that formatting should be HTML.   This can be problematic with web services because the values that people can pull from the service should not force any type of formatting. The user should be able to use those values however they want.  

I think it's important that we don't force formatting on data values.  This is why we have a separation between the field type class and the field formatter class.  All formatting happens in the formatter.   I'm not sure how to fix this though.   We may be able to override the title field class but the token substitutions for the title happens at the publishing step, so if we then want to apply HTML formatting we'd have to know how to reapply the title token string with HTML tags. 

Doug suggested that for web services we strip out the HTML elements, which could work, but then what if someone still wants some type of formatting for their end use?  They could either rebuild the title using the other fields or have to figure out how to apply the italics at the right place.   

Here's a suggestion.  In the entity YML file add a `title_plain` element that has not formatting, and insted of HTML in the `title_format` we allow markdown.  The markdown can be coverted to whatever format someone wants and is more general.

### Branch Name

tv4g[0-9]-issue\d+-[optional short descriptor]

spficklin:
Another point rasied by Doug:  Field values we save via the GUI may have formatting if we use the CKEditor, so we'll have the same problem.

dsenalik:
Though this is a tiny subset of HTML that we are talking about &lt;em&gt; and &lt;strong&gt; and maybe a few others - converting to another markup format would not be too difficult 🤞

spficklin:
Here's another issue:

```
<em>Gorilla gorilla</em> <em></em>
```

Extra `<em>` when no values for tokens. It doesn't show up on the webpage but will look funny otherwise.

spficklin:
> Though this is a tiny subset of HTML that we are talking about `<em>` and `<strong>` and maybe a few others - converting to another markup format would not be too difficult 

I agree, those are easy enough to find and markdown is just another formatting too.  I just don't like HTML as a "generic" way to define formatting and prefer something more simple and human-readable.

PR #2040, opened 2024-11-24
Decouple token parser

### LINKED ISSUE #2008: Decouple token parser service from entity/bundle.
### Task Description

We are seeing in #2006 that it would be ideal to extend the TripalTokenParser service to not require an entity/bundle.

To accomplish this, I would suggest that 
1. the replaceTokens method work directly on a values array that maps tokens to their values.
2. A method is added to make that mapping for an entity
3. A method is added to set that mapping arbitrarily
4. A method is added to make that mapping for a publication

This way we can ensure that tokens are handled consistently throughout and that all token replacements can take advantage of any improvements.

### Branch Name

tv4g8-issue2008-decoupleTokenParser

### LINKED ISSUE #2024: URL Alias some fields are not substituted
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

The token replacement for the URL alias field needs some work. Some valid tokens are not replaced as expected.
This was originally observed here https://github.com/tripal/tripal/pull/1981#issuecomment-2455656517 but it is not specific to that PR so deserves its own issue.

### Reason
In this section of TripalEntityForm.php https://github.com/tripal/tripal/blob/995da6ff16a7a92b2048c3323075d39e03659939/tripal/src/Form/TripalEntityForm.php#L203-L209
we call `setDefaultAlias()` after the call to `parent::save()` - This somehow deep in Drupal removes the entity values.
But we can't call before `parent::save()` because the entity_id will not be known.

### Steps to reproduce

1. Tripal -> Page Structure -> Organism -> Edit
2. Open the "URL alias options" tab
3. Append genus and species so it shows `organism/[TripalEntity__entity_id]-[organism_genus]-[organism_species]`
4. Save
5. Enter a new organism through the UI
6. Either look in the `path_alias` table
```
 id | revision_id |                 uuid                 | langcode |    path     |     alias     | status 
----+-------------+--------------------------------------+----------+-------------+---------------+--------
  6 |           6 | 2d4cfd46-5a8e-4276-aefb-a1cc8e82615a | und      | /bio_data/5 | /organism/5-- |      1
```
or go to `/admin/config/search/path`
Issue2024 1
Note that the genus and species were not added to the URL alias.

### Error messages

```shell
.
```

### Screenshots

.

### LINKED ISSUE #2031: Publish doesn't populate URL Paths
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### Branch

`tv4g4-issue2031-publish-url-path`

### General Description

I forgot about "URL Paths", publish needs to populate that value.

### Steps to reproduce

1. publish anything
2. check the "URL Paths" field

### Branch

`tv4g4-issue2031-publish-url-path`

dsenalik:
# New Feature

## Description
This PR undertakes the daunting task of converting the token parser into a standalone service that knows nothing about entities and the like.
The service now just takes a tokenized string, and an associative array, and does its replacement work.

This means all the entity-based functions that were in the token parser are moved to the TripalEntity class itself.
I feel that this is where they belong anyway.

This PR also fixes issue #2024. To summarize what was wrong, when an entity is created using the UI, it gains some and loses some internal values when it is saved. Basically, the fix is to load the entity after saving, then everything is there. This does mean the title and URL alias need to be set after it is saved, but that is indeed what I do here.

I have converted the token replacement tests to kernel tests for hopefully a few seconds saved there, there is now a test for the token replacement service, and the old test now tests the tripal entity token-related functions.

The template for organism title is changed to
`title_format: "<em>[organism_genus] [organism_species]</em>[ [organism_infraspecific_type]][ <em>[organism_infraspecific_name]</em>]"`
so that if there is no infraspecies, the empty <em></em> tags will no longer be present. This fixes what @spficklin noted in this comment https://github.com/tripal/tripal/issues/1891#issuecomment-2153421756

## Testing?
1. Create a project entity (to be used later)
2. Go to Tripal -> Page structure -> Contact -> Edit
3. Change "Page title options" to be
  `Entity [TripalEntity__entity_id]-[contact_type|contact_name][ ([contact_project])]`
4. Change "URL alias options" to be
  `contact/[TripalEntity__entity_id]_[contact_type|contact_name][ ([contact_project])]`
  note that we are forgetting the leading slash here, intentionally.
5. Now create a contact, filling in something for all of the fields, including project
add_contact_one
6. Check title. You filled in a contact type, so it should be present in the title, as is the project inside parentheses
contact_one_title
7. Check URL alias. The project should be present in url-escaped parentheses. And there should be a leading `/` even though we forgot to add that to the format (this is an unchanged behavior from 4.x though)
contact_one_alias
8. Create a second contact, and only enter the contact name, leave other fields blank
9. Check the title. Because there is no contact type, the token replacement should proceed to the next option, contact name. And with no project, the parentheses should not be present.
contact_two_title
10. Check the url alias. Again it uses contact name instead of type, and the project and parentheses should not be present.
contact_two_alias
11. Go back to Tripal -> Page structure -> Contact -> Edit and try to enter some invalid tokens
InvalidTokenValidation

## Some draft documentation
The TripalTokenParser class replaces tokens in strings with a corresponding value. Tokens are marked by square brackets, and the square brackets may be 1 or 2 levels deep. This allows having a prefix or suffix that is only included when the token has a defined value.

A simple token string example:
  `The title is [Title].`
An example of a token that may or may not be defined `Issue`:
  `[Journal]. [Volume][([Issue])]`
If defined the replaced string could be `Science. 123(45)` but if Issue is not defined then the parentheses are not
included `Science. 123`

In some cases the token may have several possible values, and this can be designated by a pipe "`|`" delimited list.
For example:
`[ [Journal Name|Series Name|Journal Abbreviation].]`
The first of these that has a defined value will be used for the substitution, or if none are defined, the token is removed.

Any prefix or suffix in a 2-level token is checked against the token value, and if already present, is not added a second time. For example, for a token `[[Title].]` and a title value of `Cool Science.`, then the replaced value will not include both periods, only one is retained.


tripal/src/Entity/TripalEntity.php
```diff
@@ -126,21 +134,21 @@ public function label() {
...
    * {@inheritdoc}
    */
-  public function setTitle($title = NULL, $cache = []) {
-
-    // Get the bundle object.
-    if (isset($cache['bundle'])) {
-      $bundle = $cache['bundle'];
-    }
-    else {
+  public function setTitle($title = NULL) {
```
laceysanderson, line 181: I think we should keep the $cache parameter here and the associated support for caching the bundle. This is helpful for performance reasons as there may be workflows where we already have the bundle and do not want to look it up repeatedly. For example, in a workflow where you are publishing the content for a single bundle. You may have used the bundle object in (1) the publishing service, (2) setting title, (3) setting URL alias. Your changes mean it would be loaded 3 times but leaving in the cache and passing it through reduces it to us loading it once.

tripal/src/Entity/TripalEntity.php
```diff
@@ -154,14 +162,26 @@ public function getTitle() {
...
+   * @param string $default_alias
+   *   Either an empty string if default alias is desired,
+   *   or an alias that may optionally contain tokens
+   *
    * @return string
    *   The default entity alias, e.g. "/project/1234"
    */
-  public function getDefaultAlias() {
-    // Generate an alias using the default format set by admins.
+  public function getDefaultAlias(string $default_alias = '') {
```
laceysanderson, line 215: It would be good to add the $cache here as well for consistency -that must have been missed on my part earlier 🙈

tripal/src/Entity/TripalEntity.php
```diff
@@ -174,36 +194,61 @@ public function getDefaultAlias() {
...
 
-    // If an alias does not exist, then create a default alias.
-    if (!$alias_exists) {
-      $path = \Drupal::entityTypeManager()->getStorage('path_alias')->create([
+      // If an alias already exists, we first need to remove it
+      if ($existing_alias) {
+        $alias_objects = \Drupal::entityTypeManager()->getStorage('path_alias')->loadByProperties([
+          'alias' => $existing_alias['alias'],
+        ]);
+        $alias_objects[array_key_first($alias_objects)]->delete();
```
laceysanderson: We need more error checking on this line in case something goes wrong along the way. For example, you should check that $alias_objects is a non-empty array before grabbing the first key and you should check that `$alias_objects[array_key_first($alias_objects)]` is actually an object of the correct type. If either of these are not true you should throw an exception and this exception should be caught by the post save in order to provide a useful error logged to the user.

tripal/src/Entity/TripalEntity.php
```diff
@@ -174,36 +194,61 @@ public function getDefaultAlias() {
...
+
+      // Create the alias
+      $system_path = '/bio_data/' . $this->getID();
+      $alias_object = \Drupal::entityTypeManager()->getStorage('path_alias')->create([
         'path' => $system_path,
-        'alias' => $path_alias,
+        'alias' => $new_alias,
       ]);
-      $path->save();
+      $alias_object->save();
```
laceysanderson: The same as above goes here: More Error checking!

tripal/src/Entity/TripalEntity.php
```diff
@@ -320,37 +393,84 @@ public function getFieldValues() {
...
-  protected function replaceTokens($string, $bundle) {
+  protected function processFieldValues(array $field_values): array {
+    $processed_values = [];
+    foreach ($field_values as $field => $value_array) {
+      // Token replacement only supports single-value fields,
+      // therefore, only add if the items array has a single value.
+      if (count($value_array) == 1) {
+        if (array_key_exists('value', $value_array[0])) {
+          $processed_values[$field] = $value_array[0]['value'];
+        }
```
laceysanderson, line 483: I believe the old code supported tokens specific to a given field property. For example, one a gene content type there would be a token for the organism.genus... or at least if it didn't just yet, it was designed to be able to by adding all the properies of a field to the token values list. This was done with the following which has since been removed.
```
        $props = $item->getProperties();
        foreach ($props as $prop) {
          $token_parser->addFieldValue($field_name, $prop->getName(), $prop->getValue());
        }
```
I would like to see this added back. To keep with your current flat array approach, you could keep the `$processed_values[$field] = $value_array[0]['value'];` approach for each field but then also add in the properties like this:
```
        $props = $item->getProperties();
        foreach ($props as $prop) {
          $token = $field . "." . $prop->getName();
          $processed_values[$token] = $prop->getValue();
        }
```
This way your current functionality remains unchanged but the property values are still in the list of token values for when we support that in the future.

laceysanderson (changes requested):
## Code review

I did a thorough code review of the TripalEntity.php file so far: see the inline suggestions. I will try to review the other files later today.

## Manual testing

I also did a manual test. The functionality worked as described when I followed the testing instructions exactly. However, I came across one error when I included HTML in my project title and there is one comment/suggestion/hope when it comes to multi-value fields.

### Errors when URL alias includes HTML

I come across the following error if the default alias created is invalid:
Screenshot 2024-12-02 at 12 25 56 PM

Steps to reproduce:
1. Create a project with HTML in the title (e.g. `A still <u>great project</u> but less assuming <i>then others</i>`)
2. Select this project for a contact that is set up with title/url alias as described in the testing for this PR.
3. Click save and notice it seems to be fine...
4. Try to edit and save again and notice the error I mention.

Suggested solutions:
1. Have options that can be passed to replaceTokens that indicates if the tokens should be sanitized before replacement. Use this option for URL alias to remove any html, spaces and nonalphanumeric characters with a single `-`.
2. Include a lot more error checking throughout the process so if something goes wrong we get an error drupal set message instead of a WSOD with stack trace ;-p

### Multi-value fields

If you include an unlimited cardinality field in your pattern then the results are inconsistent between individual content. For example, if you create a piece of content with only one entry in this unlimited cardinality field then it will be replaced but if you create another piece of content with multiple entries then it won't be.

Previously (maybe Tripal 3) whether the token was replaced depended on the actual cardinality of the field but here it is based on the number of values. I like this approach better but wonder if we can just go one step further at this point and actually support multi-value fields by choosing the first value for the field?

tripal/src/Entity/TripalEntity.php
dsenalik: I only removed this because it was not implemented and didn't do anything 🙈
I will work on adding it back...

tripal/src/Form/TripalEntityForm.php
```diff
@@ -199,14 +199,22 @@ public function save(array $form, FormStateInterface $form_state) {
...
+    $entity_id = $entity->id();
 
-    // If no alias was specified, then create a default alias.
-    // We couldn't do this earlier, because the entity ID was not yet known.
-    $entity->setDefaultAlias();
+    // Entity ID is only available post-save, so we have waited
+    // to set the title and path alias until after saving.
+    // Fetch the saved version of the entity.
+    $entities = \Drupal::entityTypeManager()->getStorage('tripal_entity')->loadByProperties(['id' => $entity_id]);
+    $entity = $entities[$entity_id];
```
laceysanderson: Can you just access $this->entity here? I would have thought parent::save() would update the entity attached to the form after saving. This would save us performance-wise if we can.

tripal/src/Form/TripalEntityForm.php
```diff
@@ -199,14 +199,22 @@ public function save(array $form, FormStateInterface $form_state) {
...
+
+    // Update remaining items and save.
+    $entity->setTokenValues();
+    $entity->setTitle($values['title'][0]['value']);
+    $entity->setAlias($values['path'][0]['alias']);
```
laceysanderson: This is an example of where you would take advantage of that cache. We have the $bundle already above so we can pass it in to setTitle and setAlias to save them looking it up again.

laceysanderson (commented):
More progress in code review but still not done 🙈 

I've now gone through the files in tripal/src but have not completed tripal/tests or tripal_chado

tripal/src/Form/TripalEntityForm.php
dsenalik: Sadly, no, some internal values disappear after the save 😡, so a re-load is necessary

tripal/src/Entity/TripalEntity.php
dsenalik: cache is made a class variable - let me know if this is reasonable!

tripal/src/Entity/TripalEntity.php
dsenalik: done

tripal/src/Entity/TripalEntity.php
dsenalik: This part changed significantly with commit https://github.com/tripal/tripal/pull/2040/commits/30d93853cc7122d1384aa90f8ec6f75c305ab599 - instead of fetching the alias, deleting, and creating a new one, I now (as I saw in Drupal core) just update the existing alias. But I have some error checking in there now

tripal/src/Entity/TripalEntity.php
dsenalik: same as comment above, redesigned and some error checking added

tripal/src/Entity/TripalEntity.php
dsenalik: I'm not sure I understand fully, currently we do this in `getFieldValues()`
```
            $prop_name = $prop->getName();
            $prop_value = $prop->getValue();
            $values[$field_name][$delta][$prop_name] = $prop_value;
```
so I think all of the values should be there already.
Then these get flattened in `processFieldValues()`
Here I created an organism with all the fields with values
processFieldValues 1

Here just genus and species.
processFieldValues 2

So the dbxref and publication do not show up if they don't have a value, is that what you mean?

dsenalik:
Unlimited cardinality fields are now supported, but by returning only the first value.
Still some work to do on handling errors gracefully...

dsenalik:
To document how the new exceptions @laceysanderson requested are handled, they will look like this on saving an entity. I threw a fake exception because of course the code is perfect and nothing can ever go wrong 🙄
PR2040 1204 1

The entity form validation will catch entering an alias that already exists EXCEPT if there are tokens or HTML tags, you may still get a duplicated alias. This is caught post-save (sadly can't do this in validation because we can't do token substitution until then). You can trigger this by creating a new entity, put in an existing alias, and add some random HTML tag, e.g.
PR2040 1204 2
and when you save you get an error message like this
PR2040 1204 3

tripal/src/Entity/TripalEntity.php
```diff
@@ -154,56 +205,118 @@ public function getTitle() {
...
+    // Gets and uses default template, or replaces tokens
+    // in the supplied $path_alias.
+    $new_alias = $this->getDefaultAlias($path_alias);
+
+    // Check if the specified alias already exists for a different entity.
+    // Drupal will check for this for the value from the entity form, but we
+    // need to check again for our processed value after token replacement, etc.
+    // If it is a duplicate, we remove the alias, and the entity form can complain.
+    $entities = \Drupal::entityTypeManager()->getStorage('path_alias')->loadByProperties(['alias' => $new_alias]);
+    if ($entities) {
```
laceysanderson: Shouldn't this also confirm that the new and existing are not the same so it doesn't remove the alias on editing where the alias is not changing?

tripal/src/Entity/TripalEntity.php
```diff
@@ -103,6 +119,42 @@ public static function preCreate(EntityStorageInterface $storage_controller, arr
...
+  /**
+   * Get the bundle object for the current type, and cache it.
+   *
+   * @return Drupal\tripal\Entity\TripalEntityType
+   *   The bundle object
+   */
+  protected function getBundle() {
+    $bundle_id = $this->getType();
+    $bundle = NULL;
+    if (array_key_exists($this->getType(), $this->bundle_cache)) {
```
laceysanderson:
```suggestion
    if (array_key_exists($bundle_id, $this->bundle_cache)) {
```

laceysanderson (commented):
Code is looking really good. See comments for details. I like the new test!

tripal/src/Entity/TripalEntity.php
dsenalik: Oh, an embarrassing mistake. Yes it did remove the alias on editing 😧
Fixed with commit https://github.com/tripal/tripal/pull/2040/commits/0ebdfc99b743340e0d5f13db766793c1f972701f

tripal/src/Entity/TripalEntity.php
laceysanderson: Moved into Issue #2065 since it is beyond the scope of this PR.

laceysanderson (approved):
Manual testing:
✅ Contact type and linked project are in title (even with html in the project title)
Screenshot 2024-12-17 at 3 00 20 PM
✅ URL for same contact in parenthesis and including type as expected. HTML from project title removed and spaces changed to dashes ❤️ 
Screenshot 2024-12-17 at 3 02 04 PM
✅ contact with no type or linked project shows name instead of type in the title and no parenthesis in both the title and URL
Screenshot 2024-12-17 at 3 04 18 PM
✅ contact with multiple projects linked only includes the first project in the title and URL.
Screenshot 2024-12-17 at 3 07 31 PM
✅ content can be edited without losing the alias
✅ if the alias textfield is emptied on edit then the default alias is regenerated
✅ if a custom alias is entered in the textfield then it is used instead of the default

Manual testing now works beautifully!

PR #2049, opened 2024-12-03
Fix install issue when Tripal is installed by a distribution (profile)

guignonv:
# Bug Fix

## Description
When Tripal is installed by a distribution (profile), no database parameters are set in settings.php yet, resulting in a "The specified database connection is not defined: default" error because of "\Drupal::database()->getConnectionOptions()" being called. To avoid that issue, a try-catch statement has been added to return an empty array when the $databases variable is not set.

## Testing?
Hard to test because you would need to create a Drupal distribution. Maybe (but not tested) by altering a Drupal profile such as "standard" or "demo_umami" (in core/profiles) and adding "  - tripal" in the "install:" section of the ".info.yml".
Anyway, I tested the change on my side (with a distribution I'm working on) and it works.

### DIFF
--- a/tripal/tripal.install
+++ b/tripal/tripal.install
@@ -4,6 +4,7 @@
  * Contains functions used to install/uninstall tripal.
  */
 
+use Drupal\Core\Database\ConnectionNotDefinedException;
 use \Drupal\Core\Database\Database;
 
 /**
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
It is above my abilities to create a distribution 😆 but code review looks good.
The only thing is I am unsure about is whether a message should be printed when the exception is caught, but I defer to your better knowledge.

laceysanderson (approved):
Thanks for the contribution @guignonv ❤️  Code-wise this looks solid. I would also ideally like a message logged when the exception is caught but since the requirement that we are checking (per-table prefixing not used) is now a requirement for Drupal core too as they've dropped support for per-table prefixing also -I think it's good as is ✅

PR #2060, opened 2024-12-14
Remove field debugger devel module dependency

### LINKED ISSUE #2059: Error if field debugger activated but devel module is not installed
### Task Description

As mentioned in this comment
https://github.com/tripal/tripal/issues/1823#issuecomment-2528462108
we have some `dpm()` in Tripal for the field debugger. Also some in the PubParserManager but those are (should be) temporary and be removed.
We always install the devel module in our dockers, but our official traditional install instructions do not mention this module.

I am thinking we could gray out the field debugger toggle if the devel module is not installed, then the dpm's will never be called. Plus maybe add a "requires the devel module" or something when it is grayed out.

### Branch Name

tv4g1-issue2059-field-debugger-if-no-devel-module

dsenalik:
# Bug Fix

## Description
This removes the field debugger dependency on the drupal/devel module.
When present, it will be used, however if it is not present and enabled, then a fallback of print_r and drupal messages is used. Not quite as nice, but still functional.

## Testing?
1. Create at least one entity, e.g. Project
2. Turn on field debugging for that content type Tripal -> Page Structure -> Project -> Manage fields
3. Pick one of the fields and activate "Enable Debugging"
4. View the entity you created

### With the devel module
with_devel

### Without the devel module
without_devel


reynoldtan:
I performed the test describe above and confirmed that this PR removes field debugger dependency on the drupal/devel module. Code looks good too.

With Devel enabled.
image

With Devel disabled.
image

laceysanderson (approved):
✅ Code changes are solid.

This is ready to merge.

PR #2063, opened 2024-12-14
Add character column type to string field type

### LINKED ISSUE #2062: Can't add MD5 checksum field to a new content type
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

I can't add a MD5 checksum field to a new content type based on the feature table, because neither text field is compatible with ` md5checksum` which is `character(32)`

It should be possible for the Chado String Field Type to handle this.

I forgot that we have a special field for this, but I might as well fix it anyway since the fix is super simple.

### Steps to reproduce

.

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
This just adds one more postgres column type that is supported by the Chado String Field Type

## Testing?
1. Import Genomic type collection: Tripal -> Page Structure -> + Import type collection
2. On gene go to Manage Fields
3. \+ Create a new field
4. Chado Fields
5. enter any name and select Chado String Field Type
6. On 4.x
PR2062 1

On this branch
PR2062 2


reynoldtan:
I performed the manual test describe in the Testing section and confirmed that this PR adds one more postgres column type that is supported by the Chado String Field Type. 

image

laceysanderson (approved):

PR #2064, opened 2024-12-15
Add term validation when using fixed type on Additional Type field

dsenalik:
# New Feature

## Description
This add a simple validation to the additional type field.
When it is used to define the term for a content type, the term in the field needs to be the same as the term for the bundle.

## Testing?
1. Tripal -> Page Structure -> Analysis -> Manage fields
2. \+ Create a new field
3. Chado fields -> Continue
4. Enter any label, select "Chado Type Reference" -> Continue
5. For "Type Table and Column" select "analysisprop -> type_id"
6. For "Controlled Vocabulary Term" select some random term
7. Activate "This term defines the bundle"
8. "Save settings" - on this branch you should get an error and the form field should be highlighted


Ferrisx4 (approved):
Tested this, seems to work appropriately. Throws the error on any type that is not the expected one. Thanks @dsenalik

PR #2066, opened 2024-12-18
Tv4g2 issue2041 get colum term id on null during config import

### LINKED ISSUE #2041: getColumnTermId() on null during full site configuration import
### Tripal Version

4.0.alpha3

### Drupal Version

10.3.1

### PostgreSQL Version

16.2

### PHP Version

8.3.10

### General Description

When importing a full site, including Tripal content types with custom field display layouts applied, the following error is encountered:
`Call to a member function getColumnTermId() on null" at /opt/drupal/web/modules/custom/tripal
/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php line 111`

After this occurs, every site on the page ends with the modern version of WSOD :`The website encountered an unexpected error. Try again later.`

---
*Either base_table or $base_column are NULL and I'm not sure why or which one... yet, will investigate in the morning.*

### Steps to reproduce

1. On a new site, on one of the Tripal content types, rearrange fields and add new field groups.
2. Export this configuration at `/admin/config/development/configuration`
3. Recreate the site, making sure to use cset to set the same UUID from the previous site (this can be found in the system.site.yml file in the exported configuration: `drush cset system.site uuid <uuid> -y`
4. Try to import the exported config from earlier

### Error messages

```shell
Call to a member function getColumnTermId() on null" at /opt/drupal/web/modules/custom/tripal
/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php line 111
```

### Screenshots

_No response_

### Branch
https://github.com/tripal/tripal/tree/tv4g2-issue2041-getColumTermId_on_null_during_config_import

Ferrisx4:
# Bug Fix

## Description

This PR changes how the function getColumnTermId functions within the FieldType definitions and especially how NULL $mapping is handled. This variable may be NULL during full site configuration imports using Drupal's [https://www.drupal.org/docs/administering-a-drupal-site/configuration-management/managing-your-sites-configuration](Configuration Management) system.

## Testing?

This functionality can be tested by using the above mentioned Configuration import/export system. For best results, use a new site to make the initial configuration export and yet another new site for importing, in both cases having Tripal fully installed and prepared with General content types loaded.
 1. Use the system to export the full site configuration - `/admin/config/development/configuration/full/export`
 2. On the second site, change a configuration on the site. A common example would be to switch the order of genus and species on the Organism content type page.
 3. Use the above system to re-import your changes on the second newly installed Tripal site, remembering to set the UUID from the first site: `drush cset system.site uuid <uuid> -y` which can be found on the first site using `drush cget system.site uuid`.
 4. The action of importing configurations should not throw any errors described in the initial issue, and your genus and species fields should be reverted to their original order.


Ferrisx4:
I forgot to make the changes Lacey suggested in [this comment](https://github.com/tripal/tripal/issues/2041#issuecomment-2548870920), so I will make them now and make sure that the variable usage is correct in terms of using self:: and this:: vs not.

Also, @dsenalik has pointed out a number of now-obsolete lines that can be safely deleted that would otherwise only impact performance negatively.

Ferrisx4:
After discussion, the variable and function will remain static, and we are good to go as-is with the obsolete code removed.

dsenalik (approved):
Code looks perfect, thanks also for catching MD3checksum
I tested as described and that worked.
I also created several entities through the UI without issues.
Tested publish for organism and this also worked

Ferrisx4:
Thanks @dsenalik and @laceysanderson for all your help!

PR #2075, opened 2025-01-01
Widgets should use autocompletes when there are many records

### LINKED ISSUE #2074: Dropdowns need to have an autocomplete option when there are many records
### Task Description

I have migrated my Tripal 3 chado, but when I go to enter a new record through the UI, it takes forever for the page to load.
In this case it is an organism page, and if you look at the page source there is this part
```
            <label for="edit-organism-pub-0-pub-id" class="form-item__label visually-hidden">Publication (value 1)</label>
              <select data-drupal-selector="edit-organism-pub-0-pub-id" id="edit-organism-pub-0-pub-id" name="organism_pub[0][pub_id]" class="form-select form-element form-element--type-select">
            <option value="">-- Select --</option>
                <option value="36338">&quot;Brain-specific&quot; nutrients: a memory cure?</option>
                <option value="45657">&quot;Ca. Nitrosocosmicus&quot; members are the dominant archaea associated with plant rhizospheres.</option>
                <option value="42140">&quot;Cow Healers Use It for Both Horses and Cattle&quot;: The Rise and Fall of the Ethnoveterinary Use of Peucedanum ostruthium (L.) Koch (fam. Apiaceae) in Sweden.</option>
and on and on and on for many thousands of publications
```

This would require also having an autocomplete option for all fields currently using a select.
Maybe it could automatically switch over when we get to 100 records or something like that.

Marking as high priority because I will never be able to manually add content in this current state.

A performance enhancement to include is to convert the existing autocomplete tests to kernel tests like here https://github.com/tripal/tripal/pull/1960/commits/d825602e51d75a2e3622edf9270d0f6cc1d6754b

### Branch Name

tv4g10-issue2074-more-autocompletes-for-fields

### Notes
Existing chado autocompletes only exist for CVTerm, Project, and Dbxref

dsenalik:
# New Feature

## Description
The current state of most core chado field widgets uses a dropdown select to add new records.
This is a problem when there are many records to choose from, as it can slow down page display tremendously. Currently on my migrated site, I cannot add or edit some content types through the UI because of this.

The solution proposed here is to automatically switch over to an autocomplete field when the number of records is large.
I arbitrarily chose 50 as the limit. This can be configured both globally and on a per-field basis.

Implementation of so many autocompletes was by creating a "Generic" chado autocomplete, that can be used for any chado table where the value comes from a single column. This means it can't be used for organism, so that field remains unchanged.

~~The existing project autocomplete has been removed and replaced with the generic version.~~
Tests are now kernel instead of functional, so a bit of an improvement there.

For tables with a type_id column, the type is included in the select menus in square brackets. e.g. "PI 123456 [Germplasm Accession]" for the stock field.

All instances of "-- Select --" have been changed to "- Select -" to match what Drupal uses.

## Testing?
Since most chado linking fields are updated, manual testing will require creating at least one of a content type for each core table currently supported with a field.
You can toggle select vs. autocomplete by setting the global limit at `/admin/tripal/config/widget_settings` 

As a select:
PR2075 2

As an autocomplete:
PR2075 3

## ToDo
 - [x] Configuration option for the limit
PR2075 1


laceysanderson (changes requested):
## Project autocomplete controller

You shouldn't delete the project autocomplete 😬 I have extension modules that are using it 🙈 instead you should keep it around but inherit from the Generic Autocomplete controller to reduce duplication.

## ChadoGenericAutocompleteController Query

In tripal_chado/src/Controller/ChadoGenericAutocompleteController.php you use concatenation to build the query. We should try to move away from this approach and use the query builder instead since its more readable IMO and less buggy in general.

```php
$query = $connection->select('BT', '1:' . $base_table);
$query->addField('BT', $pkey_id, 'pkey');
$query->addField('BT', $column_name, 'value');
$query->condition('LOWER(BT.' . $column_name .')', $keyword, 'LIKE');
$query->orderBy('BT.' . $column_name, 'ASC');
$query->range(0, $count);

if ($type_id > 0) {
  if ($property_table and ($property_table != $base_table)) {
    $query->join('1:'.$property_table, 'PT', 'BT.' . $pkey_id . ' = PT.' . $pkey_id);
    $query->condition('PT.' . $type_column, $type_id);
  }
  else {
    $query->condition('BT.' . $type_column, $type_id);
  }
}
```

## Restrict by type or include

> For tables with a type_id column, the type is included in the select menus in square brackets. e.g. "PI 123456 [Germplasm Accession]" for the stock field.

This part doesn't seem to be implemented 🤔 There is no adding of the cvterm name to the query and instead you just filter to the type. However, we do need to support cross-type autocompletes where we don't filter but rather add into the result. I would do this if the type_id supplied is `0`.

## Possible query mismatch

In `tripal_chado/src/TripalField/ChadoWidgetBase.php` the genericSelectElement() method allows widgets to pass in a query for select but if there are too many results it will use the controlled to create an autocomplete query. This could result in mismatched queries. Instead I would add a static `getQuery` method to the autocomplete controller that has an option indicating if the query should be limited or not. That way, `getQuery` can be used by both the `handleAutocomplete()` and `genericSelectElement()`.

## Configuration

> The solution proposed here is to automatically switch over to an autocomplete field when the number of records is large.
I arbitrarily chose 50 as the limit.

I see you added a stand alone configuration form for the limit but 

1. I think it should just have been added to the existing entity.tripal_entity_settings form.
2. I think that form sets the global default but it should also be set in the widget settings per widget. 

Also, I don't think we should rely only on the auto-switch. Count queries alone on my site will cause a page time out 🙈 As such, the widget should also have a setting for when to use select or autocomplete. Something like

> Behaviour for widget type:
> 🔘 Always Select List but limit to maximum results
> 🔘 Always Autocomplete
> 🔘 Select list if less than maximum results and autocomplete otherwise.

Since it is now configurable per content type, I would use a default of "Select list if less than maximum results and autocomplete otherwise" for the content types we know are usually small lists (like organism) and "Always Autocomplete" on most.

dsenalik:
> ## Project autocomplete controller
> 
> ...You shouldn't delete the project autocomplete 😬 ...

Sorry, it's back! Nothing in core Tripal was using it, but I didn't think about other modules 😬.

> ## ChadoGenericAutocompleteController Query
> 
> ...We should try to move away from this approach and use the query builder instead...

I have done so.
 
> ## Restrict by type or include
> 
> > For tables with a type_id column, the type is included in the select menus in square brackets. e.g. "PI 123456 [Germplasm Accession]" for the stock field.
> 
> This part doesn't seem to be implemented 🤔...

This had been implemented only for the select options, but not for the autocomplete. This is now implemented for both select and autocomplete **in cases where `type_id` is in the base table**. It is not done for cases where the bundle is defined in the property table, due to the complexity of doing this, you need to know ahead of time which terms to query for, and this seemed to be too much to do inside an autocomplete, which needs to be fast.

> ## Possible query mismatch
> 
> ...Instead I would add a static `getQuery` method to the autocomplete controller that has an option indicating if the query should be limited or not. That way, `getQuery` can be used by both the `handleAutocomplete()` and `genericSelectElement()`...

This is done, and I think makes the code clearer too.

> ## Configuration
> 
> 1. I think it should just have been added to the existing entity.tripal_entity_settings form.

Moved there now.


> 2. I think that form sets the global default but it should also be set in the widget settings per widget.

That was a job, but I did it. I copied Drupal almost verbatim here and because of that included several settings.
If set for the field, use that. If not, use the global setting. If that's not set, default to 50.
I copied the "Authored by" autocomplete settings which appear as
PR2075 1012 1
PR2075 1012 2

You can also set the number of options returned by the autocomplete now, change the field size, and specify a placeholder if you want.




> Also, I don't think we should rely only on the auto-switch. Count queries alone on my site will cause a page time out 🙈 As such, the widget should also have a setting for when to use select or autocomplete. Something like
> 
> > Behaviour for widget type:
> > 🔘 Always Select List but limit to maximum results
> > 🔘 Always Autocomplete
> > 🔘 Select list if less than maximum results and autocomplete otherwise.
> 
> Since it is now configurable per content type, I would use a default of "Select list if less than maximum results and autocomplete otherwise" for the content types we know are usually small lists (like organism) and "Always Autocomplete" on most.

We should discuss this one, I can't see a reason to always have a select list but not return everything. It seems it would confuse people if this happened.
With the current implementation, you set zero for the limit to force always autocomplete.

Another point, the count is performed on a subquery, after the LIMIT is applied, so this is much more efficient than counting every record just to see if it is over the limit. So I am curious if this works just fine in your case even with millions of records. And regardless, if you set the limit to zero, the count will not even be performed.
PR2075 1012 3

laceysanderson (commented):
So I'm a bit confused. I've built a fresh docker image container on this branch many times at this point, confirmed that your changes are in the container, cleared caches, etc and I'm still finding that most of the fields I expect to have the widget settings do not. 😕 

Do you see the widget settings cog consistently on these fields?

The following screenshots show that I can find the one analysis linker field with the settings but the contact and project linkers don't have widget settings? Also, shouldn't there be more analysis linkers? Why isn't there one on gene pages?

### Contact
Screenshot 2025-01-16 at 4 52 09 PM

### Project
Screenshot 2025-01-16 at 4 52 55 PM

### Gene
Screenshot 2025-01-16 at 4 54 29 PM

dsenalik:
This was a case of slovenly testing on my part. I forgot to add settings to most of the fields that needed them.
Now gene for example has the missing settings:
2075 gene
Contact settings shown in this screenshot have been edited to all non-default values

I will remove draft status only after properly testing on my end 🙄

laceysanderson (approved):
Cases checked:
A) Has indicated settings for all fields I would expect.
B) Shows default max records of 50 on first view
C) Shows updated global default when field-specific settings have not been updated

Regarding A, we might want to enable these settings for the "Database Reference Annotations", "Synonyms" and "Organism" too at some point --talk to Doug about this already and we will make separate issues. I was also happy to see the settings for discovered fields such as the "Sequence Feature" field on analysis.

|                             | A | B | C |
|-----------------------------|---|---|---|
| Analysis                    | ✅ | ✅ | ✅ |
| Contact                     | ✅ | ✅ | ✅ |
| Organism                    | ✅ | ✅ | ✅ |
| Project                     | ✅ | ✅ | ✅ |
| Protocol                    | ✅ | ✅ | ✅ |
| Publication                 | ✅ | ✅ | ✅ |
| Study                       | ✅ | ✅ | ✅ |
| Breeding Cross              | ✅ | ✅ | ✅ |
| Germplasm Accession         | ✅ | ✅ | ✅ |
| Germplasm Variety           | ✅ | ✅ | ✅ |
| Recombinant Inbred line     | ✅ | ✅ | ✅ |
| DNA Library                 | ✅ | ✅ | ✅ |
| Gene                        | ✅ | ✅ | ✅ |
| Genome Annotation           | ✅ | ✅ | ✅ |
| Genome Assembly             | ✅ | ✅ | ✅ |
| Genome Project              | ✅ | ✅ | ✅ |
| mRNA                        | ✅ | ✅ | ✅ |
| Phylogenetic Tree           | ✅ | ✅ | ✅ |
| Physical Map                | ✅ | ✅ | ✅ |
| Genetic Map                 | ✅ | ✅ | ✅ |
| Genetic Marker              | ✅ | ✅ | ✅ |
| Heritable Phenotypic Marker | ✅ | ✅ | ✅ |
| QTL                         | ✅ | ✅ | ✅ |
| Sequence Variant            | ✅ | ✅ | ✅ |
| Array Design                | ✅ | ✅ | ✅ |
| Assay                       | ✅ | ✅ | ✅ |
| Biological Sample           | ✅ | ✅ | ✅ |

Cases checked on a few fields but not all like above:
- ✅ Field settings are saved and override the global default
- ✅ Widget does switch between select and autocomplete when too many records and if configured to do so
- ✅ Widget does not switch regardless of number of records if configured not to.
- ✅ If widget is configured to switch at a higher number of records then global, ensure the local is respected

dsenalik:
I created issue #2106 to acknowledge that the organism field needs an autocomplete
And issue #2107 to add similar settings for dbxref and synonym fields

PR #2077, opened 2025-01-04
Implement discover() for linking fields

### LINKED ISSUE #1995: We need to implement discover functions for all remaining core fields
### Task Description

The `discover()` function has so far only been implemented for organism and property fields.
This function is needed for the "+ Check for new fields" button on the manage fields page.
For history, see PR #1869.
This issue is just a reminder to implement it for remaining core fields where it is appropriate.

### Branch Name

tv4g1-issue1995-implement-discover-functions

### PR #2077

dsenalik:
# Tripal 4 Core Dev Task

## Description
This implements discover for 12 new fields, all of these are general linking fields:
Analysis, ArrayDesign, Assay, Biomaterial, Contact, Feature, FeatureMap, Project, Protocol, Pub, Stock, Study

More details in the issue here: https://github.com/tripal/tripal/issues/1995#issuecomment-2571347927

I created new tests for Chado discover().

## Testing?
Import all the type collections and start trying discover on various content types.
The possible linking fields are listed at https://github.com/tripal/tripal_doc/issues/32#issuecomment-1837532399
There are not many direct ones to test, you can remove the analysis field from phylotree, then trying to discover it.

Example for pub
PR2077 1

## Caveat
So since I only use the "most common" term for discovered fields, you will be able to add records of a different type if from the same table. For example, the stock table has four content types.
Only those with the same term will have links to the content page
PR2077 2

I will make this a separate issue #2085, as it currently happens on 4.x with core fields. PR #2086 should fix this.


tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -755,6 +756,194 @@ protected function removeIncompatibleField(string $bundle, string $machine_name)
...
+    }
+
+    return $field_item;
+  }
+
+  /**
+   * discover() for directly linked foreign keys
+   *
+   * @see Drupal\tripal_chado\TripalField\ChadoFieldItemBase::discover()
+   */
```
laceysanderson:
```suggestion
  /**
   * Discover fields that use a foreign key in the specified base table.
   *
   * This is used by ChadoFieldItemBase::discover().
   *
   * @param \Drupal\tripal\Entity\TripalEntityType $bundle
   *   The entity type object for which new field instances should be found.
   * @param string $field_id
   *   The id of the field.
   * @param array $field_types
   *   An array where each item defines a field type that is attached to this
   *   bundle. Each item is itself an array of the annotation defined at the top
   *   of that particular field type class.
   * @param array $field_instances
   *   An array of FieldConfig objects where each object defines a field attached
   *   to this content type.
   * @param array $options
   *   Specific options from the field's discover() function. Required keys:
   *   - id: the field id, e.g. 'chado_organism_type_default
   *   - table: the field's base table, e.g. 'organism'
   *   - label: the field's label, e.g. 'Organism'
   *   - termIdSpace: The field term's DB
   *   - termAccession: The field term's accession
   *   - description: A text description for the discovered field
   *
   * @return array
   *   An associative array of fields to suggest that follows the same structure
   *   as expected by tripal.tripalfield_collection.* configuration.
   */
```

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -755,6 +756,194 @@ protected function removeIncompatibleField(string $bundle, string $machine_name)
...
+      }
+    }
+    return $field_list;
+  }
+
+  /**
+   * discover() for foreign keys through an intermediate linker table
+   *
+   * @see Drupal\tripal_chado\TripalField\ChadoFieldItemBase::discover()
+   */
```
laceysanderson:
```suggestion
  /**
   * Discover fields that link to the base table through a linking table.
   *
   * This is used by ChadoFieldItemBase::discover().
   *
   * @param \Drupal\tripal\Entity\TripalEntityType $bundle
   *   The entity type object for which new field instances should be found.
   * @param string $field_id
   *   The id of the field.
   * @param array $field_types
   *   An array where each item defines a field type that is attached to this
   *   bundle. Each item is itself an array of the annotation defined at the top
   *   of that particular field type class.
   * @param array $field_instances
   *   An array of FieldConfig objects where each object defines a field attached
   *   to this content type.
   * @param array $options
   *   Specific options from the field's discover() function. Required keys:
   *   - id: the field id, e.g. 'chado_organism_type_default
   *   - table: the field's base table, e.g. 'organism'
   *   - label: the field's label, e.g. 'Organism'
   *   - termIdSpace: The field term's DB
   *   - termAccession: The field term's accession
   *   - description: A text description for the discovered field
   *
   * @return array
   *   An associative array of fields to suggest that follows the same structure
   *   as expected by tripal.tripalfield_collection.* configuration.
   */
```

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -755,6 +756,194 @@ protected function removeIncompatibleField(string $bundle, string $machine_name)
...
+   *
+   * @param array $options
+   *   Specific options from the field's discover() function. Required keys:
+   *     'id' - the field id, e.g. 'chado_organism_type_default'
+   *     'table' - the field's base table, e.g. 'organism'
+   *     'label' - the field's label, e.g. 'Organism'
+   *     'termIdSpace' - The field term's DB
+   *     'termAccession' - The field term's accession
+   *     'description' - A text description for the discovered field
+   */
```
laceysanderson:
```suggestion
  /**
   * Finds new instances of this field for a given content type.
   *
   * Overrides TripalFieldItem::discover() to add the $options parameter so that the
   * base class can discover more fields in a consistent way.
   *
   * @param \Drupal\tripal\Entity\TripalEntityType $bundle
   *   The entity type object for which new field instances should be found.
   * @param string $field_id
   *   The id of the field.
   * @param array $field_types
   *   An array where each item defines a field type that is attached to this
   *   bundle. Each item is itself an array of the annotation defined at the top
   *   of that particular field type class.
   * @param array $field_instances
   *   An array of FieldConfig objects where each object defines a field attached
   *   to this content type.
   * @param array $options
   *   Specific options from the field's discover() function. Required keys:
   *   - id: the field id, e.g. 'chado_organism_type_default
   *   - table: the field's base table, e.g. 'organism'
   *   - label: the field's label, e.g. 'Organism'
   *   - termIdSpace: The field term's DB
   *   - termAccession: The field term's accession
   *   - description: A text description for the discovered field
   *
   * @return array
   *   An associative array of fields to suggest that follows the same structure
   *   as expected by tripal.tripalfield_collection.* configuration.
   */
```

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
```diff
@@ -581,7 +582,7 @@ public static function discover(TripalEntityType $bundle, string $field_id, arra
     ];
 
     // The parent class adds collection plugin IDs
-    $field_list = self::discoverPostprocess($field_list);
+    $field_list = parent::discoverPostprocess($field_list);
```
laceysanderson: I think we should keep it with `self` in all of these fields :thinking: If we use parent here and then one day override this method in a field, the child version will be ignored, which could be difficult to debug. What was the reason to change it to parent?

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -755,6 +756,194 @@ protected function removeIncompatibleField(string $bundle, string $machine_name)
...
+  public static function discover(TripalEntityType $bundle, string $field_id, array $field_types,
+      array $field_instances, array $options = []): array {
+
+    // The parent class only initializes an empty array, so we don't need to call it
+    $field_list = [];
```
laceysanderson: We should call the parent anyway. Right now the parent doesn't add fields but one day it might... debugging would suck if one day the parent does add fields ;-p

laceysanderson (commented):
See code suggestions :-)

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: Yes good advice

tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
dsenalik: Good catch, my mistake, not even sure why I did that 🙃

laceysanderson (commented):
Manual testing

I went through each core content type, discovered field and added all those found. I then went back and discovered again to ensure the new ones now showed as existing. 

If there is a checkmark, (1) the fields I expected were discovered, (2) existing ones were also shown as existing, (3) adding them worked without error.

- ✅ Analysis
- ✅ Contact
- ✅ Organism
- ❌ Project: works for all but assay linker.
- ✅ Protocol
- ✅ Publication
- ❌ Study: works for all but assay linker.
- ✅ Breeding Cross
- ✅ Germplasm Accession
- ✅ Germplasm Variety
- ✅ RIL
- ✅ DNA Library
- ✅ Gene
- ✅ Genome Annotation
- ✅ Genome Assembly
- ❌ Genome Project: all except for assay linker.
- ✅ mRNA
- ✅ Phylogenetic Tree
- ✅ Physical Map
- ✅ Genetic Map
- ✅ Genetic Marker
- ✅ Heritable Phenotypic Marker
- ✅ QTL
- ✅ Sequence Variant
- ✅ Array Design
- ✅ Assay
- ❌ Biosample: all but assay linker.

Screenshot from assay linker failing in a WSOD puddle consistently:
Screenshot 2025-01-23 at 4 12 38 PM

laceysanderson (commented):
Next I went through and created a piece of content for each core content type. I used every linker field that had an option as I went. I also created two contact record and referenced both of them in every non-general content type to ensure the multi-value is working as expected.

For the most part this went beautifully.
Things I noticed:
- The linker field for germplasm accession is not restricted to germplasm accessions but rather allows all stock-based content. Either we should add a linker for each germplasm type and restrict it accordingly OR we should use a generic term for the germplasm linker (not one matching a content type) and allow it to access any stock. (only Sarak is a germplasm accession in the test site below)
Screenshot 2025-01-23 at 4 35 56 PM

- NOT RELATED TO THIS PR: the obsolete field for DNA library is a text field instead of a toggle like everywhere else 😕 Furthermore it is required and gives no help for the type of value to enter.
Screenshot 2025-01-23 at 4 36 45 PM

- Why doesn't an analysis link to features and vise-versa? They are on projects.

Here's all my content :-)
image

dsenalik:
Sorry about Assay, that was just one missing `use` statement that I did not catch 🙈 - fixed with https://github.com/tripal/tripal/pull/2077/commits/3d1b07a9265d907d2dc2a7bbf9dbfc7fbeef399e

> ...OR we should use a generic term for the germplasm linker (not one matching a content type) and allow it to access any stock...

I will implement this option for consistency with other fields. Unfortunately there is not a good term which is why I used germplasm, so I will find and add a term...

>  ...the obsolete field for DNA library...

Yeah, what is the deal with this column anyway. I think I will make a Chado issue about this

```
                                                   Table "teapot.library"
      Column      |            Type             | Collation | Nullable |                      Default                       
------------------+-----------------------------+-----------+----------+----------------------------------------------------
 is_obsolete      | integer                     |           | not null | 0
```

> Why doesn't an analysis link to features and vise-versa?

Okay, this is because the linker table for this is _non-standard_ so I blame Chado
The linker table is

```
sitedb=> \d teapot.analysisfeature
                                                  Table "teapot.analysisfeature"
       Column       |       Type       | Collation | Nullable |                              Default                               
--------------------+------------------+-----------+----------+--------------------------------------------------------------------
 analysisfeature_id | bigint           |           | not null | nextval('teapot.analysisfeature_analysisfeature_id_seq'::regclass)
 feature_id         | bigint           |           | not null | 
 analysis_id        | bigint           |           | not null | 
 rawscore           | double precision |           |          | 
 normscore          | double precision |           |          | 
 significance       | double precision |           |          | 
 identity           | double precision |           |          | 
```
and as you see it is very different from other linker tables both in naming, and in columns.
It is not discovered because of the non-standard name.
We can add it manually because that checks foreign keys.
Maybe this should be another Chado issue - make a "standard" linker table!

laceysanderson:
✅ Confirmed that commit https://github.com/tripal/tripal/pull/2077/commits/3d1b07a9265d907d2dc2a7bbf9dbfc7fbeef399e fixes the issues noted above regarding the assay linker :-)

❌ However, I cleared the cache to check the above, and then when I went back to the project-base content types to check for fields a PROJECT linker is showing up on PROJECT-based content types 🫨 That was not there the first time I went through everything on the same docker container but only seems to have appeared after clearing the cache 😖 
Screenshot 2025-01-23 at 5 13 21 PM

I was worried this was a bigger problem so I checked other content types and noticed that content types with only a single type for that base table were unaffected (e.g. contact) but some with multiple types (e.g. project-based, analysis-based, featuremap-based) were. Stock-based and feature-based content types seam unaffected though 😕 
Screenshot 2025-01-23 at 5 18 00 PM
Screenshot 2025-01-23 at 5 19 16 PM

laceysanderson:
> Sorry about Assay, that was just one missing `use` statement that I did not catch 🙈 - fixed with [3d1b07a](https://github.com/tripal/tripal/commit/3d1b07a9265d907d2dc2a7bbf9dbfc7fbeef399e)

✅ Fixed

> > ...OR we should use a generic term for the germplasm linker (not one matching a content type) and allow it to access any stock...
> 
> I will implement this option for consistency with other fields. Unfortunately there is not a good term which is why I used germplasm, so I will find and add a term...

Let's use the one Stephen chose in the mapping:
Screenshot 2025-01-23 at 5 27 00 PM

> > ...the obsolete field for DNA library...
> 
> Yeah, what is the deal with this column anyway. I think I will make a Chado issue about this
> 
> ```
>                                                    Table "teapot.library"
>       Column      |            Type             | Collation | Nullable |                      Default                       
> ------------------+-----------------------------+-----------+----------+----------------------------------------------------
>  is_obsolete      | integer                     |           | not null | 0
> ```

OH YAH!!! That does need to be a chado issue 🙈 

> > Why doesn't an analysis link to features and vise-versa?
> 
> Okay, this is because the linker table for this is _non-standard_ so I blame Chado The linker table is
> 
> ```
> sitedb=> \d teapot.analysisfeature
>                                                   Table "teapot.analysisfeature"
>        Column       |       Type       | Collation | Nullable |                              Default                               
> --------------------+------------------+-----------+----------+--------------------------------------------------------------------
>  analysisfeature_id | bigint           |           | not null | nextval('teapot.analysisfeature_analysisfeature_id_seq'::regclass)
>  feature_id         | bigint           |           | not null | 
>  analysis_id        | bigint           |           | not null | 
>  rawscore           | double precision |           |          | 
>  normscore          | double precision |           |          | 
>  significance       | double precision |           |          | 
>  identity           | double precision |           |          | 
> ```
> 
> and as you see it is very different from other linker tables both in naming, and in columns. It is not discovered because of the non-standard name. We can add it manually because that checks foreign keys. Maybe this should be another Chado issue - make a "standard" linker table!

Also, OH YAH 🙈 This one shouldn't be a chado issue though because there is no way to solve it at this point in time --that would be a big breaking change.

In our case though, can't we just add a special case for the discover knowing this? Even though it has extra fields, it should still be handled fine because all the other columns are optional.

dsenalik:
Implemented the generic stock term - https://github.com/tripal/tripal/pull/2077/commits/b8990a2982e5d4b11c0ff1074fada26e0abf76b0 - I do still call it "Stock" instead of "Biospecimen" because I think that is clearer for site admins to have it match the table name.

Made issue https://github.com/GMOD/Chado/issues/147 for the `library` `is_obsolete` column

I implemented a generic way for a field to list one or more potential "custom linker tables" - https://github.com/tripal/tripal/pull/2077/commits/803b783d9a8548db269617b8dbaa2ddf97df45be - so this way `featureanalysis` can now be used


So far I haven't been able to replicate what you describe in https://github.com/tripal/tripal/pull/2077#issuecomment-2611211103
But maybe I fixed it though with https://github.com/tripal/tripal/pull/2077/commits/1d30dfab943641fe7cba0d0c849a1a727d8b92ee

tripal_chado/src/Plugin/Field/FieldType/ChadoStockTypeDefault.php
```diff
@@ -330,4 +330,28 @@ public function isCompatible(TripalEntityType $entity_type) : bool {
...
+   */
+  public static function discover(TripalEntityType $bundle, string $field_id, array $field_types,
+      array $field_instances, array $options = []): array {
+
+    // Specific settings for this field
+    // There are four stock types in core Tripal, use the most generic one
+    $options += [
+      'id' => self::$id,
+      'table' => self::$object_table,
+      'label' => 'Stock',
```
laceysanderson: I really don't agree with calling this stock as it's too chado-specific and while it's intuitive to admin it would not be to content editors / curators. What about instead using the category name as we do in the add content list and structure lists?

```suggestion
      'label' => 'Germplasm',
```

Screenshot 2025-01-24 at 9 57 11 AM
Screenshot 2025-01-24 at 9 57 25 AM

laceysanderson (approved):
> Implemented the generic stock term - https://github.com/tripal/tripal/commit/b8990a2982e5d4b11c0ff1074fada26e0abf76b0 - I do still call it "Stock" instead of "Biospecimen" because I think that is clearer for site admins to have it match the table name.

Already commented that I would prefer consistency with the existing categories. That said, this is a suggestion only and shouldn't bar merging if you disagree.


> Made issue https://github.com/GMOD/Chado/issues/147 for the library is_obsolete column

Nice 😄 

> I implemented a generic way for a field to list one or more potential "custom linker tables" - https://github.com/tripal/tripal/commit/803b783d9a8548db269617b8dbaa2ddf97df45be - so this way featureanalysis can now be used

This is beautiful! Now you actually see the analysis you had to choose when importing a GFF3 😅 
Firefox_Screenshot_2025-01-24T16-29-58 583Z

> So far I haven't been able to replicate what you describe in https://github.com/tripal/tripal/pull/2077#issuecomment-2611211103
> But maybe I fixed it though with https://github.com/tripal/tripal/commit/1d30dfab943641fe7cba0d0c849a1a727d8b92ee

This seems to be fixed 🤞 and the commit does make sense.

tripal_chado/src/Plugin/Field/FieldType/ChadoStockTypeDefault.php
dsenalik: Sounds good

PR #2080, opened 2025-01-06
Chado sequence length formatter can now hide zero length

### LINKED ISSUE #2058: Sequence length display of 0 for features with no sequence should be masked by default
### Task Description

The sequence length field did not yet get a way to mask "zero" values, this needs to be added, for example this should be present on this field
hide_seq_length_zero

Similarly, the sequence coordinates field should be able to mask this
issue2058_sequencecoord

### Branch Name

tv4g1-issue2058-configurable-hide-zero-sequence-length

dsenalik:
# Bug Fix

## Description
This allows sequence lengths of zero to be hidden.

This changes the sequence length field class to inherit from the Chado integer field, which in turn inherits from the Tripal integer field, which has support for hiding zero values.

The only thing preventing the sequence length field from inheriting 100% is that instead of "value" it uses "seqlen" for the key.

## Testing?
1. Import genomic content types
2. Create a new gene with no sequence residues. The length should not appear
3. Edit and add sequence residues. The length should now be shown.
4. Test the settings at Tripal -> Page Structure -> Gene -> Manage display
5. Under "Set display format you can add a suffix, e.g. " .b.p" and set to always show
6. Edit the gene again and remove the sequence residues. The zero should appear again
0bp

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoSequenceLengthFormatterDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldFormatter/ChadoSequenceLengthFormatterDefault.php
@@ -18,83 +18,35 @@ use Drupal\Core\Form\FormStateInterface;
  *   }
  * )
  */
-class ChadoSequenceLengthFormatterDefault extends ChadoFormatterBase {
+class ChadoSequenceLengthFormatterDefault extends ChadoIntegerFormatterDefault {
 
   /**
-   * {@inheritdoc}
-   */
-  public static function defaultSettings() {
-    $settings = parent::defaultSettings();
-    $settings['field_prefix'] = '';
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code looks good :-)

Manual testing
✅ default config, no residues hides the sequence length as described.
✅ default config, 4 residues shows the sequence length of 4 as expected
✅ suffix 'bp' rest default, 4 residues shows the sequence length of 4bp as expected
Screenshot 2025-01-08 at 12 48 17 PM
✅ suffix 'bp' rest default, 0 residues hides the sequence length as expected
✅ suffix 'bp' + never hide, 0 residues shows the sequence length of '0bp' as expected.
Screenshot 2025-01-08 at 12 49 29 PM

PR #2082, opened 2025-01-09
Add field values to drupal upon publish (#1828)

mdondrup:
# New Feature

## Description

This PR affects the way chado data is published.
The logic determining which field data are to be copied into Drupal Entity tables from the chado base tables has been changed by modifying function `getStoredTypesFilter` in `ChadoStorage.php`. Now, by default most data is copied.
Fields with storage path ending in ".residues" are now automatically exempt from redundant storage. While there was already an option to add a storage parameter "drupal_store" to annotate a field to cause it to be copied, this wasn't used anywhere and would have required extensive editing of the yaml files containing field type collections. 

These changes address #1828 to allow proper filtering and sorting of chado-derived data in Views, according to Solution 1 described there: 

> The first approach involves ensuring that while publishing Tripal content, fields like organism_common_name_value are populated with meaningful values rather than left empty. This would allow the Views filter to function correctly without modifying its logic. However, this approach could introduce redundancy and might conflict with the current design, where these fields are likely left empty intentionally.

## Testing?

 - Import new chado data, e.g. genes, mRNA, via the GFF importer.  
 - Alternatively, unpublish already published Tripal content.
 - Go to /admin/content/bio_data/publish and select the content type (e.g. gene, mRNA)
 -  Run the jobs drush trp-run-jobs

### DIFF
--- a/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
+++ b/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
@@ -958,11 +958,15 @@ class ChadoPublish extends TripalBackendPublishBase {
       }
       $args[$placeholder] = $match[$field_name][$delta][$key]['value']->getValue();
     }
-    // Non-required types never have a value stored, just a placeholder.
+    // If we want views filter this needs to be changed. Now, both required and non-required types 
+    // get their : values added.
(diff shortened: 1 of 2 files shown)


mdondrup:
Too, many failures, sorry.

PR #2083, opened 2025-01-10
Cache all chado values in drupal tables in order to support views

mdondrup:
# New Feature

## Description

This PR fundamentally affects the way chado data is published.
The logic determining which field data will be copied into Drupal Entity tables from the Chado base tables have been changed by modifying the function `getStoredTypesFilter` in `ChadoStorage.php`. Now, by default, most data is copied.
Fields with storage paths ending in ".residues" are now automatically exempt from redundant storage. While there was already an option to add a storage parameter "drupal_store" to annotate a field to cause it to be copied, this wasn't used anywhere and would have required extensive editing of the Yaml files containing field type collections. 

Note that the behavior is now the opposite of the original behavior, and most field values will now be stored redundantly.
The behavior of not storing entity values during Chado publishing was covered by unit testing in `ChadoTripalPublishTest.php`.
The tests have been adapted to reflect the new behavior.

These changes address #1828 to allow proper filtering and sorting of chado-derived data in Views, according to Solution 1 described [here](https://github.com/tripal/tripal/issues/1828#issuecomment-2541585273): 

> The first approach involves ensuring that while publishing Tripal content, fields like organism_common_name_value are populated with meaningful values rather than left empty. This would allow the Views filter to function correctly without modifying its logic. However, this approach could introduce redundancy and might conflict with the current design, where these fields are likely left empty intentionally.

### Unit testing:

- Publishing of Chado data is covered by phpunit test`testChadoTripalPublishService` in `ChadoTripalPublishTest.php`

### Manual testing:
 - Import new chado data, e.g. genes, mRNA, via the GFF importer.  
 - Alternatively, unpublish already published Tripal content.
 - Go to /admin/content/bio_data/publish and select the content type (e.g. gene, mRNA)
 -  Run the jobs drush trp-run-jobs

### DIFF
--- a/tripal/src/Services/TripalLogger.php
+++ b/tripal/src/Services/TripalLogger.php
@@ -19,9 +19,14 @@ class TripalLogger {
    */
   protected $job = NULL;
 
+  /**
+   * Global settings for $options
+   * @var array $global_options
+   */
+  protected $global_options = [];
 
   /**
-   * Intiailizes the Drupal logger.
+   * Initializes the Drupal logger.
    */
   protected function initLogger() {
     $this->logger = \Drupal::logger($this->module);
@@ -41,6 +46,40 @@ class TripalLogger {
       return TRUE;
     }
     return FALSE;
+  }
+
+  /**
+   * Returns all global options
+   *
+   * @return array
+   *   An associative array of current global options
+   */
+  public function getGlobalOptions(): array {
+    return $this->global_options;
+  }
+
+  /**
+   * Returns a single global option, or NULL if the key is not defined
+   *
+   * @param string $key
+   *   The name of the option to retrieve
+   * @return mixed
+   *   The value of the option
+   */
+  public function getGlobalOption(string $key): mixed {
+    return $this->global_options[$key] ?? NULL;
+  }
+
+  /**
+   * Sets a single global option
+   *
+   * @param string $key
+   *   The name of the option to set
+   * @param mixed $value
+   *   The value to set it to
+   */
+  public function setGlobalOption(string $key, mixed $value): void {
+    $this->global_options[$key] = $value;
   }
 
   /**
@@ -213,6 +252,8 @@ class TripalLogger {
   public function notice($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
 
+    $options = array_merge($this->global_options, $options);
+
     $this->log2Job($message, $context);
 
     if (!array_key_exists('logger', $options) or $options['logger'] !== FALSE) {
@@ -250,6 +291,8 @@ class TripalLogger {
    */
   public function info($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
+
+    $options = array_merge($this->global_options, $options);
 
     $this->log2Job($message, $context);
 
@@ -296,6 +339,8 @@ class TripalLogger {
    */
   public function error($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
+
+    $options = array_merge($this->global_options, $options);
 
     $message = 'ERROR: ' . $message;
     $this->log2Job($message, $context);
@@ -346,6 +391,8 @@ class TripalLogger {
   public function warning($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
 
+    $options = array_merge($this->global_options, $options);
+
     $message = 'WARNING: ' . $message;
     $this->log2Job($message, $context);
 
@@ -394,6 +441,8 @@ class TripalLogger {
    */
   public function emergency($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
+
+    $options = array_merge($this->global_options, $options);
 
     $message = 'EMERGENCY: ' . $message;
     $this->log2Job($message, $context);
@@ -444,6 +493,8 @@ class TripalLogger {
   public function alert($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
 
+    $options = array_merge($this->global_options, $options);
+
     $message = 'ALERT: ' . $message;
     $this->log2Job($message, $context);
 
@@ -492,6 +543,8 @@ class TripalLogger {
    */
   public function critical($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
+
+    $options = array_merge($this->global_options, $options);
 
     $message = 'CRITICAL: ' . $message;
     $this->log2Job($message, $context);
@@ -544,6 +597,8 @@ class TripalLogger {
   public function debug($message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
 
+    $options = array_merge($this->global_options, $options);
+
     // If we want to implement a toggle for debug messages in the
     // future, it could go here. Tripal 3 had an environment variable
     // TRIPAL_DEBUG to perform this function.
@@ -555,7 +610,7 @@ class TripalLogger {
         $function = $backtrace[$i];
         $message .= "  $i) " . $function['function'] . "\n";
       }
-      $this->log2job('DEBUG: ' . $message, $context);
+      $this->log2Job('DEBUG: ' . $message, $context);
       if (!array_key_exists('logger', $options) or $options['logger'] !== FALSE) {
         $message_str = $this->messageString($message, $context);
         $this->logger->debug($message_str);
@@ -606,6 +661,8 @@ class TripalLogger {
    */
   public function log($level, $message, $context = [], $options=[]) {
     if ($this->isSuppressed()) return;
+
+    $options = array_merge($this->global_options, $options);
 
     $level = strtolower($level);
     if ($level != 'info' and $level != 'notice') {
--- a/tripal/src/TripalImporter/TripalImporterBase.php
+++ b/tripal/src/TripalImporter/TripalImporterBase.php
@@ -179,6 +179,7 @@ abstract class TripalImporterBase extends PluginBase implements TripalImporterIn
   public function setJob($job) {
     $this->job = $job;
     $this->logger->setJob($job);
+    $this->logger->setGlobalOption('logger', FALSE);
   }
 
   /**


laceysanderson (commented):
We can actually predict the type_id in the publish test since we inserted them. You would do it as follows:

```patch
diff --git a/tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php b/tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
index 3a5651d77..9d471bde7 100644
--- a/tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
+++ b/tripal_chado/tests/src/Functional/Services/ChadoTripalPublishTest.php
@@ -507,36 +507,31 @@ class ChadoTripalPublishTest extends ChadoTestBrowserBase {
       'There should only be one published entity for a single organism with new properties.');
 
     // Check that the property values got published. The type_id should be
-    // 0 in the drupal field table, the default for an integer.
-
-    // With the new scheme the value of type_id isn't predictable any more
-    // Also, it does make little sense to check for a value NOT being stored
-    /*
+    // match the type of property that was created above.
     $this->checkFieldItem('organism', 'field_note', 1,
         ['record_id' => $organism_id, 'prop_id' => 1],
-        ['type_id' => 0, 'linker_id' => $organism_id,
+        ['type_id' => $note_type_id, 'linker_id' => $organism_id,
          'bundle' => 'organism', 'entity_id' => 1]);
 
     $this->checkFieldItem('organism', 'field_note', 1,
         ['record_id' => $organism_id, 'prop_id' => 2],
-        ['type_id' => 0, 'linker_id' => $organism_id,
+        ['type_id' => $note_type_id, 'linker_id' => $organism_id,
          'bundle' => 'organism', 'entity_id' => 1]);
 
     $this->checkFieldItem('organism', 'field_note', 1,
         ['record_id' => $organism_id, 'prop_id' => 3],
-        ['type_id' => 0, 'linker_id' => $organism_id,
+        ['type_id' => $note_type_id, 'linker_id' => $organism_id,
          'bundle' => 'organism', 'entity_id' => 1]);
 
     $this->checkFieldItem('organism', 'field_comment', 1,
         ['record_id' => $organism_id, 'prop_id' => 4],
-        ['type_id' => 0, 'linker_id' => $organism_id,
+        ['type_id' => $comment_type_id, 'linker_id' => $organism_id,
          'bundle' => 'organism', 'entity_id' => 1]);
 
     $this->checkFieldItem('organism', 'field_comment', 1,
         ['record_id' => $organism_id, 'prop_id' => 5],
-        ['type_id' => 0, 'linker_id' => $organism_id,
+        ['type_id' => $comment_type_id, 'linker_id' => $organism_id,
          'bundle' => 'organism', 'entity_id' => 1]);
-    */
     // Check that only the exact number of properties were published.
     $this->checkFieldItem('organism', 'field_note', 3, ['entity_id' => 1], []);
     $this->checkFieldItem('organism', 'field_comment', 2, ['entity_id' => 1], []);
```

I've tested this locally and the tests pass :-).

laceysanderson (changes requested):
Regarding this approach: 

> These changes address https://github.com/tripal/tripal/issues/1828 to allow proper filtering and sorting of chado-derived data in Views, according to Solution 1 described https://github.com/tripal/tripal/issues/1828#issuecomment-2541585273:

> > The first approach involves ensuring that while publishing Tripal content, fields like organism_common_name_value are populated with meaningful values rather than left empty. This would allow the Views filter to function correctly without modifying its logic. However, this approach could introduce redundancy and might conflict with the current design, where these fields are likely left empty intentionally.

I am in favour of this approach as a fix for Drupal views. While it does duplicate data, I think of it more as caching as compared to duplicating. Plus I appreciate that there would not be a lot of upkeep and maintenance of this approach would be relatively easy.

### Configurable

I would like to see a global configuration value for this though so individual sites could opt-in or out depending on their needs. Specifically, we could add an option in the "Tripal Entity Configuration" form as follows:

> ☑️ Cache Backend Storage field values in Drupal
> *When enabled, a copy of data from the backend storage will be stored in the Drupal field tables. **This is needed for Drupal views filtering and sorting!***

This would be enabled by default and stored as a Drupal configuration variable. That variable can then be used in chado storage to either use a default of `TRUE` or `FALSE` for $is_required depending on if caching is enabled or disabled respectively.

## Manual Testing

### ✅ Publishing content should no populate all the columns in the drupal field tables.

To test this I imported the `tripal_chado/tests/fixtures/gff3_loader/TripalusDatabasicaChr1Genes.gff3` into my development site on this branch and then published it.

When inspecting the drupal tables associated with the fields I was able to confirm that all the properties were now being populated.

### ✅ Views filter/sort should now work since the values are all in the drupal table.

I created a view of Tripal genes after importing the same GFF3 mentioned above 3 times where each time I used a different organism. 

I was then able to confirm that setting the sort to the name of the gene, sorted properly grouping all genes with the same name together even though that is not the order they were inserted and thus is different from the original sort behaviour on 4.x.
image

I was also able to confirm that entering the genus `Lens` in my example, filtered the view to all genes with an organism of `Lens culinaris` and filtering to a genus of `Tripalus` showed all genes from both `Tripalus databasica` and `Tripalus ervoides`.
image
image


### ❌ Editing the page through the UI should have the same behaviour as publish

ChadoStorage is used both for publishing chado content and for updating chado content through the UI. As such, we need to ensure that both processes produce the same values in the underlying Drupal tables. **This means any Tripal content edited through the UI would cease to filter/sort correctly in views!**

However, with the changes currently in this PR they don't. More specifically, when publishing a gene all the organism information is stored in the drupal field table as designed in this PR. However, editing that same gene through the UI does not save all the organism information in the drupal field table.

NOTE: This PR does not break editing through the UI 🎉 The original properties are still updated in the drupal table correctly (i.e. in the following example, the `gene_organism_organism_id` correctly changes from a `2` (the chado.organism.organism_id for `Tripalus ervoides` in my test site) to `4` (the chado.organism.organism_id for `Tripalus databasica subspecies postgresquelus` in my test site).

When I publish a gene, the organism field looks like this:
```
bundle | deleted | entity_id | revision_id | langcode | delta | gene_organism_record_id | gene_organism_entity_id | gene_organism_organism_id | gene_organism_organism_genus | gene_organism_organism_species | gene_organism_organis
m_infraspecific_type | gene_organism_organism_infraspecific_name | gene_organism_organism_scientific_name | gene_organism_organism_abbreviation | gene_organism_organism_common_name | gene_organism_organism_comment 
--------+---------+-----------+-------------+----------+-------+-------------------------+-------------------------+---------------------------+------------------------------+--------------------------------+----------------------
---------------------+-------------------------------------------+----------------------------------------+-------------------------------------+------------------------------------+--------------------------------
 gene   |       0 |       104 |         104 | und      |     0 |                    1392 |                      53 |                         2 | Tripalus                     | ervoides                       |                      
                     |                                           | Tripalus ervoides                      |                                     |                                    | 
```

Then when I edit that gene through the UI and use the organism field to change the organism it looks like this:

```
bundle | deleted | entity_id | revision_id | langcode | delta | gene_organism_record_id | gene_organism_entity_id | gene_organism_organism_id | gene_organism_organism_genus | gene_organism_organism_species | gene_organism_organis
m_infraspecific_type | gene_organism_organism_infraspecific_name | gene_organism_organism_scientific_name | gene_organism_organism_abbreviation | gene_organism_organism_common_name | gene_organism_organism_comment 
--------+---------+-----------+-------------+----------+-------+-------------------------+-------------------------+---------------------------+------------------------------+--------------------------------+----------------------
---------------------+-------------------------------------------+----------------------------------------+-------------------------------------+------------------------------------+--------------------------------
 gene   |       0 |       104 |         104 | und      |     0 |                    1392 |                     155 |                         4 |                              |                                |                      
                     |                                           |                                        |                                     |                                    | 
```

mdondrup:
- [x] Configurable

> I would like to see a global configuration value for this though so individual sites could opt-in or out depending on their needs. Specifically, we could add an option in the "Tripal Entity Configuration" form as follows:
☑️ Cache Backend Storage field values in Drupal
When enabled, a copy of data from the backend storage will be stored in the Drupal field tables. This is needed for Drupal views filtering and sorting!

Configuration option has been added to `/admin/tripal/config/entity_settings`. 
image


> This would be enabled by default and stored as a Drupal configuration variable. That variable can then be used in chado storage to either use a default of TRUE or FALSE for $is_required depending on if caching is enabled or disabled respectively.

The setting is stored in a tripal.settings variable: `tripal_entity_type.default_cache_backend_field_values` 

- [x]  Editing the page through the UI should have the same behaviour as publish

> ChadoStorage is used both for publishing chado content and for updating chado content through the UI. As such, we need to ensure that both processes produce the same values in the underlying Drupal tables. This means any Tripal content edited through the UI would cease to filter/sort correctly in views!

A public function has been added to ChadoStorage that provides the logic to determine field caching for one combination of $field_name and $key in a consistent way. It is used by both, batch publish and the Tripal Data edit/create Form. 

- [x] Unit testing

Unit tests have been updated to cover both configuration settings.

mdondrup:
Hi @laceysanderson and @dsenalik. My last commit should have fixed the issue with the unit tests. Surprisingly, the error did never appear during my local unit test run using the docker container.  Please let me know if there are more comments or updates you need. I can see the Test coverage process is still failing but it looks like a configuration issue.

Cheers
Michael

laceysanderson:
The expanded testing and the configuration look great!

Unfortunately the changes intended to fix the editing through the UI have not worked 🙈 Now when I edit a previously published page through the UI, make no changes and click save, I actually lose field values. I think I see why this is happening but it's difficult to explain 🤔 and I worry the solution is not straight forward. 😞 

I wonder if it might be better tackled in a separate PR? We could reverse the changes to TripalEntity.php and then we would have a fully working entity and publish system again. It would still be an intermediate since editing or creating through the UI would not support views but publishing would... but at least nothing that was working would be broken.
What do you think?

mdondrup:
Hi @laceysanderson

I am unable to reproduce this. Please give me the data types and entity_field tables where you get this behavior. I tested with Organisms and saved multiple times, with changes to fields and without changing anything. None of the field values in the tables `tripal_entity__organism_*` have disappeared. The only way to remove a complete row is when you unpublish an entry as expected. My branch is up to date with 'origin/tv4g1-issue1828-add-entity-field-values-publish'.

dsenalik:
I think I can replicate what @laceysanderson is seeing:
 - I created an organism, analysis, project, and contact.
 - I imported the genomic content collection.
 - I imported the gff from issue #1824
 - I published "gene" content
 - Edit the `PAU8` gene, add the project and the contact, and save
 - The organism is gone, and neither the project nor contact are saved either.
```
select * from tripal_entity__gene_organism;
 bundle | deleted | entity_id | revision_id | langcode | delta | gene_organism_record_id | gene_organism_entity_id | gene_organism_organism_id | gene_organism_organism_genus | gene_organism_organism_species | gene_organism_organism_infraspecific_type | gene_organism_organism_infraspecific_name | gene_organism_organism_scientific_name | gene_organism_organism_abbreviation | gene_organism_organism_common_name | gene_organism_organism_comment 
--------+---------+-----------+-------------+----------+-------+-------------------------+-------------------------+---------------------------+------------------------------+--------------------------------+-------------------------------------------+-------------------------------------------+----------------------------------------+-------------------------------------+------------------------------------+--------------------------------
(0 rows)
```

mdondrup:
Hi Douglas,

Thank you for the information, I managed to reproduce it now. There is even an easier test case where the Gene-Organism will not be saved and leads to deletion of the entry  from `tripal_entity__gene_organism;` when the new configuration option is ON. It works when it is OFF. 

That is interesting because the only relevant change in `TripalEntry.php:preSave` is this: 
```
805: - if (array_key_exists('drupal_store', $settings) and $settings['drupal_store'] == TRUE) {
     + if ($storage->isDrupalStoreByFieldNameKey($field_name, $key)) {
                $prop_values[] = $prop_value;
```
So it is simply more often true now, than it was before.

That leads me to conclude that it is not my code per se.Instead, it triggers some never previously executed code further down the line, because there were no storage types for which drupal_store was set to true. Indeed, it triggers some  delta remove logic, that has never been executed before. I propose that this part of the code  has only been safe for property_types that contain a single field. However, for content types connected by a linker table, there are multiple property values stored. 
The delta_delete logic checks if ANY of the property values is null and marks the $delta for deletion by 
```
$this->get($field_name)->removeItem($delta); // this likely deletes the whole row.
```
However, it should have checked if ALL values are null and only then can the whole entry be deleted. 

I will try to include a fix for that functionality in this PR. Or should it this be a separate issue?

dsenalik:
If we are only seeing this bug finally show up in this PR I think it is appropriate to fix it here too.

mdondrup:
The latest commit addresses the issue with the missing cached values from other tables. This seems to be working for the first time. 
Testing:

- set the Tripal entity caching option to On
- create an organism
- create a test.gene
- create a publication
- edit the gene and link the publication and organism
- the table values that are inserted during publishing are now updated
- create a gene view with filters for  fields that are found in publication and organism

Test to filter on referenced values.
Edit: Used more extensive testing with real-world data. Each data type was edited multiple time and linked to each other.

-----------

Created a project:
Screenshot 2025-01-29 at 3 33 49 PM

------------
Publications:
- 
Screenshot 2025-01-29 at 3 38 39 PM
-
Screenshot 2025-01-29 at 3 40 03 PM

------------

Gene linked to publications:
Screenshot 2025-01-29 at 3 50 14 PM

Testing Views:
----------------

Organism view:
Screenshot 2025-01-29 at 3 53 26 PM

----------------

Publication view:
Screenshot 2025-01-29 at 4 03 48 PM

-----------------

Gene view:
Screenshot 2025-01-29 at 4 05 51 PM

dsenalik:
I am testing with my migrated chado, and hit an error when publishing `project`
```
root@9f3e0377f5a3:/var/www/drupal/web# drush tripal-chado:publish project --batch-size=500 --migration-file=tripal3_entity_mapping.tsv
 [notice] Initializing publish
 [notice] Updating tripal entity sequence to start at 411926
 [notice] Loaded 405835 records of migration data, maximum entity ID is 411925
 [notice] Finding all candidate records in the "project" chado table
 [notice] Preparing to publish 293 "project" records
 [notice] Step 1 of 6: Find matching records
 [warning] Undefined array key "path_array" ChadoStorage.php:805
 [error]  TypeError: Drupal\tripal_chado\Plugin\TripalStorage\ChadoStorage::getPathValueColumn(): Argument #1 ($path) must be of type array, null given, called in /var/www/drupal/web/modules/contrib/tripal/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php on line 805 in Drupal\tripal_chado\Plugin\TripalStorage\ChadoStorage->getPathValueColumn() (line 1214 of /var/www/drupal/web/modules/contrib/tripal/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php) #0 /var/www/drupal/web/modules/contrib/tripal/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php(805): Drupal\tripal_chado\Plugin\TripalStorage\ChadoStorage->getPathValueColumn(NULL)
...
```

That line is here
```
  protected function handleStoreID(array $context, StoragePropertyValue $prop_value) {

    $base_table = $context['base_table'];
    $record_id = $prop_value->getValue();
    $value_col_info = $this->getPathValueColumn($context['path_array']);
...
```

That array key is only added if `path` is present in the storage settings for a `store_id`, and apparently for project_assay it is not
```
          // Get the path array for this field and add any joins if any are needed.
          if (array_key_exists('path', $prop_storage_settings)) {
```

This is a bug in `tripal_chado/src/Plugin/Field/FieldType/ChadoAssayTypeDefault.php` so is actually unrelated to this PR, I'll make a separate issue for it #2098 and fixed in PR #2099

```
    // Define the base table record id.
    $properties[] = new ChadoIntStoragePropertyType($entity_type_id, self::$id, 'record_id', self::$record_id_term, [
      'action' => 'store_id',
      'drupal_store' => TRUE,
//      'chado_table' => $base_table,   replace these two lines with 'path'
//      'chado_column' => $base_pkey_col,
      'path' => $base_table . '.' . $base_pkey_col,
    ]);
```

mdondrup:
Hi, I have done some more testing and refactoring, also codeclimate is happy now. All tests except Test coverage are passing now.  I won't be able to fix this test, but I don't think it's very relevant either. Could you take a look at the PR again, please?

dsenalik:
I am merging the three PR branches `tv4g1-issue2033-check_for_new_fields`, `tv4g10-issue2074-more-autocompletes-for-fields`, and `tv4g1-2098-assay-field-bug` in my testing since I am also testing migrating my Tripal 3 chado.

Yesterday I ran out of memory when publishing, but reducing batch size to 500 allowed that to work.

Right now, I cannot publish `analysis` because I am out of memory even when reducing batch size to 10
`drush tripal-chado:publish analysis --migration-file=tripal3_entity_mapping.tsv --batch-size 10`
```
 [notice] Initializing publish
 [notice] Loaded 405835 records of migration data, maximum entity ID is 411925
 [notice] Finding all candidate records in the "analysis" chado table
 [notice] Preparing to publish 978 "analysis" records in 98 batches
 [notice] Batch 1 of 98, Step 1 of 6: Find matching records
PHP Fatal error:  Allowed memory size of 1077936128 bytes exhausted (tried to allocate 20480 bytes) in /var/www/drupal/web/core/lib/Drupal/Component/Serialization/PhpSerialize.php on line 21

Fatal error: Allowed memory size of 1077936128 bytes exhausted (tried to allocate 20480 bytes) in /var/www/drupal/web/core/lib/Drupal/Component/Serialization/PhpSerialize.php on line 21
 [warning] Drush command terminated abnormally.
```
I will try and see what is happening...

dsenalik:
Ha, this is because of adding the `analysisfeature` table to field discovery. I have 1979354 records in that table!
Unrelated to this PR!
This is actually the `max_delta` problem https://github.com/tripal/t4d8/issues/310
I have tens of thousands of genes linked to a single analysis.
I shall remove that field and continue testing

(I created a new issue for this, #2101)

mdondrup:
Hi @dsenalik, ofc I would be glad to help solve any performance problems that may exist elsewhere :)
But you are raising a valid point, I do expect the patch to have some slight negative performance and memory impact.

mdondrup:
Hi @laceysanderson. I was wondering if there is any more testing that should be done before a merge.

laceysanderson:
I would like to do one more solid manual test on this one just to make sure it's not impacting things we don't expect. It is a high priority for me --sorry it's taking awhile I just have so much on my plate.

laceysanderson:
I am working on some coding standard fixes in the branch [`mdondrup-tv4g1-issue1828-add-entity-field-values-publish`](https://github.com/tripal/tripal/tree/mdondrup-tv4g1-issue1828-add-entity-field-values-publish) which I will then contribute back to this PR. I'm also really concerned with the circular dependency added in this PR where the tripal module is calling chadopublish directly 😬 Once I'm done cleaning up a few more things I'll take a look at the best way to resolve that.

You can see the current changes here: https://github.com/mdondrup/tripal/compare/tv4g1-issue1828-add-entity-field-values-publish...tripal:tripal:mdondrup-tv4g1-issue1828-add-entity-field-values-publish.

mdondrup:
Hi @laceysanderson, thank you for checking and correcting the coding style. I apologize for the deviations. Some points:

 - Regarding the circular dependency on ChadoPublish, it may be possible to refactor this code into its own module.  
 - I had problems determining the correct delta value to use. From what I have seen, it was  == 1 in all cases. What would be the method to determine the delta value for the current entity?

laceysanderson:
> Hi @laceysanderson, thank you for checking and correcting the coding style. I apologize for the deviations. Some points:
>
>  1. Regarding the circular dependency on ChadoPublish, it may be possible to refactor this code into its own module.
>  2. I had problems determining the correct delta value to use. From what I have seen, it was == 1 in all cases. What would be the method to determine the delta value for the current entity?

Regarding the circular dependancy, I think we can solve this by using the TripalBackendPublish plugin type of which chadopublish is an instance to ensure we get the right publish for the TripalStorage backend. This way we remove the assumption that all backends are chado which is currently in this PR.

Good to know about the delta, I'll look at that too :-)

I started my changes as a PR on your fork which you can see here https://github.com/mdondrup/tripal/pull/1. Right now it is a draft because I'm not done yet. Once it's complete, I'll mark it ready to review and you can take a look at it. If you approve of my changes then you can merge that PR and the changes will show up here.

laceysanderson (changes requested):
The linked PR resolves the circular dependency and removes the need to republish for a single entity.

One thing that's harder for me to resolve is that some commits were done as root. We do require signed commits in Tripal core so if you could resolve those, that would be great! I haven't done this myself but I believe you will need to use rebase to modify commits that far back. This was the [simpler tutorial](https://www.deployhq.com/git/faqs/update-author-committer-multiple-git-commits) I found but many mentioned using the interactive rebase to do multiple commits at once. One really key thing is that you will need to do a force push. Be Really careful of this, you can lose some of your work! I suggest making a copy of this branch in your fork and making the fixes on it... if it works we can make a fresh PR on that new branch pretty easily.

PR #2086, opened 2025-01-13
Entity lookup for tables with multiple content types

### LINKED ISSUE #2085: Entity lookup for tables with multiple content types
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

Content types where there is more than one bundle for a given chado base table are a problem
For example, the `analysis` table has a "plain" analysis bundle, but also two derived content types "Genome Assembly" and "Genome Annotation".
The core analysis field cannot **link** to more than one type, but will allow manual addition of all three types and will display all three types. Same with publish, it doesn't care about the type and will publish all of them.

### Steps to reproduce

1. Import the genomic content type collection
2. Create a "plain" analysis
3. Create a "Genome Assembly"
4. Create a "Genome Annotation"
5. Create a project, and add each of these three
6. Observe only the "plain" analysis gets a link
analysis_types
7. Unpublish the project, then publish project
8. Looks exactly the same

### Options

A fallback lookup already exists here
https://github.com/tripal/tripal/blob/a65073c124cee0343aa8a75657f3591e076b6d2d/tripal/src/Services/TripalEntityLookup.php#L94-L110
I'm thinking we could cache the list of bundles that `getBundles()` retrieves, and then the fallback method would not be much of a performance hit.

dsenalik:
# Bug Fix

## Description
This change uses the existing "fallback" entity lookup method, which looks up all bundles for a given base table, in all cases.
This allows us to easily support fields that link to tables that have more than one content type.
For example, the `analysis` table has a "generic" analysis, plus "Genome Assembly" and "Genome Annotation".
The `stock` table has four content types. The `project` table has two. The `feature` table has a bunch.

This involves not using the bundle as we had done, but just use the existing lookup to find all bundles for a table.
This lookup takes around 7 ms on my test docker, but it is cached in a class variable, so subsequent cached calls take only a couple of microseconds.

## Testing?
1. Import the genomic content type collection
2. Create a "plain" analysis
3. Create a "Genome Assembly"
4. Create a "Genome Annotation"
5. Create a project, and add each of these three
6. Observe that on this branch, all three types of analysis get working links
issue2085 2
7. Unpublish the project, then publish project
8. Should work here also

### DIFF
--- a/tripal/src/Services/TripalEntityLookup.php
+++ b/tripal/src/Services/TripalEntityLookup.php
@@ -17,6 +17,13 @@ use Drupal\Core\Render\Markup;
 class TripalEntityLookup {
 
   /**
+   * Caches bundle IDs used for a base table
+   *
+   * @var array $bundle_cache
+   */
+  protected array $bundle_cache = [];
+
+  /**
    * This is used by field formatters to get a ready-to-use render
    * array item to link to an entity.
    *
@@ -62,6 +69,7 @@ class TripalEntityLookup {
    *   The bundle's CV Term namespace e.g. for gene "SO"
    * @param string $termAccession
    *   The bundle's CV term accession e.g. for gene "0000704"
+   *   If IdSpace and Accession are NULL, then any bundle using the base table is checked.
    * @param string $base_table
    *   The Chado base table for the requested entity e.g. for gene "feature".
    *   Only needed if term does not map to a content type.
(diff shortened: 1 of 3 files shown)


Ferrisx4 (approved):
Tested this as described in the PR and also additionally with Organism and the various Analysis subtypes (genome annotation and genome assembly). All types of analyses are picked up and linking properly from organism pages, and each analysis page regardless of type links to the organism.

Ferrisx4:
I just tried creating the Gene Functional Annotation type manually and used all the same fields that a standard Analysis type has, and ran into trouble when publishing. Despite being configured to use the project_analysis linker table, the publish job failed with a PDO error claiming that the column `analysis.project` does not exist.

I'm not sure if it is related to this PR, I will test it on 4.x before merging this.

Issue #2087, opened 2025-01-14
GFF Importer dbxref without cvterm can cause error


dsenalik:
### Tripal Version

Both 3.x and 4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

This simple test GFF file will generate the error shown below
[issue2087.gff.txt](https://github.com/user-attachments/files/18414919/issue2087.gff.txt)

### Steps to reproduce

You need a site where the gene ontology (GO) ontology has not been loaded
Load supplied test file above with GFF importer

### Error messages

**On 4.x:**
```
...
 [notice] Step 23 of 27: Insert feature ontology terms...                    
Percent complete: 0%. Memory: 37,107,136 bytes.
 [notice] Percent complete: 0%. Memory: 37,107,136 bytes.
 [warning] Undefined array key "GO:0016021" GFF3Importer.php:2691
 [warning] Undefined array key "GO:0046872" GFF3Importer.php:2691
...
ERROR: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "cvterm_id" of relation "feature_cvterm" violates not-null constraint
DETAIL:  Failing row contains (1, 12, null, 1, f, 0).: INSERT INTO "chado"."feature_cvterm" (feature_id, cvterm_id, pub_id) VALUES 
(:feature_id_1, :cvterm_id_1, :pub_id_1),
(:feature_id_2, :cvterm_id_2, :pub_id_2); Array
(
    [:feature_id_1] => 12
    [:cvterm_id_1] => 
    [:pub_id_1] => 1
    [:feature_id_2] => 12
    [:cvterm_id_2] => 
    [:pub_id_2] => 1
)
```

**On 3.x:**
```
...
Step 23 of 27: Insert feature ontology terms...                  
Undefined array key "GO:0016021" GFF3Importer.inc:2577                                                                           [warning]
Undefined array key "GO:0046872" GFF3Importer.inc:2577                                                                           [warning]
Removing temporary cache file: /tmp/TripalGFF3Import_FdJ95b
SQLSTATE[25P02]: In failed sql transaction: 7 ERROR:  current transaction is aborted, commands ignored until end of transaction block
[site http://default] [TRIPAL ERROR] [TRIPAL_JOB] SQLSTATE[25P02]: In failed sql transaction: 7 ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

### Screenshots

.

PR #2088, opened 2025-01-14
GFF Importer dbxref without cvterm can cause error

dsenalik:
# Bug Fix

## Description
This fixes for Tripal 4.x the case where an imported GFF file has a dbxref but without an associated cvterm.

## Testing?
See Issue #2087 for how to replicate the bug.

### DIFF
--- a/tripal/src/Services/TripalEntityLookup.php
+++ b/tripal/src/Services/TripalEntityLookup.php
@@ -17,6 +17,13 @@ use Drupal\Core\Render\Markup;
 class TripalEntityLookup {
 
   /**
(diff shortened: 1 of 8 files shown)


laceysanderson (approved):
✅ Code changes is intuitive and small.
✅ Manual testing using the file supplied in the linked issue ran without error on this branch.
Screenshot 2025-01-23 at 3 36 50 PM

✅ I also went one step further and published the mRNA after discovering the two new property fields. I confirmed that the properties and dbxrefs were imported and published as expected :-)
image

PR #2091, opened 2025-01-21
remove redundant preg_split

### LINKED ISSUE #2090: Editing (and sometimes adding) Chado Type Reference fields preg_replace error
### Tripal Version

latest pull on 4.x (commit a65073c124cee0343aa8a75657f3591e076b6d2d) as of writing this issue.

### Drupal Version

10.3.1

### PostgreSQL Version

16.2 

### PHP Version

8.3.10

### General Description

[This commit](https://github.com/tripal/tripal/commit/6afa0c04706e3b073c3e279e32fcc6b6858c5d71) makes the `getTableDef()` function always return an array. The function `getTypeFKeys()` in the `ChadoAdditionalTypeTypeDefault` class expects a string here: https://github.com/tripal/tripal/blob/a65073c124cee0343aa8a75657f3591e076b6d2d/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php#L399

Since an array seems to be fine throughout the rest of Tripal, we can adjust this class as such since we don't need any further array-making:
```diff
-      $fkey_list = preg_split('/[ ,]+/', ($base_schema_def['referring_tables']??''));
+      $fkey_list = $base_schema_def['referring_tables']??'';
```
although since $fkey_list is expected to be an array later on, we'd probably want
```diff
-      $fkey_list = preg_split('/[ ,]+/', ($base_schema_def['referring_tables']??''));
+      $fkey_list = $base_schema_def['referring_tables']??[];
```

---
Currently this issue causes a "The website encountered an unexpected error. Try again later." error when trying to edit a Chado Type Reference field, such as on an Analysis, Genome Assembly, and Genome Annotation types.

### Steps to reproduce

On a content type that includes a Chado Type Reference field, try to edit it: /admin/structure/bio_data/manage/genome_annotation/fields (this requires the Genomic content collection to be installed. Alternatively, add a Chado Type Reference field to the Analysis type, then try to edit it afterwards.

I believe this may also be impacting the ability to further add other Chado Type Reference fields. While writing this issue, I tried to add one to the Organism content type and was met with the same error (and root cause, according to the log).

### Error messages

```shell
TypeError: preg_split(): Argument #2 ($subject) must be of type string, array given in preg_split() (line 399 of /opt/drupal/web/modules/custom/tripal/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php).
```

### Screenshots

_No response_

Ferrisx4:
# Bug Fix

## Description

Updates the ChadoAdditionalTypeTypeDefault class' `getTypeFKeys()` function to expect an array already since the `getTableDef()` always returns an array now as of https://github.com/tripal/tripal/commit/6afa0c04706e3b073c3e279e32fcc6b6858c5d71

## Testing?

On a content type that has a Chado Type Reference field or after adding a Chado Type Reference field to a content type that does not have this field already, and then trying to edit it, the site would WSOD.

Then switch to this branch and try the same and observe that the site does not WSOD.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoAdditionalTypeTypeDefault.php
@@ -396,7 +396,7 @@ class ChadoAdditionalTypeTypeDefault extends ChadoFieldItemBase {
 
       // Get a list of tables with foreign keys to selected $base_table.
       $base_schema_def = $schema->getTableDef($base_table, ['format' => 'Drupal']);
-      $fkey_list = preg_split('/[ ,]+/', ($base_schema_def['referring_tables']??''));
+      $fkey_list = $base_schema_def['referring_tables']??[];
       asort($fkey_list);
 
       // Include the base table at the top of the sorted list


dsenalik (approved):
I confirmed on a content type where the additional type field does not define the bundle (pub) and on a type where it does (gene) and in both cases this eliminates the WSOD. A simple fix and easy PR review as promised 😀

dsenalik:
The test coverage failure is when accessing pubmed, so not really relevant, and for a one-line change test coverage is not going to change anyway.

Issue #2033, opened 2024-11-16
Improvements for Check for new fields


dsenalik:
### Task Description

There may be some improvements needed for "+Check for new fields"

1. For content types directly mapped to a table e.g. Biosample, and even for Analysis, we offer to add a type reference field
![2024-11-16_biosample_type](https://github.com/user-attachments/assets/e8506485-3819-4dd9-b3f5-573e505293c4)
and
![2024-11-16_analysis_type](https://github.com/user-attachments/assets/05b4dcbc-2e44-40a0-963d-eface44b54fb)
If you add this, you have now broken your content type, in the sense that you can no longer republish existing content. So you won't be able to get any other fields you just added. And currently there is no way to fix this, which requires removing the `bundle_type_table` and `bundle_type_column` third party settings, which are added when this type reference field is added.
This was made unchecked by default in #2077 

2. Field detection gets confused if there happen to be two properties with the exact same name, but in different vocabularies. One gets added, but the other is suggested for addition but can never be added because it will have the same field name.
This was Fixed by #2096

### Branch Name

tv4g1-issue2033-check_for_new_fields

dsenalik:
### How to replicate Item 2

1. Create a project with two properties with two different `genotype` terms as follows
```
INSERT INTO project (name, description) VALUES ('testproject', 'fake');
INSERT INTO projectprop (project_id, type_id, value) VALUES ((SELECT project_id FROM project WHERE name='testproject'), (SELECT cvterm_id FROM cvterm T left join cv CV on T.cv_id=CV.cv_id WHERE T.name='genotype' AND CV.name='taxonomic_rank'), 'Copy1');
INSERT INTO projectprop (project_id, type_id, value) VALUES ((SELECT project_id FROM project WHERE name='testproject'), (SELECT cvterm_id FROM cvterm T left join cv CV on T.cv_id=CV.cv_id WHERE T.name='genotype' AND CV.name='sequence'), 'Copy2');
```
2. Tripal -> Page Structure -> Project -> Manage Fields +Check for new fields
3. Uncheck the type field and click "Add Fields" to add the property field
![Image](https://github.com/user-attachments/assets/3166fc97-7c71-49c0-a12c-476eb2eee664)
4. Repeat. Note there is now one field existing, and the description is of the "other" `genotype` term
![Image](https://github.com/user-attachments/assets/7a7a9329-b71e-40bc-a2b9-c27f342adaaa)
5. You get a success message, but there will still only be one of the two properties
[...]

PR #2096, opened 2025-01-23
Resolves duplicated field names during discover

dsenalik:
# Bug Fix

## Description
This resolves the problem of duplicated field names from the `discover()` process, when two properties share a CV term, but the term is in a different CV.
Previously only one could be added by discover because discover generated the exact same field name for both fields.

## Testing?
Follow procedure described in this issue comment
https://github.com/tripal/tripal/issues/2033#issuecomment-2610564548

On initial "+Check for new fields" on this branch you will now see

And doing it a second time

### DIFF
--- a/tripal/src/Form/TripalContentFieldsForm.php
+++ b/tripal/src/Form/TripalContentFieldsForm.php
@@ -138,7 +138,7 @@ class TripalContentFieldsForm implements FormInterface {
     ];
     $invalid_fields_desc = t('The following fields do not pass validation tests. '
       . 'They need correction by the module developer and cannot be added.');
-    if (empty($invalid_fields_desc)) {
+    if (empty($invalid_fields)) {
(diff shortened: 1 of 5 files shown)


dsenalik:
This is done ~~, but I will make this draft status until #2077 is merged...~~

dsenalik (commented):
Nice. Thanks

laceysanderson (approved):
✅ Code looks thorough and correct.
✅ Manual test allowed both fields to be detected as expected
Screenshot 2025-01-30 at 2 35 06 PM

After discussion with @dsenalik I made one small additional change which allows discover to more easily set the default name of a field. This would allow a field to use a more tailored approach to reduce the chance of a name collision by simply passing in a field name through the discover options. In situations like the contact by role field, this added change will ensure that the original change in this PR is needed less often.

PR #1960, opened 2024-08-28
Publication Importer Version 5

risharde:
This is the new Publication Importer V5 for Tripal 4
SupportsPub Med importing
Includes 29 unit tests

### ~~Depends on #2006~~ (merged)

PR #2105, opened 2025-02-01
Make all content types typed

### LINKED ISSUE #2097: Make all our content types "typed" by default
### Task Description

Issue #1984 is a request to not publish one table record under more than one content type.
This goes back to Tripal 3, where you could have a "Genome Assembly" analysis record, but it would also get published as a plain "Analysis".
This is because "Genome Assembly" is typed, while "Analysis" is not.
In upcoming future chado versions, we will finally have `type_id` columns in all base tables.

The task here is to add types to those bundles that currently do not have them, which means adding an "ChadoAdditionalTypeType" field to the plain "Analysis" and "Project" content types at a minimum, and any others missing them. This will solve issue #1984.
This involves:
1. Update yaml to add the types (easy)
2. Update hook to update an existing site - see `tripal_chado_update_10407` on how to add fields (not too hard)
3. Documentation on how to do this for a migrated Tripal 3 site, but specifically this primarily means adding it to the migration drush command `drush trp-check-terms` (this is the hard part)
4. Make sure all importers add the type when creating a record, if they don't already

The sql needed was discussed here https://github.com/tripal/tripal/issues/1984#issuecomment-2569579646

There are some cases that make this less than trivial. For example, `organism` has a `type_id` column, but is is used for infraspecific type. `contact` has a `type_id` column, but it does not have a fixed value, so is not available to be used to distinguish content types. So both of these will need the type in the corresponding property table. Protocol does not have a property table defined by chado, so it won't be able to support a type.

### Branch Name

tv4g1-issue2097-make-all-content-types-typed

### Core Table Status

| Table | type_id column | existing fixed_type | needs update |
| --- | --- | --- | --- |
| analysis  | no | yes | analysis |
| arraydesign | no | &nbsp; | yes |
| assay  | no | &nbsp; | yes |
| biomaterial | no | &nbsp; | yes |
| contact  | yes | no | yes |
| feature | yes | yes | no |
| featuremap | no | yes | no |
| library  | yes | yes | no |
| organism  | yes | no | yes |
| phylotree  | yes | yes | no |
| project  | no | yes | project |
| protocol  | yes | no | can't |
| pub  | yes | no | yes |
| stock | yes | yes | no |
| study  | no | &nbsp; | yes |

dsenalik:
# Tripal 4 Core Dev Task

## Description
See the linked issue for details, but basically we want to add types to bundles so that chado records do not get published under more than one content type.
We cannot include the "protocol" content type, however, because there is no `protocolprop` table in chado (chado issue https://github.com/GMOD/Chado/issues/148)

Be aware of the fact that importers must now include the type, otherwise imported content will not be able to be published. The workaround is to run the `drush trp-check-terms` command to add types to any records missing them.
The GFF and FASTA importers will be fine, as all content types like gene, mRNA already have types. But the taxonomy importer, for example would need an update. So there is a task before draft status can be removed:
 - [x] Add types to importers needing them 
 
This PR now adds the type to organisms in the taxonomy importer
The publication importer PR #1960 already included adding the type
GFF and FASTA importers already assign types because they populate the feature table
Newick tree loader already assigns types
Ontology loader doesn't populate a content type so doesn't apply here

## Testing?
1. Build a docker on the 4.x branch
2. Import all of the type collections
  Tripal -> Page Structure -> +Import type collection
3. Either migrate a Tripal 3 chado instance, or make one of each content type: Analysis, Genome Assembly, Genome Annotation, Project, Genome Project
4. Switch to this branch and run the update hook `drush updatedb`
```
root@f603ab48a8cf:/var/www/drupal# drush updatedb
 -------------- ----------- --------------- ------------------------------- 
  Module         Update ID   Type            Description                    
 -------------- ----------- --------------- ------------------------------- 
  tripal_chado   10413       hook_update_n   10413 - Adds types to content  
                                             types that don't have them.    
 -------------- ----------- --------------- ------------------------------- 


 ┌ Do you wish to run the specified pending updates? ───────────┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

>  [notice] Update started: tripal_chado_update_10413
>  [notice] Added field, "analysis_bundle_type", to content type: "analysis".
>  [notice] Added field, "contact_bundle_type", to content type: "contact".
>  [notice] Added field, "organism_bundle_type", to content type: "organism".
>  [notice] Added field, "project_bundle_type", to content type: "project".
>  [notice] Added field, "pub_bundle_type", to content type: "pub".
>  [notice] Added field, "study_bundle_type", to content type: "study".
>  [notice] Added field, "array_design_bundle_type", to content type: "array_design".
>  [notice] Added field, "assay_bundle_type", to content type: "assay".
>  [notice] Added field, "biosample_bundle_type", to content type: "biosample".
>  [notice] Update completed: tripal_chado_update_10413
>  [notice] Message: Successfully completed update
> 
```
5. Run `drush trp-check-terms --chado_schema=chado`
```
Errors
======

Differences are categorized as errors if they are likely to cause failures when preparing this chado
instance or to cause Tripal to be unable to find the term reliably.

 ┌ Would you like more details regarding the errors we found? ──┐
 │ ● Yes / ○ No                                                 │
 └──────────────────────────────────────────────────────────────┘
Missing Bundle Term Issues.
---------------------------

We have detected 8 chado table(s) having one or more records without a bundle term property. Specifically:
+-------------+-------------------+-----------------------+
| CHADO TABLE | NUMBER OF RECORDS | MISSING TERM          |
+-------------+-------------------+-----------------------+
| analysis    | 1                 | operation:2945 (2833) |
| arraydesign | 1                 | EFO:0000269 (2844)    |
| assay       | 8                 | OBI:0000070 (2960)    |
| biomaterial | 2900              | sep:00195 (2975)      |
| contact     | 159               | NCIT:C47954 (65646)   |
| organism    | 1919              | OBI:0100026 (2959)    |
| project     | 293               | NCIT:C47885 (2954)    |
| pub         | 39242             | TPUB:0000002 (106)    |
+-------------+-------------------+-----------------------+

 ┌ Would you like us to add the missing bundle terms to these records? ┐
 │ ● Yes / ○ No                                                        │
 └─────────────────────────────────────────────────────────────────────┘
                                                                                                                        
 [OK] Records with missing bundle terms have been updated.                                                              
```


Ferrisx4 (commented):
I've been testing this out, it seems to mostly work. I tested it on a site that already had Chado Type Reference field added to the Organism content type (using the same term used in this PR) and the `drush updatedb` command successfully found this and gracefully moved past it.

For testing, I also created a new contact before switching to this branch. The bundle gets the new `contact_bundle_type` Chado Type Reference field just fine. However running `drush trp-check-terms --chado_schema=chado` as indicated finds no Term Issues, which I assume is the part where it adds the record to the appropriate prop table. 

I'm trying now to test with other content types but I'm having separate issues with my machine.

dsenalik:
- [x] A bug to fix. On a 4.x docker, without adding any type collections, then switch to this branch 
```
root@9bd778428565:/var/www/drupal# drush updatedb
 -------------- ----------- --------------- ------------------------------- 
  Module         Update ID   Type            Description                    
 -------------- ----------- --------------- ------------------------------- 
  tripal_chado   10414       hook_update_n   10414 - Adds types to content  
                                             types that don't have them.    
 -------------- ----------- --------------- ------------------------------- 


 ┌ Do you wish to run the specified pending updates? ───────────┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

>  [notice] Update started: tripal_chado_update_10414
>  [notice] Added field, "analysis_bundle_type", to content type: "analysis".
>  [notice] Added field, "contact_bundle_type", to content type: "contact".
>  [notice] Added field, "organism_bundle_type", to content type: "organism".
>  [notice] Added field, "project_bundle_type", to content type: "project".
>  [notice] Added field, "pub_bundle_type", to content type: "pub".
>  [notice] Added field, "study_bundle_type", to content type: "study".
>  [error]  ERROR: The specified entity type, "array_design", for field "array_design_bundle_type", does not exist. 
> [site http://default] [TRIPAL] ERROR: The specified entity type, "array_design", for field "array_design_bundle_type", does not exist.
>  [error]  ERROR: The specified entity type, "assay", for field "assay_bundle_type", does not exist. 
> [site http://default] [TRIPAL] ERROR: The specified entity type, "assay", for field "assay_bundle_type", does not exist.
>  [notice] Update completed: tripal_chado_update_10414
>  [notice] Message: Successfully completed update
> 
```

Ferrisx4:
I am also seeing that error message now, about `array_design` and `assay`.

dsenalik:
Fixed all the bugs I hope!

Running without the expression collection loaded it looks like this now
```
>  [notice] Update started: tripal_chado_update_10414
>  [notice] Added field, "analysis_bundle_type", to content type: "analysis".
>  [notice] Added field, "contact_bundle_type", to content type: "contact".
>  [notice] Added field, "organism_bundle_type", to content type: "organism".
>  [notice] Added field, "project_bundle_type", to content type: "project".
>  [notice] Added field, "pub_bundle_type", to content type: "pub".
>  [notice] Added field, "study_bundle_type", to content type: "study".
>  [notice] Update completed: tripal_chado_update_10414
>  [notice] Message: Successfully completed update
```

Ferrisx4:
I've reviewed this in the Docker again it seems to mostly work, as least how I'm expecting it to work. In my test I added an organism on the 4.x branch, switched to the issue branch and ran `drush updatedb`. Is this the step where it is supposed to add the type to the _organismprop_ table? I also ran `drush trp-check-terms --chado_schema=chado` which updated two tables that were "having one or more records without a bundle term property", but these were `contact` and `pub`. I created a second organism which does get the _organismprop_ entry.

Interestingly, the first organism, when attempting to edit it, does have the proper type associated to it after switching to the issue branch and running the update.

I suppose I just want to make sure I'm expecting the right things when I'm testing this.

dsenalik:
Are you sure you added the organism on 4.x, and before running `drush updatedb`?
Running `drush updatedb` should add the type field to the organism bundle.
Running `drush trp-check-terms` should add the type record to the `organismprop` table.

I just tried this, made an organism on 4.x, switched to this branch, ran `drush updatedb`, and then ran `drush trp-check-terms --chado_schema=teacup`
```
We have detected 3 chado table(s) having one or more records without a bundle term property. Specifically:
+-------------+-------------------+--------------------+
| CHADO TABLE | NUMBER OF RECORDS | MISSING TERM       |
+-------------+-------------------+--------------------+
| contact     | 1                 | NCIT:C47954 (173)  |
| organism    | 1                 | OBI:0100026 (44)   |
| pub         | 1                 | TPUB:0000002 (208) |
+-------------+-------------------+--------------------+
```
Those two `contact` and `pub` that got added would be the null contact and null publication, that would be expected.

Ferrisx4:
Maybe a niche case, but I tried a few more times to recreate this and was able to - 
1. Create an organism on 4.x branch
2. Switch to issue branch
3. `drush updatedb`
4. Create a second organism. 
5. `drush trp-check-terms --chado_schema=chado`

At step 4, the second organism will have an organismprop entry created:
```sql
sitedb=> select * from organismprop;
 organismprop_id | organism_id | type_id |  value   | rank 
-----------------+-------------+---------+----------+------
               2 |           2 |      44 | organism |    0
```

When running `drush trp-check-terms --chado_schema=chado`, the already-existing Organism will not be detected, and the organismprop will not be updated:
```
We have detected 2 chado table(s) having one or more records without a bundle term property. Specifically:
+-------------+-------------------+--------------------+
| CHADO TABLE | NUMBER OF RECORDS | MISSING TERM       |
+-------------+-------------------+--------------------+
| contact     | 1                 | NCIT:C47954 (173)  |
| pub         | 1                 | TPUB:0000002 (208) |
+-------------+-------------------+--------------------+
```

Interestingly, the organismprop_id starts at 2 despite there never being an entry for organism_id=1 (I made sure to check after each action above).
For reference, my `organism` table:
```sql
sitedb=> select * from organism;
 organism_id | abbreviation |  genus  | species | common_name | infraspecific_name | type_id |                       comment                        
-------------+--------------+---------+---------+-------------+--------------------+---------+------------------------------------------------------
           1 |              | Alpha   | bravo   | First       |                    |         | <p>The first organism. Added on 4.x branch.</p>
           2 |              | Charlie | delta   | Second      |                    |         | <p>Created on issue branch, after drush updatedb</p>
```

---

It seems like something about creating a second organism while on the issue branch but before checking terms throws later detection of missing bundle term properties off.

Ferrisx4:
However, I was able to confirm that if `drush updatedb` and `drush trp-check-terms --chado_schema=chado` are run immediately after one another with no rogue organisms added in between, already-existing entities are found and their bundle term properties updated.

dsenalik:
Thanks for figuring that out. I confirmed it happens for me too, and I traced it to an error in my SQL, I used the wrong table in the condition for the subquery. Fixed with https://github.com/tripal/tripal/pull/2105/commits/37a171ba9f0d83ab7a30312b0798404ba9cdfe56

Ferrisx4:
Thanks, that works. It can now be found by `drush trp-check-terms --chado_schema=chado`, but it does not get applied when selecting (Yes) to add the missing bundle terms.
After playing around with the query in the addMissingBundleTerms() function, I was able to get it to recognize the record with the missing bundle term properties and give it one via this change in 
`tripal_chado/src/Commands/ChadoCheckTermsAgainstYaml.php`

```diff
-      $subquery1->condition('"P".' . $context['pkey'], $subquery2, 'NOT IN');
+      $subquery1->condition('"BT".' . $context['pkey'], $subquery2, 'NOT IN');
```

I'm not sure this is quite the intended solution, deciphering the underlying SQL from this query language requires a bit of mental gymnastics. When I run through the same process as my earlier comment, I end up with organism and organismprop tables that look like this:

```sql
sitedb=> select organism_id, genus, species, common_name from organism;
 organism_id |  genus  | species | common_name 
-------------+---------+---------+-------------
           1 | Alpha   | bravo   | First
           2 | Charlie | delta   | second
(2 rows)

sitedb=> select organismprop_id, organism_id, type_id, value from organismprop;
 organismprop_id | organism_id | type_id |  value   
-----------------+-------------+---------+----------
               2 |           2 |      44 | organism
               3 |           1 |      44 | organism
(2 rows)
```

I'm not sure why organismprop_id starts at 2, may be a different issue.

dsenalik:
Oh that is the exact same mistake, thanks for finding it.
The starting at `2` is because of the double save that is getting fixed in PR #2147

Ferrisx4 (approved):
I've reviewed this over the past few days and it now works as advertised. This will be very helpful going forward, thanks!

laceysanderson:
Awesome! I would like to look at this before it is merged but I'm glad to hear its working well 😍 I'm prioritizing the content save/publish but will come back to this as soon as I get where I need to be there.

I just have something twinkling in my subconscious on this one so I've been waiting for it to make itself known in case we are missing something 🙈

laceysanderson:
A comment regarding this statement in the PR description

> We cannot include the "protocol" content type, however, because there is no protocolprop table in chado (chado issue https://github.com/GMOD/Chado/issues/148)

The protocol table has a type_id so you can type it that way without needing a prop table. See https://chado.readthedocs.io/en/rtd/_static/schemaspy_integration/tables/protocol.html

laceysanderson:
Actually the above comment opens up another comment 🙈 

The following content types already had a chado_additional_type_type_default field attached to the type_id in the base table. Yet this PR adds a second chado_additional_type_type_default field which adds a type in the prop table. This may be the most backwards compatible approach but it also seems confusing to new users of Tripal (and to me ;-p)

Already had a non-fixed type field that maybe should have become fixed:
- [Contact](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml#L940-L965)
- [Protocol](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml#L1134-L1159)
- [Publication](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml#L1256-L1281)

Note: The array_design table has a number of cvterm FKs (Platform + Substrate) but none of them are the type of array design IMO. As such it is not included in the above list because the prop table approach is correct for this one. I feel the same about the organism table since the infraspecific type is not the type of organism 🤪 

The reason I think the approach taken in this PR for these three tables is confusing is because it results in the following scenario: a single contact might have a contact.type_id > "Journal" AND a contactprop type > "Contact". Personally I expect the contact to not need a type of "Contact" (Captain Obvious 🦸) when we know it is a "Journal" ;-p

tripal_chado/src/Commands/ChadoCheckTermsAgainstYaml.php
```diff
@@ -235,6 +236,123 @@ protected function chadoCheckTerms_findCVProblems(&$problems, &$solutions, &$sum
...
+      $subquery->fields('BT', [$pkey]);
+
+      // Query gets the list of records without a bundle term
+      $query = $this->chado->select('1:' . $table, 'BT');
+      $query->leftJoin('1:' . $prop_table, 'P', '"BT".' . $pkey . '=' . '"P".' . $pkey);
+      $query->condition('"BT".' . $pkey, $subquery, 'NOT IN');
+      $query->addField('BT', $pkey, 'pkey_id');
+      $query->distinct();
+      $count = $query->countQuery()->execute()->fetchField();
+
```
laceysanderson, line 316: This makes a really big assumption which may not be correct :grimaces: It assumes that the type will be in the prop table and that is SHOULD be there. What happens if it isn't? Like for example, if this site is using Chado 1.3.3.013 and there is now a project.type_id and project-based content types that have their type stored in the base table.

tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
```diff
@@ -516,13 +523,42 @@ protected function addOrganism($sci_name, $rank) {
...
+  protected function getInfraspecificTypeId(string $infraspecific_type): int {
+    $infraspecific_type = preg_replace('/ /', '_', $infraspecific_type);
+    $cvterm_id = 0;
+    if (array_key_exists($infraspecific_type, $this->infraspecific_types)) {
+      $cvterm_id = $this->infraspecific_types[$infraspecific_type];
+    }
+    else {
+      $cvterm_records = $this->cvterm_buddy->getCvterm([
+        'cv.name' => 'taxonomic_rank',
+        'cvterm.name' => preg_replace('/ /', '_', $infraspecific_type),
```
laceysanderson: Why do you do the preg_replace 2X?

tripal_chado/src/Commands/ChadoCheckTermsAgainstYaml.php
```diff
@@ -235,6 +236,123 @@ protected function chadoCheckTerms_findCVProblems(&$problems, &$solutions, &$sum
...
+   * @param array $problems
+   *   Array containing details for either errors or warnings
+   * @param array $solutions
+   *   Array containing possible solutions for either errors or warnings
+   * @param array $summary_rows
+   *   Infomation for the output table
+   * @param array $options
+   *   Options from drush command line
+   **/
+  protected function chadoCheckTerms_findBundleProblems(&$problems, &$solutions, &$summary_rows, $options) {
```
laceysanderson, line 251: I don't see anything here that excludes the null pub and null contact from being indicated as not having a bundle term and thus having one added as a solution. However the PR description/comments and Codefest convos implied that the null contact/pub would not be part of the generic bundle :confused:

tripal_chado/tests/src/Kernel/ChadoTestKernelBase.php
```diff
@@ -150,4 +150,49 @@ protected function prepareEnvironment(array $functionality) {
...
+   * @param \Drupal\tripal\TripalDBX\TripalDbxConnection $chado
+   *   A chado database object.
+   * @param string $table
+   *   The chado table name
+   * @param int $pkey_id
+   *   The primary key value for the specified $table
+   *
+   * @return void
+   */
+  public function addFixedValue($chado, $table, $pkey_id) {
```
laceysanderson: This method should be added to the ChadoTestTrait so it is included in the functional and kernel environments rather then duplicated. Assuming I haven't missed a difference?

laceysanderson (changes requested):
Some more inline comments. I'm marking this as requiring changes until we resolve the following:
1. Should we be using the base table type_id for contact, protocol and publication?
2. Is the approach in this PR still correct on chado 1.3.3+ databases where the project.type_id exists? Maybe too many assumptions are being made and we should check based on the type field with a fixed value?

Note: I did not do manual testing at this point.

laceysanderson (changes requested):
*double posted*

dsenalik:
> A comment regarding this statement in the PR description
> 
> > We cannot include the "protocol" content type, however, because there is no protocolprop table in chado (chado issue [GMOD/Chado#148](https://github.com/GMOD/Chado/issues/148))
> 
> The protocol table has a type_id so you can type it that way without needing a prop table. See https://chado.readthedocs.io/en/rtd/_static/schemaspy_integration/tables/protocol.html

The reason for this is that there is already a field using this type_id, same in Tripal 3, so repurposing it as the type that defines the bundle would likely break publish for any site that used this to distinguish types of protocols. You would need a separate bundle for every protocol type.

dsenalik:
> The following content types already had a chado_additional_type_type_default field attached to the type_id in the base table. Yet this PR adds a second chado_additional_type_type_default field which adds a type in the prop table. This may be the most backwards compatible approach but it also seems confusing to new users of Tripal (and to me ;-p)

Same as previous comment, it is because the other existing chado_additional_type_type_default field does not define the bundle type. The new one does. We could imagine a case where we wanted to distinguish between general types of contacts, people vs. institutions? And correct it is backward compatible.

dsenalik:
A summary of some slack discussions: 
To maintain full backward compatibility for migrated Tripal 3 sites, we could create two collections for each current untyped collection? Something like General (type-focused; preferred) and General (untyped; backwards compatible), with some appropriate mechanism to prevent importing both. This is probably already the case, just a tweak of the message may be necessary.

Sites wishing to later opt-in to the typed content types would have available a drush command to add necessary fields (the current update hook in this PR), and add properties to existing content (the code added to ChadoCheckTermsAgainstYaml.php), and this command could be rerun any time new records have been added to chado, perhaps by an importer that doesn't add the type properties.

For tables with an existing type_id that already has a field defined, we will now use that to define the bundle type rather than create an additional property. (contact, protocol, pub).

To be worked out, the type_id added to project with chado 1.3.3.013

tripal_chado/tests/src/Kernel/ChadoTestKernelBase.php
dsenalik: I have moved it there with commit 941645496

tripal_chado/src/Plugin/TripalImporter/TaxonomyImporter.php
dsenalik: Oops.
Fixed with commit a28097c54

PR #2108, opened 2025-02-04
Drush Commands to help manage Chado (set schema as default; register imported schema with Tripal)

### LINKED ISSUE #2104: Drush command for "Add to Tripal" and "Set Default"

Yes

Yes

### Is this feature urgent?

No

### Description

Using automation of some sort (such as Ansible) for almost the entire installation of a Drupal+Tripal site is possible with just a few exceptions. For Tripal, the one exception is the step that is required when using a dumped Chado schema from a Drupal 7 site to ingest into a Drupal 10 site.
Just after the "Prepare Chado" step, one must click two buttons on the page at `admin/tripal/storage/chado/manager`:
 1. Add to Tripal
 2. Set Default

In the solution, these functions would have Drush equivalents so that this step can be automated.

### Your specific use case

Creating a brand new Drupal 10 site using a dumped version of the previous Drupal 7 Chado schema completely automatically is the goal, either scripted or with a automation tool such as Ansible.

### Generally Applicable

Other sites may way to do the same process in an automatic way to roll out their site and subsequent versions of their sites, and generally to make the transition process easier/automatic.

_No response_

Ferrisx4:
# New Feature

## Description

Adds two additional chado commands to help a site admin automate setting up a Tripal site:
 - `drush trp-add-chado --schema-name="chado"` - Adds the specified schema to Tripal.
 - `drush trp-set-default --schema-name="chado"` - Sets the specified schema to be the default.

These drush commands can be used instead of manually performing these two steps on the 'admin/tripal/storage/chado/manager' page.
## Testing?

The official Tripal Docker does this part automatically so you'll need to spin up a new Drupal 10 site. This use-case in particular is for when you have an exported Chado schema that you import into Drupal 10's database.
After Drupal 10 is installed and running, and Tripal and its submodules (Tripal Chado, Tripal DBX, etc.) have been enabled, and the Chado schema dump has been imported into the database, run these two commands with the name of your chado schema.
Upon visiting `admin/tripal/storage/chado/manager`, you should no longer have the ability to "Add to Tripal" or "Set Default" on that particular schema. At this point, Tripal should be configured to work with your specified Chado schema, and you can proceed to continue with setting up Tripal further (adding type collections, adding fields, publishing data).


Ferrisx4:
This is a draft because, while the main functionality has been implemented, I'm thinking we may want to add a check to the setDefault function to make sure the specified schema actually exists. As it stands, this function simply adjusts the `tripal_chado.settings` config via the config.factory service. If the command is run with a typo or the wrong Chado schema, the command will not error out, e.g. `drush trp-set-default --schema-name="cahdo"`.

Conversely, this is primarily meant to be run in an automation, so likely the process is being tested multiple times before launching to production, and any such error would be picked up in testing because further interactions with Chado may produce errors.

laceysanderson:
I agree with adding some error checking :-)

Ferrisx4:
Thanks @laceysanderson, I was able to use TripalDBX's schemaExists() function to perform this check: https://github.com/tripal/tripal/commit/8440250ce26a34f54cf3ab65683f9e7fc5f10ff7
If nothing else jumps out, I'll move this from draft to ready to review.

dsenalik:
I suggest putting the check-if-exists code into a separate function, and then also using it for the ~~set default~~ add to tripal function too.

Ferrisx4:
The `schemaExists()` function lives in the TripalDbx class. Also, the addToTripal function uses the `tripal_chado.integrator` service which also calls the `schemaExists()` function automatically: https://github.com/tripal/tripal/blob/cbd19e6e81a6e76ac7b17ec09df34cb816c24a45/tripal_chado/src/Task/ChadoIntegrator.php#L66
I'm not sure we need a separate function to check if the schema exists since (so far) this is the only place we need to call it explicitly in this class.

dsenalik:
I realized I should just do the actual test. This is what the various errors look like, it looks like the cases are caught with reasonable error messages when the schema is mis-typed, although we have some duplication of messages

Ferrisx4:
In the case of the duplicated error messages, we have this code that "speaks", so to speak, the error message twice seemingly on purpose, one for the log and one for properly throwing an error:
https://github.com/tripal/tripal/blob/d5c69d36ab5ce9d9d2af6956b188f49ebd4333c6/tripal_chado/src/Task/ChadoIntegrator.php#L101-L104

It just so happens that when we're running the commands via drush, both get printed to the terminal. For when we're in a CLI context such as this, I suppose we'll need to not Log it and just throw it as normal if we can determine the context somehow. We'll need to make the same change in some of the other Tasks, such as in the ChadoPreparer: https://github.com/tripal/tripal/blob/d5c69d36ab5ce9d9d2af6956b188f49ebd4333c6/tripal_chado/src/Task/ChadoPreparer.php#L88-L91

Of course if we don't log it then the error may be lost in the stdout/terminal if it gets flushed or closed, and not in the logs. And we can't not throw the error or else the command will fail much more spectacularly a little bit further in the process.

Ferrisx4:
As discussed in the weekly Tripal Cofest, we'll move the duplicated error message to its own issue because it happens in a number of places, including several of the other tasks but also in publishing.
As for the scope of this issue, it's ready for review and merge.

dsenalik (approved):
With the duplicated message problem off in another issue, I hereby approve

PR #2112, opened 2025-02-06
Fix deprecations

### LINKED ISSUE #2045: Docker build can't handle removed files properly
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

We discovered in PR #2037 that if you delete a yaml file on a branch, then build a docker using your tripal repo branch, during the docker build what we do is copy the "branch" on top of "4.x" - so any files in 4.x that are deleted in the branch are skipped and **during the build phase**, they remain to cause hard-to-understand problems. Because in the final built docker, those files are now not there.
We are going to make @laceysanderson figure out what to do 😁 - maybe we can delete the existing tripal before copying?

### Steps to reproduce

.

### Error messages

```shell
.
```

### Screenshots

.

### LINKED ISSUE #2110: Fix deprecations showing up in automated tests.
### Task Description

I've been noticing some deprecations when running tests locally. We should fix them ;-p 

The following are showing up in our automated testing for Drupal 11.1.x and PHP 8.3:

```
1) /var/www/drupal/web/core/lib/Drupal/Core/Test/HttpClientMiddleware/TestHttpClientMiddleware.php:52
Since twig/twig 3.12: Twig Filter "spaceless" is deprecated in /var/www/drupal/web/modules/contrib/tripal/tripal/templates/tripal-entity-content-add-list.html.twig at line 18.

Triggered by:

* Drupal\Tests\tripal\Functional\Entity\TripalContentTest::testTripalEmptyContentTypes
  /var/www/drupal/web/modules/contrib/tripal/tripal/tests/src/Functional/Entity/TripalContentTest.php:62

* Drupal\Tests\tripal\Functional\TripalRoutePermissionsTest::testTripalContentPages
  /var/www/drupal/web/modules/contrib/tripal/tripal/tests/src/Functional/Permissions/TripalRoutePermissionsTest.php:241
```

These are showing up locally on Drupal 11.2.x and PHP 8.3:

```
1) /var/www/drupal/web/core/lib/Drupal/Core/Extension/ModuleHandler.php:570
Autoloading hooks in the file (modules/contrib/tripal/tripal/tripal.views.inc) is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Move the functions in this file to either the .module file or other appropriate location. See https://www.drupal.org/node/3489765
```

### Branch Name

tv4g0-issue2110-fixDeprecations

laceysanderson:
# Tripal 4 Core Dev Task

## Description

Fixes two deprecation messages that have been showing up recently.

The following are showing up in our automated testing for Drupal 11.1.x and PHP 8.3:

```
1) /var/www/drupal/web/core/lib/Drupal/Core/Test/HttpClientMiddleware/TestHttpClientMiddleware.php:52
Since twig/twig 3.12: Twig Filter "spaceless" is deprecated in /var/www/drupal/web/modules/contrib/tripal/tripal/templates/tripal-entity-content-add-list.html.twig at line 18.

Triggered by:

* Drupal\Tests\tripal\Functional\Entity\TripalContentTest::testTripalEmptyContentTypes
  /var/www/drupal/web/modules/contrib/tripal/tripal/tests/src/Functional/Entity/TripalContentTest.php:62

* Drupal\Tests\tripal\Functional\TripalRoutePermissionsTest::testTripalContentPages
  /var/www/drupal/web/modules/contrib/tripal/tripal/tests/src/Functional/Permissions/TripalRoutePermissionsTest.php:241
```

Apparently the way to fix this is to just not apply the filter 😕 From what I read the spaceless filter just ensures you don't have a lot of extra white space. I checked our template to remove any unnecessary white space and then just removed the filter 🤷 

These are showing up locally on Drupal 11.2.x and PHP 8.3:

```
1) /var/www/drupal/web/core/lib/Drupal/Core/Extension/ModuleHandler.php:570
Autoloading hooks in the file (modules/contrib/tripal/tripal/tripal.views.inc) is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Move the functions in this file to either the .module file or other appropriate location. See https://www.drupal.org/node/3489765
```

The depreaction page suggests moving these to the .module file but they are long so I really didn't want to do that. Since it's only the autoloading that is being removed, I decided to just include it in the module file. The code throwing the deprecation seemed to focus way too much on the name of the file so I had to alter that slightly to get the error to stop.

## Testing?

1. Check the Tripal entity Add content list to ensure nothing broke.
2. Check the Tripal jobs listing to ensure the views integration still works.


laceysanderson:
Tests are failing because the testing environment is still detecting the old `tripal/tripal.views.inc` file even though that no longer exists in this PR. Might this be a docker artifact that @dsenalik had mentioned previously? I thought we had fixed that?

dsenalik:
No we did not fix it yet, it is issue #2045. I think that we should fix it in this PR, because otherwise it would be really hard to test. Let me take a look...

dsenalik (approved):
So I did fix this so that it doesn't blow up due to that one file being renamed.
To document for posterity (and prosperity too) the problem was that when building the drupal docker, the `/app` directory persists, and contains a copy of tripal as it was when the drupal docker was built.
For later pull requests, this "old" `/app` still exists. It is overwritten by whatever the new branch has, but it is overwritten, not replaced, so any deleted files don't get deleted.
My changes are:
1. Use a different directory name other than `/app` for the tripal docker i.e. `/tripal_app`, which by itself solves the problem. And for tidiness, it is now removed when building the docker so it will not persist.
2. Also remove the `/app` directory from the drupal docker too, as there is no need to keep it in the docker image, and thus we save a small bit of space.

PR #2117, opened 2025-02-10
Setup a Tripal Content Type Listing View

### LINKED ISSUE #1976: TripalContent listing is not filterable
### Task Description

Currently the TripalContent listing (i.e. Tripal > Content) is sorted with the oldest content first and has no filters at all. This becomes very unusable very quickly. We should look into switching this listing to a view from a ListBuilder instance and add a lot of filters. 

Some features we want:
1. filter by content type
2. change sorting
3. search by title
4. filter by author

Note: Additionally, the term is not showing up on these listings... including that here as it doesn't make sense to fix independently.

### Branch Name

tv4g5-issue1976-tripalContentListingView

reynoldtan:
#

## Description

This Draft PR sets up a view that will display Tripal Content Type listing with filter and sort options.

**This view needs a route/menu path to become accessible in the same way as Tripal Job Listing view and replace the current non-filterable ListBuilder version.**

image

## Testing?

Load the Content Listing View:

View is setup upon Tripal Docker site creation and can be accessed at my tripal site/admin/content/bio_data

image

Manual Test:
1. Setup an account
2. In permission, set the permission to View Tripal Content Listing
3. Sign out drupaladmin and login as the user created in step 1.
4. Access my tripal site/admin/content/bio_data
5. Content Type listing should be accessible to the user.
6. Repeat steps 1-5 except in step 2 set the permission to Administer Tripal Content.
7. Test Anonymous user - this user will get access denied to the page.

### DIFF
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -281,7 +281,7 @@ display:
               past_format: '@interval ago'
               granularity: 2
               refresh: 60
-              description: ''
+              description: 'time difference'
           group_column: value
           group_columns: {  }
           group_rows: true
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -1,14 +1,11 @@
 langcode: en
 status: true
 dependencies:
-  config:
-    - user.role.administrator
-    - user.role.authenticated
   module:
     - tripal
     - user
 _core:
-  default_config_hash: D7URjeFAspIGq91Zp4pyy03_ET6SDH_UMcKVKDG70cg
+  default_config_hash: pW72ou_ok2gGi1X6U3oPn631_rvNA_Q4W_PJHDFVNTg
 id: tripal_content_type_listing
 label: 'Tripal Content Type Listing'
 module: views
@@ -381,11 +378,8 @@ display:
           sort_asc_label: Asc
           sort_desc_label: Desc
       access:
-        type: role
-        options:
-          role:
-            authenticated: authenticated
-            administrator: administrator
+        type: none
+        options: {  }
       cache:
         type: none
         options: {  }
@@ -601,7 +595,6 @@ display:
         - url.query_args
         - 'url.query_args:sort_by'
         - 'url.query_args:sort_order'
-        - user.roles
       tags: {  }
   page_1:
     id: page_1
@@ -621,5 +614,4 @@ display:
         - url.query_args
         - 'url.query_args:sort_by'
         - 'url.query_args:sort_order'
-        - user.roles
       tags: {  }
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -4,6 +4,8 @@ dependencies:
   module:
     - tripal
     - user
+_core:
+  default_config_hash: vXHzj7L1fF2AOk6gkFs-68KBr0y9W0NEDHqsBJQe0xE
 id: tripal_content_type_listing
 label: 'Tripal Content Type Listing'
 module: views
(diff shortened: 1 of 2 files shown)


reynoldtan:
Setting the path to /admin/content/bio_data which is the path to Tripal Content page
image

will override the list builder and use Tripal Content Type Listing View
image

laceysanderson:
The tests are failing as follows:

```
6.880 In UnmetDependenciesException.php line 100:
6.880                                                                                
6.880   Configuration objects provided by <em class="placeholder">tripal</em> have   
6.880   unmet dependencies: <em class="placeholder">views.view.tripal_content_type_  
6.880   listing (tripal.content_type.analysis, tripal.content_type.contact, tripal.  
6.880   content_type.organism, tripal.content_type.project, tripal.content_type.pro  
6.880   tocol, tripal.content_type.pub, tripal.content_type.study)</em>  
```

To fix this you likely just need to remove the filter by content type entirely. This is safe to do since we want all TripalEntity content anyway, regardless of type.

tripal/tripal.module
```diff
@@ -676,3 +677,32 @@ function tripal_gin_content_form_routes() {
 
   return $route_names;
 }
+
+/**
+ * Implements hook_views_pre_view().
+ */
+function tripal_views_pre_view(ViewExecutable $view, $display_id, array &$args) {
+  if ($view->id() == 'tripal_content_type_listing') {
+    // List of permission user must have to access Tripal Content Type listings.
```
laceysanderson: This comment could benefit from more detail.

```suggestion
    // The user must have 1+ of the following permissions to access the
    // Tripal Content Type listing (admin/content/bio_data).
```

tripal/tripal.module
```diff
@@ -676,3 +677,32 @@ function tripal_gin_content_form_routes() {
 
   return $route_names;
 }
+
+/**
+ * Implements hook_views_pre_view().
+ */
+function tripal_views_pre_view(ViewExecutable $view, $display_id, array &$args) {
+  if ($view->id() == 'tripal_content_type_listing') {
```
laceysanderson, line 689: Additionally I would add why we are implementing the hook here.

```suggestion
  // Override permission of Tripal Content Type listing (admin/content/bio_data)
  // in order to support OR when evaluating permissions since views UI does
  // not support this. 
  // WARNING: requires you to have NO Permissions on this view through the UI!
  if ($view->id() == 'tripal_content_type_listing') {
```

tripal/config/install/views.view.tripal_content_type_listing.yml
```diff
@@ -0,0 +1,617 @@
...
+          id: title
+          table: tripal_entity
+          field: title
+          relationship: none
+          group_type: group
+          admin_label: ''
+          entity_type: tripal_entity
+          entity_field: title
+          plugin_id: string
+          operator: '='
```
laceysanderson: Allows the title filter to match any part of the title.

```suggestion
          operator: 'contains'
```

laceysanderson (changes requested):
✅ Code review looks good --just two comment suggestions :-)

### Manual Testing

For my manual testing I created 3 organisms, 1 analysis and imported `tripal_chado/tests/fixtures/gff3_loader/TripalusDatabasicaChr1Genes.gff3` 3 times (once for each organism). I then published both genes and mRNAs. This gave me a total of 379 Tripal Content Pages to test in the content listing view.

✅ View was accessible at the Top Admin > Tripal > Content link as expected.
✅ Able to filter to each type of content properly by selecting it from the drop down.
✅ Sorted by most recent first by default
✅ Changing the sort order worked as expected
✅ Paging worked as expected
❌ Title search did not work as expected. It requires a full match but I expected a partial match (see suggestion).

#### Permissions

✅ These were solid! The screenshots below are as follows: admin user, user with only `administer tripal content`, user with only `access tripal content overview`, user without either permission. I also checked non-logged in user (same as without permission) and a user with both the previously stated permissions (same as `administer tripal content`) and the view performed as expected.

admin_pages_plus
admin_tripal_only
view_listing_only
no_permission

laceysanderson (approved):
Now that changes are applied, this is ready to merge once the tests pass again.

Issue #2116, opened 2025-02-10
Various messages printed twice: importers and Tasks.


Ferrisx4:
### Tripal Version

4

### Drupal Version

10

### PostgreSQL Version

10.3.1 

### PHP Version

8.3.10

### General Description

In several instances, there are duplicated (error) messages printed:
 - When running any of the 'Tasks', such as Installing, Preparing, and Integrating chado, via drush, errors are printed twice (see https://github.com/tripal/tripal/pull/2108#issuecomment-2637711562)
 - When performing certain loads
Most importers display double status messages, for example
![Image](https://github.com/user-attachments/assets/0785879f-8f09-4778-9de7-aea2e6c6408e)

Common code: `tripal/src/TripalImporter/TripalImporterBase.php:148:    $this->logger = \Drupal::service('tripal.logger');`
and the various messages come from `tripal/src/Services/TripalLogger.php`
```
  public function notice($message, $context = [], $options=[]) {
    if ($this->isSuppressed()) return;

    $this->log2Job($message, $context);
// ^This printed the first line in the image above

    if (!array_key_exists('logger', $options) or $options['logger'] !== FALSE) {
      $message_str = $this->messageString($message, $context);
      $this->logger->notice($message_str);
// ^ This printed the second line in the image above
    }

    if (isset($options['drupal_set_message'])) {
      $this->log2Message('notice', $message, $context);
    }
  }
```

### Steps to reproduce

 1. Drush tasks
    - Use the newly minted `drush trp-add-chado --schema-name="chado"` to try to add an already-added Chado schema to Tripal. It will print the error message twice. One is from the logging command, one is from throwing the error exception: https://github.com/tripal/tripal/blob/d5c69d36ab5ce9d9d2af6956b188f49ebd4333c6/tripal_chado/src/Task/ChadoPreparer.php#L88-L91
 2. Loading
  - Run just about any importer

### Error messages

```
```

dsenalik:
**Proposal for importers PR #2118**

Branch `tv4g3_issue2116_importer_logging`

1. In `tripal/src/Services/TripalLogger.php` add a class variable `$global_options` and some setter|getters for it
2. Merge in this global options in each of the functions that has a `$options` passed to it, for example
```
  public function notice($message, $context = [], $options=[]) {
```
[...]

PR #2118, opened 2025-02-11
Only log once for jobs

dsenalik:
# Bug Fix

## Description
This resolves **part** of issue #2116, i.e. double logging for various importers

1. In `tripal/src/Services/TripalLogger.php` add a class variable `$global_options` and some setter|getters for it
2. Merge in this global options in each of the functions that has a `$options` passed to it, for example
```
  public function notice($message, $context = [], $options=[]) {
```
3. Now we can use the `'logger'` key to suppress the logger->notice when there is a job defined. This can be done in the importer base `tripal/src/TripalImporter/TripalImporterBase.php` in `public function setJob($job)`
```
  public function setJob($job) {
    $this->job = $job;
    $this->logger->setJob($job);
    $this->logger->setGlobalOptions('logger', FALSE);  // add this line
  }
```
4. By doing it this way, we don't have to change anything in any of the importers.

## Testing?
(I messed up the branch name so the ❌ is from that)
Run one of the importers, and observe that messages are not duplicated. Probably the easiest to run is the taxonomy importer, enter some random taxid number

### On 4.x
2116-before

### On this branch
2116-after


laceysanderson (approved):
✅ Manual testing using the GFF3 importer showed that the messages are no longer duplicated.
Screenshot 2025-02-18 at 4 10 48 PM
✅ Confirmed that the messages are still logged to the job
image
✅ Code changes look clean and make sense.
✅ New methods are documented.
✅ Order of array_merge means that the global options will be overridden if the same option is passed in to the method --which is what we want 👍

PR #2124, opened 2025-02-19
Create tripal_entity_type canonical path to fix WSOD

dsenalik:
# Bug Fix

## Description
This creates a canonical path for tripal_entity_type. This resolves a WSOD under Drupal 10.x when viewing the Tripal Content Listing view.

This implements the canonical path that was missing and causing the WSOD. It is still just a placeholder, but with at least an informative message.

Thanks to @laceysanderson for telling me what to do 😵‍💫

## Testing?
Testing is easy. 
1. Build on Drupal 10.x
2. Create a tripal entity
3. View the listing without WSOD
4. Try viewing the canonical path directly. This is not a page that will normally be visited. For example, for contact, it is `/admin/structure/bio_data/contact` and there is a simple message displayed
issue2123 1

PR #2125, opened 2025-02-19
Fix escaped HTML tags in tripal content type listing view

### LINKED ISSUE #2122: HTML tags in new tripal content listing view
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

In PR #2117 we finally have a filterable listing of tripal content (yay!)
But one thing missed was formatting for content containing HTML markup in the title

### Steps to reproduce

Create an organism

### Screenshots

Image

dsenalik:
# Bug Fix
~~### Depends on #2124~~ merged
## Description
Two changes to the Tripal Content Listing
1. I added a "No results behavior" as was implemented before, so on a brand new site you will see this
issue2122 0records

2. HTML in the entity title is now rendered instead of being escaped. This primarily affects the organism content type
issue2122 1record

## Testing?
1. Create a new docker
2. Look at the empty content listing, see expected message as shown above
3. Create an organism
4. Title should be rendered in italics

### DIFF
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -1,3 +1,4 @@
+uuid: c99e6739-ccab-4512-9a1c-de5d208686c4
 langcode: en
 status: true
 dependencies:
@@ -5,7 +6,7 @@ dependencies:
     - tripal
     - user
 _core:
-  default_config_hash: pW72ou_ok2gGi1X6U3oPn631_rvNA_Q4W_PJHDFVNTg
+  default_config_hash: Bkx9Iy16THW-fIBoCMoaUpI8M_N301WjHNLpq1ztcBQ
 id: tripal_content_type_listing
 label: 'Tripal Content Type Listing'
 module: views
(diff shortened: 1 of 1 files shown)


tripal/config/install/views.view.tripal_content_type_listing.yml
```diff
@@ -383,7 +384,18 @@ display:
...
+        area_text_custom:
+          id: area_text_custom
+          table: views
+          field: area_text_custom
+          relationship: none
+          group_type: group
+          admin_label: ''
+          plugin_id: text_custom
+          empty: true
+          content: '<p>There are no tripal content entities yet.</p>'
```
laceysanderson: This is actually shown both when there are no Tripal entities yet AND when the filter criteria entered do not match 🙈 As such, I suggest a more generic message like the following. This matches what Drupal does.

```suggestion
          content: '<p>No Tripal content available.</p>'
```

laceysanderson (commented):
This looks good other then the one suggestion.

However, this PR is dependant on PR #2124 and cannot be merged until that one is.

laceysanderson (approved):
Works beautifully!

image

PR #2130, opened 2025-02-21
Allow field property types to know which ID Space plugin should be used if ID Space is missing.

### LINKED ISSUE #2129: Field PropertyTypes where term ID Space no previously created cause error.
### Task Description

When you create a field with a property type whose term ID Space has not previously been added, you receive an error:

```
Exception: Cannot create a StorageProperty object for entity type "tripal_entity", field type "chado_string_type_default", key "record_id" as IdSpace for the property term is not recognized: "SIO:000729"
```

In Tripal core or when fields are attached via the FieldCollection service, we register the ID Space with Tripal. However, this does not happen in the testing environment, nor does it happen if a custom field is only added via the UI.

A previous PR, allowed the ID Space service to register a new ID Space during loading **IF the idspace plugin id is provided**. However, we don't yet pass the idspace plugin id when creating field property types. In chado fields we know what this idspace plugin id is, so we should pass it through!

### Branch Name

tv4g1-issue2129-propertyTypesKnowIDSpace

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

When you create a field with a property type whose term ID Space has not previously been added, on 4.x you receive an error:

```
Exception: Cannot create a StorageProperty object for entity type "tripal_entity", field type "chado_string_type_default", key "record_id" as IdSpace for the property term is not recognized: "SIO:000729"
```

This PR allows chado property types to be aware of the ID Space plugin they should use to lookup an ID Space in the backend storage if it has not yet been registered with Tripal.

This is backwards compatible since passing in the idspace plugin id when creating property types is optional. Additionally, this does not require any changes to ChadoFields since the Chado property types already know they should use the chado_id_space plugin id.

This functionality is most helpful in the automated testing environment. In the past, we have had to curate a list of all the terms used in property types for every field in an automated test in order to pre-register the ID Space with Tripal. This is no longer needed as shown in test ChadoPropertyTypeSimpleCRUDTest::testCreateEntityWithField() where I removed the pre-registering of needed terms and the test still passes.

## Testing?

This is mostly going to affect automated testing. That said, since it affects field, you can simply test creating Tripal content to ensure that the property types are still created properly.

### DIFF
--- a/tripal/src/TripalStorage/BoolStoragePropertyType.php
+++ b/tripal/src/TripalStorage/BoolStoragePropertyType.php
@@ -14,22 +14,21 @@ class BoolStoragePropertyType extends StoragePropertyTypeBase {
    *
    * @param string entityType
    *   The entity type associated with this property type.
-   *
    * @param string fieldType
    *   The field type associated with this property type.
-   *
    * @param string key
    *   The key associated with this property type.
-   *
    * @param string term_id
    *   The controlled vocabulary term asssociated with this property. It must be
    *   in the form of "IdSpace:Accession" (e.g. "rdfs:label" or "OBI:0100026")
-   *
    * @param array storage_settings
    *   An array of settings required for this property by the storage backend.
+   * @param string $idspace_plugin_id
+   *   The plugin_id associated with the term. This is optional but if provided
+   *   allows a missing ID Space to be looked up in the backend storage.
    */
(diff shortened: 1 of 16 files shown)


dsenalik (approved):
- [x] Code changes look intuitive and make sense. I love that it makes it easier to write tests 😍.
- [x] Manual testing revealed issue #2131 which is unrelated to changes here, but I could create entities and view them, and create links between them. I imported a publication and published for good measure, all worked as expected.

PR #2132, opened 2025-02-22
Make the tripal content listing view again link to the content entities

### LINKED ISSUE #2131: Content listing view is no longer clickable
### Tripal Version

4.x

### Drupal Version

10.3, 11.0, 11.1

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

The tripal content view listing is no longer clickable.
I traced it down to this change which was added to allow HTML in the titles
Image

Before this change, the link worked
Image

It turns out that having the "Rewrite results" active, regardless of what is in there, prevents this option from working
Image

Also note that this option does not strip the tags (apparently they are escaped before they get here), but it also prevents the link to content from functioning
Image

### Steps to reproduce

Create any entity and then go to the content listing view

### Solution

I will add a hidden column containing the entity ID and I can use it to create an explicit link `/bio_data/{{ id }}`

dsenalik:
# Bug Fix

## Description

PR #2125 allowed rendering of HTML tags in entity titles, but we did not notice that it broke the link to the corresponding entity.
Apparently using "Rewrite results" is incompatible with also using "Link to the Tripal content"

My solution is to 
1. Add a hidden column with the entity ID
fields_with_id_added
2. Turn off "Link to the Tripal content"
3. Turn on "Output this field as a custom link" using the path `/bio_data/{{ id }}`

## Testing?
1. Create any content entity, but especially organism as it has HTML in the title
2. Go to Tripal Content Type Listing at `/admin/content/bio_data`
3. HTML should be rendered, and you should be able to click and go to that entity
now_clickable

### DIFF
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -6,7 +6,7 @@ dependencies:
     - tripal
     - user
 _core:
-  default_config_hash: Bkx9Iy16THW-fIBoCMoaUpI8M_N301WjHNLpq1ztcBQ
+  default_config_hash: mP9f-XmAQWc4QPnYqqJiE6GnTFEhd9H5qcy-Q3hxzF4
 id: tripal_content_type_listing
 label: 'Tripal Content Type Listing'
 module: views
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Thanks for the fix @dsenalik!

✅ Code changes make sense and look clean.
✅ Manual testing shows that the link is clickable, HTML is still rendered and when I generated multiple pieces of content, the link is generated accurately.
Screenshot 2025-02-24 at 10 18 15 AM

PR #2128, opened 2025-02-20
Create the Chado relationship field

dsenalik:
# New Feature

## Description
This creates the chado relationship field, the Tripal 4 implementation of the Tripal 3 sbo_relationship field. This field uses the various relationship tables in chado to link a record to another record of the same base table, using a controlled vocabulary term to describe the relationship.

This field is a hard one, because both `subject_id` and `object_id` need to be `store_link` because one of them points to the "hosting" record, the other to the "related" record. But it can be either one. Deep in chado storage now, any `store_link` properties are overwritten with the "hosting" record, even if a value is already there. This happens in two different places. This PR will include changes to chado storage and chado records to support more than one `store_link` property, by not overwriting the value if it is already set. (Commit https://github.com/tripal/tripal/pull/2128/commits/25c7a734973a027d44ee8e04461c393a30110846)

There is another difficulty with publish, again because we have two `store_link` properties. In ChadoRecords `selectItems()` these both get added as a condition, and for both the value of the condition is set to the "hosting" record id, and so nothing is found and as a result this field cannot be published. What needed to happen is these two are combined inside an `OR` condition group.

I did not add a field to organism in the yaml config, because the relationship field uses only one table column, and which one will depend on how a site uses organism. Ideally the `abbreviation` field would always be populated and that field could be used, but because it might not be, it is safer to make a site admin add the field.

A point for discussion, for discovery on publication, the column used is `title` if the default token string is present, however in the yaml, I used the `uniquename` column. I did so because that has a not null constraint, where the title can be left blank. It will be unlikely to have publications without titles, so let me know if the yaml should also use `title`.

 - [x] Implement value or rank when present = commit https://github.com/tripal/tripal/pull/2128/commits/ec27a5087d64624c4fb7549407cf237992e92b05
 - [x] Reverse toggle works
 - [x] CV Term synonyms are supported for the autocomplete = commit https://github.com/tripal/tripal/pull/2128/commits/a87ded9322929f196dfdbd267f05cb22cc3727e2
 - [x] Add field to yaml as appropriate for all content types
 - [x] Probably an update hook to add the field would be helpful
 
### Example for possible use in documentation
When a record is related to another record of the same type, you may use the "Chado Relationship" field to describe how the records are related. For example, you may have several publications that are chapters in an encompassing book, or books that are part of a series. Here we are specifying three publications that are part of a larger encompassing work
relationship_1

And once entered, they are displayed in this manner
relationship_2

On the related record, the field looks like this after republishing
relationship_3

[...]

PR #2133, opened 2025-02-24
Automated tests for property fields, widgets and formatters

### LINKED ISSUE #2119: Property Fields are brittle.
### Tripal Version

4.x

### Drupal Version

11.2

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

In our experiment page we have 15 property fields and 34 fields total. We are noticing a lot of brittleness when editing the properties. This issue will document all the ones I'm able to replicate. This is being marked high because we are experiencing data loss and I've had to stop people from using the pages.

Bugs:
- You cannot remove properties via the button or by deleting the content.
- rearranging does not change the rank in chado, just the delta in drupal.

NOTE: I Really think we need to stop deleting things from linking tables every time the page is saved! This feels at the root of some of these issues and means that when things do go wrong, the database is often left in a incongruent/inconsistent state.

### Steps to reproduce

1. Create a fresh docker on 4.x
2. Insert an organism with a large number of properties using `drush sql:cli`.
    ```sql
    INSERT INTO teapot.organism (genus, species) values ('Tripalus','databasica');
    INSERT INTO teapot.organismprop (organism_id, type_id, value) SELECT 1, cvt.cvterm_id, (SELECT string_agg(substring('0123456789bcdfghjkmnpqrstvwxyz', round(random() * 30)::integer, 1), '') FROM generate_series(1, 20)) FROM teapot.cvterm cvt WHERE cvt.cv_id=25 AND cvt.name NOT IN ('genus','species','infraspecies') ORDER BY cvt.cvterm_id ASC LIMIT 10;
    INSERT INTO teapot.organismprop (organism_id, type_id, value, rank) SELECT 1, cvt.cvterm_id, (SELECT string_agg(substring('0123456789bcdfghjkmnpqrstvwxyz', round(random() * 30)::integer, 1), '') FROM generate_series(1, 20)), 1 FROM teapot.cvterm cvt WHERE cvt.cv_id=25 AND cvt.name NOT IN ('genus','species','infraspecies') ORDER BY cvt.cvterm_id ASC LIMIT 10;
    ``` 
3. Find all the fields associated with those properties by going to Top Admin > Tripal > Page Structure > Organism > Manage Fields and clicking on "Check for new Fields" then adding all those that are discovered. There will be 73 new fields ;-)
4. Publish the organism.
    Image 
5. Now edit that organism, click remove beside one of the property values (note down the field) and save the page.
6. Observe that an error was thrown AND that property now has two copies of the non-removed value rather then only a single value (see screenshot). When you query the organismprop table you can confirm that this property is not actually duplicated. When you check the drupal table for that field, you will see it has been duplicated and that both copies have the same prop_id but a different delta.
7. Unpublish the organism and re-publish to clear any problems or start a fresh docker.
8. Edit the newly published organism, pick a field and reorder the values. Note down the new order and click save.
9. Confirm that the order looks correct on the view page but then query the chado organismprop table and notice that the order did not change according to the rank.

### Screenshots

Image

laceysanderson:
# Development Task

*Note: this was originally a bug fix but the bug is not easily reproducible anymore. Instead this has become focused on automated testing for fields to ensure that they are stable.**

## Description

Specifically, this PR will work to create tests around the PROPERTY FIELD specifically as an example for field testing in general:

- [x] Add more automated testing for the property field focusing on making sure that Chado and the Drupal field tables stay in sync. See "CRUD" tests for this.
   - Note: it doesn't check the underlying chado values.
- [x] Add automated testing for the ChadoPropertyWidgetDefault widget (textarea) to ensure that the values are being properly massaged.
   - [x] Expand the scenarios to check the form element defaults in the edit scenario
   - [x] Check the right number of form elements are available for property fields with multiple properties.
   - Note: The tests don't check scenarios that we expect to have errors for

The following scenarios are focused on:
- Ensure that removing properties happens smoothly and Drupal/Chado remain in sync.
- Ensure that the Drupal delta and Chado rank values are always in sync.
- Ensure there is no swapping between property fields
- Ensure you can add properties on create and/or edit

The following helper methods have been added to help test other fields:
- tripal/tests/src/Traits/TripalEntityFieldTestTrait:
  - assertFieldValuesMatch()
  - setupTripalEntityAddForm()
  - setupTripalEntityEditForm()
  - populateTripalEntityFormState()
- tripal_chado/tests/src/Traits/ChadoFieldTestTrait:
  - assertFieldWidgetsMatch()
  - assertArrayContainsElements()
  - setupChadoEntityFieldTestEnvironment()
  - setupChadoEntityFieldSystemUnderTest()

I will work on documenting this process in the TripalDocs repo in PR https://github.com/tripal/tripal_doc/pull/105

### Manual Testing

This PR does touch the property fields themselves so it would be good to manually test them through the interface.

#### ChadoPropertyTypeCRUDTest

This test replaces the old tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldType/ChadoPropertyTypeSimpleCRUDTest.php. It is more efficient and comprehensive plus actually tests the contents of the Drupal field tables.

It tests the following scenarios which are described in tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldType/ChadoPropertyType-TestInfo.yml

 - Simple project with no properties added on create or edit. There are two property fields attached so this checks that we are not forced to use the property fields ;-p
 - Project with no properties added on create but then 2 properties added for each of fields project_prop1 and project_prop2 on edit. This tests multiple properties per field, multiple property fields per content type and creation of properties on edit.
 - project with 2 properties for each of project_prop1 and project_prop2 on create. On edit the properties for project_prop1 are reordered but those for project_prop2 remain the same. This tests that properties can be added on create, that properties can be reordered without being changed and that reordering on one field does not affect the other field.

For each scenario
1. The entity is created with the initial user input and we check this occurs without error.
2. The entity we just created is loaded and we confirm the field values match our expectations. This checks both the values of the field in the entity and in the Drupal field table. It does not actually check chado.
3. Make changes based on user input and check that we can save without error.
4. Load the entity again and check that the changes are reflected both in the entity and in the drupal field tables.

#### ChadoPropertyWidgetFormTest

This widget test is brand new, never before done in Tripal ;-p

It tests the following scenarios which are described in tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldType/ChadoPropertyWidgetForm-TestInfo.yml

 - Simple project with no properties added on create or edit. There are two property fields attached so this checks that we are not forced to use the property fields ;-p
 - Project with no properties added on create but then 2 properties added for each of fields project_prop1 and project_prop2 on edit. This tests multiple properties per field, multiple property fields per content type and creation of properties on edit.
 - project with 2 properties for each of project_prop1 and project_prop2 on create. On edit the properties for project_prop1 are reordered but those for project_prop2 remain the same. This tests that properties can be added on create, that properties can be reordered without being changed and that reordering on one field does not affect the other field.

For each scenario
- CREATE:
  - We can generate the entire TripalEntity create form without error using the field defaults. We also check the field widget form elements match our expectations here.
  - Manipulate the form state based on user input and check that form validation executes as expected (currently expects no errors).
  - If there are no errors from validate then also check that we can submit the form and save it. Then check the resulting field values match our expectations in the same way as for the CRUD test.
- EDIT:
  - We can generate the TripalEntity edit form for the specific entity we just created without error. We check the field widget form elements match our expectations including checking the default values.
  -  Manipulate the form state based on user input and check that form validation executes as expected (currently expects no errors).
  - If there are no errors from validate then also check that we can submit the form and save it. Then check the resulting field values match our expectations in the same way as for the CRUD test.

### Code review

NOTE: This PR also fixes coding standards in the files I touched and adds coding standards checking to the docker devcontainer.

Specific updates to non-test files are:
- tripal_chado/src/Plugin/DataType/ChadoDataType.php
  - getIterator() now needs to be \Traversable (deprecation fix)
  - onchange() method parameters have changed in Drupal core; update to match.
- tripal_chado/src/Plugin/Field/FieldWidget/ChadoPropertyWidgetDefault.php:
  - formElement() support when term_id is already set in the items
  - formElement() consistently use #value for value type elements
  - massageFormValues() don't try to remove value format from empty properties.
  - settingsForm() use DI translation `$this->t()` rather than procedural `t()`.
- tripal_chado/src/TripalField/ChadoWidgetBase.php
  - Don't use `t()` in Exceptions as its not recommended.
  - massagePropertyFormValues() notify the developer if the field_name key is not present in the form values as it is now required.
  - massagePropertyFormValues() Only check each key of $value if it actually exists.


dsenalik:
I want to bring up a performance issue (as well as a nightmare for the relationship field #2128 )
When saving an entity through the UI, it gets saved twice.
https://github.com/tripal/tripal/blob/a8d662dccd8c6624ecc58c21939123c7316282d0/tripal/src/Form/TripalEntityForm.php#L237-L252
I originally inserted this second save, because we need to save the entity in case the title or URL format includes the entity ID value in it, and for a new entity it is not known until the entity is actually saved.
The perfomance issue is that the second save takes approximately 4 seconds for me right now, all just to update the entity title, which is a single field in the `tripal_entity` table. 
The nightmare part is that the "deleting properties/linking records" part actually happens twice as a result.

One option is that I hope that we can justify writing to the DB table directly, as this is 1) a performance killer and 2) a problem for the relationship field as it will have two `store_link` properties, which can get swapped on an edit.

laceysanderson:
NOTE: I also did some checking. The tests started failing in commit 5a07d1d788472421ca65fd73b0812b64663b3398 when I merged in 4.x.

laceysanderson:
NOTE: I just updated the PR description to better represent what this PR is actually now :-) Additionally, I actually have the widget testing working!!! 🎉 

https://github.com/tripal/tripal/pull/2133#issuecomment-2692965363
This was already resolved in a separate PR.

Additionally, there have been other PRs fixing bugs in fields including some of those originally described in the issue.

laceysanderson:
We are now at 100% for all but the settings of the ChadoPropertyWidgetDefault!
Screenshot 2025-06-03 at 10 46 32 PM

Additionally, we have 74.19% coverage of ChadoWidgetBase::massagePropertyFormValues() and 100% coverage of ChadoWidgetBase::saveInitialValues()

Also, this is where we are for the field type:
Screenshot 2025-06-03 at 10 51 08 PM

tripal_chado/src/TripalField/ChadoWidgetBase.php
```diff
@@ -208,14 +214,15 @@ protected function genericSelectElement(string $pkey_column, ?int $default_id,
...
-   * Default settings associated with the generic select element
+   * Default settings associated with the generic select element.
    *
    * @return array
    *   The default settings.
    */
   public static function defaultSelectSettings(): array {
     return [
-      'widget_select_limit' => '',  // Global setting applies when not set
+    // Global setting applies when not set.
```
dsenalik:
```suggestion
      // tripal_entity_type.widget_global_select_limit applies when not set here.
```

codingstandards.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<ruleset name="TripalCultivate">
```
dsenalik: I think I know where this name came from, but should we just name this "Tripal"?
```suggestion
<ruleset name="Tripal">
```

laceysanderson:
Good catch 🙈 Thanks for the review!!!

dsenalik (approved):
> This PR does touch the property fields themselves so it would be good to manually test them through the interface.

So I did import 100 publications (by chance there were 100) because those have lots of properties, then edited one adding more properties and saving, worked as expected.

Two small comments from code review

Great job @laceysanderson 😍 - and the php sniffer addition is much appreciated

PR #2139, opened 2025-02-26
Ensure composer install works on all supported versions.

### LINKED ISSUE #2138: enable Tripal 4.0.0-alpha2 on Drual 11
### Tripal Version

4.0-alpha2

### Drupal Version

11

### PostgreSQL Version

17

### PHP Version

8.3

### General Description

How do I enable Tripal v4.0.0-alpha2 on Drupal 11.  The GitHub site says Drupal 11.1.x is supported but when I tried to install tripal/tripal or tripal/tripal:4.x-dev via composer it squawked with incompatible Drupal Core issue.  I was able to install v4.0.0-alpha2, however I can't enable it.  via drush or web interface.  

When installing v4.0.0-alpha2 on Drupal v10 I can't see Tripal Layout extension.  its only available via v4.x-dev which hasn't been updated since 2023.  

Documentation is extremely confusion and at this point in time I don't know which combination of Drupal and Tripal to use so I can recreate the original site which was created in Drupal 7 and Tripal 3 (headache there is for a different post)

Thank you
Jaysheel

### Steps to reproduce

_No response_

### Error messages

```shell

```

### Screenshots

_No response_

laceysanderson:
# Bug Fix

## Description

As reported in the linked issue, install of Tripal via composer on Drupal 11 is not working as expected. This PR is working to fix this and add a github workflow to ensure it doesn't happen again. It is also adding some documentation to the README to make version compatibility more clear.

Here is the new README. Changes made include
 - Change header to say Drupal 11 instead of 10
 - Change target version to Drupal 11
 - Make it clear we currently recommend using the dev version when making new sites
 - Add table indicating versions needed for previous alpha releases
 - Updated requirements for alpha3

UpdatedReadme

## Testing?

This requires testing the install manually without using Tripaldocker or devcontainer.

Instead, you can use the Drupal docker image which does not have Tripal downloaded/installed and DO NOT mount tripal when creating the container.

The following will pull down a drupal image with D11.1.x, PHP8.3 and Pgsql11, create a container from it, use composer to pull down tripal (this branch) inside the container, install tripal inside the container.

```
docker run --publish=80:80 --name=drupaldocker -tid \
  tripalproject/tripaldocker-drupal:drupal11.1.x-dev-php8.3-pgsql17
docker exec --workdir=/var/www/drupal drupaldocker \
  composer require --dev "tripal/tripal:dev-tv4g0-2138-installIssues" \
  --no-interaction --with-all-dependencies
docker exec drupaldocker /var/www/drupal/vendor/drush/drush/drush \
  pm-install tripal tripal_biodb tripal_chado tripal_layout --yes
```

### DIFF
--- a/.github/workflows/MAIN-composerInstall.yml
+++ b/.github/workflows/MAIN-composerInstall.yml
@@ -37,8 +37,8 @@ jobs:
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
         run: |
           && if [ "$BRANCH_NAME" = "4.x" ]; then
-            docker exec --workdir=/var/www/drupal drupaldocker composer require --dev tripal/tripal:4.x-dev --no-interaction;
+            docker exec --workdir=/var/www/drupal drupaldocker composer require --dev tripal/tripal:4.x-dev --no-interaction
           else
-            docker exec --workdir=/var/www/drupal drupaldocker composer require --dev "tripal/tripal:$BRANCH_NAME" --no-interaction;
-          fi \
-          && docker exec drupaldocker drush pm-install tripal tripal_biodb tripal_chado tripal_layout --yes
+            docker exec --workdir=/var/www/drupal drupaldocker composer require --dev "tripal/tripal:$BRANCH_NAME" --no-interaction
+          fi
+          docker exec drupaldocker drush pm-install tripal tripal_biodb tripal_chado tripal_layout --yes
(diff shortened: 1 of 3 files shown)


laceysanderson:
Comments on the linked issue show a community manual test was successful (see screenshot). That in combination with the fact that this has great test coverage and is a small change, feels like enough validation to me to forgo an official review.

PR #2147, opened 2025-03-04
Streamline url alias and title save and move into the entity save directly

### LINKED ISSUE #2146: TripalEntityForm saves the entity 2X
### Task Description

TripalEntityForm saves the entity 2X currently. Not only is this a performance concern but we have been seeing bugs related to this in the new relationship field.

https://github.com/tripal/tripal/blob/a8d662dccd8c6624ecc58c21939123c7316282d0/tripal/src/Form/TripalEntityForm.php#L237-L252

@dsenalik originally inserted this second save, because we need to save the entity in case the title or URL format includes the entity ID value in it, and for a new entity it is not known until the entity is actually saved.

The performance issue is that the second save takes approximately 4 seconds for me right now, all just to update the entity title, which is a single field in the `tripal_entity` table. The nightmare part is that the "deleting properties/linking records" part actually happens twice as a result.

### Branch Name

tv4g1-issue2146-streamlineUrlAliasTitleSave

laceysanderson:
# Tripal 4 Core Dev Task

## Description
This PR will remove the need for the entity to be saved 2X when created through the UI and ensure that programatically created entities follow the same title/alias procedure as when they are created through the UI. Additionally, a test will be added to ensure the title and URL alias are being handled properly when programatically creating an entity.

Other changes to the entity that were needed in the process:
- setAlias() now throws exceptions when a duplicate alias is being set. This is caught in the form and allows us to provide more information.
- setTitle() and setAlias() are now called in the TripalEntity::postLoad() which is called after the entity is saved but before command is returned to the TripalEntityForm.
- all exceptions in TripalEntity have been updated to be more acceptable if user facing.
- a new method `TripalStorage::markPropertiesForCaching` has been added which uses TripalStorage::isDrupalStoreByFieldNameKey() to determine which properties to cache to Drupal.
- Methods and protected variables have been added to the StoragePropertyTypeBase to allow each field property to keep track of whether it should be cached or not. This is then set by `TripalStorage::markPropertiesForCaching` and used by `TripalFieldItemBase::tripalClear()` to ensure only the right properties are cleared at the beginning of TripalEntiy::preSave().
- The TripalEntityForm was updated to only call save once and not directly set the title or URL.

Specific Tasks:

- [x] Create TestTrait for testing entities with fields.
- [x] Remove assumptions from TripalStorage regarding use of ChadoStorage.
- [x] Create test to check that the title is being generated via format when programatically creating/updating an entity.
- [x] Update Tripal Entity to handle title generation when TripalEntity::save() is used.
- [x] Create test to check that the URL alias is being generated via format when programatically creating/updating an entity.
- [x] Update Tripal Entity to handle URL alias generation when TripalEntity::save() is used.
- [x] Determine why existing tests are now failing.
- [x] Test many scenarios of title/url alias in automated test.
- [x] Update TripalEntityForm to no longer save 2X or set title/url.
- [x] Specifically test/fix read action properties not being available for token replacement with the current approach.
- [x] Allow setTokenValues() values to remain through a save since the method is public. This also allows the testTripalEntityTokenParser() test to remain as it was originally despite not having any fields.

## Background

TripalEntityForm saves the entity 2X currently. Not only is this a performance concern but we have been seeing bugs related to this in the new relationship field.

https://github.com/tripal/tripal/blob/a8d662dccd8c6624ecc58c21939123c7316282d0/tripal/src/Form/TripalEntityForm.php#L237-L252

@dsenalik originally inserted this second save, because we need to save the entity in case the title or URL format includes the entity ID value in it, and for a new entity it is not known until the entity is actually saved.

The performance issue is that the second save takes approximately 4 seconds for me right now, all just to update the entity title, which is a single field in the `tripal_entity` table. The nightmare part is that the "deleting properties/linking records" part actually happens twice as a result.

## Testing?

1. Create a new docker on this branch.
2. Go to Top admin bar > Tripal > Content > "Add Tripal Content"
3. Add any content testing multiple alias combinations and ensuring the form performs as expected.

Example things to test:
- alias is auto-generated based on format
- alias is set in the form
- alias already exists for another page
- alias is not valid

### Automated testing

Changes to existing tests:
- testTripalContentPages() needed a term created since now the tokens are loaded to create the title.
- testTripalEntityCreateAllDefaults() needed the expected title changed to match that determined by the format not that which was passed in.
- test trait tripal/tests/src/Traits/TripalFieldTestTrait.php was renamed to tripal/tests/src/Traits/TripalEntityFieldTestTrait.php to make it clear it handles entity testing too.
- any tests previusly using the TripalFieldTestTrait were updated to use the new test trait instead
- the new test trait provides better helpers for testing and ensures that asserting field values also checks the drupal tables.

In addition to the code coverage on token replacement that already exists, I added the following new tests:

1. Drupal\Tests\tripal\Kernel\Entity\TripalEntityFieldTest

This test focuses on 
- ensuring that the drupal field table records for tripal fields remains in sync with the field value itself
- confirming that Tripal storage is still not dependant on chado

2. Drupal\Tests\tripal\Kernel\Entity\TripalEntityUrlAliasTest

This test focuses on
- confirming that title generation works with tripal fields
   - multi-word token value
   - optional token notation works when tripal field is left empty
   - title is updated when the token value is changed during an update
- confirming that URL generation works with tripal fields
   - multi-word token value
   - optional token notation works when tripal field is left empty
   - url alias is NOT updated when the token value is changed during an update
   - appropriate error is thrown if the alias already exists (both in the case where the alias was user submitted and in the case where it was autogenerated based on format).
   - user submitted url alias can override a format derived alias on update
   - user submitted url alias can override a previously user submitted alias on update.

3. Drupal\Tests\tripal_chado\Kernel\Entity\TripalEntityChadoFieldTest

This test focuses on
- confirming that title + url alias generation works with Chado fields
   - base table field
   - field with read_value action
- confirm that title is updated when tokens are updated.
- confirm that URL alias is not updated when tokens are updated.
- confirm that URL alias can be set through the form.


tripal/tests/src/Kernel/Entity/TripalEntityFieldTest.php
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * This will be used in combination with the data provider. It can't be
+   * accessed directly in the dataProvider due to the way that PHPUnit is
+   * setup.
+   *
+   * @var array
+   *  A list of scenarios where each one has the following keys:
+   *  - label: A human-readable label for the scenario to be used in assert
+   *    messages.
+   *  - descrition: A description of the scenario and what you are wanting to
```
dsenalik:
```suggestion
   *  - description: A description of the scenario and what you are wanting to
```

tripal/tests/src/Kernel/Entity/TripalEntityFieldTest.php
```diff
@@ -0,0 +1,180 @@
...
+    foreach ($current_scenario['edit']['user_input'] as $field_name => $new_values) {
+      $created_entity->set($field_name, $new_values);
+    }
+    // @debug print_r($created_entity->toArray());
+    $status = $created_entity->save();
+    $this->assertEquals(SAVED_UPDATED, $status, "We expected to have updated the existing entity for our " . $current_scenario['label'] . " scenario.");
+
+    // 4. Load the entity we just updated so we can check the values.
+    $updated_entity = TripalEntity::load($created_entity->id());
+    // @debug print_r($updated_entity->toArray());
```
dsenalik, line 169: What is the preference for leaving commented out debugging code like this in for the final PR? 
```suggestion
    $status = $created_entity->save();
    $this->assertEquals(SAVED_UPDATED, $status, "We expected to have updated the existing entity for our " . $current_scenario['label'] . " scenario.");

    // 4. Load the entity we just updated so we can check the values.
    $updated_entity = TripalEntity::load($created_entity->id());
```

tripal/tests/src/Kernel/Entity/TripalEntityUrlAliasTest.php
```diff
@@ -0,0 +1,229 @@
...
+   *
+   * This will be used in combination with the data provider. It can't be
+   * accessed directly in the dataProvider due to the way that PHPUnit is
+   * setup.
+   *
+   * @var array
+   *  A list of scenarios where each one has the following keys:
+   *  - label: A human-readable label for the scenario to be used in assert
+   *    messages.
+   *  - descrition: A description of the scenario and what you are wanting to
```
dsenalik:
```suggestion
   *  - description: A description of the scenario and what you are wanting to
```

tripal_chado/tests/src/Kernel/Entity/TripalEntityChadoFieldTest-TestInfo.yml
```diff
@@ -0,0 +1,151 @@
+system-under-test:
+  bundle:
+    label: Gemstones
+    termIdSpace: NCIT
+    termAccession: C47885
+    id: test_gemstone
+    title_format: 'TEST [TripalEntityType__label] <em>[project_name]</em> Entity #[TripalEntity__entity_id][ aquired by [gemologist]]'
```
dsenalik:
```suggestion
    title_format: 'TEST [TripalEntityType__label] <em>[project_name]</em> Entity #[TripalEntity__entity_id][ acquired by [gemologist]]'
```

tripal_chado/tests/src/Kernel/Entity/TripalEntityChadoFieldTest-TestInfo.yml
```diff
@@ -0,0 +1,151 @@
...
+        project_name:
+        - record_id: 1
+          value: "Amazonite"
+        metaphysical_props:
+        - record_id: 1
+          value: "Amazonite is a stone of peace, truth, harmony, and communication. It gently soothes the nerves, and it’s energies can be felt to the touch. It brings emotional balance, mental clarity and physical health benefits while encouraging spiritual growth & energy work."
+  - label: "Use format with read_value for title + URL + override url on edit"
+    description: "Create a page where every field has one value."
+    create:
+      title: 'TEST Gemstones <em>amazonite</em> Entity #1 aquired by Zhanna Beissekova'
```
dsenalik:
```suggestion
      title: 'TEST Gemstones <em>amazonite</em> Entity #1 acquired by Zhanna Beissekova'
```

tripal_chado/tests/src/Kernel/Entity/TripalEntityChadoFieldTest-TestInfo.yml
```diff
@@ -0,0 +1,151 @@
...
+          value: "soothing, calming, healing from grief"
+        gemologist:
+        - record_id: 1
+          linker_id: 1
+          link: 1
+          contact_id: 2
+          contact_name: 'Zhanna Beissekova'
+          contact_description: 'Enjoys grading and is fascinated by how each stone is different, even within the same species'
+    edit:
+      title: 'TEST Gemstones <em>Amazonite</em> Entity #1 aquired by Zhanna Beissekova'
```
dsenalik:
```suggestion
      title: 'TEST Gemstones <em>Amazonite</em> Entity #1 acquired by Zhanna Beissekova'
```

tripal_chado/tests/src/Kernel/Entity/TripalEntityChadoFieldTest.php
```diff
@@ -0,0 +1,203 @@
...
+   *
+   * This will be used in combination with the data provider. It can't be
+   * accessed directly in the dataProvider due to the way that PHPUnit is
+   * setup.
+   *
+   * @var array
+   *  A list of scenarios where each one has the following keys:
+   *  - label: A human-readable label for the scenario to be used in assert
+   *    messages.
+   *  - descrition: A description of the scenario and what you are wanting to
```
dsenalik:
```suggestion
   *  - description: A description of the scenario and what you are wanting to
```

tripal_chado/tests/src/Kernel/Entity/TripalEntityChadoFieldTest.php
```diff
@@ -0,0 +1,203 @@
...
+    foreach ($current_scenario['edit']['user_input'] as $field_name => $new_values) {
+      $created_entity->set($field_name, $new_values);
+    }
+    // @debug print_r($created_entity->toArray());
+    $status = $created_entity->save();
+    $this->assertEquals(SAVED_UPDATED, $status, "We expected to have updated the existing entity for our " . $current_scenario['label'] . " scenario.");
+
+    // 4. Load the entity we just updated so we can check the values.
+    $updated_entity = TripalEntity::load($created_entity->id());
+    // @debug print_r($updated_entity->toArray());
```
dsenalik, line 192:
```suggestion
    $status = $created_entity->save();
    $this->assertEquals(SAVED_UPDATED, $status, "We expected to have updated the existing entity for our " . $current_scenario['label'] . " scenario.");

    // 4. Load the entity we just updated so we can check the values.
    $updated_entity = TripalEntity::load($created_entity->id());
```

dsenalik (changes requested):
Code review: Very minor spelling fixes, and a question about leaving in debug code.
The new tests are a work of art.
Functional test, sorry I found a few problems:
1. I encountered issue #2148 but it turned out it was unrelated to this PR
2. Changed the URL alias format for contact to `contact/duplicate` (i.e. no tokens) and created two contacts. The second failed, and I believe this is expected and appropriate behavior
duplicate_alias
3. However, I restored the url alias format to the original version, and republished+cache rebuild. The second contact was then published (along with the null contact), but the title is blank. Doing the same, on 4.x I see the title after republish+cache rebuild.
4. I entered an invalid url alias format, a single `/` - this passed form validation, but blows up when you create an entity
invalid_alias
We could fix this by requiring at least one token in validation
Fixing the alias template and republishing doesn't fix it, I continue to get errors

I'll try more stuff...

tripal/tests/src/Kernel/Entity/TripalEntityFieldTest.php
laceysanderson: My preference is to leave it in with the `@debug` tag in automated tests --sparingly but to remove it entirely from the main codebase (i.e. not test-related files). I'm most likely to leave it in if it's debugging for an object where it might not be obvious how to print since you cant print_r() it directly due to size.

dsenalik: Okay sounds good

dsenalik:
I think the title and url format validation needs a slight change.
1. If you enter `abc` for the title format, and save, no warning is shown, just the normal status message "Saved changes Study Tripal Content Type. "
If you then edit again, you finally see the "The Page Title Format for this content type does not contain any tokens...." error message. So if you fix the format and save, even though the format is now valid, you see the error message again after saving. I think maybe we need to change the validation to the more usual, that is, change `$this->messenger()->addError` to `$form_state->setErrorByName`
2. Entering `abc` for the URL alias format never gives any kind of warning or error, it should also check for the case of no tokens

dsenalik:
Commit https://github.com/tripal/tripal/pull/2147/commits/f1e8ea26c2a583347078fffab58b8c6466083bfa does the following
1. Title and URL format validation are done in the usual way with `setErrorByName()`, with the exception that empty strings are allowed. This is because both will be set to valid defaults if they are left blank in the form.
2. Validation checks for invalid title tokens, this is the same as before, but I added similar validation for the URL alias
3. Validation checks for the case where there are no tokens defined for both title and alias
4. The existing function `validateTitleFormat()` now only checks for the one case of when the title is still the generic default, since invalid tokens are now checked in the form validation.
5. I changed the message from an error to a warning, because it will always happen when creating a new content type, and is not really a full-blown error, warning seems more appropriate.
6. This function now takes a parameter, so that the latest user input can be used to check. This prevents the case where you have the generic title format, you change it, and save the "correct" format. Before, you would still get this message shown one last time after saving a correct update. Now that doesn't happen because the value from the form is used.
7. Tested adding new content type through UI (commit https://github.com/tripal/tripal/pull/2147/commits/55c77a8c6e75d3422581e2a085774843501c3455)

laceysanderson:
The most recent commits address your concerns @dsenalik --good catches!

Code coverage of the TripalEntiy is now at 83.5% with full coverage of alias/title methods and postSave() except a few pesky untriggerable exceptions.
Screenshot 2025-03-12 at 4 27 58 PM

There are now warnings on the entity forms if the title/url formats are generic or don't have tokens.
Screenshot 2025-03-12 at 4 29 06 PM

You are now allowed to create content when the title format does not have tokens.
Additionally, you can create content with a duplicate alias. All the values will be saved but the resulting content will not have an alias and you will be warned.
Screenshot 2025-03-12 at 4 38 27 PM

I also added our own validation to catch the invalid url alias you tried since Drupal missed that one 🤪 
Screenshot 2025-03-12 at 4 38 56 PM

This should be ready for another review @dsenalik!

dsenalik:
I would like to see one more change (sorry)
The default title format is a bad thing, so we want to constantly have a message that it should be change.
The default URL alias format, however, generates an alias identical to the ones we use for content types.
For example for study the built-in alias format is `study/[TripalEntity__entity_id]`
If you delete it and save, you will then get the default format of `[TripalEntityType__term_label]/[TripalEntity__entity_id]`
Both of these generate the exact same aliases, e.g. `/study/1`, `/study/2`, ...
So for someone generating a new content type, I think it would be fine to leave it set to the default?
But now if you do you will constantly see this


I propose we remove this warning for just the URL alias
Whaddaya think @laceysanderson ?

tripal/src/Form/TripalEntityForm.php
```diff
@@ -196,33 +194,76 @@ public function form(array $form, FormStateInterface $form_state) {
...
+        'The URL Alias Format for this content type does not contain any tokens. <strong>This will result in no alias being added due to duplicate aliases.</strong>' . $url_message_suffix,
+        [':url' => $url]
+      );
+      $this->messenger()->addError($message);
+      $errors_found = TRUE;
+    } elseif ($url_format == '[TripalEntityType__term_label]/[TripalEntity__entity_id]') {
+      $message = $this->t('The URL Alias Format for this content type is the default generic format. We suggest updating it to include meaningful tokens for more readable URLs.' . $url_message_suffix,
+        [':url' => $url]
+      );
+      $this->messenger()->addWarning($message);
```
dsenalik: I propose removing these lines as described in the previous comment

dsenalik (changes requested):
Everything is working great. I triggered the duplicate alias by setting the analysis format to `duplicate/[analysis_software]` and for contact to `duplicate/[contact_description]` then creating an analysis with software=`same` and then a contact with description=`same` and the warning was triggered but the entity was saved with no alias, and shows in the content listing as `/bio_data/4` which of course works 💯.
I didn't do any real timing, but save seems faster now, as expected 😍
I just want the one change I mentioned in the earlier comment.

tripal/src/Form/TripalEntityForm.php
laceysanderson: Good point <3 I removed the URL alias warning in 07958be2b7bf9d81dea72a9a20a7dee49ad43e1a. Thanks!

dsenalik (approved):
With commit https://github.com/tripal/tripal/pull/2147/commits/07958be2b7bf9d81dea72a9a20a7dee49ad43e1a I now approve, and point out this is overall a super-dooper great improvement!

PR #2151, opened 2025-03-08
Support for calling publication importer directly from another importer

### LINKED ISSUE #2150: Allow other importers to import publications
### Task Description

Right now, to import a publication, the publication importer needs a criteria array defined in a table `public.tripal_pub_library_query`, which is retrieved inside `run()` as follows
```
    $pub_library_manager = \Drupal::service('tripal.pub_library');
    $pub_record = $pub_library_manager->getSearchQuery($query_id);
    $criteria = unserialize($pub_record->criteria);
```

The task is to create a public function where another importer can pass the criteria array directly to import one or more related publications. My use case will be to pass a pubmed ID value and get that publication added.

### Branch Name

tv4g4-issue2150-import-pub-from-criteria-array

### PR #2151

dsenalik:
# New Feature

## Description
The change primarily allows alternatively passing a criteria array to the publication importer instead of the ID value of the search query.

The motivation behind this is to allow another importer to have the ability to import a referenced publication using its pubmed ID number.

High priority just because this is needed to finish upgrading the [tripal_eutils](https://github.com/NAL-i5K/tripal_eutils) module

## Testing?
I have added an automated test for the new functionality, but nothing in core will use this new way of calling the importer.
Manual testing should confirm that the publication importer still works as before. Create a search query and import publications and publish them and make sure nothing is broken.


dsenalik:
There is a problem noted in this PR, after importing a publication, and publishing, all fields except for the title are blank.
This actually is also present on 4.x now.
After a cache rebuild all the fields show up.
So this is happening because I removed `$entity->save()` in PR #2154 to improve performance - the unexpected side effect is that somehow just the title is pre-cached, and none of the fields.
For now commit https://github.com/tripal/tripal/pull/2151/commits/e89ca2b4485ce7933e80b9a147867d29ed8cc483 reverts this one change and will just take performance back to what it was before PR #2154 and the fields are now visible after publish without need for a cache rebuild.

This is a "safe" change because it just restores previous behavior, but maybe a future PR could check if removing save + invalidating cache for the entity takes less time.

dsenalik:
Issue #2169 has been created to implement cache invalidation for republished content, so that any changed titles or field values will be updated without needing a full cache rebuild.

Ferrisx4 (approved):
Functionality seems to be unbothered by recent changes - the importer successfully either
1. Runs the query and reports back that there are no query results (when given bogus query such as "Ferrisx4" in the abstract)
2. Runs the query and adds the results to Chado when there are results found. These results can then successfully be published with no missing or empty fields.

Seems good to go!

PR #2154, opened 2025-03-09
Add bundle token values for publish

### LINKED ISSUE #2148: Missing title tokens on publish
### Tripal Version

4.x

### Drupal Version

11.1

### PostgreSQL Version

17

### PHP Version

8.4

### General Description

Some entity title tokens work entering content through the UI, but do not work when publishing

### Steps to reproduce

1. I entered an organism, worked fine. I changed the title format to `[TripalEntityType__entity_id]:[TripalEntity__entity_id];[TripalEntityType__label]|[TripalEntityType__term_namespace]` and republished. I then entered a second organism. The tokens are not working in the title on publish or republish, they do work on initial entity UI entry.
organism_titles

### Branch

`tv4g1-issue2148-missing-title-tokens`

### LINKED ISSUE #2161: No entity titles are saved when publishing
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

We broke publishing in PR #2147 in that the title is now blank.

This happens because in the Tripal Entity postSave()
https://github.com/tripal/tripal/blob/ea82f7416ea38225adb2b934085d8400e4d8453a/tripal/src/Entity/TripalEntity.php#L948-L953
The existing title gets regenerated, but the necessary token values are not present.
The entity is created with a title in publish, but during the step to add the URL Alias, the entity is loaded, the alias is set, and the entity is saved. That is where postSave() gets called without the necessary tokens, and the title is erased

### Steps to reproduce

Create an unpublished entity, either through an importer or unpublish an existing one. Then publish that content type. The title will be blank.

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
Publish never populated the values for bundle tokens. This PR fixes that.
EXCEPT for the `TripalEntity__entity_id` token. Because of the order of things in publish, the titles are created before the entity is created, so the entity ID is not know for a newly published entity. It will work for republish, however. 

~~Low priority~~ because most people won't have one of these in the title. The only likely case is when creating a new content type, the default title format is `Entity [TripalEntity__entity_id]` - but we have warnings to remind the site admin to change this.

Now this is high priority because this fixes issue #2161 where titles are blank after publish.

The fix for that incorporated here is to save all of the token values from when the titles are generated. These are used later to pass in when the URL Alias is generated.

Performance boost because we no longer need an `$entity->save()` when setting the alias.

## Testing?
Put some of the entity tokens in a title
```
TripalEntityType__entity_id
TripalEntity__entity_id
TripalBundle__bundle_id
TripalEntityType__label
TripalEntityType__term_namespace
TripalEntityType__term_accession
TripalEntityType__term_label
```
and create through the UI. They should all work.
Now try to republish, and rebuild caches after
They should work on this branch.

Also try some field value tokens in a URL alias and confirm they are replaced.


Ferrisx4 (approved):
I tested this on existing (and unpublished) and newly created content. On the current 4.x branch, published entities of content types with tokenized page titles were being published with blank titles. Upon switching to the issue branch, and performing the same task on other entities, shows that they now get their proper page title, as configured by the tokens.

Even the entities with blank titles from the 4.x test had their page title updated without having to republish them.

Ferrisx4 (approved):
Thanks for the extra checks. Still approving!

dsenalik:
We discussed during review that HTML tags are being included in entity titles even though they should be excluded by the existing tag setting - this will be dealt with in Issue #2164 
We also noted that if your title token value is blank, you can get an entity with no title - this will be dealt with in Issue #2165

Issue #1647, opened 2023-09-22
Tripal Content Listing is VERY SLOW when you have lots of content


laceysanderson:
### Tripal Version

4.x

### Drupal Version

9+

### PostgreSQL Version

n/a

### PHP Version

n/a

### General Description

Currently the more Tripal content you add to `admin/content/bio_data` the slower it gets... to the point it will time out. This page is generated by [TripalEntityListBuilder](https://github.com/tripal/tripal/blob/4.x/tripal/src/ListBuilders/TripalEntityListBuilder.php) which extends Drupal's core EntityListBuilder. 

The Drupal inherited methods of this class, load the entire entity for each and every one. I assume this is because Drupal core deals with smaller data and has more performance enhancements / caching for their fields then we do. However, it is completely unnecessary to load all the fields!

We need to override this bit of core Drupal functionality (should just be extending a specific method) to ensure we only load fields who are used as tokens in the title/URL. This will also likely need changes to TripalStorage so that we can indicate during load if we want a partial or full load.

Note: Some of the functionality for figuring out which fields/properties to load is in TripalTokenParser.

### Steps to reproduce

Add 20+ entries and you will start to notice the slow down if not earlier. This will be a really big problem once publishing is ready...

spficklin:
I'm here the night before the Codefest and itching to get this alpha release done.  @laceysanderson I'm not sure where you're at with this one. I'm going to work on it. If you've started let me know.

spficklin:
branch name: `tv4g2-issue1647-slow_content_listing`

laceysanderson:
I haven't started this yet but essentially my plan was to
1. Add a tag to TripalEntity to indicate partial loading
2. When that tag is set the entity::postLoad would know not to invoke the various storage backends. Instead only the Drupal information would be loaded.
3. Override the [load](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityListBuilder.php/function/EntityListBuilder%3A%3Aload/10) in our entity list builder to ensure that we indicate the tag for the entities on loading.

spficklin:
This is similar to what I did.  I'm not sure what you mean by tag but I used a session variable.  PR is forthcoming.

spficklin:
Fixed with PR

PR #2156, opened 2025-03-10
use route to detect if called from tripal content listing view

### LINKED ISSUE #2155: Tripal Content Listing is slow when you have lots of content
### Tripal Version

4.x

### Drupal Version

11.1

### PostgreSQL Version

17

### PHP Version

8.3

### General Description

Something happened recently, and `tripalTypes()` functions are getting called during the tripal content listing.
This is a big performance issue, as if there are lots of entities, pages can timeout.
This used to be issue #1647 but that was fixed with #1736
It has come back 😡 because the entity list builder is no longer used, the listing is now a view.

### Steps to reproduce

1. Have one or more entities entered or published which have properties
2. I put this dpm in `tripal_chado/src/Plugin/Field/FieldType/ChadoPropertyTypeDefault.php`
`dpm("CP1 property tripalTypes called");`
3. View the tripal content listing
4. I observed multiple instances of the dpm message appearing

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
The content listing is slow if there are lots of entities.
This used to be issue https://github.com/tripal/tripal/issues/1647 but that was fixed with https://github.com/tripal/tripal/pull/1736 which used the session variable `tripal_load_listing` to inhibit loading fields when coming from the content listing.
It has come back 😡 because the entity list builder is no longer used, the listing is now a view, so this is no longer getting set.
But another way to do this is by checking the route name. If we come from the route `entity.tripal_entity.collection` then don't load the fields.

## Testing?
Tested easily by observing long delays if there are lots of entities.

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -876,16 +876,14 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
     // to specify which fields to load.  By default the SqlContentEntityStorage
     // storage system we're using will always attach all fields.  But we can
     // control what fields get attached to entities with this postLoad function.
-    // In the TripalEntityListBuilder::load() function we set the
-    // `tripal_load_listing` session variable to TRUE.  If it is TRUE then
-    // we skip this. @todo: in the future if we want to only attach
-    // specific fields we can get more fancy.
-    if (\Drupal::request()->hasSession()) {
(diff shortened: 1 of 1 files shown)


Ferrisx4 (approved):
Very elegant fix, good catch @dsenalik
In my testing, the site went from timing out after 60 seconds to loading the list in 1.2 seconds (4 seconds after the initial cache rebuild).

PR #2126, opened 2025-02-20
Cache all chado values in drupal tables in order to support views

laceysanderson:
# New Feature

### PR #2083 

This PR is a remake of the one linked above because that one had a number of commits made by root/Debian. While @mdondrup tried to fix this using rebase, it was not successful. As such, I cherry picked all his commits and ammended the authorship to point to him instead of root/debian. I also had to fix a few merge conflicts along the way --specifically in the TripalEntitySettings form. The downside of this approach is that now it mentions me as the committer and him as the author, which is not ideal.

## Description

This PR fundamentally affects the way chado data is published.

The logic determining which field data will be copied into Drupal Entity tables from the Chado base tables have been changed by modifying the function `getStoredTypesFilter` in `ChadoStorage.php`. Now, by default, most data is copied.

Fields with storage paths ending in ".residues" are now automatically exempt from redundant storage. While there was already an option to add a storage parameter "drupal_store" to annotate a field to cause it to be copied, this wasn't used anywhere and would have required extensive editing of the Yaml files containing field type collections. 

Note that the behavior is now the opposite of the original behavior, and most field values will now be stored redundantly.
The behavior of not storing entity values during Chado publishing was covered by unit testing in `ChadoTripalPublishTest.php`.
The tests have been adapted to reflect the new behavior.

These changes address #1828 to allow proper filtering and sorting of chado-derived data in Views, according to Solution 1 described [here](https://github.com/tripal/tripal/issues/1828#issuecomment-2541585273): 

> The first approach involves ensuring that while publishing Tripal content, fields like organism_common_name_value are populated with meaningful values rather than left empty. This would allow the Views filter to function correctly without modifying its logic. However, this approach could introduce redundancy and might conflict with the current design, where these fields are likely left empty intentionally.

### Unit testing:

- Publishing of Chado data is covered by phpunit test`testChadoTripalPublishService` in `ChadoTripalPublishTest.php`

### Manual testing:

1. Create an organism and an analysis --fill in all the fields.
2. Go to the GFF3 importer and add `/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/fixtures/gff3_loader/TripalusDatabasicaChr1Genes.gff3` under Server Path. Select the organism and analysis you chose above and enter "chromosome (SO:0000340)" as the default landmark type. Run the submitted Tripal Job.
3. Go to Page Structure and import the Genomic content type collection. Run the submitted job.
[...]

PR #2160, opened 2025-03-14
Improve the performance of TripalEntity::load() by using the Drupal cache

### LINKED ISSUE #2158: Use Drupal cache when loading Tripal content
### Task Description

With PR #2126 all the chado values accessible to ChadoStorage are being cached in the Drupal field tables during create/update. As was discussed in the original issue (see quote below), this provides an opportunity for an additional performance boost when loading content.

[Excerpt from "Basic Search Functionality --views integration missing sorting and filtering" discussing the solution we chose](https://github.com/tripal/tripal/issues/1828#issuecomment-2552267116)
> (This) approach involves ensuring that during the publishing of Tripal content, fields like organism_common_name_value are populated with meaningful values rather than being left empty. This would allow the Views filter to function correctly without modifying its logic. However, this approach could introduce redundancy and might conflict with the current design, where these fields are likely left empty intentionally.
> 
> The current design leaves those columns blank to reduce redundancy. That said, the current design should not assume that they are empty. As such, this solution should not break/conflict with the current design :-) Or more specifically, we can simply populate all columns without breaking the current Tripal Content edit.
> 
> This solution would also provide a significant performance boost if paired with an upgrade to the current design. Specifically, if we updated our current design to use the Drupal tables as a cache then it would allow us to take advantage of the Drupal caching system more effectively and reduce the number of longer Chado queries performed. We could do this by adding logic to either [TripalStorage::getStoredValues](https://github.com/tripal/tripal/blob/4.x/tripal/src/TripalStorage/TripalStorageBase.php#L182), make a base TripalStorage::loadValues or [ChadoStorage::loadValues](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php#L265) methods. In the case of updating ChadoStorage::loadValues the logic would
> 
> - check a state variable indicating whether the cache is still valid or not.
> - if the cache is valid then returns the values loaded by Drupal as they are.
> - if the cache is not valid then does what it currently does which is to query Chado for the information based on the field tripalTypes property types.

This issue provides the jumping off point for using these cached values during loading of the TripalEntity.

### Branch Name

tv4g9-issue2158-improveTripalEntityLoad

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR allows the TripalEntity::postLoad() and ChadoStorage::loadValues() to ingnore loading TripalFields where all properties are cached in Drupal. This provides an insane speed boost 😍 especially when loading many entities like in a search or view.

How does it work?
1. A previous PR (#2126) ensured that by default almost all TripalFields cache their chado values in the drupal field table. This was done for views integration but will also be used here. As such, during save:
    1. The values submitted by the fields are inserted into Chado
    2. These values are then loaded from Chado into the Drupal fields using ChadoStorage::loadValues with $ignore_cached_fields set to FALSE.
    3. Drupal then saves the loaded values into the drupal field tables.
2. Drupal then uses the SqlContentEntityStorage to load all the field values that were cached in Drupal during TripalEntity::load().
3. During TripalEntity::postLoad() we used to look up the values in chado for all fields and replace the ones set by Drupal but now we only do that for fields which contain some properties that are not cached (i.e. ChadoSequenceTypeDefault). We simply leave any fields are fully cached in Drupal alone. Additionally, we don't even setup the TripalStorage object for fields that are fully cached and all entities with only fully cached fields are essentially skipped during the postLoad().

## Automated testing

This change is fully tested by the tests added in PR #2147 which confirms that all the values are being saved to chado and drupal as expected and being loaded as expected.

## Manual testing

1. Create a new docker on this branch.
2. Go to Top admin bar > Tripal > Page Structure > Import type collections and import the genomic collection.
3. Go to Top admin bar > Tripal > Content > "Add Tripal Content"
4. Add a few contacts, then add an project making sure to relate the contacts you created.
5. Confirm that creating and editing the project and contacts works as expected by looking at the underlying chado and drupal field tables. This checks the cases where all fields are 100% cached and checks a variety of different fields including linker fields. This should all function the same as on 4.x but with a small speed increase. The main speed increase is in loading.
6. Next create an organism (fill in the infraspecific type) and a gene (fill in the sequence residues).
7. Confirm that creating and editing works as expected -especially the sequence residues! This tests the case where a field is not 100% cached since the residues should not be saved in the drupal field table but should still appear on the page during view and prefilled in the form on edit. Also confirm the computed style fields for the checksum and length also work.
8. Finally lets test the publish. For this we want to publish the contacts. This should populate the project link which was added to chado by the project. Run the job and then view the contact fields to ensure the project is now listed and is a link.
9. For extra completeness, feel free to import a GFF3 and fasta for the genes and publish them. This job should complete without error and the sequence should be viewable through the gene pages.

### DIFF
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -278,7 +278,7 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
 
     // We can actually skip this entire process if all the fields are 100% cached.
     // The cached_fields indicates for each field (key) if it is fully cached (TRUE)
-    // or not (FALSE). Thus is not fields in this variable are false then
+    // or not (FALSE). Thus if no fields in this variable are false then
     // all the fields are cached!
     if ($ignore_cached_fields AND (array_search(FALSE, $this->cached_fields, TRUE) === FALSE)) {
       return TRUE;
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -270,7 +270,7 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
   /**
    * @{inheritdoc}
    */
-  public function loadValues(&$values) : bool {
+  public function loadValues(&$values, bool $ignore_cached_fields = TRUE) : bool {
 
     // Setup field debugging.
     $this->field_debugger->printHeader('Load');
@@ -280,16 +280,21 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
     // The cached_fields indicates for each field (key) if it is fully cached (TRUE)
     // or not (FALSE). Thus is not fields in this variable are false then
     // all the fields are cached!
-    if (array_search(FALSE, $this->cached_fields, TRUE) === FALSE) {
+    if ($ignore_cached_fields AND (array_search(FALSE, $this->cached_fields, TRUE) === FALSE)) {
       return TRUE;
     }
 
     // Build the ChadoRecords object.
     $this->records = new ChadoRecords($this->field_debugger, $this->logger, $this->connection);
 
-    // Parameters indicate this is not a find action and that we should not add
-    // any properties for fields which are 100% cached in drupal to ChadoRecords.
-    $this->buildChadoRecords($values, FALSE, TRUE);
+    if ($ignore_cached_fields) {
+      // Parameters indicate this is not a find action and that we should not add
+      // any properties for fields which are 100% cached in drupal to ChadoRecords.
+      $this->buildChadoRecords($values, FALSE, TRUE);
+    }
+    else {
+      $this->buildChadoRecords($values);
+    }
 
     $transaction_chado = $this->connection->startTransaction();
     try {
--- a/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
+++ b/tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
@@ -276,9 +276,20 @@ class ChadoStorage extends TripalStorageBase implements TripalStorageInterface {
     $this->field_debugger->printHeader('Load');
     $this->field_debugger->summarizeChadoStorage($this, 'At the beginning of ChadoStorage::loadValues');
 
+    // We can actually skip this entire process if all the fields are 100% cached.
+    // The cached_fields indicates for each field (key) if it is fully cached (TRUE)
+    // or not (FALSE). Thus is not fields in this variable are false then
+    // all the fields are cached!
+    if (array_search(FALSE, $this->cached_fields, TRUE) === FALSE) {
+      return TRUE;
+    }
+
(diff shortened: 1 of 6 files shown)


tripal_chado/src/Plugin/TripalStorage/ChadoStorage.php
```diff
@@ -270,15 +270,31 @@ public function updateValues(&$values) : bool {
...
-  public function loadValues(&$values) : bool {
+  public function loadValues(&$values, bool $ignore_cached_fields = TRUE) : bool {
 
     // Setup field debugging.
     $this->field_debugger->printHeader('Load');
     $this->field_debugger->summarizeChadoStorage($this, 'At the beginning of ChadoStorage::loadValues');
 
+    // We can actually skip this entire process if all the fields are 100% cached.
+    // The cached_fields indicates for each field (key) if it is fully cached (TRUE)
+    // or not (FALSE). Thus is not fields in this variable are false then
```
dsenalik:
```suggestion
    // or not (FALSE). Thus if no fields in this variable are false then
```

dsenalik:
I imported a gff file and when publishing, neither gene nor mRNA had a title 🙈. Edit and save restored the missing title

laceysanderson:
> I imported a gff file and when publishing, neither gene nor mRNA had a title 🙈. Edit and save restored the missing title

This appears to be a problem on 4.x as well 🙈 I suspect we missed this in #2147...

dsenalik:
Missing titles will be fixed by PR #2154

dsenalik:
It is quite possible this is not related to this PR, but
I imported a Tripal 3 chado instance
Loaded all type collections
Added new fields to organism type
Published organism
Saw this error
`ERROR: The organism_cvterm.record_id property type uses the store_id action type but is not associated with the base table of the field. Either change the base_table of this field or use store_pkey instead.  featuremap != organism`
Never mind, confirmed on 4.x
Issue #2177

dsenalik:
Another rabbit hole while testing this was #2179
But I want to mention that I am noting that on this branch, publish memory usage is LOWER than on 4.x 😃

Ferrisx4 (commented):
I tested this and got the results as expected. Particularly, the sequence in fact is available on view and edit displays, showing that even non-cached fields are being populated properly.

Also, I did try to import a GFF and Fasta but borked my site because it was a puny Docker and I was loading tree data and it ran out of memory fairly quickly. Trying again with more sensible data but for now I'll approve it (without officially approving it), since I do want to try imported data as well.

The main goal of this issue and PR was a speedup through cache usage and that has been witnessed. Loading the Tripal content list is very speedy even with 20k+ features loaded (see above).

Ferrisx4 (approved):
Used more sensible data for testing gff and fasta imported data, didn't run out of memory, and the sequences are visible as expected :)

Did notice a quirk in the GFF Loader - the Default Landmark Type field is marked as required but the help text notes that it is only required if the landmark features are not already in the database. Unrelated to this PR though so I'm ready to approve.

dsenalik (approved):
I only found those unrelated issues, but I did not encounter any bugs while testing.
Nor do I see anything from code review.
Awesome PR!

laceysanderson:
Awesome! Thanks both of you! 😍 🎉

Issue #2121, opened 2025-02-18
Derived feature sequences: gff loader/page display cannot automatically generate gff sequences


nasongyan:
### Discussion or Question

Tripal Version:3
Drupal Version:7
PHP Version:7.1.27
PostgreSQL Version:9.6.12
GFF file(test):
`##gff-version 3
Maban_01        Gmove   gene    7740    25393   3397.02 -       .       ID=Maban_01g000010;Name=Maban_01g000010;Note=SEC7 domain-containing protein
Maban_01        Gmove   mRNA    7740    25393   3397.02 -       .       ID=Maban_01g000010.1;Parent=Maban_01g000010;Dbxref=TrEMBL:A0A804HNK2,Gene3D:1.10.1000.11,Gene3D:1.25.10.10,KEGG:K18442,InterPro:IPR016024,InterPro:IPR023394,InterPro:IPR000904,InterPro:IPR032629,InterPro:IPR032691,InterPro:IPR035999,InterPro:IPR015403,InterPro:IPR011989,SUPERFAMILY:SSF48371,Pfam:PF01369,Pfam:PF16213,Pfam:PF12783,SUPERFAMILY:SSF48425,Pfam:PF09324;Name=Maban_01g000010.1;Note=SEC7 domain-containing protein;Ontology_term=GO:0005085,GO:0032012
Maban_01        Gmove   exon    7740    8533    .       -       1       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     7922    8533    .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     14657   15324   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     15421   15769   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     15932   16253   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     16365   16870   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     18487   18760   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     18839   19131   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     19208   19630   .       -       .       Parent=Maban_01g000010.1
Maban_01        Gmove   CDS     19752   20082   .       -       .       Parent=Maban_01g000010.1
[...]

Issue #1836, opened 2024-03-29
Sequence Field infers sequence based on location


laceysanderson:
### This feature attempts to solve an existing problem

None

### I am open to developing or collaborating on an extension module if this is not a good fit for Tripal Core

None

### Is this feature urgent?

None

### Description

In previous versions of Tripal, the sequence field would look-up the sequence for feature-based content with featureloc records if the srcfeature had associated sequence.

This is part of what @mdondrup is describing in issue #1823 and in this excerpt from a recent PR conversation:
> I think importing of genes sequences separately from the GFF fasta file is a good work-around to populate the sequence field. however, it would require quite some pre-processing on behalf of the user. I was hoping that either the GFF importer or the UI itself would extract the gene sequence from alignment, ideally as it was the case in Tripal 2 (I cannot say how this worked in Tripal 3 because I never migrated).
Example: https://licebase.org/feature/Tigriopus/kingsejongensis/gene/maker-scaffold693_size110418-snap-gene-0.25?pane=sequences
> 
> The sequence of the gene and related records is extracted from the coodinates imported in the GFF file.

This functionality was not ported to Tripal 4 because I don't think it was ported to Tripal 3 🙈 That is likely because it is only trivial if there is a single location...  once you add in multiple locations for different assemblies it becomes a bit of a mess. That said, my team (KnowPulse) did attempt this in Tripal 3 with our sequence field alternative that also shows any variants. You can see an example of this field here: [LcC00003p567](https://knowpulse.usask.ca/bio_data/917920) which is located on the original contigs [LcRBContig00003:567](https://knowpulse.usask.ca/bio_data/917920#) and then the next assembly [LcChr3:136769246](https://knowpulse.usask.ca/bio_data/917920#). The code that does this is here: https://github.com/UofS-Pulse-Binfo/nd_genotypes/blob/7.x-3.x/includes/TripalFields/local__sequence_with_variants/local__sequence_with_variants.inc#L187-L294.

Note: this code makes the assumption that data is stored at is is in KP which is [described here under method 2](https://nd-genotypes.readthedocs.io/en/latest/data_storage.html) for pulling variants. Also, it does not support records which are aligned to the negative strand... this is due to difficulties converting variants when the strand may not be known.

### Your specific use case

See Issue #1823 for @mdondrup's use case

### Generally Applicable

The variant part of the example field may not be universally applicable due to differences in storing genotypic data.

[...]

PR #2162, opened 2025-03-16
Implement sequence fields that display sequence from genomic coordinates

mdondrup:
#New Feature

### Issue #1836 #2121
(possibly more)

## Description

This PR implements sequence fields that compute a sequence for genomic features derived from their genomic location (featureloc in Chado) against their landmark sequence. This new feature eliminates the need to extract feature residues upon import for any genomic feature located on a landmark (chromosome, scaffold, contig).
Features (most already supported by the helper code):

 - extract gene sequences from coordinates and reverse complements if located on reverse strand
 - splices transcripts based on exon information
 - can translate polypeptide sequences based on CDS fragments
 - supports and preserves masking information (capitalization of sequence)
 - supports IUPAC ambiguity codes in revcomp and translation
 - supports all genetic codes including mitchondrial, codes can be inferred from Chado
 - supports multiple alignments of one feature to multiple landmarks (e.g. genome builds)

The interface is still relatively basic and supports only genes. Support for different sequence types and displaying of related and sub-features including coloring, interactive highlighting, graphical display of gene models, are work in progress.

Handling and displaying sequence data is essential functionality for a genomics database. This PR attempts to lift functionality towards the level that was implemented in Tripal 2.

A more comprehensive discussion is here: https://github.com/tripal/tripal/issues/2121#issuecomment-2681425901

## Testing?

Best to start with an empty container or fresh install.
check out mdondrup:tripal:tv4g1-issue1836-sequence-field-infers-sequence-from-location

 - If running in the tripal container or if the tripal module has been activated on a different branch (even if the genomic fields collection has never been loaded):
run `drush  config:delete tripal.tripalfield_collection.genomic_chado; drush cr` to reload the field collection from yaml
 - Delete the Gene field from tripal data collections if present
 - import Genomic fields collection
 - Create organism (S. cerevisiae) and Assembly (S288c_R64)
 - Import PAU8 GFF file 
 - Publish Tripal gene content
 - Visit PAU8 gene
Genomic sequence is displayed and is identical to the residue sequence

Screenshot 2025-03-15 at 10 40 08 PM

[PAU8-withgeneseq.txt](https://github.com/user-attachments/files/19270394/PAU8-withgeneseq.txt)

## License
Part of the helper code tripal/tripal_chado/src/Services/GenomicFeatureHelper.php is based on BioPHP by Joseba Bikandi under GPL v2 and  can be used in Tripal under the same license:

// author    Joseba Bikandi
// license   GNU GPL v2
// source code available at  biophp.org

 - https://www.biophp.org/minitools/sequence_manipulation_and_data/
 - https://www.biophp.org/minitools/dna_to_protein


dsenalik:
This looks great so far, one thought I have at this early stage is that instead of `ChadoAlignedSequenceTypeDefault` this field be called `ChadoDerivedSequenceTypeDefault`. To me an alignment implies two sequences somehow showing their similarity, and this is more of deriving (either by lookup or through codon translation) one sequence from another. Opinions requested.

And I wanted to help you add the `discover()` function but discovered it is currently difficult because we have never done one for a base table column as with the original sequence field - see Issue #2163

mdondrup:
Indeed, the term "alignment" in the genomic sequence context is misleading. I think I was confused by the way this field was labeled in the interface previously. I have been looking at how other databases label this part, and they simply call it "genomic sequence" or just "sequence". Given the new functionality, a gene having separate residues different from the would be the absolute exception, because that would either mean the underlying landmark sequence is wrong or the gene model is wrong and those should be corrected. Of course it might still happen in very rare cases when a gene sequence is re-sequenced, but then again, this sequence should rather be saved as a new unplaced gene. As an example of how the gene display could look, I have had my eyes on SGD, see https://www.yeastgenome.org/locus/S000028510#sequence (selected a gene with introns).

Regarding the discover() functionality, I am not sure I completely follow. The derived sequence field doesn't have a direct source column in any chado table. (I have used "dummy_derived_sequence" because the field definition syntax requires it).

dsenalik:
> Regarding the discover() functionality, I am not sure I completely follow. The derived sequence field doesn't have a direct source column in any chado table. (I have used "dummy_derived_sequence" because the field definition syntax requires it).

Well I am thinking the "dummy" column would also work for discover, but first I have to get discover to work for the "regular" sequence field

mdondrup:
Now the derived sequence type has its own formatted and a dummy widget.

image


------------------------------------------------------------------------------------------------
Same sequence in SGD:
image

Todo: 
- add configuration dropdown
- add download buttons

PR #2166, opened 2025-03-17
Implement field discovery for fields using a base table column

### LINKED ISSUE #2163: Field discovery for base table columns
### Task Description

We need to implement a parent helper function for discovery of chado fields that handle the value of a base table column, for example the `residues` column of a feature.
Currently we have `discoverDirect()` for fields that use a foreign key in the specified base table, and `discoverLinked()` for fields that link to the base table through a linking table.
We could have a new function `discoverBase()` to help handling this.

### Branch Name

tv4g1-issue2163-field-discover-base

### PR #2166

dsenalik:
# New Feature

## Description
Add support for discovery of fields that obtain their value from a column in the base table of the content type.
Examples are the `residues` column of the feature table, `seqlen`, and `md5checksum`.

## Testing?
Basic automated tests for discovery are provided for these three fields.
Manual testing can be done by removing the sequence, sequence length, and checksum fields from the Gene content type, then discovering and adding these fields back.
Alternatively create a new content type from scratch using the `feature` table


laceysanderson (approved):
✅ Code looks solid 😍 
✅ Manual testing shows the sequence length, checksum, residues fields are found after they are deleted, they are created with the same term they had originally, and adding content after still works as expected.
Screenshot 2025-03-24 at 1 53 12 PM
✅  Manual testing shows the unit field is found on the physical map after it is are deleted, it is are created with the same term they had originally, and adding content after still works as expected.
Screenshot 2025-03-24 at 2 04 02 PM

PR #2171, opened 2025-03-20
Clear cache for republished entities

### LINKED ISSUE #2169: Clear cache for any republished entities
### Task Description

When republishing chado content, a field value may be changed if the underlying chado record was updated, or if some field configuration was changed.
We should clear the Drupal cache for any updated entities after publishing them, otherwise a site admin needs to remember to clear caches.
A simple way to do this would be to pass an array of entity tags, for example
`\Drupal::service('cache_tags.invalidator')->invalidateTags(['tripal_entity:' . $entity_id]);`

An optimization that was attempted in PR #2154 by removing `$entity->save()` as it is not needed to update the URL alias.
However, this was added back in PR #2151 because without the save, no fields appeared on newly published entities until after a full cache rebuild.
Experiments are showing that a cache invalidation is not a functional substitute for this save, but in this issue we can explore what is going on to prevent this optimization from working.

### Branch Name

tv4g1-issue2169-clear-cache-after-publish

### PR #2171

dsenalik:
# New Feature

## Description

When republishing content, the title may be updated if the title format was changed, fields may have been added to the content type, or field values may be updated if values in chado were updated.
This PR adds cache invalidation for new and for republished entities, because otherwise you need a full cache rebuild after publishing to see the changes for republished content.

I was able to also remove the `$entity->save()` by replacing it with a cache clear for the new entities.

My one concern is that this `\Drupal::service('cache_tags.invalidator')->invalidateTags(['rendered']);` will, I assume, invalidate all render tags, but it will still be faster than a full cache rebuild.

### Timings

I created a publication importer for all pubs with "carrot" in the title in the last 1000 days, which loaded ~~233~~ 10 publications (see bug in #2172 why only ten)

On 4.x:
Cumulative time doing `$entity->save()` = 0.059466 second

On this branch:
Initial publish clear cache new entities: 0.000350 second
Republish, clear cache existing entities: 0.011120 (no content had been viewed)
If I view ten publications, and republish a second time, the republish time actually decreased to 0.003930

None of these times are particulary long, but removing save does seem to be beneficial.

## Testing?

Example of testing on publication.
1. Create a publication importer and import at least one publication.
2. Publish - verify that fields are present and not just the title.
3. Now edit the publication title format, add another field or just some fixed text
4. Check for new fields, hopefully there will be several
5. And change a value in chado, e.g. `update pub set publisher='tripal' where pub_id=2;`
6. Publish with republish set
7. View and confirm expected changes appear without needing a cache rebuild.

### DIFF
--- a/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
+++ b/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
@@ -666,6 +666,16 @@ class ChadoPublish extends TripalBackendPublishBase {
       }
     }
 
+    // If we are going to republish, then clear the cache for all of the
+    // existing entities because we may change titles or field values.
+    if ($this->republish) {
+      $tags = [];
+      foreach ($this->existing_published_entities as $entity_id) {
+        $tags[] = 'tripal_entity:' . $entity_id;
+      }
+      \Drupal::service('cache_tags.invalidator')->invalidateTags($tags);
+    }
+
     return $titles;
   }
 
--- a/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
+++ b/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
@@ -671,9 +671,12 @@ class ChadoPublish extends TripalBackendPublishBase {
     if ($this->republish) {
       $tags = [];
       foreach ($this->existing_published_entities as $entity_id) {
-        $tags[] = 'tripal_entity:' . $entity_id;
+        $tags[] = 'values:tripal_entity:' . $entity_id;
       }
-      \Drupal::service('cache_tags.invalidator')->invalidateTags($tags);
+      if ($tags) {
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code looks clean and makes sense to me :-)
✅ Manual testing produced the expected results! Title updated on republish, fields updated on republish, title/url alias formed perfectly on first publish

Manual Testing

- Imported genes from GFF3 on one organism and published; checked pages and titles ✅ 
- Made a bunch of changes
    - Added a new property to all the genes at the database level, checked for new field to have it added.
    - Updated the name of one of the genes (name is used in the title)
    - Changed the URL alias to have the gene name in it too.
    - Changed the infraspecific name of the existing organism
- Imported the same gene set using the GFF3 importer but attached to a second organism.
- Run publish genes again with republish toggled. Confirm that 
    - the new genes published right with the correct title, url alias and fields
    - the feature name is updated and the title when it is changed in chado
    - the new field is published and the field is populated in existing genes
    - the organism name is updated in all genes.

PR #2174, opened 2025-03-23
Import pubs in batches

### LINKED ISSUE #2172: Publication importer only imports the first ten publications
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### Branch

`tv4g4-issue2172-import-pubs-in-batches`

### General Description

The publication importer is only importing the first ten publications

This section of code is at fault, there is no loop
https://github.com/tripal/tripal/blob/6fbc8afc2d04170f1fca3378e23aaaa8fe62e761/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php#L96-L123

In tripal 3 here there is a loop, retrieving pages of 100 publications each
https://github.com/tripal/tripal/blob/f651120829d8229805dfa3a63428ff92c7f37cda/tripal_chado/api/modules/tripal_chado.pub.api.inc#L532-L557

### Steps to reproduce

Try a publication query that is general enough to return more than ten publications. Testing the query will show the correct number of publications.

### Error messages

```shell
Step 3 of 9: Retrieving publication data from remote database ...
  🗸 Found publications: 10
```

### Screenshots

.

### LINKED ISSUE #2173: Reference to chado in tripal part of publication importer
### Task Description

This function
https://github.com/tripal/tripal/blob/6fbc8afc2d04170f1fca3378e23aaaa8fe62e761/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php#L919-L920
uses Chado but this function is in the `tripal` module rather than the `tripal_chado` module. 
The chado CV term lookup should be deferred and moved to 
./tripal_chado/src/Plugin/TripalImporter/PubSearchQueryImporter.php
and in the process use a chado buddy for better performance

### Branch Name

~~tv4g4-issue2173-pub-importer-move-to-tripal-chado~~
This will be fixed in #2174 as it is a relatively small change

dsenalik:
# Bug Fix + New Feature

## Description
The bug fixed here is that the publication importer only imports the first ten publications. This is the justification for "High" priority.
The new feature is to load publications in batches of 100 to match the batch size used for remote queries, and to minimize memory usage.
Added injection of tripal logger service.
Fix issue #2173 as it was a very small change and too hard for me not to just fix that too.

## Testing?
1. Create a query that will match just a few publications and import. e.g. "tripal" in title or in abstract
```
Initializing publication importer
Step 1 of 7: Retrieving publication data from remote database tripal_pub_library_PMID
  🗸 Importing 23 publications
Step 2 of 7: Check for already imported publications
  🗸 New publications to be inserted: 22
Step 3 of 7: Insert new publications
  🗸 Inserted: 22
Step 4 of 7: Insert database crossreferences
  🗸 Inserted: 22
Step 5 of 7: Link publications to database crossreferences
  🗸 Inserted: 22
Step 6 of 7: Insert publication properties
  🗸 Inserted: 520
Step 7 of 7: Insert authors
  🗸 Inserted: 230
Done.
``` 
2. Create a query that will import more than 100 publications and import. A prefix is added to the log messages when there is more than one batch
```
Step 1 of 7: Retrieving publication data from remote database tripal_pub_library_PMID
  🗸 277 publications will be imported in 3 groups of 100 publications each
  🗸 Importing 100 publications
Group 1 of 3, Step 2 of 7: Check for already imported publications
...
```
3. Create a query that will match no publications and import.
```
Initializing publication importer
Step 1 of 7: Retrieving publication data from remote database tripal_pub_library_PMID
  🗸 No records matched the search criteria
Done.
```
4. Import this pubmed accession 38234546, which is a retracted publication, and its type is not present in the tripal_pub vocabulary. It will be skipped.
`WARNING: Publication with accession 38234546 has a type "Retraction Notice" which is not present in the tripal_pub vocabulary, so will not be imported. Consider adding a term for this type.`

### DIFF
--- a/tripal/src/Form/PubSearchQueryForm.php
+++ b/tripal/src/Form/PubSearchQueryForm.php
@@ -837,6 +837,17 @@ class PubSearchQueryForm extends FormBase {
           }
         }
       }
+
+      // Call plugin-specific form validation
+      $plugin_id = $user_input['plugin_id'] ?? NULL;
+      if ($plugin_id) {
+        // Instantiate the selected plugin
+        // Pub Library Manager is found in tripal module:
+        // tripal/tripal/src/TripalPubLibrary/PluginManagers/TripalPubLibraryManager.php
+        $pub_library_manager = \Drupal::service('tripal.pub_library');
+        $plugin = $pub_library_manager->createInstance($plugin_id, []);
+        $plugin->formValidate($form, $form_state);
+      }
     }
   }
 
--- a/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php
+++ b/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php
@@ -158,16 +158,14 @@ class TripalPubLibraryPubMed extends TripalPubLibraryBase {
    * @ingroup tripal_pub
    */
   public function remoteSearchPMID($search_array, $num_to_retrieve, $page, $row_mode = 1): ?array {
+    $api_key = $search_array['ncbi_api_key'] ?? $search_array['form_state_user_input']['ncbi_api_key'] ?? '';
     // Only initialize for page zero, subsequent pages use the established query
     if ($page == 0) {
+      $days = $search_array['days'] ?? '';
+
       // convert the terms list provided by the caller into a string with words
       // separated by a '+' symbol.
       $num_criteria = $search_array['num_criteria'];
-      $days = NULL;
-      if (isset($search_array['days'])) {
-        $days = $search_array['days'];
-      }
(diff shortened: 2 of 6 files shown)


laceysanderson (approved):
I tested a number of different searches including those suggested. One imported 1441 publications in a single run, another only 12. I confirmed that if the term is missed then that property is reported as skipped. I also published all the publications I imported (1742 total) and looked at a random subset of 10 of them. They look correct and did not provide any errors.

I also did a cursory code review (not enough focus for an indepth code review right now) and nothing juped out at me as concerning.

Good work @dsenalik!

PR #2178, opened 2025-03-25
Fix discovery on unit type field

### LINKED ISSUE #2177: Unit type should only be discovered for featuremap content types
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

I messed up `discover()` for the ChadoUnitTypeDefault field.
It should only be discovered for the `featuremap` base table.
If it is discovered on other tables, for example `pub`, it causes errors.

### Steps to reproduce

Check for new fields for publication. 
You will be able to add 
"Unit Type (pub_cvterm): A unit of measurement is a standardized quantity of a physical quality."

### Error messages

```shell
ERROR: The pub_cvterm.record_id property type uses the store_id action type but is not associated with the base table of the field. Either change the base_table of this field or use store_pkey instead.  featuremap != pub
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
I am sorry, I messed up the ChadoUnitTypeDefault field's `discover()` function. It should only be discovered for content types based on the `featuremap` table. Currently it is getting discovered for any table with a fkey to cvterm_id 😱

The fix is simple

## Testing?
1. On this branch, import the genetic and genomic content types
2. Remove the Unit Type field from Genetic Map
3. Confirm that "Check for new fields" discovers it and adds it*
4. Do the same for the Physical Map content type
5. Discover for the Publication content type. The field should NOT be discovered

* Note that the original field machine name is `genetic_map_unit_type` and after removal and discovery it is `genetic_map_unittype_id` because the discovered machine name is based on the table column name `unittype_id`, and the original machine name is from the yaml.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoUnitTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoUnitTypeDefault.php
@@ -119,7 +119,8 @@ class ChadoUnitTypeDefault extends ChadoFieldItemBase {
     // Specific settings for this field
     $options += [
       'id' => self::$id,
-      'table' => 'cvterm',
+      'base_table' => 'featuremap',
+      'base_column' => 'unittype_id',
       'label' => 'Unit Type',
       'termIdSpace' => 'UO',
       'termAccession' => '0000000',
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
✅ Code changes make sense
✅ Confirmed that it is no longer discovered on publications.
✅ Confirmed it is discovered on genetic map
✅ Confirmed it is discovered on physical map
✅ Confirmed I could create a genetic map after using discover to add the unit type field
✅ Confirmed I could create a physical map after using discover to add the unit type field

PR #2181, opened 2025-03-26
Persistent publish form defaults

### LINKED ISSUE #2176: Remember "last state" for a toggle
### Task Description

Some forms have toggles that, for some users, will be annoying because they need to click them every time.
The example inspiring this is the "Republish existing content" toggle - some sites are more "archival" and existing content would not often be republished. Others, however, would prefer to republish every time.

The task is to store the last selected state in a persistent way and use this to set the default for the next time the form is used

### Branch Name

tv4g0-issue2176-persistent-form-defaults

### PR #2181

dsenalik:
# New Feature

## Description
This provides a quality of life improvement for the publish form.
When publishing a content type, the previously used settings will be used as the current defaults.
The most obvious is "republish" and "batch size", however the tripal migration settings "migration_file" and "lenient" are also saved as the defaults. And if a migration file is specified, then the "Tripal 3 Migration Options" fieldset will be open by default. This will be super helpful when migrating a Tripal 3 site, the migration file will not need to be specified every time.

## Testing?
Interact with the publish form (you don't even need to create content) trying various combinations of settings, and verify that the previous state is the current default.

### DIFF
--- a/tripal/src/Form/TripalEntityPublishForm.php
+++ b/tripal/src/Form/TripalEntityPublishForm.php
@@ -339,7 +339,7 @@ class TripalEntityPublishForm extends FormBase {
       $options[$key] = $values[$key];
     }
 
-    // Store the current republish setting as the default for the next time publish is run
+    // Store the current form settings as the default for the next time publish is run
     \Drupal::state()->set('tripal_publish_form_defaults', $options);
 
     // Run the form submit for the storage backend.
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code changes are clean and make sense. I agree with using the state api for this
✅ Manually tested and the defaults were remembered seamlessly. I confirmed they are remembered through a cache clear and changed them numerous times to ensure they kept updating.

This is ready to merge once the tests pass again.

PR #2185, opened 2025-03-31
ChadoIdSpace/ChadoVocabulary Missing Values

### LINKED ISSUE #1354: ChadoIdSpace/ChadoVocabulary Missing Values
If the user creates a Vocabulary or IDSpace it will automatically add the records to the `tripal_vocabulary_collection` table.   This in turn will call `ChadoIdSpace::create()` and `ChadoVocabulary::create()` function respectively and those will  add the appropriate record to the `cv` and `db` tables of Chado.  However, if for some reason the `cv` or `db` records get deleted outside of Tripal ( for example the Chado schema is dropped and recreated) then the `ChadoIdSpace` and `ChadoVocabulary` plugins can't get a record so the getters return nothing.  

The current way to test if an IDspace or Vocabulary exists is to call the loadCollection function for the plugin manager. If it does not return NULL then we know the collection exists.  For example:

```php
    $idsmanager = \Drupal::service('tripal.collection_plugin_manager.idspace');
    $idSpace = $idsmanager->loadCollection($name, 'chado_id_space');
    if (!$idSpace) {
      $idSpace = $idsmanager->createCollection($name, 'chado_id_space');
    }
```

Can we add a new `TripalCollectionPluginInterface::load` function that can be called with the `loadCollection` function is called and allow the plugin to report if the collection exists internally and if not return NULL so it can be recreatd?

laceysanderson:
# Bug Fix

## Description

If the user creates a Vocabulary or IDSpace it will automatically add the records to the `tripal_vocabulary_collection` table.   This in turn will call `ChadoIdSpace::create()` and `ChadoVocabulary::create()` function respectively and those will  add the appropriate record to the `cv` and `db` tables of Chado.  However, if for some reason the `cv` or `db` records get deleted outside of Tripal ( for example the Chado schema is dropped and recreated) then the `ChadoIdSpace` and `ChadoVocabulary` plugins can't get a record so the getters return nothing.  

In the linked issue we were going to return NULL if you tried to load a collection that no longer exists in the underlying chado. However, while working on this PR I discovered it has since been updated to create the backend record during load if it doesn't exist. As such, this PR now confirms that through tests.

Additionally, I discovered that my expectations of what happens when you create a Tripal ID Space / Vocab where the backend record already exists is not currently TRUE. I expected that the backend record would be loaded and the internals of the collection set appropriately but they were not. This PR will now fix that.

Tasks:
- [x] Move existing term test to kernel instead of functional
- [x] Add test for the expectations mentioned above.
- [x] Fix create/load collection where backend record already exists to pull backend values into the object.

### Automated Testing

The original Functional test was broken into 4 kernel tests:

- ChadoVocabTermsTest::testTripalIDSpaceClass()
- ChadoVocabTermsTest::testTripalVocabClass()
- ChadoVocabTermsTest::testTripalTermClass()
- ChadoVocabTermsTest::testIssue2032()

They were left pretty similar with updates focused on moving some helpers to ChadoTestTrait and adding error message checks.

The new test focused on this issue is ChadoVocabTermsTest::testIssue1354(). More specifically it does the following:

1. Insert the cv/db records into the test chado.
2. Creates a vocab/ID Space collection for the records inserted above. Checks the name and description are set in the object and match the original chado record.
3. Delete the chado records. Check the name is still available but the description is now empty.
4. Loads the vocab/ID Space collection despite the missing chado records. Check that the name matches, the description is empty and the chado record was recreated.

## Manual Testing

I'm not sure how to manually test this 🙈 That said there are extensive tests for this and if I had broken backwards compatibility the pre-existing tests would have failed and likely the docker would not have even built since terms are so central to Tripal.

### DIFF
--- a/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
+++ b/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
@@ -36,7 +36,7 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
   /**
    * A record in the chado.db table for this ID Space.
    *
-   * @see @see Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::loadIdSpace().
+   * @see Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::loadIdSpace().
    *
    * @var array|null
    */
@@ -181,7 +181,7 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
    */
   protected function loadIdSpace(bool $force_reload = TRUE) {
 
-    if ($force_reload OR $this->chado_record === NULL) {
+    if ($force_reload or $this->chado_record === NULL) {
       // Get the Chado `db` record for this ID space.
       $query = $this->connection->select('1:db', 'db')
         ->condition('db.name', $this->getName(), '=')
--- a/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
+++ b/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
@@ -33,6 +33,14 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
    */
   protected $db_def = NULL;
 
+  /**
+   * A record in the chado.db table for this ID Space.
+   *
+   * @see @see Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::loadIdSpace().
+   *
+   * @var array|null
+   */
+  protected array|null $chado_record = NULL;
 
   /**
    * An instance of the TripalLogger.
@@ -119,7 +127,7 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
    * {@inheritdoc}
    */
   public function recordExists() {
-    $db = $this->loadIdSpace();
+    $db = $this->loadIdSpace(TRUE);
     if ($db and $db['name'] == $this->getName()) {
       return True;
     }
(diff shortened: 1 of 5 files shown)


tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
```diff
@@ -33,6 +33,14 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
    */
   protected $db_def = NULL;
 
+  /**
+   * A record in the chado.db table for this ID Space.
+   *
+   * @see @see Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::loadIdSpace().
```
dsenalik:
```suggestion
   * @see Drupal\tripal_chado\Plugin\TripalIdSpace\ChadoIdSpace::loadIdSpace().
```

tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
```diff
@@ -163,21 +171,35 @@ public function destroy(){
...
-    // Get the Chado `db` record for this ID space.
-    $query = $this->connection->select('1:db', 'db')
-      ->condition('db.name', $this->getName(), '=')
-      ->fields('db', ['name', 'url', 'urlprefix', 'description']);
-    $result = $query->execute();
-    if (!$result) {
-      return NULL;
+  protected function loadIdSpace(bool $force_reload = TRUE) {
+
+    if ($force_reload OR $this->chado_record === NULL) {
```
dsenalik:
```suggestion
    if ($force_reload or $this->chado_record === NULL) {
```
fussy, but in the other case of this test, the OR is lower case

tripal_chado/tests/src/Kernel/Plugin/TripalVocabTerms/ChadoVocabTermsTest.php
```diff
@@ -961,4 +1051,76 @@ public function testTripalVocabularyClasses() {
     $this->assertTrue($is_valid, "Term is not valid but should be");
 
   }
+
+  /**
+   * Testing idSpace and Vocabulary collection with missing db or cv.
+   * This can happen when a TripalTerm is created and then the underlying
+   * cd/db is deleted.
```
dsenalik:
```suggestion
   * cv/db is deleted.
```

dsenalik (approved):
Tiny little changes from code review.
I imported a GFF and published and edited a gene just to be able to say I did some sort of manual test 😁

PR #2186, opened 2025-04-01
Organism autocomplete

### LINKED ISSUE #2106: Create an organism autocomplete
### Task Description

We do not yet have an organism autocomplete, and this is because it is more complicated to do, since several columns will need to be included.
The task is to generate such an autocomplete.

### Branch Name

tv4g1-issue2106-organism-autocomplete

dsenalik:
# New Feature

## Description
This creates an autocomplete for links to chado.organism
It is the last autocomplete because it is the most complicated to implement, we need to concatenate genus, species, infraspecific type, and infraspecific name

It is a performance issue because if a site has a large number of organisms, the organism field can cause the page to time out loading all organisms into a select.

This autocomplete will work the same as the other autocompletes based on the generic autocomplete, and has settings for switching between select for a few records vs. autocomplete for many records, and the same other settings (size, etc.)

One new feature is, **for selects only**, if the `abbreviation` field is populated for an organism, that is used in place of concatenating the four values. This means you can have infraspecific type abbreviations if you want or the genus abbreviated if you want 😁
organism_select

## Testing?
1. Import the genomic type collection
2. You will need to create several organisms. Populate the `abbreviation` for some but not all.
3. Now create a Gene, and select an organism
4. Go to the gene widget settings `/admin/structure/bio_data/manage/gene/form-display` and change the settings so autocomplete will be triggered
organism_settings
5. Now you can test using the autocomplete
n.b. If you notice that the "Autocomplete matching" setting has no effect, that is true for all autocompletes and is issue #2187


laceysanderson (changes requested):
Manual testing

Created 5 organism with 2 having abbreviations. Widget showed a select as expected and worked on create/save.
Screenshot 2025-04-15 at 4 18 04 PM

Created an additional 100 organism and confirmed that it does force the autocomplete and the autocomplete shows the records as expected. 
Screenshot 2025-04-15 at 4 36 57 PM

I can select one and save and it seems to be saved right as well.
Screenshot 2025-04-15 at 4 46 36 PM

But then when I edit the page, it pukes 😢 
Screenshot 2025-04-15 at 4 39 17 PM

dsenalik:
> But then when I edit the page, it pukes 😢 

Well yes, I can replicate that, it turns out I did not do very good testing there 😳 - sorry.
Simple to fix with commit https://github.com/tripal/tripal/pull/2186/commits/c90077b55e65a3a26f993d38114d9035368ea31c

laceysanderson (approved):
Works perfectly now! I tested create/edit with autocomplete and select and all saves as expected with no more puke 😁 
Screenshot 2025-04-23 at 10 37 35 AM
Screenshot 2025-04-23 at 10 37 12 AM
Screenshot 2025-04-23 at 10 36 46 AM
Screenshot 2025-04-23 at 9 56 31 AM

PR #2188, opened 2025-04-01
Resolve high memory consumption during publish

### LINKED ISSUE #2179: Memory usage during publish
### Task Description

I'm creating this issue to note that there is some cumulative usage of memory during publish when publishing many records. Publish uses batch size of 1000. This chart shows memory usage above background while publishing a big bunch of publications. Eventually the job failed because the docker ran out of memory.

Image

### Branch Name

tv4g10-issue2179-publish-memory-usage

dsenalik:
# Bug Fix

## Description
When publishing a large number of entities, e.g. genes, you may run out of memory eventually.
This is because Drupal caches every new entity in memory.
The resolution is to clear the memory cache after each batch during publish, after we are done with the entity objects.

## Testing?
Either migrate a Tripal 3 site, or import a lot of genes, but try to publish them.
A good diagnostic is to put a memory usage message in
`tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php`
such as
```
$memory = number_format(memory_get_usage());
$this->logger->notice("Batch $batch_num memory: $memory");
```
before line 1443 on this branch

Example with publications, more memory-intensive because there are lots of fields
```
root@8f6bd6f2910f:/var/www/drupal# drush trp-chado-publish pub     
 [notice] Initial memory: 31,141,600
...
 [notice] Preparing to publish 39,242 "pub" records in 40 batches
...
 [notice] Batch 0 memory: 149,249,328
...
 [notice] Batch 1 memory: 196,258,784
...
 [notice] Batch 2 memory: 194,211,928
...
 [notice] Batch 3 memory: 137,605,048
...
```
and so on. There may be a slow average increase but nothing like it was before. In the graph you can see where we ran out of memory on 4.x
memoryusage

### DIFF
--- a/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
+++ b/tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
@@ -810,10 +810,14 @@ class ChadoPublish extends TripalBackendPublishBase {
       $tags[] = 'values:tripal_entity:' . $entity_id;
       $index++;
     }
-    // Clear cache so that fields will appear on new entities
     if ($index) {
+      // Clear cache so that fields will appear on new entities
       \Drupal::service('cache.entity')->invalidateMultiple($tags);
       \Drupal::service('cache_tags.invalidator')->invalidateTags(['rendered']);
+
+      // The Drupal memory cache can get quite large with large
+      // publish jobs. Clear it since we are done with these entities.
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
These are the numbers I'm seeing across 7 batches:

Screenshot 2025-04-04 at 10 18 07 AM

While it is still increasing over time it is not as pronounced and I see no downside to this change ❤️

PR #2190, opened 2025-04-06
Repair widget remove button

### LINKED ISSUE #2020: Dbxref widget not deleting removed dbxrefs from chado
### Tripal Version

4.x

### Drupal Version

Probably any

### PostgreSQL Version

Probably any

### PHP Version

Probably any

### General Description

The Database Reference Annotations field when set to multiple cardinality should allow removal of existing references with the "Remove" button. Right now it only removes from the drupal field table, not from the chado table.

This appears to be limited to this field, other linking fields e.g. analysis remove the reference from chado also.

### Steps to reproduce

1. Create a new project, and include any two dbxrefs
2. Check the chado table
```
sitedb=> select * from project_dbxref;
 project_dbxref_id | project_id | dbxref_id | is_current 
-------------------+------------+-----------+------------
                 1 |          1 |      3496 | t
                 2 |          1 |      2129 | t
(2 rows)
```
3. Edit and click the "Remove" button on one of them and save
4. Verify it no longer appears on the entity
5. Verify it was removed from the chado table. Currently seeing it not removed
```
sitedb=> select * from project_dbxref;
 project_dbxref_id | project_id | dbxref_id | is_current 
-------------------+------------+-----------+------------
                 2 |          1 |      2129 | t     <-- This one should have been removed
                 3 |          1 |      3496 | t
(2 rows)
```

### LINKED ISSUE #2157: Widget Remove button problems
### Tripal Version

4.x

### Drupal Version

11.1

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

I am noting problems with widgets for unlimited cardinality fields when trying to remove content. Chado is updated correctly but drupal is not.

### Steps to reproduce

1. Create two projects
2. Create a contact and link the two projects
Image
3. Edit the contact and click the remove button on the first one, then save
Image
4. Observe a blank entry in the list
Image

### Error messages

```
sitedb=> select * from project_contact;
 project_contact_id | project_id | contact_id 
                  3 |          2 |          2
(1 row)

sitedb=> select * from tripal_entity__contact_project;
 bundle  | deleted | entity_id | revision_id | langcode | delta | contact_project_record_id | contact_project_entity_id | contact_project_linker_id | contact_project_link | contact_project_project_id | contact_project_project_name | contact_project_project_description 
---------+---------+-----------+-------------+----------+-------+---------------------------+---------------------------+---------------------------+----------------------+----------------------------+------------------------------+-------------------------------------
 contact |       0 |         3 |           3 | und      |     0 |                         2 |                         2 |                         3 |                    2 |                          2 | Project Two                  | <p>D1</p>
 contact |       0 |         3 |           3 | und      |     1 |                         2 |                        -1 |                         1 |                    2 |                          0 |                              | 
(2 rows)
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
On Drupal 10.3, the widget remove button is leaving a record in the Drupal field table. Chado is updated correctly.
On Drupal 11.1 the same thing, but also setting a record back to "-Select-" also has the same problem.
I figure this is high priority because it means editing content is broken if you want to remove a link.

~~To fix this, I needed to get information from the widget into Tripal `preSave()`. The widget only returns a `$values` array, and this gets significantly processed before `preSave()` sees it, and you can't just add additional information to it and expect to see it in `preSave()`. I realized I could use the `entity_id` property to pass a value to tell `preSave()` to remove the record (using the existing mechanism) so that Drupal won't see it and try to create a field table entry.
The only problem with this is that a field not linked to a content type won't be able to use this method, and specifically I am thinking of the dbxref field. There is an existing issue for this very problem in that field #2020~~
~~So for now this PR is draft~~

To fix this, in `preSave()` I check for the properties with an action of `save` and get their values. For example where we have projects on a contact entity, the property will be for `project_id`.
This property is an integer zero when either the "Remove" button was used, or the value was removed from the form.
When this happens, then the existing mechanism (the `$delta_remove` array) is populated with this delta, and downstream Drupal will not see this delta 🙈 so won't create an empty record. YAY!

And with this now working, commit https://github.com/tripal/tripal/pull/2190/commits/86f85ba480b8d83919ae22604d54bb20c5b56e29 fixes issue #2020

## Testing?
See Issue #2157 for steps to reproduce the bug, do the same to test this PR
Also test setting an existing record to `- Select -`
project-to-select

Also test the dbxref field specifically, I did that on the project content type.

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -915,8 +915,8 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
           // Determine whether the property values are to be cached in the
           // Drupal Entity Field tables.
           if ($storage->isDrupalStoreByFieldNameKey($field_name, $key)) {
-            $prop_values[$key] = $prop_value;
-            $prop_types[$key] = $prop_type;
+            $prop_values[] = $prop_value;
+            $prop_types[] = $prop_type;
           }
         }
 
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -908,10 +908,22 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
         // Load into the entity the properties that are to be stored in Drupal.
         $prop_values = [];
         $prop_types = [];
+        $store_values = [];
         foreach ($values[$tsid][$field_name][$delta] as $key => $prop_info) {
           $storage = $tripal_storages[$tsid];
           $prop_type = $storage->getPropertyType($field_name, $key);
           $prop_value = $prop_info['value'];
+
+          // Store the values of any properties with a "store" action.
+          // There will usually only be one, exceptions are dbxref, relationship.
+          $action = '';
+          if ($prop_type) {
+            $action = $prop_type->getStorageSettings()['action'] ?? '';
+          }
+          if ($action == 'store') {
+            $store_values[$key] = $prop_value->getValue();
+          }
+
           // Determine whether the property values are to be cached in the
           // Drupal Entity Field tables.
           if ($storage->isDrupalStoreByFieldNameKey($field_name, $key)) {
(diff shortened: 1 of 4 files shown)


tripal/src/Entity/TripalEntity.php
```diff
@@ -929,7 +940,22 @@ public function preSave(EntityStorageInterface $storage): void {
           $item->tripalLoad($item, $field_name, $prop_types, $prop_values, $this);
           // Keep track of elements that have no value.
           // A given delta should only be present once here.
-          if ($this->allNull($prop_values) and (!array_key_exists($field_name, $delta_remove) or !in_array($delta, $delta_remove[$field_name]))) {
+          if ($this->allNull($prop_values) && !($delta_remove[$field_name][$delta] ?? NULL)) {
```
laceysanderson: I think that `$delta_remove[$field_name]` is a list of `$delta` --it is not keyed by `$delta`. That is why the original if criteria checks that `$delta` is in the array not the key exists... The new change is working by coincidence I think, since delta is numeric and is often added in order.

laceysanderson:
Commit 09a5e74dcb40ecf5ed79d8c31cbbfdfb81e558aa fixes a small bug appearing on the project page if no dbxref was entered (see screenshot). It just adds a check that assigns an empty array as a default if no initial values are stored for a field.
PR2190-removeContactOnCreate

laceysanderson (changes requested):
Code-wise I only have the one question (see inline comment below).

Manual testing wise this is working perfectly 😍 

The following situations performed as expected:

- ✅ Removing a contact linked to a project on create. Checked when the removed record is the first, last or middle.
- ✅ Removing a contact linked in a previous save on edit.
- ✅ Removing multiple contacts but not all on create or edit.
- ✅ Removing all contacts on create or edit.
- ✅ Removing a dbxref on create or on edit.
- ✅ Removing multiple dbxref.
- ✅ Removing all dbxref.
- ✅ Remove by selecting the empty record in a number of the above combinations.

This looks great! Thanks @dsenalik ❤️

tripal/src/Entity/TripalEntity.php
dsenalik: Oh good catch. My misguided simplification shall be reverted 🙈

laceysanderson (approved):
With the suggested change made, I approve this! It is ready to merge assuming tests pass when complete :-)

PR #2191, opened 2025-04-07
Use phpunit 9.6 for any 10.x drupal version

### LINKED ISSUE #2170: Wrong phpunit config in 10.x docker images
### Tripal Version

4.x

### Drupal Version

10.3, 10.4, 10.5

### PHP Version

8.1, 8.2, 8.3

### General Description

I saw this in a failing test, but it is present in passing tests too. Example below is from https://github.com/tripal/tripal/actions/runs/13940262233/job/39015640379 which is Drupal 10.4 PHP 8.3 Postgresql 17

### Steps to reproduce

Observe automated test results for any of the Drupal 10.x tests

### Error messages

```shell
Run docker exec tripaldocker service postgresql restart
Restarting PostgreSQL 17 database server: main.
PHPUnit 9.6.22 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 20:
  - Element 'phpunit', attribute 'displayDetailsOnTestsThatTriggerErrors': The attribute 'displayDetailsOnTestsThatTriggerErrors' is not allowed.
  - Element 'phpunit', attribute 'displayDetailsOnTestsThatTriggerWarnings': The attribute 'displayDetailsOnTestsThatTriggerWarnings' is not allowed.
  - Element 'phpunit', attribute 'displayDetailsOnTestsThatTriggerDeprecations': The attribute 'displayDetailsOnTestsThatTriggerDeprecations' is not allowed.
  - Element 'phpunit', attribute 'cacheDirectory': The attribute 'cacheDirectory' is not allowed.

  Line 61:
  - Element 'bootstrap': This element is not expected. Expected is ( extension ).

  Line 95:
  - Element 'source': This element is not expected.

  Test results may not be as expected.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
Hopefully this will resolve the PHP Unit problem described here https://github.com/tripal/tripal/issues/2170#issuecomment-2783515163
Because the docker now builds with the correct version of `phpunit.xml`, the `select_phpunit.sh` can be eliminated since it is no longer needed. I feel like this was the problem, we were effectively changing the config twice for 10.x runs.

## Testing?
This is an automated testing only problem, so only code review, and verify that tests pass.
But we can only really test on the nightly run after merge 😕


laceysanderson (approved):
🤦‍♀️ Opps!

Code change makes perfect sense. I will merge this so we can confirm the testing tonight.

PR #2193, opened 2025-04-17
Implement "AND" instead of "OR" for tripal content listing filter

### LINKED ISSUE #2192: Tripal content type listing filter uses OR instead of AND
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

The view implemented for the tripal content listing uses an OR instead of an AND.
The result is that if you select any content type, then the filter by title returns every record.

This filter
Image

Returns every publication

We need to fix here

Image

### Steps to reproduce

.

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
This is a very simple change to the tripal content listing, where conditions in the filter are now combined with "AND" instead of "OR"
The problem was that if you tried to filter by both a content type, and some text, the existing "OR" condition meant that you would potentially see records from a different content type than what was in the filter.

## Testing?
1. Create any two different content types with "Tripal" in the title e.g. Project and Contact
2. In the content listing first filter by "Project" - you will now only see one
3. Then type "T" in the "Title" field. The other one will again appear on 4.x, but not on this branch

### DIFF
--- a/tripal/config/install/views.view.tripal_content_type_listing.yml
+++ b/tripal/config/install/views.view.tripal_content_type_listing.yml
@@ -6,7 +6,7 @@ dependencies:
     - tripal
     - user
 _core:
-  default_config_hash: mP9f-XmAQWc4QPnYqqJiE6GnTFEhd9H5qcy-Q3hxzF4
+  default_config_hash: MFwmxO1-UgyEec22hL2SmulHUG_sl8LvW2_nX5MybdQ
 id: tripal_content_type_listing
 label: 'Tripal Content Type Listing'
 module: views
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code change is small and makes perfect sense
✅ I agree that this approach is the correct one over the "OR" 🙈 
✅ Manual testing performs as expected (see screenshots)
Screenshot 2025-04-22 at 11 32 21 AM
Screenshot 2025-04-22 at 11 32 36 AM
Screenshot 2025-04-22 at 11 32 53 AM

PR #2194, opened 2025-04-18
Warn if entity title is blank

### LINKED ISSUE #2165: Warn site admin if entity title is blank
### Task Description

If you set up tokens for an entity title that are not required, you could hit a case where you get an entity with no title.
For example, set an organism title to just `[organism_infraspecific_type]` then create an organism witout an infraspecific type.

The task is to
1. If this happens during publish, set a watchdog error to notify the site admin
2. If this happens during entry or edit through the UI, set an appropriate warning message

### Branch Name

tv4g1-issue2165-warn-if-title-blank

dsenalik:
# New Feature

## Description
This presents warnings if an entity title is blank during either publish or entry through the UI
Blank here means an empty string, a string with only white space, or a string with only HTML tags e.g. `<p></p>`

## Testing?
1. Enter a new publication and leave the title blank
NoTitleGUI
2. Unpublish this entity and then publish "Pub" content type. The existing "NULL" publication will also have no title.
```
# drush trp-chado-publish pub
...
 [warning] WARNING: 2 entities have blank titles. You should either update the source records, or modify
 the title format tokens, and then republish.
[site http://default] [TRIPAL] WARNING: 2 entities have blank titles. You should either update the source
 records, or modify the title format tokens, and then republish.
 [notice] Publish completed. Published 2 new entities, added 26 new field values.
```


laceysanderson (approved):
✅ Code changes make sense and look good to me :-)
✅ Entering a publication with no title does show the error and also save the other content added.
✅ Editing to add a title does make the error go away on save :-)
✅ I also see the warning on publish :-)
image
Screenshot 2025-04-23 at 11 13 30 AM

PR #2203, opened 2025-05-22
Implement max delta

### LINKED ISSUE #2101: Implement Max Delta
### Task Description

"Max Delta" is a concept relating to publishing, where, when there are too many linked records, we don't publish every single one.

For example, one might have a "Genome Annotation" analysis where 40,000 genes were annotated in a genome.
These can be linked together through the `analysisfeature` table, but when we publish this analysis, we don't want to have 40,000 rows of genes listed, bad things will happen. Even if we used a pager, this is not a reasonable approach. We should have a mechanism to limit the number of records of a specific type for a given bundle, and ideally convert to a summary view or a link to a search page when the limit is exceeded.

This was referenced in https://github.com/tripal/t4d8/issues/310 and was implemented for Tripal 3 in the fields provided by the https://github.com/tripal/chado_linker_fields module.
However, Tripal 4 publish is very different than Tripal 3 publish, so the mechanism will also have to be different.

### Branch Name

tv4g1-issue2101-max-delta

dsenalik:
# New Feature

## Description
"Max Delta" is a concept relating to publishing, where, when there are too many linked records, we don't publish every single one.

For example, one might have a "Genome Annotation" analysis where 40,000 genes were annotated in a genome.
These can be linked together through the `analysisfeature` table, but when we publish this analysis, we don't want to have 40,000 rows of genes listed, bad things will happen. Even if we used a pager, this is not a reasonable approach. We should have a mechanism to limit the number of records of a specific type for a given bundle, and ideally convert to a summary view or a link to a search page when the limit is exceeded.

This was referenced in https://github.com/tripal/t4d8/issues/310 and was implemented for Tripal 3 in the fields provided by the https://github.com/tripal/chado_linker_fields module.
However, Tripal 4 publish is very different than Tripal 3 publish, so the mechanism here is very different.

Linked issue for **documentation**: https://github.com/tripal/tripal_doc/issues/106

The limit is implemented on a chado table basis, rather than on a content type basis. So for example gene + mRNA are lumped together. However, fields work this way too, for example the feature field displays both.

These need to be addressed:
 - [x] Override Drupal's formMultipleElements() in TripalWidgetBase so that if cardinality is set to e.g. 100, you don't see 100 empty fields, only those populated plus one are displayed, and an add button is provided (See comment https://github.com/tripal/tripal/pull/2203#issuecomment-2937285555 )
 - [x] Populate the object_table setting during publish when it is missing (See comment https://github.com/tripal/tripal/pull/2203#issuecomment-2949439882 ) commit https://github.com/tripal/tripal/commit/e9b55d4dcec73fae637206b79a105fad4b9f6a3e
 - [x] There is no feed back when the limit restricts publishing. https://github.com/tripal/tripal/pull/2203/commits/9aedc30e9ad51c18038ab9d9f2bf8280bb6d63e9
 - [x] Option for when  more than max delta records exist to either: a. publish max delta records, or b. publish **no** records. https://github.com/tripal/tripal/pull/2203/commits/565f10109dd5af329ef5e3904e1f61117d9abec0
 - [x] For option a. above, the formatters could include a message that not all records are shown.
 - [x] Use the existing cardinality setting as a method to override this for individual fields, if cardinality is a positive integer greater than one, this overrides the global setting for publishing. Mostly in commit https://github.com/tripal/tripal/pull/2203/commits/e9b55d4dcec73fae637206b79a105fad4b9f6a3e
 - [x] Tests!

And there is no way to switch to an "overview" mode, but this is beyond the scope of this PR

There is a field on Tripal -> Configuration -> Tripal Entity Settings where the global value may be configured:
PR2203 2

## Testing?
Let's use different tables for manual testing than the unit tests.
  1. create a publication
`INSERT INTO pub (title, uniquename, type_id) VALUES ('Pub1', 'p000001', 5);`
  2. create lots of organisms
```
DO $$
BEGIN
    FOR count IN 1..200 LOOP
        INSERT INTO organism (genus, species) VALUES ('Tripalus', count);
        INSERT INTO teacup.organism_pub (organism_id, pub_id) VALUES (count, 2);
    END LOOP;
END; $$
;
```
  3. Put the organism field on the pub content type, can be done with "Discover new fields"
  4. Publish the organisms first so that they have entity ID values.
`drush trp-chado-publish organism`
...
`Publish completed. Published 200 new entities, added 600 new field values.`
  5. publish publications (the null publication will also be published)
`drush trp-chado-publish pub`
` [warning] WARNING: The number of organism_pub records exceeds the configured limit for pub_id=2, only 101 records will be returned`
...
`  Published 101 items for field "pub_organism"`
` [warning] WARNING: 1 entities have blank titles. You should either update the source records, or modify the title format tokens, and then republish.`  (this is from the NULL publication and is expected)
`
This reported 101 published, this is because we publish one extra so we can distinguish between being exactly at the limit, and being above that and some records are not published. The formatter will not display the 101st record.
PR2203 2

  6. Verify in the field table
```
  sitedb=> SELECT COUNT(*) FROM tripal_entity__pub_organism;
 count 
-------
   101
(1 row)
```

  6. Try updating the global setting at Tripal -> Configuration -> Tripal Entity Settings
  Set it to **105**, save, and republish
  `drush trp-chado-publish pub --republish`
```
  sitedb=> SELECT COUNT(*) FROM tripal_entity__pub_organism;
 count 
-------
   106
(1 row)
```
  7. Tripal -> Page Structure -> Publication -> Manage Fields
  8. Organism -> Edit
  9. Change cardinality to limited, and set a value higher than the global max delta
PR2203 3
  10. Republish again and verify the count is now higher than the global max delta
  ```
sitedb=> SELECT COUNT(*) FROM tripal_entity__pub_organism;
 count 
-------
   121
(1 row)
```
  11. Change organism back to unlimited cardinality
  12. Change global setting to 150, but also set the "Inhibit publish when many records are present"
  13. Republish again. Existing records should remain, but no new ones added. Count should still be 121.
  This message should be displayed
`[warning] WARNING: The number of organism_pub records exceeds the configured limit for pub_id=2, no records will be returned`

### CODE REVIEW

**Comment by dsenalik** [comment]
I am considering the possibility of using the existing cardinality setting to override the global max_delta setting on a per-field basis.
This has the advantage of already being present, however the one disadvantage is that for fields with a fixed cardinality, the widget will display a field for every delta, so with max_delta=100 there will be 100 (likely empty) fields on the widget form.

@laceysanderson 's comment from slack:
This is a drupal thing. There is code in the base drupal field that checks this value and if unlimited uses the form with add more and if limited generates exactly that number of forms. To change this functionality we would override the formMultipleElements method I believe. See https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21WidgetBase.php/function/WidgetBase%3A%3AformMultipleElements/11.x

There is a drupal issue about it https://www.drupal.org/project/drupal/issues/1156338 and although relatively current, it dates back to 2011 so I am not sure how soon it might be incorporated.
Current state of code changes on that merge request: https://git.drupalcode.org/project/drupal/-/merge_requests/6522/diffs?diff_head=true#b35173aafbefc4909e17d0ad418458d2c9b02d91

**Comment by dsenalik** [comment]
Some notes about a situation I have noticed, I wanted to use the `object_table` setting in the storage plugin settings for a field, but it is never defined for any field.
All of the linker fields have this function
https://github.com/tripal/tripal/blob/989379de8243b398507183582fb4ff1b87c861be/tripal_chado/src/Plugin/Field/FieldType/ChadoProjectTypeDefault.php#L40-L47
and line 45 defines the `object_table` setting.
However because `object_table` is not in any of the yaml used to define fields, it is not available for use.
And when a field is added manually, it does not get included there either.

Publish will have code to look up and add this and store it permanently. Most of this is in commit https://github.com/tripal/tripal/pull/2203/commits/e9b55d4dcec73fae637206b79a105fad4b9f6a3e

File: tripal/src/Form/TripalEntitySettingsForm.php [hunk 1]
```diff
@@ -97,6 +107,39 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+                        . " field's cardinality to a value greater than 1."),
+      '#default_value' => $publish_global_max_delta,
+      '#required' => FALSE,
+    ];
+    $form['publish']['publish_global_max_delta_inhibit'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Inhibit publish when many records are present'),
+      '#description' => $this->t('When enabled, no linked records will be published on an entity'
+          . ' if the total number of records of one type exceeds the value set above.'
+          . ' If not enabled, then records will be published, but only up to this limit.'),
```
laceysanderson: This doesn't feel clear to me that it's only the specific field with too many values that is being inhibited. :thinking:

What about something like:

> This setting applies to each field individually on an entity. For a given field with more then the max delta number of records: if this setting is enabled then no records will be published for this field and if not, then max delta + 1 records will be published for this field.

tripal/src/Form/TripalEntitySettingsForm.php
```diff
@@ -116,16 +159,21 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
 
     $widget_global_select_limit = trim($form_state->getValue('widget_global_select_limit'));
+    $publish_global_max_delta = trim($form_state->getValue('publish_global_max_delta'));
 
     // Non-negative integers or an empty string are valid
     if (!preg_match('/^\d*$/', $widget_global_select_limit)) {
       $form_state->setErrorByName('widget_global_select_limit',
         $this->t('This field must contain an integer value, or be left blank.'));
     }
+    if (!preg_match('/^\d*$/', $publish_global_max_delta)) {
```
laceysanderson: Regular expressions are pretty slow so I default to other methods where possible. In this case, I would use `ctype_digit(strval($publish_global_max_delta))` which checks that the value is a positive integer and works when the value is a string.

tripal/src/Form/TripalEntitySettingsForm.php
```diff
@@ -144,13 +192,17 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     $allowed_title_tags = preg_replace('/ +/', ' ', $allowed_title_tags);
 
     $widget_global_select_limit = trim($form_state->getValue('widget_global_select_limit'));
+    $publish_global_max_delta = trim($form_state->getValue('publish_global_max_delta'));
```
laceysanderson, line 166: Shouldn't you cast the value here to ensure it is an integer as you put in the settings yml?

tripal/src/TripalField/TripalFormatterBase.php
```diff
@@ -25,4 +32,79 @@ abstract class TripalFormatterBase extends FormatterBase {
...
+  }
+
+  /**
+   * Converts a list of multiple renderable items into a list.
+   * A single item is passed through unchanged.
+   *
+   * @param array $list
+   *   A list of renderable items
+   * @return array
+   *   A render array. Will be empty if there were no items.
```
laceysanderson: Maybe expand this a bit to include mention of the warning?

```suggestion
   * HELPER: Makes multiple renderable items into a list.
   *
   * Notes: If the number of items exceeds the max delta for this field
   * then a warning will be added right before the list.
   *
   * @param array $list
   *   A list of renderable items
   * @return array
   *   A render array where (1) if no items then an empty element is
   *   returned, (2) if there is one item it is passed through unchanged,
   *   (3) if `$items < $max_delta` then a list of items is returned, (4) 
   *   if `$items > $max_delta` then two elements are returned where the
   *   first is a warning string and the second is a list of items.
```

tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php
```diff
@@ -1316,6 +1398,35 @@ protected function publish_summarize(bool $success, array $stats): void {
...
+   */
+  protected function getFindValuesOptions(): array {
+    // We add one to max_delta to use the extra record as a flag that
+    // at least one record was not published. The formatters will not
+    // display this last extra record.
+    $max_deltas = [];
+    foreach ($this->field_info as $field_name => $field_info) {
+      if (array_key_exists('object_table', $field_info)) {
+        if ($field_info['cardinality'] > 1) {
+          $max_deltas[$field_info['object_table']] = $field_info['cardinality'] + 1;
```
laceysanderson: This is going to cause inconsistent behaviour where there are multiple fields with the same object table and different cardinality :see_no_evil: For example, with property fields, if field A has a cardinality of 5 and field B has a cardinality of 10 then depending on order we might get a max delta of 6 or 11... I wonder if we can overcome this for such fields by finding the table alias here since property fields will alias the tables to keep them separate.

tripal/src/TripalField/TripalFormatterBase.php
```diff
@@ -3,26 +3,113 @@
...
     return preg_replace('/[^\w]/', '_', $key);
   }
+
+  /**
+   * {@inheritdoc}
+   *
+   * Initializes value of max_delta.
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $max_delta = \Drupal::config('tripal.settings')->get('tripal_entity_type.publish_global_max_delta');
```
laceysanderson: I've found in testing that grabbing from config can be slow. Since this is called for every field on every page load --maybe we should use the Drupal cache? We can then also use the cached value in the publish :thinking:

tripal/src/TripalField/TripalFormatterBase.php
```diff
@@ -3,26 +3,113 @@
...
+    }
+
+    // If more than one value has been found, display all values in an
+    // unordered list.
+    elseif (count($list) > 1) {
+      $this->addMaxDeltaWarning($elements, $list);
+      $elements[] = [
+        '#theme' => 'item_list',
+        '#list_type' => 'ul',
+        '#items' => $list,
```
laceysanderson, line 85: I don't see anything anywhere that removes the extra element from the end added by publish?

tripal/src/TripalField/TripalFormatterBase.php
laceysanderson: Maybe we can move it to the field type class tripalTypes() method and then it can be used in the formatter, widget and publish from the same cached value? That would also allow us to have access to the table alias which might help an issue I commented on in publish.

laceysanderson (commented):
I didn't make it quite all the way through code review but here are some thoughts/suggestions on what I did cover. Specifically, I finish TripalWidgetBase, ChadoRecords and testing.

tripal/src/TripalField/TripalFormatterBase.php
dsenalik: It's one line earlier, line 74
`      $this->addMaxDeltaWarning($elements, $list);`
if the limit is exceeded, it does an array_slice()

laceysanderson: Aaah! There it is 😅 Thanks!

tripal/src/TripalField/TripalFormatterBase.php
dsenalik: Great, source is outdated now, but I will add this.

tripal/src/Form/TripalEntitySettingsForm.php
dsenalik: I don't think so, because integer OR blank is allowed here, and this is enforced even before validation by Drupal because the field is set as 
```
      '#type' => 'number',
      '#min' => 0,
      '#step' => 1,
```


If we store as a blank string, then the default of 100 built into the code is used (which is the same as an existing site updating to this)

tripal/src/Form/TripalEntitySettingsForm.php
dsenalik: It turns out this validation is redundant, because the field has an integer enforced by Drupal because it uses this
```
      '#type' => 'number',
      '#min' => 0,
      '#step' => 1,
 ```
so only non-negative integers or an empty box are valid. I will remove these validations as any value they detect has already been filtered by Drupal.

This only allows through one additional case, prefixing with a `+`. Since `+100` gets stored identically as `100` I think this is okay to allow.

File: tripal/src/Form/TripalEntitySettingsForm.php (hunk 1 above)
Comment (dsenalik) [reply tripal/src/Form/TripalEntitySettingsForm.php]: Good, except the term "max delta" is not user-facing, so I will change to this
`'#description' => $this->t('This setting applies to each field individually on an entity. For a given field with more than the specified maximum number of records: if this setting is enabled then no records will be published for this field and if not, then only the specified maximum number of records will be published for this field.'),`
I took off the + 1 because the formatter removes that last one.

File: tripal/src/TripalField/TripalFormatterBase.php (hunk 6 above)
Comment (dsenalik) [reply tripal/src/TripalField/TripalFormatterBase.php]: I did a simple test to evaluate this, and I am getting very similar times for the config lookup and the cache retrieval, sometimes one is faster, sometimes the other, but always in sub-millisecond range
Uncached 100: 0.000244
Cached 100: 0.000257
If this holds, I see no reason to cache this, what do you think?

```
$t1 = microtime(TRUE);
$max_delta = \Drupal::config('tripal.settings')->get('tripal_entity_type.publish_global_max_delta');
$t2 = microtime(TRUE);
print "Uncached $max_delta: ".sprintf('%0.6f', $t2-$t1)."\n";

$cache_id = 'tripal_global_max_delta';

$t1 = microtime(TRUE);
if ($cache = \Drupal::cache()->get($cache_id)) {
  $cached_max_delta = $cache->data;
  $t2 = microtime(TRUE);
  print "Cached $cached_max_delta: ".sprintf('%0.6f', $t2-$t1)."\n";
}

// Cache the value, specifying expiration in 24 hours.
// Remember to store also in settings form!!!!
\Drupal::cache()->set($cache_id, $max_delta, \Drupal::time()->getRequestTime() + (3600*24));
```

File: tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php (hunk 5 above)
Comment (dsenalik) [reply tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php]: Hmm, I kind of forgot about properties, and max delta is not implemented for those because they are "single hop". There is more to do...

File: tripal/src/TripalField/TripalFormatterBase.php (hunk 6 above)
Comment (laceysanderson) [reply tripal/src/TripalField/TripalFormatterBase.php]: Ok, maybe it's the mocking in tests that makes it really slow :-) In this case maybe the current approach is good though I would still be tempted to move it to the field type class and the widget/formatter pull it from there.

File: tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php (hunk 5 above)
Comment (laceysanderson) [reply tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php]: Another example of this would be the contact by role field. This will alias the linking table so as to keep the fields apart based on the type of link.

File: tripal/src/TripalField/TripalFormatterBase.php (hunk 6 above)
Comment (dsenalik) [reply tripal/src/TripalField/TripalFormatterBase.php]: Would you then recommend to make a function in TripalFieldItemBase.php to return the value?
getGlobalMaxDelta() or similar?

Comment (laceysanderson) [reply tripal/src/TripalField/TripalFormatterBase.php]: Maybe it should be `getFieldMaxDelta()` and then the logic for comparing with the field cardinality can live there too? Alternatively if we take into account the other comment of respecting the table alias', maybe `getTableMaxDeltas()` and it can be an array of table alias to field max delta. This way you can also use it in the publish and just merge all the results to get your find options.

File: tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php (hunk 5 above)
Comment (dsenalik) [reply tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php]: I am working on this, commit https://github.com/tripal/tripal/pull/2203/commits/afbd30995d254d61d9a4f61b47a980d6440aa2fe stores the alias during publish,
and commit https://github.com/tripal/tripal/pull/2203/commits/fd2f329d334715b2af93e530602d3d5247aff30e uses the alias in ChadoRecords.
I think this resolves this problem

**Comment by dsenalik** [comment]
I wish to describe a new problem.
The current mechanism for indicating that not all records were published is to publish one extra record as a flag, and the formatters then remove the extra record and add a message to indicate that not all records were published.

Unfortunately this fails when we use the cardinality to set to a finite number to override the global max delta value for a particular field, because now Drupal enforces the cardinality.
For example, we set cardinality to 5. Publish added 6 records to the Drupal field table. But now the formatter only receives 5 records, and doesn't know about the 6th. This happens even before chado storage gets access. ChadoStorage loadValues() gets called by TripalEntity postLoad() which gets called by Drupal postLoad(), and at this point there are only 5 records available.

It's not really what we wanted, but commit https://github.com/tripal/tripal/pull/2203/commits/90f26595081c3ea47283ecc7964e1f5f2aab548b will print the "more records" message for fields with finite cardinality when the have a number of records matching the full cardinality. We indicate there "may be more records" instead of "there are more records".

File: tripal/src/TripalField/TripalFormatterBase.php (hunk 6 above)
Comment (dsenalik) [reply tripal/src/TripalField/TripalFormatterBase.php]: Slack discussion: Move `getFieldMaxDelta()` to TripalFieldItemBase, then hopefully field item list will be the appropriate object that we can call this from. Likewise in publish.

Comment (dsenalik) [reply tripal/src/TripalField/TripalFormatterBase.php]: I have made this move, and call this from both the formatter and publish.
However in publish I have left in the global_max_delta lookup and still pass it to chado storage,  because my head started to hurt. I wasn't sure if this looked up value would be available in every case, so probably a bit more checking is in order.

Comment (dsenalik) [reply tripal/src/TripalField/TripalFormatterBase.php]: The global value is no longer passed, just the values from `getFieldMaxDelta()`, so this should be ready for review again

Comment (laceysanderson) [reply tripal/src/TripalField/TripalFormatterBase.php]: I've looked through the changes and am marking this as resolved.

File: tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php (hunk 5 above)
Comment (laceysanderson) [reply tripal_chado/src/Plugin/TripalBackendPublish/ChadoPublish.php]: Based on code-review this looks resolved as well.

**Comment by laceysanderson** [comment]
Update on my progress review-wise:

- I went back through my unresolved comments and based on the changes they look resolved according to our discussion.
- Also I looked through the changes in general since my last review and those look good.

What is next:

- Do another full code review of the current state of things with a focus on TripalFieldWidget and ChadoRecords which I didn't do the first time around.
- Manually test the functionality with a focus on multiple property fields as well as the linking table tests described in the PR.

File: tripal/src/Form/TripalEntitySettingsForm.php [hunk 8]
```diff
@@ -139,18 +157,23 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+    $publish_global_max_delta_inhibit = $form_state->getValue('publish_global_max_delta_inhibit');
 
-    // Update configuration
+    // Update configuration.
     \Drupal::configFactory()
       ->getEditable('tripal.settings')
```
laceysanderson: The config factory is available in all forms via dependency injection! See https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Form%21FormBase.php/property/FormBase%3A%3AconfigFactory/11.x

That means this can be

```suggestion
    $this->configFactory->getEditable('tripal.settings')
```

tripal/tests/src/Kernel/Entity/TripalEntitySettingsFormTest.php
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Tests the tripal entity settings form.
+ *
+ * @group Tripal
+ * @group TripalEntity
+ */
+class TripalEntitySettingsFormTest extends TripalTestKernelBase {
+//@needed?
+  //protected $defaultTheme = 'stark';
```
laceysanderson: Technically this line is only needed in (1) Functional tests, (2) Kernel tests that render the page. That said, we often just include it all the time because it doesn't hurt and will make life easier if a test changes to include rendering at some point.

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoAnalysisFormatterDefault.php
```diff
@@ -73,23 +73,8 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
-      $elements[0] = [
-        '#theme' => 'item_list',
-        '#list_type' => 'ul',
-        '#items' => $list,
-        '#wrapper_attributes' => ['class' => 'container'],
-      ];
-    }
-
+    // If multiple items, convert to a list.
+    $elements = $this->createListMarkup($list);
```
laceysanderson: I confirmed that `createListMarkup()` does not create a list out of singular items but this comment is still misleading.


```suggestion
    // Will convert $list to a markup list if there are > 1 items.
    $elements = $this->createListMarkup($list);
```

laceysanderson (commented):
Some more code comments. Still haven't made it all the way through the changes.

tripal/tests/src/Kernel/Entity/TripalEntitySettingsFormTest.php
dsenalik: Oops yes I forgot to remove this it seems

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoAnalysisFormatterDefault.php
dsenalik: Resolved with commit https://github.com/tripal/tripal/pull/2203/commits/3af1f12299c5ed08ca5d08b93c05a7971ee26a55

tripal/tests/src/Kernel/Entity/TripalEntitySettingsFormTest.php
dsenalik: Resolved with commit https://github.com/tripal/tripal/pull/2203/commits/fb7fefe774e0ab2c1f968bf995dbe565a8956865

laceysanderson (approved):
✅ Code review complete

## Manual Testing

Image

✅ Max delta left at default; organism_pub only publishes 101 of 200 organism on a publication; shows 100 on the page
✅ Max delta left at default; organism prop field only publishes 101 of 200 properties on an organism; shows 100 on the page
✅ Field-specific delta of 10 for one prop field and 150 for another prop field; both prop fields publish the correct number of properties (11 and 151 respectively); both fields show the right number of properties on the page (10 and 150 respectively)

Thus confirmed that:
- max delta is respected for both single-hop (property) and double-hop (organism_pub) fields.
- publish warns that max delta was reached and indicates the number of records published.
- publish adds one more record then max delta but the page only shows max delta.
- if two fields for the same table (organismprop) each have their own field-specific max delta using cardinality then the correct limit is applied to each field.
- field-specific max delta can be higher then the global max delta and the field-specific number is respected.

PR #2206, opened 2025-05-27
Fix gff loader inconsistent form

### LINKED ISSUE #2180: GFF Loader Inconsistent Form
### Tripal Version

4.x

### Drupal Version

10.4

### PostgreSQL Version

16

### PHP Version

8.3

### General Description

On the GFF Loader form page, the Default Landmark Type field's description says that it is only needed if the "landmark features (first column of the GFF3 file) are not already in the database", however this field is marked as required. This is an easy fix but I'm not sure if we should change the field not not be required (likely) or if we should change the description.

### Steps to reproduce

Try to use the GFF Importer (admin/tripal/loaders/chado_gff3_loader) and witness the above inconsistency between description and field configuration.

### Error messages

```shell

```

### Screenshots

_No response_

dsenalik:
# Bug Fix

## Description
This fixes the issue in the linked issue, _i.e._ the GFF importer form no longer requires a Default Landmark Type.

Making this optional has revealed that in some GFF files there was a failure, specifically in using an existing landmark (loaded with the FASTA importer) when processing the GFF file. This PR adds a lookup step for the existing landmark.

## Testing?
For setup, create an analysis and an organism

Case 1: Landmark is already defined in chado
1. Load this test fasta file to create a landmark. Use the term `chromosome (SO:0000340)`
[issue2180.1.fna.txt](https://github.com/user-attachments/files/20490233/issue2180.1.fna.txt)
2. Load this test fasta file to create a landmark of a different type. Use the term `contig (SO:0000149)`
[issue2180.2.fna.txt](https://github.com/user-attachments/files/20494275/issue2180.2.fna.txt)

3. Import this gff3 file which has these two different types of landmarks
[issue2180.3.gff3.txt](https://github.com/user-attachments/files/20494276/issue2180.3.gff3.txt)
and leave the "Default Landmark" Type blank

4. Check chado
```
sitedb=> select F.feature_id, F.name, F.uniquename, T.name, L.srcfeature_id from feature F left join cvterm T on F.type_id=T.cvterm_id left join featureloc L on F.feature_id=L.feature_id;
 feature_id |           name            |         uniquename         |    name     | srcfeature_id 
------------+---------------------------+----------------------------+-------------+---------------
          1 | NC_001133.9               | NC_001133.9                | chromosome  |              
          2 | Contig0                   | Contig0                    | contig      |              
          3 | I                         | NC_001133.9:1..2200        | region      |             1
          4 | PAU8                      | gene-YAL068C               | gene        |             1
          5 | NM_001180043.1            | rna-NM_001180043.1         | mRNA        |             1
          6 | exon-NM_001180043.1-1     | exon-NM_001180043.1-1      | exon        |             1
          7 | NP_009332.1               | cds-NP_009332.1            | CDS         |             1
          8 | DifferentLandmarkTypeTest | Bogus0                     | gene        |             2
          9 | rna-NM_001180043.1        | rna-NM_001180043.1-protein | polypeptide |             1
(9 rows)
```

### DIFF
--- a/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
+++ b/tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
@@ -378,7 +378,7 @@ class GFF3Importer extends ChadoImporterBase implements ContainerFactoryPluginIn
     $form['landmark_type'] = [
       '#title' => t('Default Landmark Type'),
       '#type' => 'textfield',
-      '#required' => TRUE,
+      '#required' => FALSE,
       '#description' => t("Use this field to specify a Sequence Ontology type
        for the default landmark sequences in the GFF fie (e.g. 'chromosome'). This is only needed if
        the landmark features (first column of the GFF3 file) are not already in the database."),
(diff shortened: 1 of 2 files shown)


dsenalik:
Discussion in today's cofest at first went in the direction of leaving as-is and changing the description, and making the default term "Chromosome".
However, the use case came up of a GFF3 file with more than one landmark type, such as chromosomes + scaffolds + contigs, which needs to have this looked up from existing landmarks in chado.
So I am continuing in that direction, the field will become optional. I added a function to do the chado query, and modified some of the existing tests to have this looked up rather than using the default term.

Ferrisx4 (approved):
I've looked at the code and it seems OK and clear, and run the tests as described in the PR.
The feature table is looking good afterwards, and the type_id links to the appropriate type as expected:
Screen Shot 2025-05-29 at 1 50 07 PM

PR #2214, opened 2025-06-06
Remove final keyword from classes to allow extension and testing

vidu-wil:
# Task

## Description

`ChadoBuddyPluginManager`, `ChadoBuddy` and `ImportTripalEntityTypeCollection` classes are currently declared as final, which prevents them from being extended or mocked in unit tests. This restricts test coverage for any service or class that depends on it. Removing the final keyword improves flexibility and make it easier to write isolated tests that simulate plugin behavior.

## Testing?

None


dsenalik (approved):
These changes should have no functional effect, but just in case I built a docker on drupal 11.0 php 8.3 postgres 16 and imported type collections, and imported a couple of NCBI taxa

PR #2219, opened 2025-06-23
Drupal coding standards for publishing 📰

dsenalik:
# Tripal 4 Core Dev Task

## Description
This brings parts of tripal and tripal_chado modules involved in publishing in compliance with Drupal coding standards.

## Testing?
Mostly code review is needed. Some dependency injection has been added on the publish form, so publish/unpublish should get some testing.


laceysanderson (approved):
✅ Code changes look great!

## Manual Testing

✅ DI for listing publish backends worked
✅ DI tracking the state of the republish button worked. It remembered that I had it turned on the next time I loaded the form and then when I turned it off again it remembered that as well.
✅ DI for grabbing the entity types worked great. I imported the genomic collection but not the others and confirmed that only the general and genomic content types were listed
✅ DI for Tripal job and user worked, confirmed by submitting a job without error and confirming it was owned by me :-)
✅ DI for file + drupal database stuff works, confirmed by uploading a tsv file and running the job to confirm that the job could find it. Also then reloaded the form to confirm that the file was now listed in the system path.

That's all of the new dependency injections specifically tested! Thanks for this @dsenalik! I find form dependency injection is often more involved then I expect as well 🤪 

Screenshot 2025-10-07 at 9 46 29 AM
image

PR #2221, opened 2025-06-30
Add a publish option for importers in the base class 📰

### LINKED ISSUE #2056: Add an option to publish to the base importer class
### Task Description

It would be awesome if we add an option to publish to importers. For example, in the Tripal 3 publication importer, all the newly imported publications were automatically published.
We can have a checkbox on the base importer form "[x] Publish imported records" or something like that.
The GFF importer would be a more complicated case, which content types get published? Gene, mRNA, others, so the importer could extend this into multiple checkboxes or a select list 🙄

Because we needed the publish service, this issue was expanded to fully implement dependency injection in all of the importers, which means the class interface has additional parameters, and failing to provide the needed services will be deprecated.

Specifically, the TripalImporterBase class now takes four additional parameters, which for now are optional, but failing to provide them triggers the deprecation notices.
```   
   * @param ?Drupal\Core\Messenger\Messenger
   *   The Drupal messenger service.
   * @param ?Drupal\tripal\Services\TripalLogger
   *   The Tripal logger service.
   * @param ?Drupal\tripal\Services\TripalFileRetriever
   *   The Tripal file retrieval service.
   * @param ?Drupal\tripal\TripalBackendPublish\PluginManager\TripalBackendPublishManager
   *   The Tripal publish manager service.
```

The ChadoImporterBase adds one additional parameter
```
   * @param Drupal\tripal_chado\Database\ChadoConnection $connection
   *   A connection to the Chado database.
```
which is inserted in front of these four to allow them to remain optional for now.

### Branch Name

tv4g4-issue2056-publish-option-for-importers

dsenalik:
# New Feature

## Description
**Quality of Life Improvement**: This adds an option to automatically publish content loaded by various importers.
An annotation (to be deprecated) or attribute specifies the bundle or bundles that are appropriate to publish after importing.
A form element is added to opt out of publishing the bundles that were just loaded, the default is to publish.
The actual publishing is done in the `postRun()` function in the chado importer base class.
~~There is a new dependency injection there to get the publish service.~~ In the interests of backwards compatibility, the publish service is not injected into the base importer class. It can be injected into a specific importer, and the base class receives it through a setter function. commit https://github.com/tripal/tripal/pull/2221/commits/3abea3f4c247101f648af6c2bf760e57d3b05f4b This means existing non-core importers from other modules will not be broken! 😲

### How to add publish to your importer

1. Add an array of attributes to your importer specifying what content type or types to publish. For example
```
  publish: [
    'pub',
  ],
```
or
```
  publish: [
    'gene',
    'mrna',
  ],
  ```
2. Add the publish manager class to your importer use statements
`use Drupal\tripal\TripalBackendPublish\PluginManager\TripalBackendPublishManager;`
3. Inject the publish service into your importer, and pass it to the parent class with the setter function. For example
```
public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
    return new static(
      $configuration,
      $plugin_id,
      $plugin_definition,
      $container->get('tripal_chado.database'),
      $container->get('tripal.backend_publish')  // Add this
    );
  }

public function __construct(array $configuration, $plugin_id, $plugin_definition, ChadoConnection $connection,
  TripalBackendPublishManager $publish_manager  // Add this
) {
    parent::__construct($configuration, $plugin_id, $plugin_definition, $connection);
    $this->setPublishManager($publish_manager);  // Add this to pass the publish service with the setter method
...
```

## Testing?
Import genomic type collection.
Run every importer to make sure nothing has been broken.
However, the only importers that have an annotation to publish are pub, GFF, taxonomy, and phylotree.
~~Note that imported phylotrees cannot be published, issue #2222~~ fix has been merged

**New form element for publication importer**
pub

**New form element for taxonomy importer**
organism

**New form element for GFF importer**
genemrna

**New form element for Newick Tree Loader**
tree

**New form element for taxonomy tree generator**
speciestree

### CODE REVIEW

**Comment by Ferrisx4** [comment]
I see that @laceysanderson plans to review this, I also took a look at it. For the few content types that I've tried (pub, taxonomy) it seems to have worked (or at least there are no obvious errors).
after pub and taxonomy import

I wonder if it might be less awkward if the checkbox was enabled by default with just the word "Publish" next to it, or a similar positive phrase, instead of "enabling" the checkbox to disable the functionality. It reminds me of those email unsubscribe pages that are confusing on purpose.

**Comment by dsenalik** [comment]
You have a very good point @Ferrisx4 - I think the code was simpler this way, but a simpler UI is more important. I will make it so!

**Review by laceysanderson — COMMENTED** [review]
Some questions/thoughts I had when going through the code:

## Support for non-core bundles

Currently you list the machine name of the bundle in the importer attribute. This makes things easy for publish but restricts the importers to hard-coded support for specific bundles. Often for chado-based importers it may make more sense to publish for all content types with for a specific chado base table.

Maybe it would be better to either:
1. Use a property instead of attribute to support dynamically populating the list. Although then I'd worry about it being changed between when we make the form and when the job is submitted 🙈 
2. Make the attribute keyed by value type so that more complicated look-up could be supported later. Right now we would only support the `bundle` key but a later PR could add support for the `chado_base_table` key :-)
   ```php
   #[TripalImporter(
     id: 'chado_gff3_loader',
     label: new TranslatableMarkup('Chado GFF3 File Loader'),
     description: new TranslatableMarkup('Import a GFF3 file into Chado'),
     publish: [
       'bundle' => [
         'gene',
         'mrna',
       ],
       'chado_base_table' => [
         'organism',
       ],
     ],
   )]
   ```

## ChadoImporterBase vs TripalImporterBase

I think more of this functionality is data backend agnostic then it is implemented here.

I would suggest 
- moving the `$publish_manager` property and its getter/setter into TripalImporterBase.
- making the TripalImporterForm::submitForm() process the `publish_` elements into a list of bundles to publish that gets passed when the job is created. That way ChadoImporterBase::postRun() doesn't have to do any processing and no other data backends will need to do any processing.
- create a `publishBundles` method in `TripalBackendPublishManager` that can handle publishing regardless of data backend. This is essentially the code you have in `ChadoImporterBase::postRun()` if we define the method as follows. Then `ChadoImporterBase::postRun()` literally becomes `$this->publish_manager->publishBundles(['chado_storage' => $run_args['publish']]);`
   ```php
   /**
    * Publish a number of bundles for specific publish instances.
    * 
    * @param array $bundles
    *   An associative array keyed by the id for the 
    *   TripalBackendPublish plugin instance (e.g. chado_storage)
    *   where the value is an array of bundles to be published.
    */
   public function publishBundles (array $bundles): void;
   ```

**Comment by dsenalik** [comment]
* Use a property instead of attribute = I went with option 2:
 * Make the attribute keyed by value type so that more complicated look-up could be supported later = commit https://github.com/tripal/tripal/pull/2221/commits/eb6532841cd0a912e700f3f2396abb7b42cd0e40
 * moving the $publish_manager property and its getter/setter into TripalImporterBase. = commit https://github.com/tripal/tripal/pull/2221/commits/edafa4c463759248d7ca9373d2a57de0768b8174
 * making the TripalImporterForm::submitForm() process the publish_ elements into a list of bundles to publish that gets passed when the job is created. = commit https://github.com/tripal/tripal/pull/2221/commits/8741d0c61be7b28974e79f23b9bf8b9101dbe411
 * create a publishBundles method in TripalBackendPublishManager that can handle publishing regardless of data backend. = commit https://github.com/tripal/tripal/pull/2221/commits/451510fe45646615fac391fc37822a9dcf60fbe9 although a bit different than your proposal.

File: tripal/src/TripalImporter/Attribute/TripalImporter.php [hunk 1]
```diff
@@ -70,6 +70,8 @@ class TripalImporter extends Plugin {
    *   necessary to require that the user provide a file for uploading if the
    *   loader keeps track of previous files and makes those available for
    *   selection.
+   * @param array|null $publish
+   *   An array containing a list of content types to be published after import.
```
laceysanderson: This needs to be updated now :-)

```suggestion
   *   An associative array indicating what Tripal Content should be published
   *   after import. It is keyed by the list type and the value indicates what
   *   to publish using that approach. List Types supported:
   *   - bundle (array): a list of bundle machine names to publish.
```

tripal/src/TripalImporter/TripalImporterBase.php
```diff
@@ -584,6 +602,16 @@ protected function setItemsHandled($total_handled) {
...
+   * Performs tasks after the importer has completed.
+   *
+   * @return void
+   *   No return value.
+   */
+  public function postRun() {
+
+  }
+
+    /**
```
laceysanderson: This somehow got mis-indented in this PR?

```suggestion
  /**
```

tripal/tests/src/Kernel/TripalImporter/TripalImporterFormBuildTest.php
```diff
@@ -200,6 +201,13 @@ public function testTripalImporterForm() {
     // a submit button.
     $this->assertArrayHasKey('button', $form,
       "The form should have a submit button since we indicated a specific importer.");
+    // Do not publish checkboxes.
+    $this->assertArrayHasKey('publish', $form,
+      "The form should include a pubish form element.");
```
laceysanderson:
```suggestion
      "The form should include a publish form element.");
```

tripal/tests/src/Kernel/TripalImporter/TripalImporterFormBuildTest.php
```diff
@@ -200,6 +201,13 @@ public function testTripalImporterForm() {
     // a submit button.
     $this->assertArrayHasKey('button', $form,
       "The form should have a submit button since we indicated a specific importer.");
+    // Do not publish checkboxes.
```
laceysanderson: I'm confused by this comment. Maybe you meant

```suggestion
    // Publish checkboxes.
```

tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
```diff
@@ -343,8 +351,10 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function __construct(array $configuration, $plugin_id, $plugin_definition,
-                              ChadoConnection $connection, ChadoBuddyPluginManager $buddy_manager) {
-    parent::__construct($configuration, $plugin_id, $plugin_definition, $connection);
+                              ChadoConnection $connection, TripalBackendPublishManager $publish_manager,
+                              ChadoBuddyPluginManager $buddy_manager) {
```
laceysanderson: Coding standards update because we changed the line :-)
```suggestion
  public function __construct(
    array $configuration,
    $plugin_id,
    $plugin_definition,
    ChadoConnection $connection,
    TripalBackendPublishManager $publish_manager,
    ChadoBuddyPluginManager $buddy_manager,
  ) {
```

tripal/src/Form/TripalImporterForm.php
```diff
@@ -119,6 +123,25 @@ public function buildForm(array $form, FormStateInterface $form_state, $plugin_i
...
 
+    // Adds a checkbox to indicate whether or not to publish the imported
+    // content. The default will be to publish (checked), uncheck to opt-out.
+    // An importer attribute can specify a list of bundles to publish,
+    // e.g. both 'gene' and 'mrna' for GFF importer.
+    if ($importer_def['publish']['bundle'] ?? FALSE) {
+      $bundles = (array) $importer_def['publish']['bundle'];
+      $form['publish'] = [
+        '#type' => 'fieldset',
+        '#title' => $this->t('Specify publishing behavior after importing'),
```
laceysanderson, line 134: We want to make sure it moves to just above the advanced fieldset.

```suggestion
        '#title' => $this->t('Specify publishing behavior after importing'),
        '#weight' => 8,
```

Screenshot 2025-09-23 at 10 51 40 AM

tripal_chado/src/TripalImporter/ChadoImporterBase.php
```diff
@@ -244,4 +244,33 @@ public function addAnalysis($form, &$form_state) {
...
+      if ($bundles) {
+        $publish_options = [
+          'chado_storage' => [
+            'schema_name' => $arguments['run_args']['schema_name'],
+            'bundles' => $bundles,
+          ],
+        ];
+        $this->publish_manager->publishBundles($publish_options);
+      }
+    }
```
laceysanderson: I suggest we switch this around a bit to ensure we don't sublit a publish job when no bundles are to be published. This will also help us if we decide to set the manager more globally.

```suggestion
    $arguments = $this->getArguments();
    $bundles = $arguments['run_args']['publish'] ?? [];
    if ($this->publish_manager and !empty($bundles)) {
      $publish_options = [
        'chado_storage' => [
          'schema_name' => $arguments['run_args']['schema_name'],
          'bundles' => $bundles,
        ],
      ];
      $this->publish_manager->publishBundles($publish_options);
    }
```

laceysanderson (changes requested):
There are code review comments in-line; however, the biggest issue is that only the GFF3 importer actually publishes now 🙈 This is because ChadoStorageImporterBase will only publish if the manager is set and none of the other importers set the manager...

I wonder if we can update the TripalImporterBase to set it in the constructor as long as we make the manager parameter optional? This way 
- all importers that do not override the create/constructor will get the manager, 
- any existing/new importers that do override but pass in the manager can have the parent set it.
- any existing importers that do override but do not pass in the manager do not end up with an error since the manager param is optional.

Note: We could also add a deprecation notice so that eventually we do not have to make the manager param optional. See https://www.drupal.org/about/core/policies/core-change-policies/how-to-deprecate#how-constructor-additions

tripal_chado/src/Plugin/TripalImporter/GFF3Importer.php
dsenalik: Couldn't commit suggestion - done separately

laceysanderson (commented):
Code changes look good :-) I'll do another manual test next.

laceysanderson (approved):
I've done manual testing on all of the importers and it all seems good. I confirmed

- publishes on all importers when default is not toggled
- does not publish when I select the toggle to turn off publishing
- fails gracefully when the content type doesn't exist but you ask to publish (e.g. publish gff3 but haven't imported genomic collection.
- fasta importer doesn't fail and doesn't publish --good because publishing is not enabled on this importer
- no errors in the drupal log

tripal/src/TripalImporter/TripalImporterBase.php
```diff
@@ -149,14 +216,33 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
     $this->prev_update = 0;
     $this->reported = 0;
 
-    // Initialize the logger.
-    $this->logger = \Drupal::service('tripal.logger');
+    // Initialize the Drupal messenger
+    if ($messenger === NULL) {
+      @trigger_error('Calling ' . __METHOD__ . '() without the $messenger argument is deprecated in tripal 4.0.0 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\Core\Messenger\Messenger from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```
laceysanderson:
```suggestion
      @trigger_error('Calling ' . __METHOD__ . '() without the $messenger argument is deprecated in tripal 4.0.0-alpha3 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\Core\Messenger\Messenger from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```

tripal/src/TripalImporter/TripalImporterBase.php
```diff
@@ -149,14 +216,33 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
...
+      $messenger = \Drupal::messenger();
+    }
+    $this->messenger = $messenger;
 
-    // Initialize messenger
-    $this->messenger = \Drupal::messenger();
+    // Initialize the Tripal logger.
+    if ($logger === NULL) {
+      @trigger_error('Calling ' . __METHOD__ . '() without the $logger argument is deprecated in tripal 4.0.0 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\Services\TripalLogger from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```
laceysanderson:
```suggestion
      @trigger_error('Calling ' . __METHOD__ . '() without the $logger argument is deprecated in tripal 4.0.0-alpha3 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\Services\TripalLogger from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```

tripal/src/TripalImporter/TripalImporterBase.php
```diff
@@ -149,14 +216,33 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
...
+      $logger = \Drupal::service('tripal.logger');
+    }
+    $this->logger = $logger;
 
     // Initialize file retrieval service.
-    $this->fileretriever = \Drupal::service('tripal.fileretriever');
+    if ($fileretriever === NULL) {
+      @trigger_error('Calling ' . __METHOD__ . '() without the $fileretriever argument is deprecated in tripal 4.0.0 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\Services\TripalFileRetriever from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```
laceysanderson:
```suggestion
      @trigger_error('Calling ' . __METHOD__ . '() without the $fileretriever argument is deprecated in tripal 4.0.0-alpha3 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\Services\TripalFileRetriever from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```

tripal/src/TripalImporter/TripalImporterBase.php
```diff
@@ -149,14 +216,33 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
...
+      $fileretriever = \Drupal::service('tripal.fileretriever');
+    }
+    $this->fileretriever = $fileretriever;
+
+    // Initialize the publish manager.
+    if ($publish_manager === NULL) {
+      @trigger_error('Calling ' . __METHOD__ . '() without the $publish_manager argument is deprecated in tripal 4.0.0 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\TripalBackendPublish\PluginManager\TripalBackendPublishManager from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```
laceysanderson:
```suggestion
      @trigger_error('Calling ' . __METHOD__ . '() without the $publish_manager argument is deprecated in tripal 4.0.0-alpha3 and it will be required in tripal:4.1.0. To resolve this, make sure the create() method in your importer grabs the Drupal\tripal\TripalBackendPublish\PluginManager\TripalBackendPublishManager from the container and supplies it to the parent in the constructor.', E_USER_DEPRECATED);
```

laceysanderson (commented):
Last minute thought --maybe we should have this as deprecated in the alpha since the message is being thrown now.

laceysanderson (approved):
Changes look great and the change record is now merged so this is ready to go! This allowed me to confirm the URL added to the deprecations in this PR and they work great!

PR #2227, opened 2025-07-05
Some widget fields should not default to HTML formatting

### LINKED ISSUE #2226: Some widget fields should not default to HTML formatting
### Task Description

Some widget fields default to HTML formatting. This was added by PR #1862, fields based on a text type of database column can be configured with various levels of HTML formatting when editing the field in a widget.
The Chado String Widget does not have any settings to support HTML.
The Chado Text Widget does, and it will default to "Basic HTML".
In many cases, this default is overkill, and we should set the default in the appropriate yaml files.
Example of how to do this
https://github.com/tripal/tripal/blob/9797278ceb7eeb040bd125c3d13d3e695219be83/tripal_chado/config/install/tripal.tripalfield_collection.expression_chado.yml#L33-L34

So we can avoid this
Image

Likewise property fields seem to default to "Basic HTML" with three lines, so in some cases adding this may be appropriate:
```
              settings:
                filter_format: plain_text
                num_rows: 1
```
and I am thinking that for discovery, this should be the new default.

### Branch Name

tv4g1-issue2226-yaml-form-text-field-defaults

### List of fields to change

Pub: DOI, journal abbreviation, EISSN, Publication Date, ISSN, Language Abbr, Publication Type, Publication Model, Language, ELocation, URL, Journal Country, PII?
Array Design: Array Version, Array Dimensions, Element Dimensions
Assay: Array Identifier, Array Batch Identifier

dsenalik:
# New Feature

## Description
The widget form settings following fields have been updated from Basic_HTML to plain_text, 1 row
Pub: DOI, journal abbreviation, EISSN, Publication Date, ISSN, Language Abbr, Publication Type, Publication Model, Language, ELocation, URL, Journal Country, PII
Array Design: Array Version, Array Dimensions, Element Dimensions
Assay: Array Identifier, Array Batch Identifier

A setting has been added to the TripalTextTypeWidget for number of rows for the textarea. This is the same setting that the ChadoPropertyTypeDefault field already has.
new_text_setting

This setting is now set to 1 for some fields
Biosample name, Array design name, Assay name, Analysis Data Source URI, Protocol URI, 

At this point I will not add an update hook for existing sites, as I don't want to risk messing with existing settings unexpectedly.

 - [x] For discussion: Should the chado property field discovery continue to default to HTML or should it default to plain text, 1 row? Gathertown discussion July 7: we will change the default for discovery from "Chado Property: Long Text" to "Chado Property: Short Text" (ChadoPropertyStringWidgetDefault)

## Testing?
Build a docker on this branch.
Import Expression type collection
Try adding Pub, Assay, or Array Design and note that there are no more large fields for simple values.
Image
becomes
new_array_version_field

Discovered and updated properties should default to
pub_doi


laceysanderson (changes requested):
The existing changes look good in code review and reflect in the site as I expect 😍 

That said, it would be good to make the following text fields also plain text with 1 row:

- [x] protocol_instrument
- [x] protocol_software
- [x] pub_volumetitle
- [x] germplasm_unique_name
- [x] breeding_cross_unique_name
- [x] germplasm_variety_unique_name
- [x] ril_unique_name
- [x] gene_uniquename
- [x] mrna_uniquename
- [x] dna_library_uniquename
- [x] genome_annotation_uri
- [x] genome_assembly_uri
- [x] qtl_unique_name
- [x] sequence_variant_unique_name
- [x] genetic_marker_unique_name
- [x] phenotypic_marker_unique_name

laceysanderson (approved):
This looks perfect now. I did a code review and looked at the add content page for all the content types ✅ MUCH Better!

PR #2233, opened 2025-07-10
Escape special characters in usernames that cause issues in generated drush commands

### LINKED ISSUE #2232: Generated drush commands must escape spaces in usernames
### Tripal Version

4.x

### Drupal Version

10

### PostgreSQL Version

17

### PHP Version

8

### General Description

When submitting jobs that provide a drush command to run, for instance when publishing a content type, if the user's username has spaces in it, the generated command also includes these spaces. These spaces, as can be expected, wreak all kinds of havoc on the command line.

### Steps to reproduce

As an admin, have a username that has spaces in it such as "John Doe". The command that Tripal generates for the user to run will have spaces:
`drush trp-run-jobs --job_id=13 --username=John Doe --root=/var/www/html/drupal_test/wwwroot`
which will of course result in syntax errors due to "Doe".

As per Drupal's "Add user" page, the special characters allowed in a username are **space, period (.), hyphen (-), apostrophe ('), underscore (_), and the @ sign.**

In my testing, the apostrophe (') also causes trouble in generated commands.

### Error messages

```shell
If the username has a space, the error is
No arguments expected for "trp-run-jobs" command, got "doe". 

If the username has an apostrophe, the command includes an unpaired apostrophe and the terminal will await further input after hitting return.
```

### Screenshots

_No response_

---
Branch: [tv4g3-issue2232-generated_drush_commands_must_escape_spaces_in_usernames](https://github.com/tripal/tripal/tree/tv4g3-issue2232-generated_drush_commands_must_escape_spaces_in_usernames)

Ferrisx4:
# Bug Fix

## Description

This fix will escape special characters in usernames when generating drush commands which ordinarily break/cause issues in the command line.
The fixed characters are space `( )` and apostrophe `(')`.

## Testing?

Change your username to include either a space or an apostrophe (or both). Then create a new Tripal job, such as publishing a Tripal content type. The generated command, when run in the terminal, will be interpreted correctly and will execute the job.


tripal/src/Services/TripalJob.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\Core\Url;
 use Drupal\Core\Link;
+use Exception;
```
dsenalik:
```suggestion
```
I ran phpcs and it says that this is unused. Was there a reason you added it? Same with the one in the other file...

tripal/src/api/tripal.jobs.api.php
```diff
@@ -3,6 +3,7 @@
 use Drupal\Core\Link;
 use Drupal\Core\Url;
 use Drupal\tripal\Services\TripalJob;
+use Symfony\Component\HttpFoundation\RedirectResponse;
```
dsenalik:
```suggestion
```
same as other comment

dsenalik (changes requested):
I approve except for those two use statements that appear to be not needed.

I created a new admin user `Signore@Molto_molto-molto' Particolare2025`

The generated drush command
`drush trp-run-jobs --job_id=1 --username=Signore@Molto_molto-molto\'\ Particolare2025 --root=/var/www/drupal/web`
ran perfectly

tripal/src/Services/TripalJob.php
Ferrisx4: I'm on board with these two suggested changes. I haven't the foggiest idea why those two use statements are there. Very mysterious...

dsenalik (approved):
Perfect, approved.

Ferrisx4:
Thanks for the review and merge @dsenalik, or should I call you Signore@Molto_molto-molto' Particolare2025?
😄

PR #2231, opened 2025-07-10
Updating our repo to use Qlty Cloud instead of CodeClimate

laceysanderson:
# Tripal 4 Core Dev Task 

## Description

CodeClimate has actually split into two companies: one which provides the quality functionality we use (QLTY Cloud) and velocity which is more for businesses tracking progress (CodeClimate). We have started getting a failure on our PRs since CodeClimate is being phased out. This PR will migrate us to QLTY Cloud and thus remove this failure.

## NOTES

- The maintainability has gone down from A to C. This is expected and valid 🙈 although I did try to make it a lot less stringent so we are not so overwhelmed.
- Code coverage is not showing up because 4.x doesn't yet have coverage on QTLY Cloud for it to make comparisons to. This means we can't get an idea of what it will look like until this is merged. I did confirm that the reports are being pushed to QltyCloud :-)
- I will want to add more checks in, especially regarding security but we'll wait for another PR in order to get the coverage working again and stop failing all PRs.

## Testing?

1. Check the [Tripal QLTY Cloud page](https://qlty.sh/gh/tripal/projects/tripal) and confirm that the Maintainability is evaluated.
2. Confirm that all the workflows and tests are passing on this PR.
3. Confirm that configuration for CodeClimate has been removed.
4. Confirm that the readme has been updated to mention QltyCloud instead of code climate.

PR #2234, opened 2025-07-11
Starting configuring plugins for QLTY Cloud

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This is a follow-up to PR #2231 with a focus on enabling plugins to further check our code.

### Ena ble the following plugins and configure.

- [x] YMLlinter to check the YAML files.
- [x] ActionLint to check our github workflows for security issues.
- [x] checkov to check our dockerfile for security concerns.

### Fix the issues brought up by the above plugins in non-php files.

- [x] YMLlinter to check the YAML files.
- [x] ActionLint to check our github workflows for security issues.
- [x] checkov to check our dockerfile for security concerns.

The following lists the fixes that were made:
- modify all YAML files to quote strings with single quotes and strip any trailing whitespace.
- add `permissions: read-all` to our workflows to ensure they do not default to write permissions.
- Remove SIMPLETEST variables from our workflows since we don't use them anymore.
- Add HEALTHCHECK to our dockerfiles to ensure that the container is functional.
- install the Qlty CLI in our dockerfiles so that we can check problems on the command-line.

NOTE: The checkov plugin also tells us we should use a non-root user and not store basic credentials but these are medium concerns and not easily fixed so I've triaged them.

## Testing?

All of the changes are either in the YAML or Dockerfiles. Additionally, I believe they should all be covered via automated tests. That said, it might be good to check some of the routing as it would have the least test coverage.


tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldType/ChadoPropertyWidgetForm-TestInfo.yml
```diff
@@ -1,87 +1,88 @@
...
-          value: "Simple Project"
+          value: 'Simple Project'
         project_description:
         - record_id: 0
           value:
-            format: basic_html
-            value: "This is a very simple project. We haven't even added any metadata to expand it's purpose."
+            format: 'basic_html'
+            value: |
+              This is a very simple project. We haven't even added any metadata to expand it's purpose.
```
dsenalik, line 77: Nothing wrong, marking just because I am curious about this one. A litter farther down it is double quoted.

dsenalik (approved):
I scrolled all the way through those changes 😵‍💫, looks good.

Built a docker and tried various things. Add analysis, unpublish it, publish it, import taxon with taxonomy importer, publish organism, check for new fields, import publications, populate a mview, edit tripal settings.
Looks great!

PR #2251, opened 2025-08-01
Make Publication type property field unlimited cardinality

### LINKED ISSUE #2250: YAML for field publication type should be unlimited cardinality
### Task Description

The publication importer will sometimes add more than one type for a publication, e.g.
```
 type_id |              property_value              |   cvterm_name    
---------+------------------------------------------+------------------
     109 | Journal Article                          | Publication Type
     109 | Research Support, Non-U.S. Gov't         | Publication Type
     109 | Research Support, U.S. Gov't, Non-P.H.S. | Publication Type
```
so the cardinality for this field should be unlimited.
https://github.com/tripal/tripal/blob/54ceebeaae10bed1af4c462a3e0d8c84ff8ba5ac/tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml#L1909-L1914

### Branch Name

tv4g1-issue2250-publication-yaml-for-type

dsenalik:
# Tripal 4 Core Dev Task

## Description
The publication importer often adds more than one publication type **property** to an individual publication.
For example for PMID [30333150](http://www.ncbi.nlm.nih.gov/pubmed/30333150)
```
 type_id |              property_value              |   cvterm_name    
---------+------------------------------------------+------------------
     109 | Journal Article                          | Publication Type
     109 | Research Support, Non-U.S. Gov't         | Publication Type
     109 | Research Support, U.S. Gov't, Non-P.H.S. | Publication Type
```
Currently only one of these will be published, because the field is cardinality 1. No warning will be shown that some were not published (until PR #2203 is merged - then you will see the warning)

The simple fix is to make it unlimited cardinality.

Note that the `pub` table also has a `type_id` column (not null) which has a field "Publication Type". This pair of fields is inherited from Tripal 3, but to avoid confusion, let's rename the property field to "Additional Publication Type", the same as we do when we have two dbxref fields.

I decided not to do an update hook for this.

However, it did seem appropriate to adjust the publication importer to remove duplicated information, so now the first publication type, whether there is just one, or there are more, is used to populate `type_id` in the pub table. This is the same behavior as before. With this PR, this type is **not** saved as a property, but any **additional** publication types are saved as properties.

If you want to remove duplicates in an existing site, this should do it because the first property is rank 0 (but be careful, newly imported records will now start with rank 0):
`DELETE FROM pubprop WHERE rank=0 AND type_id=(SELECT cvterm_id FROM cvterm WHERE name='Publication Type' AND cv_id=(SELECT cv_id FROM cv WHERE name = 'tripal_pub'));`

## Testing?

1. Import publication with PMID [30333150](http://www.ncbi.nlm.nih.gov/pubmed/30333150)

2. Publish publications

3. View the publication

**On 4.x**
pub_type_1
...
pub_type_1

**On this branch**
PR2251_20250814


tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
```diff
@@ -1514,6 +1514,33 @@ fields:
               region: content
               weight: 12
 
+    -   name: 'pub_publication_type'
+        content_type: 'pub'
+        label: 'Additional Publication Type'
```
laceysanderson: Since this is an unlimited field, shouldn't it be optionally plural?
```suggestion
        label: 'Additional Publication Type(s)'
```

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
```diff
@@ -1514,6 +1514,33 @@ fields:
...
+        type: 'chado_property_type_default'
+        description: 'Additional descriptors indicating the type of publication'
```
laceysanderson: What about something like this to make it more clear.
```suggestion
        description: 'Additional terms further describing the type of publication. This should not include the primary term which is indicated in the "Publication Type" field.'
```

laceysanderson (approved):
Code-wise this looks good! Just a few picky suggested improvements in wording.

Manual testing worked exactly as described and I like that the two types are now located right beside each other in the form!
Screenshot 2025-09-16 at 4 48 35 PM
Screenshot 2025-09-16 at 4 49 28 PM

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
dsenalik: Nice improvement

PR #2260, opened 2025-08-15
Allow core fields to work on custom tables

### LINKED ISSUE #2253: Cannot use core chado fields on new content types based on custom tables
### Task Description

When trying to add a core chado field, e.g. ChadoTextTypeDefault, to a new content type based on a custom table, an error like this occurs
`Undefined array key "primary key" ChadoTextTypeDefault.php:61`
This happens because the custom table is not in the schema returned earlier by
`$base_schema_def = $schema->getTableDef($base_table, ['format' => 'Drupal']);`

By default, the schema is either cached or **read from a static file** here
https://github.com/tripal/tripal/blob/54ceebeaae10bed1af4c462a3e0d8c84ff8ba5ac/tripal_chado/src/Database/ChadoSchema.php#L17-L20
so new custom tables are not present in this schema.

Changing how that works is a massive undertaking, but we can have a ChadoFieldItemBase function to retry looking up from the database instead of file.

We have a similar problem publishing, so a similar function can be added to Chado storage.

### Branch Name

tv4g1-issue2253-core-fields-on-custom-tables

dsenalik:
# Bug Fix

## Description

This makes some significant changes to fully support using custom chado tables for both creating new content types, and support for fields linking from or to custom tables.
This is being developed in conjunction with the porting of the tripal_file module, PR #2273.

 - [x] To do: materialized view tables can now be found for field discovery, but if they don't have a primary key, e.g. analysis_organism, you have an error. Fix this https://github.com/tripal/tripal/pull/2260/commits/4fe9e9d5e90885dd01f3c3f757ddfc40a10fc0ef

## Testing?

I fleshed out the unit tests for custom tables, so that will no longer be marked incomplete.
100% coverage on ChadoCustomTableManager class.
87% coverage on ChadoCustomTable class, only unexpected database errors and similar are not covered.
I included test coverage of all of three 'source' modes for getTableDef() in TripalDBX on both a "real" chado table and a custom one.

This was needed to port the tripal_file module, so if you want to do manual testing it is best done from that branch `tv4g1-issue1542-tripal-file`.


dsenalik:
back to draft because all the linker fields need to be updated

reynoldtan:
Reviewed and confirmed items outlined in the test documentation:

- [x] All @todo test comments in ChadoCustomTableTest.php have been resolved.
- [x] 100% test coverage of the service class ChadoCustomTableManager, including test for create(), loadById(), loadByName(), findByName() and getTables() methods.
- [x] getTableDef() method has been thoroughly tested across data sources/modes - file, tripal and database - covering both Chado and custom table scenarios.
- [x] All tests in ChadoCustomTableTest are passing cleanly.
Screenshot 2025-09-25 at 11 04 29 AM

tripal_chado/src/ChadoCustomTables/ChadoCustomTable.php
```diff
@@ -152,82 +161,87 @@ public function setLocked($lock = False) {
...
+    $query = $public->select('tripal_custom_tables', 'tct');
     $query->fields('tct', ['schema']);
     $query->condition('tct.table_name', $this->table_name);
     $query->condition('tct.chado', $this->chado_schema);
     $table_schema = $query->execute()->fetchField();
     if (!$table_schema) {
       return [];
     }
-    return unserialize($table_schema);
+    return unserialize($table_schema, ['allowed_classes' => FALSE]);
```
laceysanderson, line 215: Was there a bug that caused adding this option? It may be good to document the case that made this necessary if there was one that popped up.

tripal_chado/src/api/ChadoSchema.php
```diff
@@ -218,7 +218,7 @@ public function getTableNames($include_custom = FALSE) {
     // now add in the custom tables too if requested
     // @todo change this to the variable once custom tables are supported.
     if (FALSE) {
-      $sql = "SELECT table FROM {tripal_custom_tables}";
+      $sql = "SELECT table FROM [tripal_custom_tables]";
       $resource = $this->connection->query($sql);
```
laceysanderson: Don't we need to make sure we actually get here? And I think it's best to use 0 here to indicate Drupal rather then the older [ ] approach :thinking:

That said, this is the old ChadoSchema class that will be deprecated. Not the new one that is part of Tripal DBX.

```suggestion
    // Now add in the custom tables too if requested.
    if ($include_custom) {
      $sql = "SELECT table FROM {0:tripal_custom_tables}";
      $resource = $this->connection->query($sql);
```

tripal_chado/src/api/ChadoSchema.php
```diff
@@ -307,7 +307,7 @@ public function getTableSchema($table) {
    */
   public function getCustomTableSchema($table) {
 
-    $sql = "SELECT schema FROM {tripal_custom_tables} WHERE table_name = :table_name";
+    $sql = "SELECT schema FROM [tripal_custom_tables] WHERE table_name = :table_name";
```
laceysanderson:
```suggestion
    $sql = "SELECT schema FROM {0:tripal_custom_tables} WHERE table_name = :table_name";
```

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -1170,4 +1179,74 @@ protected static function getColumnTermId(string $table, string $column, string
...
+   *
+   * @param $schema
+   *   The schema to query for the table.
+   * @param string $table_name
+   *   The table name.
+   *
+   * @return array
+   *   The table schema.
+   */
+  public static function getChadoTableDef($schema, string $table_name): array {
```
laceysanderson: I think the schema needs to be a second, optional parameter. I have in mind some caching changes to the field that may make passing it around so much less ideal and if that happends then its easier to deprecate a second optional parameter ;-p. Additionally, it would be great if we could get the table schema of more then one table with a single call since we have often dealing with more then one.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -1170,4 +1179,74 @@ protected static function getColumnTermId(string $table, string $column, string
...
+   *
+   * @param Drupal\tripal_chado\Database\ChadoSchema $schema
+   *   The chado schema definition.
+   * @param string $table_name
+   *   The chado table to look up the primary key for.
+   *
+   * @return string
+   *   The table primary key name.
+   */
+  protected static function getPrimaryKey(ChadoSchema $schema, string $table_name): ?string {
```
laceysanderson: Same as above, it would be good to move the schema as the second optional parameter and make this support more then one call at a time. Think: `getPrimaryKeys(array $table_names, ?ChadoSchema $schema)` with a return value of primary key names keyed by the table they are the primary key for.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -1170,4 +1179,74 @@ protected static function getColumnTermId(string $table, string $column, string
...
+   *   The name of the table the foreign key resides in. E.g. 'feature' for
+   *   the feature.type_id => cvterm.cvterm_id foreign key.
+   * @param string|null $right_table
+   *   The name of the table the foreign key refers to. For the example
+   *   above it would be cvterm. If NULL, then all foreign keys are returned.
+   *
+   * @return array
+   *   The the foreign key definition, or an empty array if none exists.
+   */
+  public static function getChadoForeignKeyDef($schema, string $left_table, ?string $right_table = NULL): array {
```
laceysanderson: For this one we still need to switch the schema to the last option, but allowing multiple tables doesn't make as much sense. I'd like to see a future PR allow people to pull all the keys/table defs with one static method that uses these functions but that's a task for another day. :zany:

Edited to add: self:getLinkerTables() may add a case for allowing multiple tables :thinking:

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -2209,4 +2209,44 @@ protected function isEmptyRecord($record) {
...
+   *
+   * @param $schema
+   *   The schema to query for the table.
+   * @param string $table_name
+   *   The table name.
+   *
+   * @return array
+   *   The table schema.
+   */
+  public function getChadoTableDef(ChadoSchema $schema, string $table_name): array {
```
laceysanderson: Definitely don't pass the schema here at all but rather use $this->connection->schema() in these methods or add another property to the class for the schema.

Leaving these singular feels fine as it's not often we are processing more then one table in chado records.

tripal_chado/src/TripalStorage/ChadoRecords.php
```diff
@@ -2209,4 +2209,44 @@ protected function isEmptyRecord($record) {
...
+   *
+   * @param Drupal\tripal_chado\Database\ChadoSchema $schema
+   *   The chado schema definition.
+   * @param string $table_name
+   *   The chado table to look up the primary key for.
+   *
+   * @return string|null
+   *   The table primary key name.
+   */
+  public function getPrimaryKey(ChadoSchema $schema, string $table_name): ?string {
```
laceysanderson: Same comment as above: no passing in schema here.

Leaving these singular feels fine as it's not often we are processing more then one table in chado records.

laceysanderson (changes requested):
Results of code review. This PR is awesome :heart: excited for the improvements.

tripal_chado/src/TripalStorage/ChadoRecords.php
dsenalik: Updated in commit cae987409

tripal_chado/src/TripalStorage/ChadoRecords.php
dsenalik: Updated in commit cae987409

tripal_chado/src/api/ChadoSchema.php
dsenalik: Can't commit suggestion "Applying suggestions on deleted lines is currently not supported." but I will add this with commit c60ee0b

tripal_chado/src/ChadoCustomTables/ChadoCustomTable.php
dsenalik: comment added with commit https://github.com/tripal/tripal/pull/2260/commits/1f2d779d6c79cc45efed64840d08151845d41baa

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: After discussion we opted for two functions, `getChadoTableDef()` and `getChadoTableDefs()` to handle one or multiple tables, respectively.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: After discussion we opted for two functions, `getPrimaryKey()` and `getPrimaryKeys()` to handle one or multiple tables, respectively.

tripal_chado/src/TripalField/ChadoFieldItemBase.php
dsenalik: We can introduce `getChadoForeignKeyDefs()` in the future if we so decide, similar to `getPrimaryKey()` + `getPrimaryKeys()`

tripal_chado/src/api/ChadoSchema.php
dsenalik: I have reverted this to never run, because tests are failing, but it's not worth the time to figure out why for a deprecated function that never executed anyway

tripal_chado/src/TripalField/ChadoFieldItemBase.php
laceysanderson: Thanks for all the changes! 💞

tripal_chado/src/TripalField/ChadoFieldItemBase.php
laceysanderson: Sounds good!

laceysanderson (approved):
✅ All my changes have been addressed and the code looks great.
✅ Automated testing is awesome.
✅ I manually tested entering a number of content types through the UI to test that the fields are still working perfectly on our core content types and they are.

Image

Ready to merge! Thanks @dsenalik ❤️ 🎉

PR #2262, opened 2025-08-18
Tv4g10 issue2195 tripal content views inefficient

### LINKED ISSUE #2195: Views of Tripal Content are slow and inefficient
### Tripal Version

4.x 

### Drupal Version

10.3.1

### PostgreSQL Version

16.2

### PHP Version

8.3.10

### General Description

I am using Drupal views to create an "Organism Glossary", a paginated list of specifically organisms with the columns `Scientific name`, which uses the id field to link to the content and combines the genus and species fields (species hidden), as well as the common name:

Image

The configuration looks as such:

Image

We have noticed significant slowness when this view is being generated (uncached) with load times upwards of 30 seconds and occasionally beyond the default 60 second timeout. When using Views' option to display the query, we noticed an exceedingly inefficient query:

[View queries.txt](https://github.com/user-attachments/files/19851923/View.queries.txt)

It seems to be doing the same queries and subqueries over and over, and a lot do not seem to be directly responsible for creating the view.
I suspect that there is some miscommunication between Drupal/Drupal Views and Tripal content that is causing these queries to be so complex and repetitive, but I am not sure what mechanism is responsible for this.

Edit:
We are also testing an Elasticsearch module to work with Tripal 4 and we are seeing positive results. Search is functional, and is quick to index the Tripal content. When querying the Elasticsearch server itself, results are also very quick. However, this module utilizes Drupal Views and, similar to the Organism Glossary above, we are seeing basically identical slowness caused by repetitive queries and subqueries. 

### Steps to reproduce

Create a view with the above configurations, and have Drupal Views' configuration set to show the full SQL queries.

Issue: [tv4g10-issue2195-tripal_content_views_inefficient](https://github.com/tripal/tripal/tree/tv4g10-issue2195-tripal_content_views_inefficient)

Ferrisx4:
# Bug Fix

## Description

This change reroutes some database calls by checking the cache first. Even on basic views of contents, such as a paginated organism glossary that lists the Common Name, Genus, Species, and a link to the organism (20 per page), this will avoid tens of thousands of database calls.

*There are still additional changes to the basic Tripal content loading, especially for views. While the cache is much faster than raw database calls, we should still try to avoid tens of thousands of calls to the cache for such a simple page. However, for the time being, I am looking to merge these relatively simple changes to open the door to supporting Elasticsearch.
## Testing?

Create a new view on your site by visiting `admin/structure/views/add`, and recreate the View described in the [original issue](https://github.com/tripal/tripal/issues/2195), or any regular Drupal view featuring Tripal content. When viewing the freshly made View, compare the speed of page generation with this branch against the 4.x branch, making sure to rebuild the cache each time. The page generation should be measurably and noticeably faster.


dsenalik:
I tried a docker on Drupal 11.2 and couldn't test because for some reason the "Tripal Content" view type is not listed
no_tripal_entity

Update: **Also happens on 4.x branch** 😱 so not related to this PR. Made a new issue #2263

Ferrisx4:
Just saw your update, I can also confirm it working on 10 and not 11 :(

tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
```diff
@@ -422,6 +427,9 @@ public function getTerm($accession, $options = []) {
       }
     }
 
+    // Cache the term for an hour.
+    \Drupal::cache()->set($cache_id, $term, \Drupal::time()->getRequestTime() + (3600));
```
dsenalik, line 435: FWIW this is what is triggering the testing errors.

tripal/src/TripalDBX/TripalDbxSchema.php
```diff
@@ -628,7 +628,18 @@ public function primaryKeyExists(
     // If they didn't supply the column, then we can look it up.
     if ($column === NULL) {
       $parameters = ['source' => 'database', 'format' => 'Drupal',];
-      $table_schema = $this->getTableDef($table, $parameters);
+
+      // Cache the table schema so we don't fet it thousands of times.
```
dsenalik:
```suggestion
      // Cache the table schema so we don't fetch it thousands of times.
```

tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
dsenalik: Thanks @laceysanderson

dsenalik (approved):
I tested with a view with 31 organisms and the difference is quite noticeable, thanks @Ferrisx4 
And thanks to @laceysanderson for the coding standards.
Looks good to me, pending tests passing of course!

dsenalik (approved):
Code review of new changes looks good! And still functions as expected.

laceysanderson (approved):
I also tested this using a view of ontology terms and the speed increase was noticeable and appreciated ❤️ 

Code-wise this also looks good. I think this is ready to be merged now that the tests are passing. Thanks @Ferrisx4!

PR #2239, opened 2025-07-13
Support Drupal 11.2, remove 10.3 and 11.0 - Part 2 of 2

dsenalik:
# Tripal 4 Core Dev Task

## Description

Add support for Drupal 11.2 - this was released in June so making this high priority.

Removes support for Drupal 10.3 and 11.0 because security updates for these versions ended in June 2025 when Drupal 11.2 was released.

This will have to be a two-part PR, because we need to have the docker image in place to be able to run tests.

 - [x] @laceysanderson will have to change the required versions for testing to allow the update of three of them
 - [x] Fix the four errors = commit https://github.com/tripal/tripal/commit/be222ffa0fe44e6fd6214d303919a88895566613
 - [x] Fix the deprecations (Epic Task!) - Most deprecations are mentioned in issue #2209
 - [x] `Error: Unknown named parameter $valid_tokens in ReflectionAttribute->newInstance()` - Add a new annotation class to support the `valid_tokens` attribute = commit https://github.com/tripal/tripal/pull/2239/commits/10719a0365d19f0911b8c1272ca48ed8811cf813
 - [x] Attribues are supported since Drupal 10.2, so when everything else is fixed, remove the annotations! For backward compatibility, the annotation classes should remain in case anyone has used only annotations in their module, such as for an importer. Note: Some removals caused failures on all 10.x versions, so they are retained, with `@todo` notes to remove them when we no longer support Drupal 10.x. See commits https://github.com/tripal/tripal/pull/2239/commits/ba5087b56b2dab6d2349fa8f8ee29c9233b4fff9 and https://github.com/tripal/tripal/pull/2239/commits/d3eafa5a99b522e11fc772b7362a59db5181a384

 - [x] The deprecation `The update to add a default table CSS class for view "chado_mviews" is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Profile, module and theme provided configuration should be updated.  See` https://www.drupal.org/node/3499943 took a long time to fix. Simply adding a placeholder as I did in commit https://github.com/tripal/tripal/pull/2239/commits/747ce4494b4a10df7887a744d03d2bb9e12f50c8 breaks every version before 11.2 because of a schema error. `Exception: Exception when installing config for module tripal_chado, message was: Schema errors for views.view.chado_custom_tables with the following errors: views.view.chado_custom_tables:display.default.display_options.style.options.class missing schema` 
[...]

PR #2264, opened 2025-08-21
Fix missing TripalEntityViewsData attribute

### LINKED ISSUE #2263: "Tripal Content" view type missing on Drupal 11.2
### Tripal Version

4.x

### Drupal Version

11.2

### PostgreSQL Version

17

### PHP Version

8.3

### General Description

"Tripal Content" does not show up in the list when making a view through the UI. Seems to be 11.2-specific, 11.1 is okay

### Steps to reproduce

Structure -> Views -> +Add View 
The select list under "View Settings" "Show" is missing "Tripal Content"
no_tripal_entity

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
One annotation to attribute conversion was missed in PR #2239 
The result is that on Drupal 11.2 you can't make a view for any tripal content because the select item is missing.
This adds the attribute and fixes the bug.

## Testing?
Build a Drupal 11.2 docker
Structure -> Views -> +Add View
The select list under "View Settings" "Show" should now include "Tripal Content".

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -14,6 +14,7 @@ use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Component\Utility\Xss;
 use Drupal\user\UserInterface;
 use Drupal\tripal\Access\TripalEntityAccessControlHandler;
+use Drupal\tripal\Entity\TripalEntityViewsData;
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
Tested it and it works great. Even created some test content to make sure that it was pulling the data properly.

Image
image

PR #2266, opened 2025-08-22
Create tripal-specific attribute classes

### LINKED ISSUE #2255: Create tripal-specific attribute classes for field type and field widget
### Task Description

PR #2239 Introduced an attribute class for field formatters, to support the `valid_tokens` attribute, which is the list of valid tokens for an entity title.
For consistency and good practices, we should create tripal-specific classes for the type and widgets too, currently they are using the base Drupal attribute classes.

### Branch Name

tv4g1-issue2255-tripal-specific-attribute-classes

dsenalik:
# Tripal 4 Core Dev Task

## Description

PR #2239 Introduced an attribute class for field formatters, to support the `valid_tokens` attribute, which is the list of valid tokens for an entity title.
For consistency and good practices, we should create tripal-specific attribute classes for the type and widgets too, currently they are using the base Drupal attribute classes.
With this PR, all tripal and tripal_chado fields now use these tripal attribute classes.

## Testing?

There should be no functional changes, as the new attribute classes do not yet add any functionality. Mainly testing is making sure nothing has unexpectedly broken.

### DIFF
--- a/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalBooleanTypeFormatter.php
+++ b/tripal/src/Plugin/Field/FieldFormatter/DefaultTripalBooleanTypeFormatter.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\tripal\Plugin\Field\FieldFormatter;
 
-use Drupal\Core\Field\Attribute\FieldFormatter;
+use Drupal\tripal\TripalField\Attribute\TripalFieldFormatter;
 use Drupal\Core\Field\FieldItemListInterface;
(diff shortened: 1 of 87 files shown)


laceysanderson (approved):
✅ Code-wise these simple though widespread changes look good 👍 
✅ I added a number of organisms, genes, germplasm and projects as well as editing them to check that fields seem to be fine. All looks good.

Note: There are a lot of properties in the field types and widgets that we will want to move into the annotation but that can wait until another PR :-)  see Issue #2276

Thanks @dsenalik!

PR #2271, opened 2025-09-03
Editing single-hop property field fixed

### LINKED ISSUE #2269: Bug with the "Remove" button on any property field with unlimited cardinality
### Tripal Version

4.x

### Drupal Version

11.1, 10.4

### PostgreSQL Version

17

### PHP Version

8.3

### General Description

There is a bug somewhere in the property field widget when dealing with the "Remove" button. 

Branch: `tv4g1-issue2269-property-remove-button`

### Why

The property field is a "single hop" field. The property table has a foreign key column to the pkey of the base table.
The 'linker_id' property therefore is still getting assigned a value even for a deleted record, because it is the pkey of the base table, rather than the pkey of a linker table.
And because of that value, when `setPropValues()` is called following the actual update of the property table, there is only a single condition, and as a result it finds some random property linked to the base table.

### Steps to reproduce

 - Import the germplasm type collection
 - Create an organism 

This bug affects **any property field of unlimited cardinality**
 - Create a germplasm collection, and when creating, add three entries for germplasm collection type
Image

 - Edit it and hit the "Remove" button on the middle one, and save.
 - Chado is updated correctly. However the Drupal field table has the third entry as a duplicate of the first one.
Image

```
select * from stockcollectionprop;
 stockcollectionprop_id | stockcollection_id | type_id | value | rank 
------------------------+--------------------+---------+-------+------
                      4 |                  1 |     228 | abc   |    0
                      5 |                  1 |     228 | ghi   |    1
(2 rows)

select * from tripal_entity__germcollection_addtltype;
     bundle     | deleted | entity_id | revision_id | langcode | delta | germcollection_addtltype_record_id | germcollection_addtltype_prop_id | germcollection_addtltype_linker_id | germcollection_addtltype_value | germcollection_addtltype_rank | germcollection_addtltype_type_id 
----------------+---------+-----------+-------------+----------+-------+------------------------------------+----------------------------------+------------------------------------+--------------------------------+-------------------------------+----------------------------------
 germcollection |       0 |         3 |           3 | und      |     0 |                                  1 |                                4 |                                  1 | abc                            |                             0 |                              228
 germcollection |       0 |         3 |           3 | und      |     1 |                                  1 |                                5 |                                  1 | ghi                            |                             1 |                              228
 germcollection |       0 |         3 |           3 | und      |     2 |                                  1 |                                4 |                                  1 | abc                            |                             0 |                              228
(3 rows)
```
This bug does **not** affect the dbxref field
 - Create a entity of any of the germplasm content types. When creating, add three database cross references
 - Edit it and hit the "Remove" button on the middle one, and save 
 - The record is removed from both the drupal table and from chado

This bug hits also if you just remove the contents of one field. Here I removed the contents of the middle item and saved.

Image

dsenalik:
# Bug Fix

## Description

The changes here fix a bug when editing unlimited cardinality properties on any content type.

The fix for the "Remove" button involves updating the ChadoRecords setColumnValue() function to indicate when we are clearing a value in deleteRecord(), don't set the `has_values` flag.

The fix for deleting the property value by clearing the field in the widget involves knowing the main property name for a field. If the value for this is either NULL or an empty string, we remove this delta in TripalEntity preSave even though other properties may have values.

## Testing?

Test on any field that has properties with unlimited cardinality. A handy field to test with is the Germplasm Collection in the "Germplasm Content Types" collection. Both the "Germplasm Collection Type" and "Description" properties are currently unlimited cardinality.
Create several properties and save. Then edit, and remove the property with the "Remove" button, and by clearing the value from the field.

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -878,6 +878,33 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
   }
 
   /**
+   * Returns the name of the main property for a field.
+   *
+   * The main property name defaults to 'value', but a field can define
+   * a function mainPropertyName() to indicate a different name.
+   *
+   * @param string $field_name
+   *   The machine name of the field.
+   *
+   * @return string
+   *   The main property name for this field.
+   */
+  public function getMainPropertyName(string $field_name): string {
+    $main_property_name = 'value';
+    /** @var \Drupal\Core\Field\FieldItemList $items **/
+    $items = $this->get($field_name);
+    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem $item **/
+    foreach ($items as $item) {
+      if (method_exists($item, 'mainPropertyName')) {
+        $main_property_name = $item->mainPropertyName();
+      }
+      // We only need to examine the first item.
+      break;
+    }
+    return $main_property_name;
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function preSave(EntityStorageInterface $storage): void {
@@ -935,6 +962,7 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
     $delta_remove = [];
     $fields = $this->getFields();
     foreach ($fields as $field_name => $items) {
+      $main_property_name = $this->getMainPropertyName($field_name);
       foreach($items as $item) {
 
         // If it is not a TripalField then skip it.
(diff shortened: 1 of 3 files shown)


laceysanderson (approved):
✅ Code-wise this change looks good :-)
✅ Manual testing performed as expected.

Specifically, I tested on the project content type. I created two fields through the UI with unlimited cardinality. Then I performed the following:

- CREATE: added many values to each field and removed one from each during create to ensure there was no swapping between fields and it worked on create.
- EDIT: removed 2 from one field and 1 from the other
- EDIT: both removed and added to both fields
- EDIT: added two values that were the same and only removed one of them
- EDIT: removed one value, rearranged all values, added a new value and rearranged again

PR #2274, opened 2025-09-04
Better caching of TripalEntity object

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR is working to greatly improve TripalContent loading by ensuring it is actually cached in the Drupal persistent cache.

More specifically,

- moves loading from backend storage into the storage loading hook (TripalEntityHooks::tripalEntityStorageLoad) in order to ensure it is run before the entity is added to the persistent cache. It used to be done in the TripalEntity::postLoad which was called after the entity was added to the persistent cache. This provides a huge performance boost the second time and entity is loaded in the site!
- adds a helper method `saveValuesArray()` to centralize processes that used to be duplicated between the `postLoad()` and `preSave()` and additional helper methods to reduce code complexity.

## Tasks

- [x] Determine the actual flow of entity loading to diagnose why our entities are not being cached when we think they should be. See [Issue #2195-comment](https://github.com/tripal/tripal/issues/2195#issuecomment-3251068935).
- [x] Move TripalStorage loading to the storage loading hook so the result is cached in the persistent cache. See TripalEntityHooks::tripalEntityStorageLoad().
- [x] Fix title/url token bugs resulting from this change.
- [x] Update preSave() to use saveValuesArray() and reduce duplication.
- [x] Make a diagram showing the new flow of tripal entity loading including the role the backend storage plays and how it interacts with caching.

The following tasks will be completed in a future PR since it impacts the code of TripalStorage and ChadoStorage which have not been brought up to coding standards yet. Furthermore, it impacts TripalEntity::validate() which is not covered in the current automated testing -making it difficult to ensure we are not breaking anything 🙈 

- Look into additional caching of the setup TripalStorage objects.
- Look into caching the status of which fields are saved in the drupal tables.

### Loading the Entity

EntityLoad

### Saving the Entity (both create and update)

SaveDiagram

NOTE: Currently TripalStorage is setup THREE TIMES when updating an entity: during load, during validate and during save 🙈

## Testing?

You will want to test 
- creating an entity through the UI (confirm page, chado and drupal tables match expectations)
- reload that page multiple times and enjoy the speed on subsequent loads!
- edit the entity through the UI (confirm page, chado and drupal tables match expectations)
- delete the entity and republish it to make sure there are no unforseen consequences with publish.
- republish an entity with a newly added field with content to ensure that invalidates the cache and the new field appears.
- check that bug #2271 has not returned.


tripal/src/Entity/TripalEntity.php
```diff
@@ -879,6 +838,211 @@ public static function getValuesArray($entity, $ignore_cached_fields = FALSE) {
...
+   *
+   * @param TripalEntity $entity
+   *   The entity that we want to update.
+   * @param array $values
+   *   Values returned from TripalStorage mapping to fields of this entity.
+   * @param array $tripal_storages
+   *   Array of TripalStorage objects.
+   * @param bool $do_save
+   *   TRUE indicates this is being called within the save workflow and
+   *   when FALSE if is being called in the load workflow.
```
dsenalik:
```suggestion
   *   FALSE when it is being called in the load workflow.
```

tripal/src/Entity/TripalEntity.php
```diff
@@ -879,6 +838,211 @@ public static function getValuesArray($entity, $ignore_cached_fields = FALSE) {
...
+        }
+        [$delta, $tsid] = $storage;
+
+        // Create a new properties array for this field item.
+        $prop_values = [];
+        $prop_types = [];
+        $store_values = [];
+        foreach ($values[$tsid][$field_name][$delta] as $key => $info) {
+
+          // Get the specific prop type and it's corresponding value.
```
dsenalik:
```suggestion
          // Get the specific prop type and its corresponding value.
```
sorry 😁

tripal/src/Entity/TripalEntity.php
```diff
@@ -879,6 +838,211 @@ public static function getValuesArray($entity, $ignore_cached_fields = FALSE) {
...
+   *   The name of the field this item is for.
+   * @param Drupal\Core\Field\FieldItemInterface $item
+   *   The item whose backend storage we want to retrieve.
+   *
+   * @return array|bool
+   *   FALSE if this is not a TripalFieldItem or if it doesn't indicate its
+   *   TripalStorage plugin. Otherwise, an associative array describing the
+   *   backend storage for this item. Specifically,
+   *   - delta: the delta of this item in the fielditemlist it came from.
+   *   - tsid: the tripalstorage id for it's storage backend.
```
dsenalik:
```suggestion
   *   - tsid: the tripalstorage id for its storage backend.
```

dsenalik (approved):
Only fussy grammar fixes.
I confirmed manual testing on the stockcollection field as that is easiest to re-confirm the property bug has not come back. Everything worked and so much faster! 🏃‍💨

I also tested on my Tripal File branch, which uses content types on custom tables, and everything is working there too.

PR #2277, opened 2025-09-08
Update Test Coverage Workflow to run on 4.x

laceysanderson:
# Bug Fix

## Description

When analyzing the change in test coverage for my recent performance PR I noticed that QLTY Cloud was reporting a decrease in coverage for files that should not be impacted by my PR. Needless to say I was concerned and did some digging which caused me to notice that coverage had not been run on 4.x in 2 months!!! 🙈 We need coverage run on 4.x since it is used in the comparison to say how much our PRs are affecting the diff and total coverage!

This PR adds the 4.x branch into those test coverage should be run on.

NOTE: I manually triggered the test coverage workflow for 4.x tonight as shown in this screenshot from our github action logs.

Screenshot 2025-09-07 at 9 54 25 PM

## Testing?

This is literally a workflow only change. As such it's really just a code review. GitHub should scream if there is any syntax issues in the workflow.

### DIFF
--- a/.github/workflows/ALL-testCoverage-qltycloud.yml
+++ b/.github/workflows/ALL-testCoverage-qltycloud.yml
@@ -1,6 +1,9 @@
 name: 'Test Coverage on Drupal 11.1.x-dev + PHP 8.3'
 
 on:
+  push:
+    branches:
+      - '4.x'
   workflow_dispatch:
   pull_request:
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
Well I noticed some strange coverage changes too. How did I touch the GFF importer for example 😕. Thanks for figuring it out!

PR #2280, opened 2025-09-08
Update OBO importer to work for EDAM

### LINKED ISSUE #2278: Problems importing the EDAM ontology
### Tripal Version

4.x

### Drupal Version

11.2

### PostgreSQL Version

17

### PHP Version

8.4

### General Description

There are problems importing the full EDAM ontology.
Errors shown below.
There are also two calls to `tripal_insert_obo()` in the code, this is a tripal 3 API function that does not exist in tripal 4.

Also a serious bug, the `type_id` column of the cvtermsynonym table is NOT being populated for ANY ontology.
```
SELECT * FROM cvtermsynonym where type_id IS NOT NULL;
 cvtermsynonym_id | cvterm_id | synonym | type_id 
------------------+-----------+---------+---------
(0 rows)
```

### Steps to reproduce

- Tripal
- Data Loaders
- OBO Vocabulary Loader  
- Add a New Ontology OBO Reference
New Vocabulary Name: EDAM
Remote URL: https://edamontology.org/EDAM.obo
- Click "Import OBO File"
- Run the job

### Error messages

```shell
...
 [warning] Undefined array key 1 OBOImporter.php:2412
...
ERROR: Unhandled exception downloading URL http://www.ebi.ac.uk/ols/api/ontologies/http: Server error: `GET http://www.ebi.ac.uk/ols/api/ontologies/http` resulted in a `500 Internal Server Error` response:
{
  &quot;status&quot; : 500,
  &quot;message&quot; : &quot;Expected at least 1 result for solr getFirst&quot;
}

[site http://default] [TRIPAL] ERROR: Unhandled exception downloading URL http://www.ebi.ac.uk/ols/api/ontologies/http: Server error: `GET http://www.ebi.ac.uk/ols/api/ontologies/http` resulted in a `500 Internal Server Error` response:{  "status" : 500,  "message" : "Expected at least 1 result for solr getFirst"}
ERROR: Unable to get response from http://www.ebi.ac.uk/ols/api/ontologies/http when trying to retrieve data for HTTP:.
[site http://default] [TRIPAL] ERROR: Unable to get response from http://www.ebi.ac.uk/ols/api/ontologies/http when trying to retrieve data for HTTP:.
 [warning] Undefined array key 1 OBOImporter.php:2412
...
Percent complete: 58.03 %. Memory: 51,892,760 bytes.
ERROR: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "cvtermsynonym_c1"
DETAIL:  Key (cvterm_id, synonym)=(3224, Quality) already exists.: INSERT INTO "chado"."cvtermsynonym" ("cvterm_id", "synonym") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1) RETURNING cvtermsynonym_id; Array
(
)

[site http://default] [TRIPAL] ERROR: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "cvtermsynonym_c1"DETAIL:  Key (cvterm_id, synonym)=(3224, Quality) already exists.: INSERT INTO "chado"."cvtermsynonym" ("cvterm_id", "synonym") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1) RETURNING cvtermsynonym_id; Array()
```

dsenalik:
# Bug Fix

## Description

 1. `Undefined array key 1` - Blank line handling has been reordered so that they are always detected. Tripal 3 version had this same bug. See commit https://github.com/tripal/tripal/pull/2280/commits/2ffb0da54abeaf47198c4e89437ef46a99146174
 2. EDAM does not have the ontology name in the header, rather it has a URL. Ignore URLs as ontology name, and use an alternate header "default-relationship-id-prefix"
 3. Update some of the logging, especially upgrade the remaining `logMessage()` call. This is mostly in commit https://github.com/tripal/tripal/pull/2280/commits/7a56f917ee754fda59374cf343dd444bfd49f270
 4. The unique constraint for the `cvtermsynonym` table only allows one combination of `cvterm_id` + `synonym`, so if we try to add the same synonym but of a different syn_type_id, the insert will fail. Reference: https://github.com/GMOD/Chado/issues/149 For example this section in the EDAM obo file broke the import:
 ```
synonym: "Quality" BROAD []
synonym: "Quality" RELATED []
```
 5. The `type_id` column of the `cvtermsynonym` table was not being populated. This has been corrected. This happened for ANY obo file that was imported. On this branch the import can be repeated to populate these missing values.
 6. The EDAM ontology has four ID spaces, but it also includes relationships to the parent 'EDAM' namespace. These are stored in the `cvterm_relationship` table. Because of this relationship, code here https://github.com/tripal/tripal/blob/1acd05b695ac8d63c8b73ebc0812b5dad599e5ed/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php#L396-L432 will try to retrieve the tripal term for EDAM, the parent when adding a field, but it may not exist. This breaks adding any fields that use an EDAM term. The simple fix is to call `$this->getIdSpace()` for any added DB records, which will add it to the collection if it isn't already present.
 7. `Cannot add Alt ID "topic_0197" without an accession`
 This is because they are stored in the obo file as `topic_0197` instead of as `topic:0197`
 There are only three, and the same thing happened in tripal 3, so I am ignoring this.
 8. I significantly expanded the "marked as incomplete" obo importer test. The obo files loaded by chado get effectively tested, so remove the TCONTACT test and replaced with a plant trait ontology subset. This one also tests the external database lookup, I limited to just doing this once in the subset obo file.

## Testing?
 1. Build a docker on this branch to make sure ontology import for chado completes.
 1. Go to Tripal -> Data Loaders -> OBO Vocabulary Loader
 2. Under "Ontology OBO File Reference" select "EDAM"
 3. Click "Import OBO File"
 4. Run the job
 5. Job should complete successfully.
Some changes in the output resulting from changes here
```
...
Skipping addition of synonym "Quality" because it already exists under a different synonym type
...
Skipping addition of synonym "Function" because it already exists under a different synonym type
...
Skipping addition of synonym "Quality" because it already exists under a different synonym type
...
Cannot add Alt ID "topic_0197" without an accession
...
Cannot add Alt ID "topic_3076" without an accession
...
Cannot add Alt ID "topic_3040" without an accession
...
```
8. Check the `cvtermsynonym`, there should be no NULL `type_id` records
 ```
SELECT * FROM cvtermsynonym WHERE type_id IS NULL;
  cvtermsynonym_id | cvterm_id | synonym | type_id 
------------------+-----------+---------+---------
(0 rows)
```


laceysanderson (approved):
✅ Code changes look good :-)
✅ The added tests look great and are Much Appreciated! PS. I loved seeing the use of the YAML file for your scenarios!
✅ Manual testing went great.

 - Confirmed that I saw the warnings mentioned regarding EDAM
 - Confirmed by querying the cv, db, cvterm, dbxref tables that the terms were inserted and associated with the existing idspaces 🎉 
 - Inserted other ontologies including taxrank and sequence to confirm that they also still loaded without error and terms were created as expected
Thanks for this fix @dsenalik! This looks ready to merge!

(Assuming the tests pass again now that I updated the branch 🙏 )

PR #2294, opened 2025-09-24
Actually deprecate the Chado Query API in favour of Tripal DBX.

### LINKED ISSUE #1341: Deprecate existing Chado Query API in favour of new BioDB API
The new BioDB API extends the Drupal Database API to handle multiple PostgreSQL schema and cross-schema querying -something which is obviously needed for Tripal for ideal Chado integration. Specifically, this allows us to use the Drupal style of querying with Chado as shown in the following example:

```php
// Initiate the new Tripal BioDB service with Chado integration.
$biodb = \Drupal::service('tripal_chado.database');
// Start a select query on the feature table.
$query = $biodb->select('feature', 'x');
// Add JOINS to the organism and cvterm tables.
$query->join('organism', 'o', 'x.organism_id = o.organism_id');
$query->join('cvterm', 'cvt', 'x.type_id = cvt.cvterm_id');
// Exclude obsolete features.
$query->condition('x.is_obsolete', 'f', '=');
// Add the columns you want to be returned.
$query->fields('x', ['name', 'uniquename']);
$query->fields('o', ['genus', 'species', 'common_name']);
$query->addField('cvt', 'name', 'feature_type');
// Indicate the range of results you want.
$query->range(0, 10);
// Finally execute the above build query.
$result = $query->execute();
foreach ($result as $record) {
  // Do something with each record from the chado database.
}
```

Which would be the equivalent of the following Chado Query:

```sql
SELECT x.name, x.uniquename, cvt.name as feature_type, o.genus, o.species, o.common_name
FROM chado.feature x
LEFT JOIN chado.organism o ON x.organism_id = o.organism_id
LEFT JOIN chado.cvterm cvt ON x.type_id = cvt.cvterm_id
WHERE x.is_obsolete = f
LIMIT 10;
```

Tripal will be moving to this new Drupalized way of building queries and thus the old function-based Chado Query API will become deprecated. That said, it will remain around for multiple versions of Tripal 4.x to provide ample time for extension modules to upgrade to the new API.

Additionally, the insides of the old API will be replaced with this new Drupalized way of doing things in order to streamline maintenance of Tripal 4.x.

laceysanderson:
# Tripal 4 Core Dev Task

## Description

We need to actually deprecate this very old API as mentioned in https://github.com/tripal/tripal_doc/pull/109. This PR will accomplish this.

More specifically, this PR

- Adds deprecation notices to the Chado Query API methods and their direct helpers and hooks. This does not touch the Chado Variables API but since it calls the now deprecated Chado Query API you will see deprecation messages when using it.
- Updates all usage of chado_query outside of the importers since this is relatively trivial. Does not update usage of chado_*_record() methods.
- Adds `IgnoreDeprecations` to API specific tests so that they don't fail due to the deprecations which we expect. This allows us to continue testing these until we remove them in Tripal 4.1.0
- Updates the config for PHPUnit 9.6 to allow tests to pass when deprecations are seen. This is because the `IgnoreDeprecations` is not available in that version. This seems ok since we will see the deprecations in Drupal 11+

NOTE: PHPUnit 11.5 is still configured not to fail on deprecations. I did not fix this in this PR because there are remaining deprecations triggered by this PR that will be fixed in #2297. We should make sure to fix this in that PR to ensure that deprecations triggered in tests outside of those specifically testing the deprecated API get caught and cause failure.

## Testing?

I'm not sure how to test this. There are automated tests that check this functionality. Since the importers have not yet been updated in this PR one way to test may be run the importers and make sure they haven't broken in this PR.

### DIFF
--- a/tripal_chado/migration/export_tripal3_entity_mapping.php
+++ b/tripal_chado/migration/export_tripal3_entity_mapping.php
@@ -16,7 +16,7 @@ function find_all_published(string $output_file_name) {
 
   // Get a list of defined bio_data bundles
   $sql = "SELECT name, label FROM [tripal_bundle] ORDER BY term_id";
-  $results = \Drupal::service('tripal_chado.database')->query($sql, []);
+  $results = chado_query($sql, []);
   $bundle_names = [];
   while ($obj = $results->fetchObject()) {
     $bundle_names[$obj->name] = $obj->label;
@@ -152,7 +152,7 @@ function query_one_bundle(string $bundle_name) {
   }
 
   $sql = "SELECT * " . $from . $where;
-  $results = \Drupal::service('tripal_chado.database')->query($sql, $args);
+  $results = chado_query($sql, $args);
   return [$table, $pkey_field, $results];
 }
 
--- a/tripal_chado/migration/export_tripal3_entity_mapping.php
+++ b/tripal_chado/migration/export_tripal3_entity_mapping.php
@@ -16,7 +16,7 @@ function find_all_published(string $output_file_name) {
 
   // Get a list of defined bio_data bundles
   $sql = "SELECT name, label FROM [tripal_bundle] ORDER BY term_id";
-  $results = chado_query($sql, []);
+  $results = \Drupal::service('tripal_chado.database')->query($sql, []);
   $bundle_names = [];
   while ($obj = $results->fetchObject()) {
     $bundle_names[$obj->name] = $obj->label;
@@ -152,7 +152,7 @@ function query_one_bundle(string $bundle_name) {
   }
 
   $sql = "SELECT * " . $from . $where;
-  $results = chado_query($sql, $args);
+  $results = \Drupal::service('tripal_chado.database')->query($sql, $args);
   return [$table, $pkey_field, $results];
 }
 
--- a/tripal_chado/src/api/tripal_chado.db.api.php
+++ b/tripal_chado/src/api/tripal_chado.db.api.php
@@ -666,7 +666,7 @@ function chado_autocomplete_dbxref($db_id, $string = '') {
   $results = \Drupal::service('tripal_chado.database')->query($sql, [
     ':db_id' => $db_id,
     ':accession' => $string . '%',
-  ])->execute();
+  ]);
   $items = [];
   foreach ($results as $ref) {
     $items[$ref->accession] = $ref->accession;
--- a/tripal_chado/src/api/tripal_chado.db.api.php
+++ b/tripal_chado/src/api/tripal_chado.db.api.php
@@ -159,12 +159,9 @@ function chado_get_db_select_options($schema_name = NULL) {
     $schema_name = \Drupal::config('tripal_chado.settings')->get('default_schema');
   }
 
-  $dbs = chado_query(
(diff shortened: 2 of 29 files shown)


tripal_chado/migration/export_tripal3_entity_mapping.php
```diff
@@ -16,7 +16,7 @@ function find_all_published(string $output_file_name) {
 
   // Get a list of defined bio_data bundles
   $sql = "SELECT name, label FROM [tripal_bundle] ORDER BY term_id";
-  $results = chado_query($sql, []);
+  $results = \Drupal::service('tripal_chado.database')->query($sql, []);
```
dsenalik:
```suggestion
  $results = chado_query($sql, []);
```
This is a special case, this script is designed to be run on a Tripal 3 site, so it won't have the new service!

tripal_chado/migration/export_tripal3_entity_mapping.php
```diff
@@ -152,7 +152,7 @@ function query_one_bundle(string $bundle_name) {
   }
 
   $sql = "SELECT * " . $from . $where;
-  $results = chado_query($sql, $args);
+  $results = \Drupal::service('tripal_chado.database')->query($sql, $args);
```
dsenalik:
```suggestion
  $results = chado_query($sql, $args);
```

tripal_chado/src/api/tripal_chado.db.api.php
```diff
@@ -661,15 +658,15 @@ function chado_autocomplete_dbxref($db_id, $string = '') {
...
     WHERE db_id = :db_id and lower(accession) like lower(:accession)
     ORDER by accession
     LIMIT 25 OFFSET 0
   ";
-  $results = chado_query($sql, [
+  $results = \Drupal::service('tripal_chado.database')->query($sql, [
     ':db_id' => $db_id,
     ':accession' => $string . '%',
-  ]);
+  ])->execute();
```
dsenalik:
```suggestion
  ]);
```
In this case we don't want execute.

tripal_chado/src/api/tripal_chado.property.api.php
```diff
@@ -670,13 +670,15 @@ function chado_get_record_with_property($record, $property, $options = [], $sche
...
+  $chado_connection->setSchemaName($schema_name);
 
   // Make sure this is a prop table.
   if (!preg_match('/prop$/', $prop_table)) {
     throw new Exception('Please provide a valid Chado property table');
   }
-  $sql = 'SELECT DISTINCT type_id FROM {' . $prop_table . '}';
-  $results = chado_query($sql, [], [], $schema_name);
+  $sql = 'SELECT DISTINCT type_id FROM {1:' . $prop_table . '}';
+  $results = $chado_connection->query($sql, [])->execute();
```
dsenalik:
```suggestion
  $results = $chado_connection->query($sql, []);
```

dsenalik (changes requested):
A few changes, mainly the script for entity mapping is designed to be run on a Tripal 3 site so no updates there!

I ran all the importers and no problems there.

dsenalik (approved):
Approved when tests complete!

PR #2297, opened 2025-09-24
Remove deprecated functions from importers

### LINKED ISSUE #2295: Remove deprecated functions from importers
### Task Description

This will remove deprecated api functions inherited from Tripal 3 from importers.
E.g. `chado_generate_var`, `chado_get_cv`, `chado_query` and the like.

### Branch Name

tv4g0-issue2295-remove-deprecated-functions-from-importers

dsenalik:
# Tripal 4 Core Dev Task

## Description

PR #2294 will deprecate a number of `chado_` api functions inherited from Tripal 3.

This PR replaces them in various importers.

## Testing?

~~Before testing `git merge origin/tv4g2-1341-deprecateChadoQueryAPI` so that deprecated functions show up in testing.~~
Automated tests should now catch any missed deprecations

Run each of the importers to verify that I have not broken anything.
To test obo importer on a local file you can use `modules/contrib/tripal/tripal_chado/tests/fixtures/obo_loader/test_EDAM.obo`

### DIFF
--- a/tripal_chado/src/api/tripal_chado.phylotree.api.php
+++ b/tripal_chado/src/api/tripal_chado.phylotree.api.php
@@ -774,7 +774,7 @@ function chado_phylogeny_import_tree(&$tree, $phylotree, $options, $vocab = [], 
   // Get the vocabulary terms used to describe nodes in the tree if one
   // wasn't provided.
   if (count($vocab) == 0) {
-    $vocab = chado_phylogeny_get_node_types_vocab($chado, $options, $schema_name);
+    $vocab = chado_phylogeny_get_node_types_vocab($options, $schema_name);
   }
 
   if (is_array($tree) and array_key_exists('name', $tree)) {
--- a/tripal_chado/src/api/tripal_chado.phylotree.api.php
+++ b/tripal_chado/src/api/tripal_chado.phylotree.api.php
@@ -997,7 +997,9 @@ function chado_phylogeny_lookup_organism_by_name($name, $schema_name = 'chado') 
  *
  * @ingroup tripal_phylotree_api
  */
-function chado_phylogeny_get_node_types_vocab($chado, $options, $schema_name = 'chado') {
+function chado_phylogeny_get_node_types_vocab($options, $schema_name = 'chado') {
+  $chado = \Drupal::service('tripal_chado.database');
+  $chado->setSchemaName($schema_name);
   // Get the three default vocabulary terms used to describe nodes in the tree.
   $terms = ['leaf' => 'phylo_leaf', 'internal' => 'phylo_interior', 'root' => 'phylo_root'];
   $vocab = [];
--- a/tripal_chado/src/api/tripal_chado.phylotree.api.php
+++ b/tripal_chado/src/api/tripal_chado.phylotree.api.php
@@ -252,12 +252,19 @@ function chado_validate_phylotree($val_type, &$options, &$errors, &$warnings, $s
         'name' => $db_name,
       ],
     ];
-    $dbxref = chado_generate_var('dbxref', $values, [], $schema_name);
+    $query = $chado->select('1:dbxref', 'x');
+    $query->join('1:db', 'db', '"x".db_id = "db".db_id');
+    $query->condition('x.accession', $accession, '=');
+    $query->condition('db.name', $db_name, '=');
+    $query->addField('x', 'dbxref_id', 'dbxref_id');
+    $dbxref_id = $query->execute()->fetchField();
 
-    if (!$dbxref) {
-
-      $db = chado_generate_var('db', ['name' => $db_name], [], $schema_name);
-      if (!$db) {
+    if (!$dbxref_id) {
+      $query = $chado->select('1:db', 'DB');
+      $query->condition('DB.name', $db_name, '=');
+      $query->addField('DB', 'db_id', 'db_id');
+      $db_id = $query->execute()->fetchField();
+      if (!$db_id) {
         $errors['dbxref'] = t(
             'dbxref could not be created for %dbname:%dbxref, this DB does not exist.',
             ['%dbname' => $db_name, '%dbxref' => $dbxref]);
@@ -265,11 +272,11 @@ function chado_validate_phylotree($val_type, &$options, &$errors, &$warnings, $s
       }
 
       // Here we create the new dbxref for the specified new accession.
-      $dbxref = $chado->insert('1:dbxref')->fields([
+      $dbxref_id = $chado->insert('1:dbxref')->fields([
         'accession' => $values['accession'],
-        'db_id' => $db->db_id
+        'db_id' => $db_id,
       ])->execute();
-      if (!$dbxref) {
+      if (!$dbxref_id) {
         $errors['dbxref'] = t(
             'dbxref could not be created for %dbname:%dbxref.',
             ['%dbname' => $db_name, '%dbxref' => $dbxref]);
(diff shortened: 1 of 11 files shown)


laceysanderson:
I made this PR dependant on #2294 so that we can check the deprecations right now. I also added a setting to PHPUnit 11.5 config to ensure that our tests will fail when deprecations are encountered outside of tests with `#[IgnoreDeprecations]`. This settings should not fail when the deprecations are caused by third-party code such as PHPunit so the long standing mock deprecations (See Issue #2249) can remain unaddressed for now 😅

tripal_chado/tests/src/Kernel/api/ChadoPhylogenyAPITest.php
```diff
@@ -83,15 +85,19 @@ public function testChadoPhylotreeAPIFunctions() {
             'common_name' => 'False Tripal',
             'abbreviation' => 'T. ' . $species . ' subsp. sativus',
            ];
-    $dbq = chado_insert_record('organism', $org, [], $this->schemaName);
-    $this->assertNotNull($dbq, 'Unable to insert test organism 1.');
-    $organism_ids[0] = $dbq['organism_id'];
+    $query = $this->chado_connection->insert('1:organism');
+    $organism_id = $query->fields($org);
```
laceysanderson:
```suggestion
    $query = $this->chado_connection->insert('1:organism')
      ->fields($org);
```

tripal_chado/tests/src/Kernel/api/ChadoPhylogenyAPITest.php
```diff
@@ -83,15 +85,19 @@ public function testChadoPhylotreeAPIFunctions() {
...
+    $this->assertEquals('1', $organism_id, 'Unable to insert test organism 1.');
+    $organism_ids[0] = $organism_id;
 
     $org['infraspecific_name'] = 'selvaticus';
     $org['abbreviation'] = 'T. ' . $species . ' subsp. selvaticus';
-    $dbq = chado_insert_record('organism', $org, [], $this->schemaName);
-    $this->assertNotNull($dbq, 'Unable to insert test organism 2.');
-    $organism_ids[1] = $dbq['organism_id'];
+    $query = $this->chado_connection->insert('1:organism');
+    $organism_id = $query->fields($org);
```
laceysanderson:
```suggestion
    $query = $this->chado_connection->insert('1:organism')
      ->fields($org);
```

tripal_chado/src/api/tripal_chado.phylotree.api.php
```diff
@@ -998,26 +997,25 @@ function chado_phylogeny_lookup_organism_by_name($name, $schema_name = 'chado')
  *
  * @ingroup tripal_phylotree_api
  */
-function chado_phylogeny_get_node_types_vocab($options, $schema_name = 'chado') {
+function chado_phylogeny_get_node_types_vocab($chado, $options, $schema_name = 'chado') {
```
laceysanderson: This is not backwards compatible and we have not added deprecation messages indicating the change :see_no_evil:

If this change was made because of the default of schema name for chado then we can either update the importer to pass in the schema name or we can add a section here that if the schema name is not passed in then do not set it or grab the set default schema name.

```suggestion
function chado_phylogeny_get_node_types_vocab($options, $schema_name = 'chado') {
  $chado = \Drupal::service('tripal_chado.database');
  $chado->setSchemaName($schema_name);
```

laceysanderson (changes requested):
Code review is looking mostly good :-) Just one backwards incompatible change that we should resolve.

laceysanderson (commented):
## Manual Testing

### GFF3 importer.

The change was in insertFeatureSeqs() so I tested 
- using this test fixture `tripal_chado/tests/fixtures/gff3_loader/small_gene_with_fasta.gff`
- imported it with a default srcfeature type of scaffold
- created a scaffold content type and published it
- ✅ The residues and associated metadata showed up as expected
Screenshot 2025-09-29 at 12 04 46 PM

### Newick Tree Loader

The change here was in the form when the sequence ontology is not present so I tested
- by updating the name of the sequence ontology to `Fred` which did trigger the code... However, apparently the original code was checking too late and other errors appeared. 
- As such I make the code change in commit f73a8e61b5edded2e6653b7037d36a5ad6306fc7 with @dsenalik blessing
- ✅ Now loading the form shows the warning as expected.
Screenshot 2025-09-29 at 12 21 00 PM

laceysanderson (changes requested):
## Manual Testing

### OBO Importer

The change here was in the stanza caching. This should be impacted on most obo loads so I tested by 
- I used the local EDAM subset suggested by Doug
- On the obo imporer I selected "Add a New Ontology", included the path provided by Doug and set the same to "Local EDAM"
- ✅ I then ran the job and confirmed there were no errors as expected. If the changes had not worked, there would have been PDO errors.
   Screenshot 2025-09-29 at 1 07 10 PM

### Taxonomy Importer
Changes here were made in the following methods: findOrganism(), addOrganism(), and importRecord(). These will be hit on all jobs for this importer that has at least one organism. As such I tested by
- Created the organism `Daucus muricatus` which matches one of the ids in the next step through the Add Organism form (to test find when one exists).
- Imported the following ids `48026 48106 79170 79196 79200 180796` using the "NCBI Taxonomy Importer" form and selected "Import details for existing species"
- When I ran the job it failed with the following error showing the updated query in addOrganism() did not work.
   Screenshot 2025-09-29 at 1 22 06 PM
- ❌ I fixed this by adding the field to be selected to one of the queries (see commit 37bbd1795181f3ca4cbb04c9d771f49be013451f) but it still failed as follows. Note: While I did add the one organism ahead of time and it does have an organism_id of 5, it did not have a dbxref_id assigned via organism_dbxref before the import.
   ```sql
   sitedb=# select * from teapot.organism;
    organism_id | abbreviation |  genus   |  species   | common_name | infraspecific_name | type_id | comment 
   -------------+--------------+----------+------------+-------------+--------------------+---------+---------
              1 |              | Tripalus | databasica |             |                    |         | 
              5 |              | Daucus   | muricatus  |             |                    |         | 
   (2 rows)

   sitedb=# select * from teapot.organism_dbxref;
    organism_dbxref_id | organism_id | dbxref_id 
   --------------------+-------------+-----------
   (0 rows)
   ```

   Screenshot 2025-09-29 at 1 39 41 PM

laceysanderson (approved):
## Manual Testing

✅ GFF3 Importer
✅ Newick Tree Loader
✅ OBO Importer
✅ Taxonomy importer
✅ Tree Generator
✅ FASTA Importer

### Taxonomy Importer
Changes here were made in the following methods: findOrganism(), addOrganism(), and importRecord(). These will be hit on all jobs for this importer that has at least one organism. As such I tested by
- Created the organism `Daucus muricatus` which matches one of the ids in the next step through the Add Organism form (to test find when one exists).
- Imported the following ids `48026 48106 79170 79196 79200 180796` using the "NCBI Taxonomy Importer" form and selected "Import details for existing species"
- When I ran the job, it completed without error as expected :-)
- ✅ After finding the new fields and re-publishing, I was able to confirm all the details are present both on the pre-existing `Daucus muricatus` and one other organism.
   image
   image

### Taxonomy Tree Generator

Changes were made in the run(), retrieveLineage(), removeTree(), and queryPhylonode(). I believe these changes will be best fit by regenerating an existing tree and trying to generate a tree that doesn't have an appropriate root organism.
- Tried to generate tree where the root taxon was "fabaceae" despite there being no organisms in the site for this family.
- ✅ Confirmed that the job ran with the expected error and that the initial record added by the importer was removed leaving an empty `phylotree` table.
   Screenshot 2025-09-29 at 3 44 03 PM
- Used the importer to generate a tree for `Daucus` genus which there are 4 organism in the site.
- ✅ the job ran without error, confirmed there were 4 organism which should be included and 2 which should not. Selecting from the database shows there are the expected phylonode records 🎉 
   Screenshot 2025-09-29 at 3 54 01 PM

### FASTA Importer

Changes are made to loadFastaFeature(), specifically with reference to updating an existing feature. As such we are testing by
- Import features using the GFF3 importer and `/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/fixtures/gff3_loader/gff_Citrus_sinensis-orange1.1g015632m.g.gff3`
- Import the sequence for orange1.1g015632m using the FASTA importing using the `/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/fixtures/fasta_loader/Citrus_sinensis-orange1.1g015632m.g.fasta`
- ✅ I confirmed that the sequence was imported and assigned to the residues field of the existing gene.
   image

PR #2293, opened 2025-09-23
New Controller Class to replace the ChadoOrganismAutocompleteController

vidu-wil:
# New Feature

## Description

This PR introduces a new controller class, ChadoOrganismFormElementController, which replaces and extends the functionality of the existing ChadoOrganismAutocompleteController. It retains all existing methods while adding methods for generating the autocomplete/select elements used in the organism widget. Additionally, it includes a method to get the select options as a direct replacement for chado_get_organism_select_options() method.

### Automated testing
ChadoOrganismFormElementControllerTest class containing the following methods:
- testChadoOrganismFormElementController() - tests if the suggestions returned by the handleAutocomplete using the url request matches the expected number of options and whether the organism_id returned by getPkeyId is valid.
- testFormElement() along with its data provider provideDataForTestFormElement() - tests the following test case scenarios to see if the returned element contains the expected keys and values:
                1. Select element with valid default id given
                2. Autocomplete with valid default id given
                3. Autocomplete with valid default value given
                4. Select element with valid default value given
                5. Autocomplete with invalid default id given

PR #2301, opened 2025-10-02
Add missing optional parameter to getChadoForeignKeyDef()

### LINKED ISSUE #2253: Cannot use core chado fields on new content types based on custom tables

dsenalik:
# Bug Fix

## Description

This bug was introduce in PR #2260 when we moved the position of the `$schema` parameter. In this one case, the optional second table parameter was not present, and so the function call now fails.

## Testing?

1. Go to Tripal → Page Structure
2. For any content type, select "Manage Fields".
3. For any field, select "Edit".
4. On this branch the error should not occur. On 4.x we had:

`TypeError: Drupal\tripal_chado\TripalField\ChadoFieldItemBase::getChadoForeignKeyDef(): Argument #2 ($right_table) must be of type ?string, Drupal\tripal_chado\Database\ChadoSchema given, called in /var/www/drupal/web/modules/contrib/tripal/tripal_chado/src/TripalField/ChadoFieldItemBase.php on line 474 in Drupal\tripal_chado\TripalField\ChadoFieldItemBase::getChadoForeignKeyDef() (line 1293 of modules/contrib/tripal/tripal_chado/src/TripalField/ChadoFieldItemBase.php).`

### DIFF
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -471,7 +471,7 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
     $object_pkey_col = self::getPrimaryKey($linked_table, $schema);
     $all_tables = $schema->getTables(['type' => 'table']);
     foreach (array_keys($all_tables) as $table) {
-      $foreign_keys = self::getChadoForeignKeyDef($table, $schema);
+      $foreign_keys = self::getChadoForeignKeyDef($table, NULL, $schema);
       if ($foreign_keys) {
         // For "single-hop" logic, we add this table if there is a
         // foreign key to our linked_table.


laceysanderson (approved):
✅ Code changes makes perfect sense --oops 🙈 
✅ Confirmed all calls to getChadoForeignKeyDef() now include 3 parameters.
✅ Manually tested by editing a number of fields including making changes and saving and there is not an error any more.

PR #2293 is in the merge queue first but once it's merged, then I will merge this as well.

Good catch @dsenalik!!!

PR #2305, opened 2025-10-03
Fix discover bug by checking for foreign key in both directions

### LINKED ISSUE #2304: Field discovery bug
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

This happens because the analysis field discover() options suggests an additional linking table `analysisfeature` which incorrectly gets added as a potential field on the contact bundle.
https://github.com/tripal/tripal/blob/8acb7cefb998a9d32e9f9b2bed2ed1f7f717917b/tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php#L309

It is then not correctly checked here for a foreign key to contact, only to analysis.
https://github.com/tripal/tripal/blob/8acb7cefb998a9d32e9f9b2bed2ed1f7f717917b/tripal_chado/src/TripalField/ChadoFieldItemBase.php#L996-L998

### Steps to reproduce

Tripal -> Page Structure -> **Contact** -> Manage Fields
+Check for new fields
Add this field: Analysis (contact_analysis): Apply analytical methods to existing data of a specific type.

### Error messages

```shell

TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in array_keys() (line 117 of modules/contrib/tripal/tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php).

Drupal\tripal_chado\Plugin\Field\FieldType\ChadoAnalysisTypeDefault::tripalTypes(Object) (Line: 430)
Drupal\tripal\TripalField\TripalFieldItemBase::schema(Object) (Line: 483)
Drupal\field\Entity\FieldStorageConfig->getSchema() (Line: 2335)
...
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description

During field discovery, a linking table needs to have a foreign key both to the linked table, but also to the base table.
This second foreign key was never actually checked, so in the case of the "special" linking table `featureanalysis` a bug showed up and this table was used to link contact and analysis which resulted in a wsod.

## Testing?
Tripal -> Page Structure -> **Contact** -> Manage Fields
+Check for new fields
You should **not** see this field in the list any more: Analysis (contact_analysis): Apply analytical methods to existing data of a specific type.

### DIFF
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -994,8 +994,10 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
         // Some materialized views may look like a linker table, but if
         // there is no primary key, we can't use it, e.g. analysis_organism.
         $linker_pkey = self::getPrimaryKey($linking_table, $options['chado']->schema());
+        // Make sure a foreign key exists in both directions.
         $fk_def = self::getChadoForeignKeyDef($linking_table, $options['table'], $options['chado']->schema());
-        if ($linker_pkey && $fk_def) {
+        $bt_fk_def = self::getChadoForeignKeyDef($linking_table, $options['base_table'], $options['chado']->schema());
+        if ($linker_pkey && $fk_def && $bt_fk_def) {
           $linker_fkey_column = array_keys($fk_def['columns'])[0];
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Code changes make complete sense
✅ Tested and confirmed that the field no longer appears.

Image

PR #2315, opened 2025-10-06
Add php 8.4 support - 1 of 2

dsenalik:
# Tripal 4 Core Dev Task

## Description

This only adds building of php 8.4 dockers, so that part 2 of this PR will have dockers available for tests.

## Testing?
?
8.4 images should be present after merge and workflow has run at https://hub.docker.com/r/tripalproject/tripaldocker-drupal/tags


laceysanderson (approved):
Both workflow updates are good and the dockers have been updated ✅ Nod coding related changes so I'm skipping the automated testing.

PR #2320, opened 2025-10-09
Add Field::mainDisplayPropertyName() for use as the title/url token value and fix Field::mainDisplayPropertyName()

### LINKED ISSUE #2319: Field::mainPropertyName() is not suitable as the title/url token
### Task Description

While working on #2281 I came to realize that the property returned by `TripalFieldItem::mainPropertyName()` in our core fields is being used as the value for token replacement in title/URLs.

This is in direct conflict with the fact that this value is also used to detect if a field is empty and is populated by the backend storage for many of our linker fields. 🙈 

Also, according to api.drupal.org `TripalFieldItem::mainPropertyName()` is defined as

> Returns the name of the main property, if any.
> Some field items consist mainly of one main property, e.g. the value of a text field or the target_id of an entity reference. If the field item has no main property, the method returns NULL.

### Example

Let's use the ChadoContactTypeDefault field item as an example.

 - `ChadoContactTypeDefault::mainPropertyName()` return `contact_name`
 - the `contact_name` property has an action of `read_value` so it is only populated when loading the entity
 - the `ChadoContactWidgetDefault` sets the `contact_id` based on the user input to the form element but does not set the `contact_name`
 - this means that all contact elements would be filtered as empty when saving because the `contact_name` is always empty until after loading

## Design

We propose to fix this by adding a `mainDisplayPropertyName()` which returns the property we have been mistakenly putting in `mainPropertyName()`. This can then be used for the title/url substitution. Then the `mainPropertyName()` can be set to the id like it should have been.

### Branch Name

tv4g1-2319-mainProperty

laceysanderson:
# Bug Fix

## Description
This PR defines a new method `mainDisplayPropertyName()` in Tripal/Chado Fields for use populating field-specific tokens in Title and URL formats. This is needed because the previously used `mainPropertyName()` should actually be the key property in order to properly detect empty fields.

- [x] Add the `mainDisplayPropertyName()` method to the Tripal Field Base + Interface. We return 'value' as a default since that is what is used for single property fields such as boolean, integer, string.
- [x] For every field with multiple properties, have `mainDisplayPropertyName()` return the property name previously defined by `mainPropertyName()` and update `mainPropertyName()` to return the name of the property set directly by the widget.
- [x] Update the entity and publish use of `mainPropertyName()` to `mainDisplayPropertyName()` instead. This should mean that there is no change to what to expect as the value of the title/url tokens.
- [x] Add a test for better test coverage of fields.

## Testing?

There are automated tests for the title/url substitutions and these tests should continue to pass without needing to be changed. This confirms that the value returned for the tokens has not changed despite being determined by a different method.

Manual testing should include 
- making a number of Tripal Content pages and confirming that the titles and URLs still match what you expect based on the format.
- when editing content with unlimited fields, make sure to add multiple instances of each field and confirm that they are not filtered out when they are not empty ;-p also confirm that they are removed properly when they are empty although this is currently a bit fragile in the main branch.

### DIFF
--- a/tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldStaticMethodTest.php
+++ b/tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldStaticMethodTest.php
@@ -66,9 +66,9 @@ class FieldStaticMethodTest extends ChadoTestKernelBase {
    * setup.
    *
    * @var array
-   *  A list of scenarios where each one has the following keys:
-   *  - id: the machine name of the field to be tested.
-   *  - class: the class defining the field to be tested.
+   *   A list of scenarios where each one has the following keys:
+   *   - id: the machine name of the field to be tested.
+   *   - class: the class defining the field to be tested.
    */
   protected array $scenarios;
 
--- a/tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldStaticMethodTest.php
+++ b/tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldStaticMethodTest.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace Drupal\Tests\tripal_chado\Kernel\ChadoField;
+
+use Drupal\Tests\tripal_chado\Kernel\ChadoTestKernelBase;
+use Drupal\Tests\tripal_chado\Traits\ChadoFieldTestTrait;
+use PHPUnit\Framework\Attributes\Group;
+
+/**
+ * Tests the static methods of Chado Fields.
+ *
+ * @group TripalField
+ * @group ChadoField
+ */
+#[Group('TripalField')]
+#[Group('ChadoField')]
+class FieldStaticMethodTest extends ChadoTestKernelBase {
+
+  use ChadoFieldTestTrait;
+
+  /**
+   * The theme to use when rendering content in the test environment.
+   *
+   * @var string
+   */
+  protected $defaultTheme = 'stark';
+
+  /**
+   * The modules that this test depends on.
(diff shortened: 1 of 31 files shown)


tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldStaticMethodTest.php
```diff
@@ -0,0 +1,143 @@
...
+   * Describes the scenarios to test.
+   *
+   * This will be used in combination with the data provider. It can't be
+   * accessed directly in the dataProvider due to the way that PHPUnit is
+   * setup.
+   *
+   * @var array
+   *  A list of scenarios where each one has the following keys:
+   *  - id: the machine name of the field to be tested.
+   *  - class: the class defining the field to be tested.
```
dsenalik:
```suggestion
   *   A list of scenarios where each one has the following keys:
   *   - id: the machine name of the field to be tested.
   *   - class: the class defining the field to be tested.
```

dsenalik (approved):
One minor change from code review.
Awesome tests added 😍
- [x] Created organism and title is correct
- [x] Imported some publications, added multiple "Additional publication type" records, then removed them either by clearing field or using the "Remove" button = success
- [x] Imported some organisms
- [x] Created a project and linked multiple publications and dbxref annotations. Was able to edit, reorder, and remove them also. (for another issue, we may want to revisit the width of the widget for publications)
- [x] Added other tokens to title format for project, worked as expected.
- [x] Changed organism title format to `[organism_common_name]` and republished, worked as expected.
- [x] All URL aliases as expected.

PR #2324, opened 2025-10-14
Embedded Entity Formatter for Chado Linker Table fields

### LINKED ISSUE #2317: Embedded Entity Formatter for Chado Linker Table fields.
### Task Description

We would like a formatter to embed a view of a related entity similar to the Drupal Entity Reference field. 

This formatter would be available to all Linker Fields in Tripal Chado (e.g. ChadoOrganismTypeDefault, ChadoPubTypeDefault) and would use the entity_id which is already looked up in the field type. It would then allow the admin to configure the view mode they want to use and it would embed a rendered version of the linked entity using that entity_id and view mode.

## Here's an example of the Entity Reference field on a Tripal content type

The following screenshot is on a project page. The content within the coloured box is actually a rendered entity that was linked to this page using the entity reference field.
Image

When you edit this project, you are provided with a form to select the related entity. This would be handled by the existing widgets for Linker-style fields as is beyond the scope of this task.

On the "Manage Display" page for the project content type, the "Funding Grant" is shown like the following screenshot. In this case, the "View mode" is "Embedded in Child".

Image

When you configure the display of this field you see the following options:

Image

Where the options in this list are the "view modes" for the linked entity type. You can configure view modes in Drupal by going to Admin > Structure > Display Modes > View Modes and adding one to Tripal Content. There are the ones I have configured which supported the above field.

Image

### Reference

- [EntityReferenceEntityFormatter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21Plugin%21Field%21FieldFormatter%21EntityReferenceEntityFormatter.php/class/EntityReferenceEntityFormatter/11.x): the formatter to base our implementation from.
- [ChadoProjectTypeDefault](https://github.com/tripal/tripal/blob/4.x/tripal_chado/src/Plugin/Field/FieldType/ChadoProjectTypeDefault.php): an example field type of a linker field. All these fields that link one type of Tripal content to another will have the `entity_id` property which is what the formatter will use to determine which entity to render.
- Note: the entity may not always be present so we will want to have a fallback render in the case it is not 🤔 Maybe we will simply act like the field is empty?

### Branch Name

tv4g1-issue2317-embeddedEntityFormatter

vidu-wil:
# New Feature

## Description

We needed a formatter to embed a view of a related entity similar to the Drupal Entity Reference field.

This formatter is available to all Linker Fields in Tripal Chado (e.g. ChadoOrganismTypeDefault, ChadoPubTypeDefault) and would use the entity_id which is already looked up in the field type. It allows the admin to configure the view mode they want to use and it embeds a rendered version of the linked entity using that entity_id and view mode.

### Manual Testing
1. Configure view modes in Drupal by going to Admin > Structure > Display Modes > View Modes and adding one to Tripal Content as shown below.
Screenshot 2025-10-15 at 9 29 19 AM
2. On the "Manage Display" page for the Project content type, configure the display of the "Contact" field and "Analysis" field as shown in the following screenshot. 
Screenshot 2025-10-15 at 10 04 49 AM
3. Create a test contact using Tripal > Content > Add Tripal Content > Contact.
Screenshot 2025-10-15 at 10 05 50 AM
4. Create a test analysis using Tripal > Content > Add Tripal Content > Analysis.
Screenshot 2025-10-15 at 10 06 42 AM

Screenshot 2025-10-15 at 10 07 03 AM

5. Create a Project using Tripal > Content > Add Tripal Content >Project and link the created Analysis and the contact to the project as shown.
Screenshot 2025-10-15 at 10 07 41 AM

6. Once the project is created it should show the following view instead of links.
Screenshot 2025-10-15 at 3 35 49 PM


tripal_chado/src/Plugin/Field/FieldFormatter/ChadoEmbeddedEntityFormatter.php
```diff
@@ -0,0 +1,355 @@
...
+  ],
+)]
+class ChadoEmbeddedEntityFormatter extends ChadoFormatterBase {
+
+  /**
+   * The number of times this formatter allows rendering the same entity.
+   *
+   * @var int
+   */
+  const RECURSIVE_RENDER_LIMIT = 20;
```
dsenalik: I set up a recursive pair of fields, and I got project with contact with project with contact with project... and the limit did indeed work and eventually stopped it at 20, but at around 11 the field is displayed so compressed that only one or two characters appeared per line. So I am thinking this limit is too large. But I also wonder if we could even just set it to 1. In that case I have a project with embedded contact which in turn has an embedded project, and the recursion ends there. Is there any use case needing more than that?

dsenalik (changes requested):
This looks great! In addition to my one comment about recursion limit being too large, I note that an embedded entity does not have any clickable link to go to that entity. In many use cases this may be fine, I just wonder if that is something we might want to support. This is not a request that you do that, just a question for discussion.
Another thought is whether we should adjust the css - the background color of the embedded entity is almost indistinguishable from the surrounding background color. I wouldn't mind a thin border as an alternative option. Again, not an official request for a change.
For example, here is
```
  background-color: #f0f0f0;
  border-style: solid;
  border-width: 1px;
  border-color: #303030;
```

dsenalik:
I discussed with @laceysanderson and my suggestions about css changes don't need any changes on this PR.
We will make an issue in tripal docs https://github.com/tripal/tripal_doc/issues/128 on how a site admin can override css if they want to change the appearance, for example if you wanted a different colored background or border as I presented above.

vidu-wil:
> This looks great! In addition to my one comment about recursion limit being too large, I note that an embedded entity does not have any clickable link to go to that entity. In many use cases this may be fine, I just wonder if that is something we might want to support. This is not a request that you do that, just a question for discussion. 

I believe it’s already possible to support clickable links for embedded entities through some UI configurations.

- Go to the "Manage Display" page on the entity that needs to be embedded. (For above scenario go to manage display pages in both "Contact" and "Analysis")
- Move name to Disabled section and move the "Title" from Disabled to the top section.
- Click on the gear icon for the title section.
Screenshot 2025-11-04 at 11 01 50 AM

- Turn on the toggle for "Link to Tripal Content".
Screenshot 2025-11-04 at 11 02 04 AM

- Click "Update" and then "Save".
- Once you create new "Project" content the titles for linked "Contact" and "Analysis" should be clickable as shown below. 
Screenshot 2025-11-04 at 11 07 08 AM

tripal_chado/src/Plugin/Field/FieldFormatter/ChadoEmbeddedEntityFormatter.php
vidu-wil: I have changed the recursive limit to 1 as you have suggested. :)

dsenalik (approved):
Thanks, that worked perfectly to make the embedded entity have a clickable link.
So with the recursive limit reduced I hereby approve, thanks for all the work on this!

PR #2327, opened 2025-10-15
Add missing "1:" prefix in select in ChadoRecords et al.

### LINKED ISSUE #2326: Missing 1: in ChadoRecords
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

I am encountering this in a test I am working on, where I get the message shown below.

I think we are missing the `1:` prefix here. If I add it then the error goes away.

https://github.com/tripal/tripal/blob/596aebd73d9102156e14f1f6f419d6eba4258d4f/tripal_chado/src/TripalStorage/ChadoRecords.php#L1286

Change to
`$query = $this->connection->select('1:' . $fk_table, 'fk');`

### Steps to reproduce

I'm not sure of an easy way...
This is happening in a widget test during form validation.

### Error messages

```
   │ Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "test19825614cvterm" does not exist
   │ LINE 3: "test19825614cvterm" "fk"
   │         ^: SELECT "fk"."cvterm_id" AS "cvterm_id"
   │ FROM
   │ "test19825614cvterm" "fk"
   │ WHERE "cvterm_id" = :db_condition_placeholder_0; Array
   │ (
   │     [:db_condition_placeholder_0] => 15
   │ )
```

dsenalik:
# Bug Fix

## Description

The affected select statement is missing the `1:` prefix. During automated tests it may fail to find the table as a result.

Then I thought I should check if this is the case anywhere else. I found two widgets lacking the `1:` prefix, and one comment.

## Testing?
Code review should show this is a correct change.
Automated testing will hopefully cover this later as we add widget tests 🙈

### DIFF
--- a/tripal/src/TripalDBX/TripalDbxConnection.php
+++ b/tripal/src/TripalDBX/TripalDbxConnection.php
@@ -41,7 +41,7 @@ use Drupal\tripal\TripalDBX\Exceptions\ConnectionException;
  * query builder as shown in this next example:
  *
  * $dbxdb = \Drupal::service('tripal_chado.database');
- * $query = $dbxdb->select('feature', 'x');
+ * $query = $dbxdb->select('1:feature', 'x');
  * $query->condition('x.is_obsolete', 'f', '=');
  * $query->fields('x', ['name', 'residues']);
  * $query->range(0, 10);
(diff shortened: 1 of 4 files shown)


laceysanderson (approved):
✅ Code changes are great :-)
✅ Quick manual test of creating a genetic map with the unit, dbxref and a contact (foreign key) showed that everything works as it should and no errors are logged.

Image

PR #2330, opened 2025-10-20
Add idspace name to cache key so it is unique

### LINKED ISSUE #2328: Bug in additional type autocomplete term lookup
### Tripal Version

4.x

### Drupal Version

11.2

### PostgreSQL Version

17

### PHP Version

8.4

### General Description

Getting the wrong cvterm_id value from the additional_type field autocomplete.

Here is the bug, we don't put the idSpace in the cache key! We are returning any cached term with the same accession 🙊
https://github.com/tripal/tripal/blob/596aebd73d9102156e14f1f6f419d6eba4258d4f/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php#L284

Proposed fix

`$cache_id = 'chado_id_space_term_' . $this->getName() . '_' . $accession;`

### Steps to reproduce

Functional test is to create a contact of type "Collective (TCONTACT:0000002)" and it is saved as "Publication".

You can also test by evaluating this code:
```
$idspace = 'TCONTACT';
$accession = '0000002';
$idSpace_manager = \Drupal::service('tripal.collection_plugin_manager.idspace');
$idSpace = $idSpace_manager->loadCollection($idspace);
$term = $idSpace->getTerm($accession);
print "Internal ID " . $term->getInternalId() . "\n";
print "Name        " . $term->getName() . "\n";
print "ID Space    " . $term->getIdSpace() . "\n";
print "Accession   " . $term->getAccession() . "\n";
```
Output is
```
Internal ID 212
Name        Publication
ID Space    TPUB
Accession   0000002
```
But should be
```
Internal ID 2990
Name        Collective
ID Space    TCONTACT
Accession   0000002
```

dsenalik:
# Bug Fix

## Description
The caching in this situation is returning **any cached term with the same accession** 😱.
The fix is to add the namespace to the cache key.
This may cause infrequent and non-visible use of incorrect terms, so this is high priority.

## Testing?
See issue for code to evaluate

An additional phpunit test is added to catch the original issue. Without the change to the cache key made by this PR, the test catches this:
```
   │ Did not retrieve expected term ID
   │ Failed asserting that two strings are equal.
   │ --- Expected
   │ +++ Actual
   │ @@ @@
   │ -'some_db_2:0000042'
   │ +'some_db_1:0000042'
```

### DIFF
--- a/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
+++ b/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
@@ -555,7 +555,7 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
     $term->setInternalId($cvterm->cvterm_id);
 
     // Invalidate the cache for this term.
-    $cache_id = 'chado_id_space_term_' . $this->getName() . '_' . $term->getAccession();
+    $cache_id = 'chado_id_space_term_' . $term->getIdSpace() . '_' . $term->getAccession();
     \Drupal::cache()->invalidate($cache_id);
 
     return TRUE;
--- a/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
+++ b/tripal_chado/src/Plugin/TripalIdSpace/ChadoIdSpace.php
@@ -280,7 +280,6 @@ class ChadoIdSpace extends TripalIdSpaceBase implements ContainerFactoryPluginIn
     if (!$this->is_valid) {
       return NULL;
     }
-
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
✅ Coding standard changes look great and are appreciated :-)
✅ Code changes that fix the bug are clean. I agree that the IDspace + accession should be unique according to chado constraints.
✅ Manual testing as described in the issue ensures that the correct term is now being saved!
✅ The new automated test and the fix in a previous automated test make me very happy!

Thanks @dsenalik ❤️ This is ready to merge 🎉 once tests pass.

PR #2333, opened 2025-10-20
Per Tripal Content Type permissions

### LINKED ISSUE #2390: Tripal Content Listing view not supporting permissions
After implementing Content type permissions on PR #2333, when accessing Tripal Content Listing page (`admin/content/bio_data`), users are able to view a list of all the content regardless of their assigned permissions. For example, if a user can access an Analysis but not Organism and there exists one of each, the listing should be filtered to only show the analysis. But currently it's listing both the Organism and Analysis contents on the `Tripal Content Listing` page as shown below in the example. 

Image

We need to make it respect those dynamic permissions and allow the users to only view those contents that they have permissions to access. (Should only list 'Analysis' content type in the example above)

vidu-wil:
# New Feature

## Description

This PR adds the following permissions for each Tripal Entity:
- _Content Type:_ Create new content
- _Content Type:_ Edit any content
- _Content Type:_ Edit own content
- _Content Type:_ Unpublish any content
- _Content Type:_ Unpublish own content
- _Content Type:_ Delete any content
- _Content Type:_ Delete own content
- _Content Type:_ View all content
- _Content Type:_ View own content

Additionally it cleans up all permissions surrounding TripalEntity

- Tripal Content (TripalEntity).
  - Tripal Content Listing: 'access tripal content overview'
  - Publish Tripal Content: 'publish tripal content'
  - Add Tripal Content Listing: ANY 1+ 'create TYPE content' permission(s)
  - The following forms will have TYPE in the route and need the matching dynamic permission with that TYPE as follows:
    - Tripal Entity Add Form: 'create TYPE content'
    - Tripal Entity Edit Form: 'edit TYPE content'
    - Tripal Entity Unpublish Form: 'unpublish TYPE content'
    - Tripal Entity Delete Form: 'delete TYPE content'
  - Any user with the 'administer tripal content' can access any of the above pages without needing any other permission.

And adds a custom views access handler to allow us to easily indicate that a view should respect the Tripal Content permissions!

Screenshot 2025-10-23 at 5 16 13 PM

## Tasks

- [x] Add a permissions class to generate the above permissions for all Tripal Content Types available.
- [x] Update permissions YAML with full plan surrounding TripalEntity + TripalEntityType.
- [x] Update TripalEntityAccessControlHandler to handle per content type TripalEntity CRUD permissions
- [x] Update TripalEntityAccessControlHandler to handle administer override
- [x] Update routing to match other TripalEntity permission expectations
- [x] Resolve permissions for the "Add content" page that lists the types of content you can add using the handler to give you access if you have permission to add any content types.
- [x] Resolve permissions for the "Tripal Content" view and remove the hook implementation for this from the .module file
- [x] Ensure the "Add content" page respects permissions and only shows the content types you have permission to add.
- [x] Generate the options for configuring the View access (remove hard coded-subset). See TripalContentViewAccessHandler::buildOptionsForm().
- [x] Update the automated tests to respect these changes.
- [x] Add additional kernel tests to check the per content type permissions specifically.
- [x] Add additional kernel tests to check TripalContentViewAccessHandler.

### Content type permissions checking

- Create a fresh docker container and login as an admin. 

- Navigate to Admin -> People -> Permissions. New permissions for each content type should appear under 'Tripal' category as shown below.
Screenshot 2025-11-03 at 2 56 39 PM

- Grant 'create', 'view all', 'view own', 'edit any', 'edit own', 'delete any' and 'delete own' permission to 'content editor' for some of the content types.

- Create two users with ‘content editor’ permission (People -> Add user)

- Login to one of the newly created users and go to 'bio_data/add'. It should only show the content types with 'Create new content' permission.

- Create some test content and test whether they behave as expected according to the granted permissions.

### Tripal Content Listing View access checking

1. On a fresh site, go to the Tripal content listing (`admin/content/bio_data`) and confirm that you can access it as long as you have 'view' permissions for any TripalContent type.
2. Edit the Tripal content listing view (`admin/structure/views/view/tripal_content_type_listing/edit/page_1`) and confirm that the access now indicates our special handler!
   Screenshot 2025-10-23 at 5 16 13 PM
3. Click on "view at least 1 of the existing Tripal Content Type(s)" to open up the options and confirm you see the full list of content types for your current site.
   Screenshot 2025-10-23 at 5 18 28 PM
4. Change some of these options and confirm that viewing the listing responds as expected.
5. Import any content type collection then repeat step 3 to confirm that the new content types are now listed.

**Testing the `Tripal Content Listing` page's listing (tests if it respect the assigned permissions**

- Add a curator role by navigating to Admin -> People -> Roles -> Add role.
- Create a curator account by navigating to Admin -> People -> List -> Add user.
- Assign view permissions to analysis content type for the 'Curator' role.
- Create an 'Analysis' content and 'Organism' content.
- Log in to the newly created curator account.
- Navigate to `admin/content/bio_data`.  You should be able to see the 'Analysis' content you created but not the 'Organism' content as shown below.
Screenshot 2025-12-12 at 10 07 17 AM

## Automated Testing

- Added more test cases on `testTripalEntityAccessControlHandler()` and `testTripalContentPages()` for checking new permissions.
- Added the test class TripalContentViewAccessTest to test the new TripalContentViewAccessHandler with the following methods:
     1. testAccess() : Tests the access() method
     2. testBuildOptionsForm() : Tests the buildOptionsForm() method
     3. testSummary(): tests the summaryTitle() method


reynoldtan:
I have completed the manual test and confirmed the following:

1. New permission for each content type appear in the permission page under Tripal.

Screenshot 2025-11-05 at 11 19 33 AM

2. The permission restricts user to content they are allowed to create, view, edit and delete.

image
another user with Organism Tripal Content type only permission.

3. Tripal Content Listing display configuration has the special access handler
image

4. Page access contains full list of content types
image

5. Modifying options causes expected outcome.
image

dsenalik:
1. So before this PR, an anonymous user is not allowed by default to view any tripal content
no_tripal_permission

In this PR it is still the default to not allow anonymous view, the admin has to check each of these "Anonymous User" boxes for every content type.
pr2333 3
I feel the default should be anonymous users CAN access all tripal content by default. Otherwise this creates an extra step for an admin to enable access after importing a type collection, and they may not yet know how to do that when first setting up a site.


2. As shown as I configured here
pr2333 1
my logged-in user cannot access an analysis even though an anonymous user can.
Maybe this is design behavior? I feel like the permission should "cascade" and look like this
pr2333 4
but it looks like this is how drupal normally works, so probably you can ignore this.

tripal/src/Access/TripalAccessOwnContentCheck.php
```diff
@@ -40,7 +41,7 @@ public function access(UserInterface $user, AccountInterface $account) {
     }
 
     return AccessResult::forbidden();
-    // return ($account->hasPermission('administer tripal') && $this->someOtherCustomCondition()) ? AccessResult::allowed() : AccessResult::forbidden();
-
+    // Return ($account->hasPermission('administer tripal') && $this->someOtherCustomCondition()) ? AccessResult::allowed() : AccessResult::forbidden();
```
dsenalik: I think this would be a good time to remove this commented out line, and remove "First" from the comment above, unless there is a reason to leave it.
```suggestion
```

dsenalik:
3. We lose the tripal shortcut menu for a content creator user, and get some blank menus on the left sidebar.
I am wondering if there is any way to get the shortcut menu back, otherwise such a user will have to manually create shortcuts to replace them.
pr2333 2

These empty menus on the left are Drupal behavior, though, nothing to do with this PR.

vidu-wil:
> 1. So before this PR, an anonymous user is not allowed by default to view any tripal content
> 
> no_tripal_permission
> In this PR it is still the default to not allow anonymous view, the admin has to check each of these "Anonymous User" boxes for every content type. pr2333 3 I feel the default should be anonymous users CAN access all tripal content by default. Otherwise this creates an extra step for an admin to enable access after importing a type collection, and they may not yet know how to do that when first setting up a site.
> 
I verified that Drupal, by default, grants permission to the anonymous user to view published content. So, as @dsenalik suggested, I set the default permissions for the anonymous role to include all “view all CONTENT_TYPE content” permissions.



> 3. We lose the tripal shortcut menu for a content creator user, and get some blank menus on the left sidebar.
>    I am wondering if there is any way to get the shortcut menu back, otherwise such a user will have to manually create shortcuts to replace them.
> 
> pr2333 2
> These empty menus on the left are Drupal behavior, though, nothing to do with this PR.

As you mentioned, it is Drupal’s default behaviour for the side bar items to disappear in content editor user’s view. However, for tripal shortcut to appear you can grant the following permission to content editor 
Screenshot 2025-11-21 at 10 01 00 AM

and then grant which permissions you want to grant as shown below. 
Screenshot 2025-11-21 at 10 13 17 AM

The side bar will display ‘Tripal’ and its shortcuts depending on the permissions you grant.

dsenalik:
View Tripal Administration Pages permission grants menu access, thanks for that!
Tested the update hook and anonymous permissions granted as expected, thanks for that too!

tripal/src/Access/TripalEntityAccessControlHandler.php
```diff
@@ -41,15 +130,312 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+      }
+      return FALSE;
+    }, ARRAY_FILTER_USE_BOTH);
+
+    // Finally, return just the permission names.
+    return array_keys($filtered_permissions);
+  }
+
+  /**
+   * Returns an array of content types the user has 'create TYPE content'.
```
dsenalik:
```suggestion
   * Returns an array of content types the user can 'create TYPE content'.
```

tripal/src/Plugin/views/access/TripalContentViewAccessHandler.php
```diff
@@ -0,0 +1,196 @@
...
+    // TripalEntity access controller.
+    $route->setRequirement('_custom_access', static::class . '::checkManyPermissions');
+    // Sets the options that this method expects.
+    $route->setOption('_content_types', $this->options['content_types']);
+    $route->setOption('_comparison_mode', $this->options['mode']);
+    $route->setOption('_permission_operation', $this->options['operation']);
+  }
+
+  /**
+   * Grabs paramters from route for use with the TripalEntity Access Controller.
```
dsenalik:
```suggestion
   * Grabs parameters from route for use with the TripalEntity Access Controller.
```

tripal/tripal.install
```diff
@@ -1070,3 +1071,33 @@ function tripal_update_10403() {
     ])->save();
   }
 }
+
+/**
+ * Implements hook_update_10404().
+ *
+ * Updates hook to grant new permissions to anonymous users.
```
dsenalik:
```suggestion
 * Update hook to grant new permissions to anonymous users.
```

tripal/src/Access/TripalEntityAccessControlHandler.php
```diff
@@ -2,37 +2,126 @@
...
+    // the following permissions.
     if ($account->hasPermission('administer tripal content')) {
       return AccessResult::allowed();
     }
 
+    // For view, update, delete we want to use bundle-specific permissions
+    // so lets get the bundle now.
+    $entity_bundle = $entity->getType();
+
+    // Now, lets check the permission based on the bundle and operation.
```
dsenalik:
```suggestion
    // so let's get the bundle now.
    $entity_bundle = $entity->getType();

    // Now, let's check the permission based on the bundle and operation.
```
Minor fussy change 🙄

dsenalik (dismissed):
A few trivial changes from code review.

I created a user with these permissions


and this user was able to import an organism, and to generate a taxonomy tree, but from the permissions I would think that the tree should not be allowed.

The organism imported cannot be edited, however an organism manually created through the UI can be edited, unpublished, etc. ~~I suspect the owner is not handled properly by the importer, so this might be worthy of a separate issue.~~ It is simply that the user running the tripal job is the user of the published content. If I run the job as admin (as I did) then that is the owner, if I run the job as the user, then it works as expected. No problem here.

✅ Permissions to add new content work as expected.
✅ Importing a type collection assigns default permissions as expected.

Maybe I am not clear how this should work, but I have turned off "Access Tripal Content Listing" and have all permissions denied for Analysis and Breeding Cross Progeny. I set the views as follows

My content editor user can view the full tripal content listing. From your description I wonder if access should be denied?

tripal/src/Plugin/views/access/TripalContentViewAccessHandler.php
laceysanderson: Fixed in c44dfdccbab6aa2572f29d8d96ea183cc43135ab because this suggestion brought it over the maximum line length.

tripal/tripal.install
laceysanderson: This is actualy what gets shown to the admin when they run an update so I'd be more likely to say something like 'Grants anonymous user permission to view Tripal Content by default.'

laceysanderson:
> Maybe I am not clear how this should work, but I have turned off "Access Tripal Content Listing" and have all permissions denied for Analysis and Breeding Cross Progeny. I set the views as follows
> 
> My content editor user can view the full tripal content listing. From your description I wonder if access should be denied?

Yes, you should have gotten access denied. There was a bug here that I believe has now been fixed. For some reason, the 'all' option was staying in the options array even though it wasn't visibly selected. I've also now added some helpful information to that form.

laceysanderson:
Current changes to be addressed:

- [x] Users should only see the content they have permission to see in the Tripal Content listing regardless of the settings chosen in the access handler. For example, if a user can access an Analysis but not Organism and there exists one of each, the listing should be filtered to only show the analysis. I suspect we need to implement `hook_query_TAG_alter` so that views respects our permissions and only includes entities the user has permission to see.
- [x] Vidusha: Remove the `access tripal content overview` from tripal.permissions.yml
- [x] Vidusha: Create an update hook that migrates an existing sites permissions to the new approach. For example, if a role has `view all tripal content` then the update hook should assign every `view all CONTENT TYPE content` permission to that role. This needs to be done for all existing roles and view, add, edit, and delete. We will not assign permissions to any of the `own` permissions or unpublish. 
- [x] We need to fix the tests 🙈 These are the changes that broke the tests: [commits](https://github.com/tripal/tripal/pull/2333/files/39c39f9f43987c03d1579201b84eb33f90c57602..c44dfdccbab6aa2572f29d8d96ea183cc43135ab). We suspect it is regarding `$this->messenger->addMessage($rendered_message, 'warning');` or `$bundle_entities = TripalEntityAccessControlHandler::checkCreateAccessForTypes($this->currentUser());`. However, the add Content page does work when manually tested so it's a weird test environment thing.

vidu-wil:
> Current changes to be addressed:
> 
> * [x]  Users should only see the content they have permission to see in the Tripal Content listing regardless of the settings chosen in the access handler. For example, if a user can access an Analysis but not Organism and there exists one of each, the listing should be filtered to only show the analysis. I suspect we need to implement `hook_query_TAG_alter` so that views respects our permissions and only includes entities the user has permission to see.

This is described in a separate issue (see issue #2390), but it has been implemented as part of this PR as well.

tripal/src/Hook/TripalViewsHooks.php
```diff
@@ -70,4 +34,45 @@ public function viewPresave(ViewEntityInterface $view): void {
...
+   *
+   * Alters the Tripal Content Type Listing view to filter content types
+   * based on the current user's permissions.
+   */
+  #[Hook('views_query_alter')]
+  public function viewsQueryAlter(ViewExecutable $view, QueryPluginBase $query) {
+    // Only alter the Tripal Content Type Listing view.
+    if ($view->id() !== 'tripal_content_type_listing') {
+      return;
+    }
```
laceysanderson: We don't only want to do this on this specific view but rather all views of Tripal Content. We should be able to use the base_table setting in the view to determine if it shows Tripal Content. Maybe something like `$view->storage->get('base_table') == 'tripal_entity'`?

tripal/src/Hook/TripalViewsHooks.php
```diff
@@ -70,4 +34,45 @@ public function viewPresave(ViewEntityInterface $view): void {
...
+
+    // Load all Tripal entity types.
+    $entity_types = \Drupal::service('entity_type.manager')
+      ->getStorage('tripal_entity_type')
+      ->loadByProperties([]);
```
laceysanderson: It's not performant to load all the entity types for this. It's better to use `TripalEntityAccessControlHandler::getTripalContentPermissionsList('view');` to get the list of permissions. This also allows us to not make any assumptions about the format of the permission string.

tripal/src/Hook/TripalViewsHooks.php
```diff
@@ -70,4 +34,45 @@ public function viewPresave(ViewEntityInterface $view): void {
...
+        $allowed_types[] = $entity_id;
+      }
+    }
+
+    if (empty($allowed_types)) {
+      // If the user doesn't have permission to view any entity type,
+      // return no rows.
+      $query->addWhereExpression(0, 'null');
+      return;
+    }
```
laceysanderson: I suspect this is not needed and would be tempted to not include it. While it would be faster to use `$query->addWhereExpression(0, 'null');` I worry it would be less obvious to any developers debugging views queries where this is coming from.

tripal/src/Hook/TripalViewsHooks.php
```diff
@@ -32,4 +34,45 @@ public function viewPresave(ViewEntityInterface $view): void {
     }
   }
 
+  /**
+   * Implements hook_views_query_alter().
+   *
+   * Alters the Tripal Content Type Listing view to filter content types
+   * based on the current user's permissions.
```
laceysanderson:
```suggestion
   * Adds a where clause to any view showing Tripal Content which
   * restricts the results to Tripal Content Types that the current
   * user has permission to view.
```

tripal/src/Hook/TripalViewsHooks.php
laceysanderson: Let's also ensure we do not add a where clause at all if the current user has `'administer tripal content'` just like we do in our AccessControlHandler.

laceysanderson (commented):
Results of code review :-) Getting closer!

laceysanderson (commented):
✅ Code looks good to me 👍 

## Views functionality

✅ "Content editor" can see organism and contacts but not projects in the Tripal content listing --as expected.
✅ "Content editor" can see contacts only in my test view (not projects) --as expected.
✅ "Researcher" can go to my test view but does not see any rows (not contacts or projects) --as expected.
✅ "Researcher" can go to the tripal content listing but only sees the organism --as expected.
✅ Based on the above two observations on my test view, the logic to filter private content is now working on any tripal content view not just those using the handler and not just the listing.

## Setup description

Screenshot 2026-01-12 at 4 22 01 PM

Screenshot 2026-01-12 at 4 23 31 PM

Screenshot 2026-01-12 at 4 28 30 PM

dsenalik:
I am thinking we need to adjust the default authenticated user view permissions.
Drupal default permissions for node looks like this, i.e. anyone can see published content by default
pr2333 0114 1

The tripal permission by default is not giving authenticated users view permissions, I think we should match the Drupal default here. If we add authenticated user then we automatically get content editor too.
pr2333 0114 2

dsenalik:
There is probably something wrong with importer permissions.
I created a user with content editor permission.
I modified importer permissions to grant the content editor permission to run them all:
pr2333 0114 3
I logged in as the editor user
Only one importer is listed
pr2333 0114 4
And if I try to create a new query here
pr2333 0114 5
I get denied permission
pr2333 0114 6

1. The fact that the pub importer shows in the listing is maybe a bug
2. But the other bug is that permission is denied for all importers.

laceysanderson:
The importer issue exists in 4.x too --just confirmed. As such we should create a separate issue for that.

> The tripal permission by default is not giving authenticated users view permissions, I think we should match the Drupal default here. If we add authenticated user then we automatically get content editor too.

Good catch on the authenticated user! @vidu-wil can you add that ASAP?

laceysanderson:
See issue https://github.com/tripal/tripal/issues/2402 for the importer bug described by @dsenalik.

dsenalik (approved):
I have gone through this one more time and everything looks good to me. (I did come up with an unrelated issue #2406 which is not related to this PR.)

laceysanderson (approved):
Thanks for all the work on this one! Ready to merge once the branch is updated and status' all pass.

PR #2338, opened 2025-10-23
Method to update a table schema without data loss

### LINKED ISSUE #2337: Method to update a table schema without data loss

Yes

Yes

### Is this feature urgent?

No

### Branch

`tv4g2-issue2337-update-table-schema`

### Description

It would be useful to have a way to update a table schema without having to do it through SQL commands.
Currently chado custom tables only support this by destroying and recreating the table, which means all data is lost 😲.
My first thought is to create a new function in `tripal/src/TripalDBX/TripalDbxSchema.php`
`updateTableSchema($table_name, $schema)`

### Your specific use case

When migrating the Tripal File module from Tripal 3 to Tripal 4, I have added `type_id` columns to the linker tables. I would like to be able to apply the Tripal 4 schema for tables migrated from Tripal 3, which are full of data, which means adding the `type_id` columns, as well as updating unique constraints.

### Generally Applicable

Chado custom tables could use this, so that the table schemas could be updated by a site admin without data loss.

I would like to see the custom table form would change from

Image

to

Image

dsenalik:
# New Feature

## Description

This provides a way to update a table schema without having to do it through SQL commands.
Currently chado custom tables only support this by destroying and recreating the table, which means all data is lost 😲.
This PR adds a new function in tripal/src/TripalDBX/TripalDbxSchema.php
updateTableSchema($table_name, $schema)

This function is then used in `ChadoCustomTable.php` when a schema is changed.

Additional code in the test is added to cover this new functionality.

Coding standards were (mostly) limited to the first commit 4a40a1f - this has been transferred to PR #2358

This does not support adding a column with a NOT NULL constraint if there is data in the table. If you really wanted to do that, you could add first without the constraint, then populate the column, then update a second time with the constraint.

## What is horribly wrong?

With the method used here, if there are foreign keys, they would have to be found and removed at the start, and regenerated at the end. This is not implemented, and this is already starting to get really ugly, so I am doubtful that this is even a good approach.

## Testing?

`analysis_organism` is a chado custom table present in a default tripal installation

1. Create an analysis
2. Create an organism
3. `INSERT INTO analysis_organism (analysis_id, organism_id) values (1, 1);`
4. `SELECT * FROM analysis_organism;`
` analysis_id | organism_id`
`-------------+-------------`
`           1 |           1`
`(1 row)`
5. Tripal -> Data Storage -> Chado -> Chado Custom Tables
6. For `analysis_organism` select "Edit"
7. Check "Re-create table"
8. Add a `type_id` column
```
...
    'organism_id' => array (
      'size' => 'big',
      'type' => 'int',
      'not null' => true,
    ),
    'type_id' => [
      'type' => 'int',
      'not null' => false,
    ],
  ),
  'indexes' => array (
...
```
9. Save
10. `SELECT * FROM analysis_organism;`
` analysis_id | organism_id | type_id`
`-------------+-------------+---------`
`           1 |           1 |`
`(1 row)`
Note new column is added, but data is retained.

### DIFF
--- a/tripal/src/TripalDBX/TripalDbxSchema.php
+++ b/tripal/src/TripalDBX/TripalDbxSchema.php
@@ -276,7 +276,7 @@ abstract class TripalDbxSchema extends PgSchema {
    *     - 'blob_fields' that lists all the blob fields in the table.
    *     - 'sequences' that lists the sequences used in that table.
    */
-  public function queryTableInformation(string $table) {
+  public function queryTableInformation($table) {
 
     // Generate a key to reference this table's information on.
     $key = $this->connection
@@ -376,7 +376,7 @@ EOD;
    * @return bool
    *   TRUE if the given index exists, otherwise FALSE.
    */
-  public function indexExists(string $table, string $index_name, bool $exact_name = FALSE) :bool {
+  public function indexExists($table, $index_name, bool $exact_name = FALSE) :bool {
     if ($exact_name) {
       return (bool) $this->connection
         ->query("SELECT 1 FROM pg_indexes WHERE indexname = '{$index_name}'")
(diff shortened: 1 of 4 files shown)


dsenalik:
Closing this PR because this should be implemented differently than the current method.

Issue #2339, opened 2025-10-23
Tests utilizing the getTestSchema() method in ChadoTestTrait get PHP deprecation error on PHP 8.4


carolyncaron:
### Tripal Version

4.0.alpha4

### Drupal Version

11.3-dev

### PostgreSQL Version

17.6

### PHP Version

8.4.13

### General Description

I'm in the process of updating our docker images and testing grids in our Tripal extension module TripalCultivate to PHP 8.4 (see PR https://github.com/TripalCultivate/TripalCultivate/pull/90). Our workflows with PHP 8.4 are failing and giving multiple errors that refer to line 314 in ChadoTestTrait.php: https://github.com/tripal/tripal/blob/96967cb5ffd9f5b228f299959ffb8c4cc37d597a/tripal_chado/tests/src/Traits/ChadoTestTrait.php#L314

This line should be updated to:
```
protected function getTestSchema(?int $init_level = NULL, string $version = '1.3') {
```
and $init_level parameter in the docblock should be typed `int|NULL`.

Also, potentially a simple test calling this method can be added to ensure the method gets called by Tripal workflows.

... I suspect there may be a few more methods like this that could use updating and tests 🙈 

### Error messages

```shell
Drupal\Tests\tripal_chado\Traits\ChadoTestTrait::getTestSchema(): Implicitly marking parameter $init_level as nullable is deprecated, the explicit nullable type must be used instead in /var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Traits/ChadoTestTrait.php on line 314
Deprecated: Drupal\Tests\tripal_chado\Traits\ChadoTestTrait::getTestSchema(): Implicitly marking parameter $init_level as nullable is deprecated, the explicit nullable type must be used instead in /var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Traits/ChadoTestTrait.php on line 314
```

dsenalik:
This is also caught by `phpcs` (#2215) so if you are ambitious you can fix all coding standard issues in that file 😁.
`FOUND 50 ERRORS AND 4 WARNINGS AFFECTING 51 LINES`
`PHPCBF CAN FIX THE 40 MARKED SNIFF VIOLATIONS AUTOMATICALLY`

carolyncaron:
[...]

PR #2341, opened 2025-10-23
Update ChadoTestTrait to resolve PHP 8.4 deprecations and meet coding standards

carolyncaron:
# Bug Fix

## Description

This PR applies the code change brought up in issue #2339 to the ChadoTestTrait in order to comply with PHP 8.4 and remove deprecation errors in methods that use the `getTestSchema()` method.

 https://github.com/tripal/tripal/blob/96967cb5ffd9f5b228f299959ffb8c4cc37d597a/tripal_chado/tests/src/Traits/ChadoTestTrait.php#L314

This line has been updated to:
```
protected function getTestSchema(?int $init_level = NULL, string $version = '1.3') {
```
and $init_level parameter in the docblock is now typed `int|null`.

In a separate commit, fixes have been made to the "easier" PHP coding standard issues. Some issues have been left alone due to probable downstream consequences - in particular, renaming the `getCvtermID()` method. 😅 

## Testing?

This PR is missing automated tests. I gave it a try but I'm feeling very out of element here. Please feel free to contribute a test if you know how to write one and where to put it. Even if it is *just* for `getTestSchema()` method 😁

Also, ensure all existing automated tests in Tripal's workflows still pass and have not been affected by the code changes made in this PR!

### DIFF
--- a/tripal_chado/tests/src/Traits/ChadoTestTrait.php
+++ b/tripal_chado/tests/src/Traits/ChadoTestTrait.php
@@ -298,10 +298,11 @@ trait ChadoTestTrait  {
 
   /**
    * Gets a new Chado schema for testing.
-   * Retrieves the current test schema.
-   * If there is not currently a test schema set-up then one will be created.
    *
-   * @param int $init_level
+   * Retrieves the current test schema. If there is not currently a test schema
+   * set-up then one will be created.
+   *
+   * @param int|null $init_level
    *   One of the constant to select the schema initialization level.
    *   If this is supplied then it forces a new connection to be made for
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
Thank you for fixing all these.
I very much agree with not changing `getCvtermID()` because it will surely break things. We can mark this so phpcs will ignore that, I went ahead and pushed a commit to do so. And I went ahead and added the `use` statements, so now phpcs doesn't complain about everything.
If you also approve, this can be merged once the tests pass.

carolyncaron:
Thank you for doing those last few fixes! They look great to me 😄

PR #2346, opened 2025-10-27
Add two missing links to tripal admin menu

### LINKED ISSUE #2323: We have two different "Tripal" admin menus
### Task Description

There are two different admin Tripal menus that have the same name, but different options.

**One at the top right**
Image
This is defined in `tripal/tripal.module function tripal_toolbar()`

**The other in the left sidebar**
Image
This is defined in `tripal/tripal/tripal.links.menu.yml`

The "more" option on the first menu leads to the same options as the second one.

This makes it hard to write clear documentation as just writing "Tripal" is ambiguous, because the two menus have the same icon and name, but contain different options.

I propose renaming one or both.
The other option is to make both menus contain the same options.

### Plan

After discussing in cofest and slack, we will just add the two missing menu options to the left sidebar menu.

### Branch Name

tv4g0-issue2323-tripal-admin-menus

dsenalik:
# Bug Fix

## Description

This PR adds two menu items to the Tripal administrative menu. These two items were formerly only available from the shortcut bar menu (upper right), they are now added to the tripal admin menu (left side menu). The new items are shown with an asterisk here:

## Benefit

This will make writing documentation easier. If a particular item is present in both "Tripal" menus, we don't have to go into such detail ("Upper right", "Left side") to distinguish the menus.

## Testing?

If you build a docker on this branch, the two new menu items should be available, and you can just make sure that they go to the desired destinations.

If you build a docker on 4.x, then switch to this branch, and rebuild caches, the new menu items should be present.

### DIFF
--- a/tripal/tripal.links.menu.yml
+++ b/tripal/tripal.links.menu.yml
@@ -37,7 +37,7 @@ tripal.cv_lookup:
 # -- The root Tripal administration menu item.
 tripal.admin:
   route_name: 'tripal.admin'
-  title: 'Tripal Administration'
+  title: 'Tripal'
   description: 'The core administrative menu for Tripal. This menu listing provides tools to manage the behavior or Tripal and its various modules and extensions.'
   parent: 'system.admin'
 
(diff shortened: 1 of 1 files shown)


laceysanderson (approved):
✅ Coding changes are clean and make perfect sense :-)
✅ Manual testing shows that "Content" and "Page Structure" are now listed in both menus. Thanks @dsenalik for this quality of life fix!

Image

PR #2349, opened 2025-10-28
Update chado buddy associateDbxref to not fail if already associated

### LINKED ISSUE #2300: Update chado buddy associateDbxref to not fail if already associated
### Task Description

This is a reminder to move the fix from commit 23a4454 from PR #2297 into the chado buddy function so anything can benefit from it.
And to add a test for this very situation.

### Branch Name

tv4g2-issue2300-associateDbxref

dsenalik:
# Bug Fix

## Description

This moves the fix made in commit https://github.com/tripal/tripal/commit/23a44545f5b86e9218e17b4b4af042cb6569b5ec from PR #2297, where trying to creating a dbxref that already exists caused an error. The check is now done in the buddy itself, commit https://github.com/tripal/tripal/pull/2349/commits/538f2f303966a27a7c97ec164ccda161c7a6a8ac.
The check in the taxonomy importer is now no longer needed, so that is removed in commit https://github.com/tripal/tripal/pull/2349/commits/1b03722b097dd9688d9e04048a5b6684c6e4eb21

Low priority because there is no known case where this is causing an actual problem, it is just a potential bug.

## Testing?

A phpunit test for this has been added (Commit https://github.com/tripal/tripal/pull/2349/commits/01d7f4ae4f565b8d3c292f38d017ce75f0f667bd), so probably just code review needed. I added the test first so I could be sure it would catch this, before I fixed it.


dsenalik:
I notice that we should treat `associateCvterm()` similarly, to not fail if already associated. Done in commit https://github.com/tripal/tripal/pull/2349/commits/aa4cd131a589acce8deba313ac66b0ec6178dc51

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
```diff
@@ -627,20 +627,31 @@ public function associateDbxref(string $base_table, int $record_id, ChadoBuddyRe
...
+          'dbxref_id' => $dbxref->getValue('dbxref.dbxref_id'),
+          $base_pkey_col => $record_id,
+        ];
+
+        // Add in any of the other columns for the linking table.
+        foreach ($options as $key => $value) {
+          if ($key != 'pkey') {
+            $fields[$key] = $value;
+          }
+        }
```
laceysanderson, line 650: Why do you move this part inside the `if (!$count) {` in DbxrefBuddy but not in CvtermBuddy? Both are likely right but maybe lets be consistent unless there's a reason not to?

laceysanderson (approved):
Code review looks pretty good on this one. Only thing I noticed was that the cvterm/dbxref buddies are slightly different when they compile the field values. (see inline comment).

I don't require this to be changed but thought I'd mention it in case consistency would be preferable :-)

Marking approved since it doesn't really need to be changed.

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
dsenalik: Doh! Yes they should be consistent!

PR #2350, opened 2025-10-28
Chado buddy should support custom tables

### LINKED ISSUE #2347: Make sure chado buddies support chado custom tables
### Task Description

I see this

https://github.com/tripal/tripal/blob/4c4f29ecfc9f0a0b897618589d00f54bfc0be800/tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php#L605

and the task is to make this compatible with chado custom tables or migrated chado tables by using all three possible `source`s, wherever it occurs in the buddies.

### Branch Name

tv4g2-issue2347-chado-buddy-custom-tables

dsenalik:
# New Feature

## Description

In a few places in the Chado Buddy code, `getTableDef()` is called. Calling this way does not support chado custom tables, or non-core chado tables. We fixed this for fields in PR #2260, the changes here make sure that the chado buddy functions also support this.

Low priority because we don't yet have a known case where this is causing a problem.

## Testing?

I added a non-core chado table to the tests, which caught a bug due to difference when getting unique keys from core schema vs. from database (commit https://github.com/tripal/tripal/pull/2350/commits/414368bc5187255ccf1d52b2ba8b44e444eed857)

Manual testing could involve running one or more importers that use chado buddies, _i.e._ GFF3 importer and TaxonomyImporter. PR #2206 has a sample GFF you could use, and the first part of #1927 has taxonomy importer test steps.


tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
```diff
@@ -73,6 +73,28 @@ public function description(): string {
...
+  public function getChadoTableDef(string $table_name): array {
+    $parameters = [
+      'format' => 'drupal',
+      'source' => [
+        'file',
+        'tripal',
+        'database',
+      ],
+    ];
+    $def = $this->chado_connection->schema()->getTableDef($table_name, $parameters);
```
laceysanderson: Instead of making the change in TripalDbx which results in a substring check on the exception, I would be tempted to wrap this call in a check for the table.

```suggestion
    $def = NULL;
    if ($this->chado_connection->schema()->checkTableExists($table_name)) {
      $def = $this->chado_connection->schema()->getTableDef($table_name, $parameters);
    }
```

Alternatively, if you see reason to ensure that getTableDef() doesn't throw an exception for missing table then I would use the check for the table existing around the call there so an exception is not thrown.

tripal/src/TripalDBX/TripalDbxSchema.php
```diff
@@ -1002,15 +1003,25 @@ public function getTableDdl(
...
+            AS \"definition\";
+        ";
+        $result = $this->connection->query(
+            $sql_query,
+            [':schema' => $schema_name, ':table' => $table_name, ]
+        );
+      }
+      catch (DatabaseExceptionWrapper $e) {
+        // We will only catch the case of "table does not exist",
+        // and in that case we return an empty string.
```
laceysanderson: Comment doesn't match the default set on line 1006. Should it be this instead?
```suggestion
        // and in that case we return NULL.
```

laceysanderson (commented):
Code review brings up the following thoughts. Otherwise the code looks good. I did not yet do a manual test.

tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
dsenalik: I have done very similar to your suggestion with commit https://github.com/tripal/tripal/pull/2350/commits/fc31b41f644c12b5218717eb5e201b2114435e2c

tripal/src/TripalDBX/TripalDbxSchema.php
dsenalik: No longer relevant, changes to TripalDbx were reverted with commit https://github.com/tripal/tripal/pull/2350/commits/ed8d17622655b20304f7f2b0d8039b70c1eeb664

laceysanderson (approved):
✅ Code looks great 👍 Thanks for the changes @dsenalik 
✅ Manually tested the GFF3 importer and no issues related to ChadoBuddies came up :-)
✅ Verified the automated test locally to confirm this works for a custom table.

This is ready to merge once tests pass! 🎉

PR #2356, opened 2025-11-04
Deprecated in Drupal 11.3 and removed in Drupal 12

dsenalik:
# Bug Fix

## Description

This PR addresses deprecations mentioned in issue #2352.

It is important that we don't break Drupal 10.x though, so we need to retain the deprecated hooks until Drupal 10 end-of-life (Dec. 2026), but we can include `#[LegacyHook]` attributes so Drupal knows not to use them for Drupal 11. By creating simple hook services, the legacy hooks can call the code in the new hook so there is no need to duplicate any code.

### Updated
 - [x] hook_theme https://github.com/tripal/tripal/pull/2356/commits/8ecfc3d0839596c070d8650071fe07caef8449d7 and https://github.com/tripal/tripal/pull/2356/commits/40963c4e2e4101da91cead2851dd79e3626dd0d9
 - [x] hook_theme_suggestions_HOOK https://github.com/tripal/tripal/pull/2356/commits/8ecfc3d0839596c070d8650071fe07caef8449d7 and https://github.com/tripal/tripal/pull/2356/commits/40963c4e2e4101da91cead2851dd79e3626dd0d9
 - ❌ **Exclude deprecations in external modules**. This is a bit involved because Drupal is already hiding some deprecations, we need to append to that list. This is a bit of cause for worry if these are never fixed in their respective modules. https://github.com/tripal/tripal/pull/2356/commits/cf2163647319754bdbeefb9e269b5810b70673b0 - **This will be moved to a separate PR #2361**

Now I am assuming there is no test converage on the following because at this point there are no more deprecations, but these are the additional hooks that will of course be big problems in Drupal 12:
 - [x] hook_help https://github.com/tripal/tripal/pull/2356/commits/03fb41cede00ff04bdf1103e097be3c9a3d73ffc
 - [x] hook_page_attachments https://github.com/tripal/tripal/pull/2356/commits/91272924319aaf8e99095d1f1610bd021d03a0f6
 - [x] hook_rebuild - for this one we create new rebuild services https://github.com/tripal/tripal/pull/2356/commits/a3cc2fed71603b5b4544b1e92256cd3bb14b275f
 - [x] hook_form_alter https://github.com/tripal/tripal/pull/2356/commits/312f3173a0a8295d860fd9fd9ef173fef1e6a227 and https://github.com/tripal/tripal/pull/2356/commits/a20f647ababf2b3f20f03b87ae521682db91ad7a
[...]

PR #2361, opened 2025-11-17
Deprecated in Drupal 11.3 and removed in Drupal 12 additional exclude rules

### LINKED ISSUE #2352: Deprecated in Drupal 11.3 and removed in Drupal 12
### Task Description

We need to update our codebase to remove deprecations from Drupal 11.3. These are currently being highlighted in the Drupal 11.x test run and are shown when running tests on our devcontainer.

Feel free to start a branch focused on one or more of these deprecations. They need not all be fixed at once.

See the linked Drupal page for the specific change record showing before and after examples of how to fix this. Then check the most recent test run on Github for Drupal 11.x to see which test files are triggering that deprecation.

### Deprecations

- [ ] field_group_module_implements_alter without a #[LegacyModuleImplementsAlter] attribute is deprecated in drupal:11.2.0 and removed in drupal:12.0.0. See https://www.drupal.org/node/3496788
- [ ] Method "Twig\Extension\ExtensionInterface::getFunctions()" might add "array" as a native return type declaration in the future. Do the same in implementation "Drupal\devel\Twig\Extension\Debug" now to avoid errors or add an explicit @return annotation to suppress this message.
- [ ] Providing a file for theme hook tripal_entity_type is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3549500
- [ ] Providing template_preprocess_tripal_entity_type() is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3504125
- [ ] Providing a file for theme hook tripal_entity is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3549500
- [ ] Providing template_preprocess_tripal_entity() is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3504125
- [ ] Providing a file for theme hook tripal_entity_content_add_list is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3549500

### Branch Name

tv4g9-2352-deprecations

dsenalik:
# Bug Fix

## Description

This adds some deprecation exclusion rules from external modules. This was moved from PR #2356 to separate out the types of things being done.

## Testing?

Automated tests under 11.x should pass without deprecations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ tripaldocker/dev/.env
 vendor
 .phpunit.result.cache
 .phpunit.cache
+.deprecation-ignore.txt
 clover.xml
 cc-test-reporter
 coverage/
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,7 +5,7 @@
  https://www.drupal.org/node/2116263 for details.
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         bootstrap="../../../core/tests/bootstrap.php"
+         bootstrap="../../../modules/contrib/tripal/tripal/tests/tripal-bootstrap.php"
          colors="true"
          beStrictAboutTestsThatDoNotTestAnything="true"
          beStrictAboutOutputDuringTests="true"
@@ -36,6 +36,7 @@
       correctly.
     -->
     <!-- <env name="SYMFONY_DEPRECATIONS_HELPER" value="ignoreFile=.deprecation-ignore.txt"/> -->
+    <env name="SYMFONY_DEPRECATIONS_HELPER" value="ignoreFile=../../../modules/contrib/tripal/.deprecation-ignore.txt"/>
     <!-- Example for changing the driver class for mink tests MINK_DRIVER_CLASS value: 'Drupal\FunctionalJavascriptTests\DrupalSelenium2Driver' -->
     <env name="MINK_DRIVER_CLASS" value=""/>
     <!-- Example for changing the driver args to mink tests MINK_DRIVER_ARGS value: '["http://127.0.0.1:8510"]' -->
--- a/tripal/tests/tripal-bootstrap.php
+++ b/tripal/tests/tripal-bootstrap.php
@@ -1,6 +1,7 @@
 <?php
 
 /**
+ * @file
  * Tripal preprocessing of bootstrap.php.
  *
  * The purpose of this file is to append a few additional phpunit deprecation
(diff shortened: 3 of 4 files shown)


laceysanderson:
So I figured out what was lurking in my brain about "a better way to do this"! This actually came up when I was working on deprecating some of our own code and wanted to keep the tests checking it.

PHPUnit has the concept of three scopes of deprecations being triggered:

- [The <ignoreSelfDeprecations> Attribute](https://docs.phpunit.de/en/11.5/configuration.html#appendixes-configuration-source-ignoreselfdeprecations) setting can be used to ignore deprecations triggered by first-party code in first-party code
- [The <ignoreDirectDeprecations> Attribute](https://docs.phpunit.de/en/11.5/configuration.html#appendixes-configuration-source-ignoredirectdeprecations) setting can be used to ignore deprecations triggered by first-party code in third-party code
- [The <ignoreIndirectDeprecations> Attribute](https://docs.phpunit.de/en/11.5/configuration.html#appendixes-configuration-source-ignoreindirectdeprecations) setting can be used to ignore deprecations triggered by third-party code

Where "first-party" code includes the files defined within the `<source>` tags in our `phpunit.xml`.

This means we can stop our tests failing on the two deprecations mentioned above (and any others by external modules) by adding `ignoreIndirectDeprecations="true"` into our source tag like follows:

```xml
  <source
    ignoreSuppressionOfDeprecations="true"
    ignoreIndirectDeprecations="true">
    <include>
      <directory suffix=".php">tripal/src</directory>
```

I tested this locally on `tv4g9-2352-deprecations` branch which does not include the code in this PR. Before making the change I see it fail just like we do in #2356 and after this change the tests do not fail and do not populate our logs.

Do we want to switch to this config change rather then overriding drupal bootstrap? It is a more maintainable change but it is also not as specific as your current implementation.

dsenalik:
I think I favor this new method, mainly because it is far less complex and as you say more maintainable. The downside of course is we won't see new deprecations in external modules.
I will leave this PR as it is, but close it, just in case we ever want to use this method. Though I do want to let the tests finish just to know the result.
I will create a new PR for this method.

PR #2362, opened 2025-11-17
Deprecated in Drupal 11.3 and removed in Drupal 12 for external modules

### LINKED ISSUE #2352: Deprecated in Drupal 11.3 and removed in Drupal 12

dsenalik:
# Bug Fix

## Description

This deals with two deprecations described in issue #2352, i.e.
 - field_group_module_implements_alter without a #[LegacyModuleImplementsAlter] attribute is deprecated in drupal:11.2.0 and removed in drupal:12.0.0. See https://www.drupal.org/node/3496788
 - Method "Twig\Extension\ExtensionInterface::getFunctions()" might add "array" as a native return type declaration in the future. Do the same in implementation "Drupal\devel\Twig\Extension\Debug" now to avoid errors or add an explicit @return annotation to suppress this message.

We do not actually fix them! Because they are in external modules, what we do is ignore them 🙈 by using `ignoreIndirectDeprecations`.
This is of course not ideal, but they do need to be dealt with in their respective modules.
There is precedent, Drupal is currently ignoring several deprecations. Look in a docker at the file
`/var/www/drupal/web/core/.deprecation-ignore.txt` to see them. They reference https://www.drupal.org/node/3285162 for more details. This is a more fine-grained approach to know exactly what you are ignoring. PR #2361 was an attempt to use this approach, but is a lot more complicated and maybe less maintainable than this approach.

## Testing?

Phpunit should pass for 11.x without any deprecation messages.

### DIFF
--- a/.github/workflows/ALL-phpunit.yml
+++ b/.github/workflows/ALL-phpunit.yml
@@ -96,10 +96,14 @@ jobs:
           docker exec tripaldocker service postgresql restart
           docker exec tripaldocker bash -c 'export drupalversion="'${{ matrix.drupal-version }}'"; \
             if $(dpkg --compare-versions "${drupalversion}" "le" "10.9"); then \
-            echo "Drupal version is ${drupalversion}, copying phpunit config for 10.x" \
-            && mv web/modules/contrib/tripal/phpunit.xml web/modules/contrib/tripal/phpunit.10.5.xml \
+            echo "Drupal version is ${drupalversion}, copying config for phpunit 9" \
+            && mv web/modules/contrib/tripal/phpunit.xml web/modules/contrib/tripal/phpunit.11.xml \
             && mv web/modules/contrib/tripal/phpunit.9.6.xml web/modules/contrib/tripal/phpunit.xml; \
-            else echo "Drupal version is ${drupalversion}, using default phpunit config for 11.x"; fi'
+            elif $(dpkg --compare-versions "${drupalversion}" "le" "11.1"); then \
+            echo "Drupal version is ${drupalversion}, copying config for phpunit 10" \
+            && mv web/modules/contrib/tripal/phpunit.xml web/modules/contrib/tripal/phpunit.11.xml \
+            && mv web/modules/contrib/tripal/phpunit.10.xml web/modules/contrib/tripal/phpunit.xml; \
+            else echo "Drupal version is ${drupalversion}, using config for phpunit 11"; fi'
       # Runs the PHPUnit tests.
       - name: 'Run PHPUnit Tests'
         run: |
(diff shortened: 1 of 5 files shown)


laceysanderson:
GAH! Failing test is caused by a different version of PHP unit 😒 apparently... how many PHPunit configurations do we need to maintain!?!? 😭

- Drupal 10.4 and 10.5 use PHPunit 9
- Drupal 11.1 uses PHPunit 10
- Drupal 11.2 and 11.3 use PHPunit 11

The functionality we are using was added in PHPunit 11 and it's not available in PHPunit 10 which is causing the Drupal 11.1 test to fail.

laceysanderson (approved):
This is exactly what I had in mind and now all the tests are passing! Thanks @dsenalik <3

PR #2364, opened 2025-11-18
Reduce external site dependencies - obo files

### LINKED ISSUE #2363: Reduce build dependency on external sites
### Task Description

It so happens that there is a cloudflare outage today, which for us means we can't build a docker because we can't download http://purl.obolibrary.org/obo/so.obo

The proposal here is to keep a copy of all essential obo files in the tripal repo and use them if external sites are not available.

### Branch Name

tv4g0-issue2363-reduce-external-site-dependencies

dsenalik:
# Bug Fix

## Description

This PR provides support for more than one URL|path for downloading an OBO file.
The purpose is to have a fallback local copy of the obo files critical to installing Chado
If the URL is not available for any reason, then load from the local copy.

## Testing?

To test, you can temporarily edit `tripal_chado/src/Task/ChadoPreparer.php` and "Break" one or all of the external URLs e.g.
`'path' => 'http://purl.obolibrary.org/obo/go.obo||{tripal_chado}/files/so.obo',`
becomes
`'path' => 'http://xxxxxx.obolibrary.org/obo/go.obo||{tripal_chado}/files/so.obo',`

Then build a docker. Expected result is build completes successfully, but with notices such as
```
...
#9 17.47  [notice] Importing The Sequence Ontology
#9 17.70  [notice] Downloading URL "http://xxxxxx.obolibrary.org/obo/so.obo", saving to "/tmp/obo_nlGeeu"
#9 17.72  [error]  ERROR: Invalid hostname in URL http://xxxxxx.obolibrary.org/obo/so.obo: cURL error 6: Could not resolve host: xxxxxx.obolibrary.org (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://xxxxxx.obolibrary.org/obo/so.obo 
#9 17.72 [site http://default] [TRIPAL] ERROR: Invalid hostname in URL http://xxxxxx.obolibrary.org/obo/so.obo: cURL error 6: Could not resolve host: xxxxxx.obolibrary.org (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://xxxxxx.obolibrary.org/obo/so.obo
#9 17.73  [notice] Unable to download the remote OBO file at http://xxxxxx.obolibrary.org/obo/so.obo. Could a firewall be blocking outgoing connections? If you are unable to download the file you may manually download the OBO file and use the web interface to specify the location of the file on your server.
#9 17.73  [notice] Retrying with local OBO file: '/var/www/drupal/web/modules/contrib/tripal/tripal_chado/files/so.obo'
#9 17.73  [notice] Importing into schema "chado"
#9 17.74  [notice] Step 1: Preloading File "//var/www/drupal/web/modules/contrib/tripal/tripal_chado/files/so.obo"...
...
```
You can now have both url and local file in the OBO Vocabulary Loader. try editing the one you wrecked above and you should be able to correct the url and save it.
obourlandfile

### Blank entry fix
The blank entry as described in comment https://github.com/tripal/tripal/issues/2363#issuecomment-3548979037 should now appear as "GO" on this PR as I have added a fallback if a description is missing.
After this PR is merged, and new docker images build, there should be a proper description and then it will appear as "Gene Ontology" since I added this DB to `tripal.tripal_content_terms.chado_content_terms.yml`
gohasname

### Update hook
Unfortunately the blank entry on an existing site can't be manually edited to fix it, so we need a simple update hook, see commit https://github.com/tripal/tripal/pull/2364/commits/b974c76cbd25b7b74c6152791d20fc945989177a


vidu-wil:
I have done the manual testing for this PR as mentioned on the steps above:

✅ I changed the path for both the sequence and taxonomic_rank and built the docker container. The results were as expected. 
Screenshot 2025-12-04 at 11 19 25 AM
Screenshot 2025-12-04 at 11 19 59 AM

✅ I was able to change and update the link to the correct one on OBO Vocabulary Loader.
Screenshot 2025-12-04 at 12 49 25 PM

✅ Blank entry has been replaced by 'GO' in the newly built docker.
Screenshot 2025-12-04 at 12 58 16 PM

✅ Update hook on an existing docker is working as expected and 'go' has replaced the blank entry.
Screenshot 2025-12-04 at 12 19 54 PM

### ‼️**Note:** 
When replacing the blank entry, the new entry 'GO' in the newly built docker is in upper case vs when it's updated using the update hook, 'go' is in lower case. (Refer to the two screenshots above).

dsenalik:
Thanks!

> When replacing the blank entry, the new entry 'GO' in the newly built docker is in upper case vs when it's updated using the update hook, 'go' is in lower case. (Refer to the two screenshots above).

This is sort of by design, the update hook doesn't have access to the "real" name "GO" so bases the value it uses on the name of the obo file minus extension. It's not a functional issue, just cosmetic.

vidu-wil (approved):
All the changes look great and tests work as expected! 👍

dsenalik:
Thank you @vidu-wil

PR #2365, opened 2025-11-19
Deprecated in Drupal 11.3 and removed in Drupal 12 - final fix

dsenalik:
# Bug Fix

## Description

There was one more place we need to specify using the phpunit version 10 config.

## Testing?

This gets tested after merge which is how I missed it.


laceysanderson (approved):
Ahhh, I thought there was one more place 🙈 See https://github.com/tripal/test-tripal-action/issues/10 for a thought on how we could make this easier to manage in the future. We can even use the test action for the matrixed and test coverage workflows like I do in all our extension modules.

Thank you for the quick catch!

Issue #2215, opened 2025-06-17
Follow Drupal Coding Style Standards


dsenalik:
### Task Description

With the merging of PR #2119 we now have VS Code built-in support for [coding standards](https://www.drupal.org/docs/develop/standards) when using a dev container.
This issue is a place to submit changes that are just for coding standard compliance.
Related is existing issue #1699 for removing unneeded `use` statements. (Merged)
Related is do not translate exception messages #2229 (Merged)

### Branch Names

This task will involve a huge number of changes, so we can break this task up into sections.

 - `tv4g0-issue2215-coding-standards` #2218 (Merged)
 - `tv4g0-issue2215-coding-standards-publish` #2219 (Merged)
 - `tv4g0-issue2215-coding-standards-tripal_layout` #2220 (Merged)
 - `tv4g0-issue2215-coding-standards-chado-buddy` #2348 (Merged) 
 - `tv4g0-issue2215-coding-standards-tripaldbx` #2358 (Merged)
 - `tv4g0-issue2215-coding-standards-biotask`
 - `tv4g3-issue2398-drush-commands` #2399
 - `tv4g0-issue2285-tripal-chado-services-coding-standards` #2408

### Methods

Reference for using phpcs for Drupal coding standards: https://www.drupal.org/docs/extending-drupal/contributed-modules/contributed-module-documentation/coder/php-codesniffer-command-line-usage

#### Using VS Code

1. Access settings.json https://code.visualstudio.com/docs/configure/settings
2. I have these customizations
```
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
    "files.trimTrailingWhitespaceInRegexAndStrings": false,
    "editor.showUnused": true,
    "workbench.colorCustomizations": {
      "editorUnnecessaryCode.border": "#dd7aab"
    },
```
3. Build a docker
4. When prompted, reopen in a dev container
5. Coding style violations should be highlighted, e.g. 
![Image](https://github.com/user-attachments/assets/82e72423-bbdb-4228-bdbc-d575e39e3d77)

#### Command line (for a complete list)

1. Build a docker 
2. Open a command prompt inside the docker
3. To check using the tripal standards file, for example the tripal_biodb module:
```
root@ae9290092a34:/var# cd /var/www/drupal/web/modules/contrib/tripal
[...]

PR #2375, opened 2025-11-23
Renamed default chado schema remains default

### LINKED ISSUE #2371: If you rename the default chado schema, it is no longer default
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

Renaming the default chado schema causes it to no longer be the default. Even if it is the only one.
We should update the default too.

This could happen if you migrate a tripal 3 chado, and need to get the current one out of the way first.
Current workaround is to remember to set as default after renaming. If you forget this step, you get the error below, and probably many others.

### Steps to reproduce

Tripal -> Data Storage -> Chado -> Manage Chado Schemas -> Rename

### Error messages

```shell

Drupal\tripal\TripalDBX\Exceptions\SchemaException: Invalid or unsupported Chado schema version ''. in Drupal\tripal_chado\Database\ChadoSchema->getSchemaDef() (line 60 of modules/contrib/tripal/tripal_chado/src/Database/ChadoSchema.php).
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description

On this branch, if a chado schema is being renamed, and it happens to be the default schema, then the schema remains default but under the new name.

I have added a service that could be injected, `\Drupal::service('config.factory')`. I did not do so here because the entire BioTaskBase -> ChadoTaskBase -> various classes all need this done, and the changes would be much more involved than the actual task of this PR so will be deferred for now. I added branch `tv4g0-issue2215-coding-standards-biotask` to issue #2215 for this.

## Testing?

Rename chado to teacup and observe it remains the default, rename it back and it is still default.


Ferrisx4 (commented):
Code looks good, setter and getter methods for the configured default Chado schema are useful.

Change performs as advertised - a renamed Chado instance retains its 'default' designation whereas previously it didn't.

Ferrisx4 (approved):
Meant for the last comment to be "Approve".

Ferrisx4:
Looks like the setter was mostly borrowed from https://github.com/tripal/tripal/blob/f3e89aed0723d23b0bde1301ce936482dee88de0/tripal_chado/src/Form/ChadoManagerForm.php#L470 without the message, which I guess is the intention since we don't want this action to be chatty in this context.

vidu-wil (approved):
✅ Code looks good!
✅Tests work as expected!
- Before this PR when default chado is renamed, errors appear as mentioned. 
Screenshot 2025-12-08 at 9 37 13 AM
- After the changes, it works as expected. 
Screenshot 2025-12-08 at 9 46 56 AM

PR #2386, opened 2025-12-07
Rename and fix mview destroy and add phpunit tests

### LINKED ISSUE #2385: Can't delete a materialized view using destroy()
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

To delete a chado custom table, the method `delete()` is used.
To delete a materialized view, the method `destroy()` is used, which in turn should therefore call `delete()`.
This inconsistency needs to be fixed here
https://github.com/tripal/tripal/blob/f3e89aed0723d23b0bde1301ce936482dee88de0/tripal_chado/src/ChadoCustomTables/ChadoMview.php#L287

However, the fact that this function does not work means no one is using it, and we could also rename it to be consistent!

### Steps to reproduce

`drush php:eval '$mview_service = \Drupal::service("tripal_chado.materialized_views"); $mview = $mview_service->loadbyName("organism_stock_count", "chado"); $mview->destroy();'`

### Error messages

`[error]  Error: Call to undefined method Drupal\tripal_chado\ChadoCustomTables\ChadoCustomTable::destroy() in Drupal\tripal_chado\ChadoCustomTables\ChadoMview->destroy() (line 287 of .../tripal/tripal_chado/src/ChadoCustomTables/ChadoMview.php)`

### Screenshots

.

dsenalik:
# Bug Fix

## Description

The `destroy()` function for materialized views never worked. Fix the bug, and rename to `delete()` to be consistent with chado custom tables. Renaming is safe because it didn't work so nobody was using it.
But this is an important function for a module uninstall hook!

Note that the UI delete form called `delete()` instead of `destroy()` so that was also not working.

## Testing?
Test deleting a mview through the UI.

Brand new phpunit tests now cover `delete()` and other mview functions.
F-A


tripal_chado/src/ChadoCustomTables/ChadoMview.php
```diff
@@ -270,11 +274,11 @@ public function populate() {
    * @return bool
    *   True if successful. False otherwise.
    */
-  public function destroy() {
+  public function delete(): bool {
     $logger = \Drupal::service('tripal.logger');
     if (!$this->getTableId()) {
-      $logger->error('Cannot destroy the materialized view. Please, first run the init() function.');
-      return False;
+      $logger->error('Cannot destroy the materialized view, ChadoMview object was not properly initialized.');
```
vidu-wil:
```suggestion
      $logger->error('Cannot delete the materialized view, ChadoMview object was not properly initialized.');
```

vidu-wil (commented):
I have tested deleting the mview `'organism_stock_count'` through UI `before` and `after` the code changes in this PR.
- Before this PR, the mview was not deleted successfully and appeared on the view even after the successful deletion message, as shown below.
Screenshot 2025-12-09 at 2 38 10 PM

>
- After the code changes in this PR, it was deleted successfully. ✅
Screenshot 2025-12-09 at 2 41 06 PM

All the code changes look good including automated testing. Just a minor error message wording suggestion.

dsenalik:
> Just a minor error message wording suggestion.

Good catch! Change has been made.

tripal_chado/tests/src/Kernel/ChadoCustomTables/ChadoMviewTest.php
```diff
@@ -0,0 +1,285 @@
...
+      ->insert('1:organism')
+      ->fields(['genus' => 'Tripalus', 'species' => 'databasica'])
+      ->execute();
+  }
+
+  /**
+   * Kernel test for ChadoMview create/populate/delete lifecycle.
+   */
+  #[Test]
+  public function createPopulateDeleteMviewTest(): void {
```
laceysanderson: Switch to the standard testTESTNAME format for the method name so that `#[Test]` is not required.

```suggestion
  public function testCreatePopulateDeleteMview(): void {
```

tripal_chado/tests/src/Kernel/ChadoCustomTables/ChadoMviewTest.php
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * Kernel test for ChadoMview create/populate/delete lifecycle.
+ *
+ * @group tripal-chado
+ * @group service-chado-custom-table
+ */
+#[CoversClass(ChadoMview::class)]
+#[Group('tripal-chado')]
+#[Group('service-chado-custom-table')]
```
laceysanderson: We no longer use the tripal-chado group. We could benefit from an mview group though 🤔 


```suggestion
#[Group('service-chado-mview')]
#[Group('service-chado-custom-table')]
```

laceysanderson (commented):
Code review looks good and yay for the test! Just some minor suggestions for consistency.

laceysanderson (approved):
Awesome, thanks! Looks good to me!

PR #2397, opened 2025-12-16
Fixes missing column when adding linking method for field

### LINKED ISSUE #2396: Adding field linking method column missing
### Tripal Version

4.x

### Drupal Version

11.x-dev

### PostgreSQL Version

18

### PHP Version

8.4

### General Description

Missing column name when adding a field

### Branch

`tv4g1-issue2396-missing-column-adding-field`

### Steps to reproduce

Create a new content type "Primer"
Add a name field and select the feature table
Now add an organism field. When we go to select linking method, the column is missing

### Screenshots

Image

dsenalik:
# Bug Fix

## Description

The logic to get a list of possible linking methods when manually adding a field was broken and it was all my fault.
This PR should fix that.

In addition, I found some bugs due to the lack of accessing the storage settings subform (introduced in Drupal 10.2).
With that fix, I dropped some of the code supporting drupal 10.1 without the subform since our lowest supported version is 10.2 (soon to be 10.3).
But not all of it because I don't want to change the purpose of this PR.

## Testing?

You don't have to complete adding the field, just view the list of options. In particular check on fields that will have "irregular" column names, i.e. type_id->cvterm_id or manufacturer_id->contact_id:
1. Import the expression type collection.
2. Tripal -> Page Structure -> Array Design -> Manage Fields
3. +Create a new field
4. Chado Fields
5. Enter any label, click Chado Contacts: All, then click "Continue"
6. Should see only these two options (assay is operating like a linker table here just because it has the two needed foreign keys)
array_design_1

On 4.x we had this monstrosity and you could not add the field
array_design_4 x

### DIFF
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -658,7 +658,7 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
           // reference to the object table. If the current table is a
           // linker table, then we also need to check that we also have
           // a foreign key to the base table.
-          if ( ($table_name == $base_table) or ($all_fkeys[$base_table][$base_pkey_col] ?? FALSE)) {
+          if (($table_name == $base_table) or ($all_fkeys[$base_table][$base_pkey_col] ?? FALSE)) {
             $linker_tables[] = [$table_name, $all_fkeys[$object_table][$object_pkey_col]];
           }
         }
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -653,22 +653,28 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
         // Map for easy lookup.
         $all_fkeys = [];
         foreach ($table_schema_def['foreign keys'] as $foreign_key) {
-          // For example, a table with a type_id column, here we would store
-          // $all_fkeys['a_table'] = ['type_id' => 'cvterm_id']
-          $all_fkeys[$foreign_key['table']] = $foreign_key['columns'];
+          // We reverse the order here for easy lookup. For example,
+          // array_design table has an operator_id column which is a foreign
+          // key to contact.contact_id. We want the key to be contact_id.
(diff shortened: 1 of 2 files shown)


tripal_chado/src/TripalField/ChadoFieldItemBase.php
```diff
@@ -653,22 +641,28 @@ protected function getLinkerTables($object_table, $base_table) {
...
-            $linker_tables[] = [$table_name, $all_fkeys[$base_table][$base_pkey_col]];
-          }
-          else if ($all_fkeys[$object_table][$object_pkey_col] ?? FALSE) {
+        if ($all_fkeys[$object_table][$object_pkey_col] ?? FALSE) {
+          // At this point we found a foreign key to the object table.
+          // If the current table is the base table, this is a direct
+          // reference to the object table. If the current table is a
+          // linker table, then we also need to check that we also have
+          // a foreign key to the base table.
+          if ( ($table_name == $base_table) or ($all_fkeys[$base_table][$base_pkey_col] ?? FALSE)) {
```
laceysanderson: Coding standards change:

```suggestion
          if (($table_name == $base_table) or ($all_fkeys[$base_table][$base_pkey_col] ?? FALSE)) {
```

laceysanderson (changes requested):
So I had a sneaking suspicion that this form had actually changed again recently... so I started PR #2400 to add a kernel test to 1) confirm my suspicion that this PR won't work on earlier versions and 2) get some coverage of this form that Drupal core keeps changing 🤦‍♀️ 

Unfortunately, my suspicions were correct and after way to much effort trying to find the change record, here it is: [FieldStorageAddForm is split into FieldStorageAddController and FieldStorageAddForm](https://www.drupal.org/node/3503549).

I'm going to expand my test PR now but wanted to get this discovery posted.

laceysanderson (approved):
Ok, I have confirmed that this is working in Drupal 10.5, PHP 8.3 and further inspection of the change mentioned previously confirms that it shouldn't affect the specific changes in this PR. As such, I am approving this :-) Thanks @dsenalik! 

Note: The testing added in #2400 is still needed to confirm this form is functioning fully as we expect and that it continues to function in the future. However, it will become an independent body of work so as not to hold this fix back.

PR #2410, opened 2026-01-20
Add getter and setter for chado buddy schema name

### LINKED ISSUE #2409: ChadoBuddies should support multiple schema

Yes

Yes

### Is this feature urgent?

Yes

### Description

Currently chado buddies are created via their plugin manager which uses the current chado connection. There is no built in way to set the schema to anything other then the main chado schema.

There should be an approach to set the schema for a specific chado buddy object. For example, I would expect to be able to create a cvterm buddy, set its chado schema specifically and then do my work. So something like:

```php
$cvterm_buddy = $buddy_manager->createInstance('chado_cvterm_buddy', []);
$cvterm_buddy->setChadoSchema($staging_schema);
```

### Your specific use case

My specific use case is when working on an extension module that attempts to maintain support for multiple schema. We have services and importers which allow the schema the selected.

See [[Bug] TripalCultivatePhenotypesTermsService broken by recent ChadoBuddy update](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/201). Our services got around this by accessing the connection property directly which caused them to be more brittle then one would like 🙈 We would like to fix our current issue by calling a new method rather then just changing the property name and hoping it doesn't break again. 🙈 🫠 

### Generally Applicable

An example use case is when checking a staging schema has a specific cvterm.

_No response_

dsenalik:
# New Feature and Bug Fix

## Description

This adds a getter and setter function to Chado Buddies to change the chado schema name.

## Testing?

I added phpunit test assertions for the new functions. @laceysanderson has a good use case that can test this

### DIFF
--- a/tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
+++ b/tripal_chado/src/ChadoBuddy/ChadoBuddyPluginBase.php
@@ -74,6 +74,32 @@ abstract class ChadoBuddyPluginBase extends PluginBase implements ChadoBuddyInte
   }
 
   /**
+   * Returns the currently active chado schema name.
+   *
+   * @return string
+   *   The name of the active chado schema.
+   */
+  public function getSchemaName(): string {
+    return $this->chado_connection->getSchemaName();
+  }
+
+  /**
+   * Sets the current chado schema name.
+   *
+   * @var string $schema_name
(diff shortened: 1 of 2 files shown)


laceysanderson (approved):
Perfect! Thank you @dsenalik ❤️ 🎉

PR #2413, opened 2026-01-29
Ensure that TripalTerms created by TripalDefaultIdSpace know their id/vocab plugin

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR doesn't actually solve the problem in #2281 that I thought it would 🙈 That said, it still updates the default Tripal IDSpace plugin to match the Chado IDSpace plugin in ensuring that created terms know their id/vocab plugin. It also fixes coding standards.

## Testing?

Only needs a code review.


dsenalik (approved):
The one change looks appropriate!
All coding standards look very standard.

PR #2415, opened 2026-02-01
Do not load config if module not active

### LINKED ISSUE #2414: Cache rebuild installs configuration for modules that are not active
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

This bug shows up in the case where an extension module is installed which uses any of these three configuration types: `tripal.tripalentitytype_collection.*`, `tripal.tripal_content_terms.*`, or `tripal.tripalfield_collection.*`.
If the module is present, but not yet enabled, then when the cache is rebuilt, the module configuration is loaded. Then when you later go to enable the module, it fails with an error like 

Configuration objects (tripal.tripal_content_terms.trpcultivate_terms, etc.) provided by trpcultivate already exist in active configuration

### Steps to reproduce

You need an extension module with appropriate config.
See https://github.com/TripalCultivate/TripalCultivate/issues/99 for an example.

### Error messages

```shell

```

dsenalik:
# Bug Fix

## Description

This bug shows up in the case where an extension module is installed which uses any of these three configuration types: `tripal.tripalentitytype_collection.*`, `tripal.tripal_content_terms.*`, or `tripal.tripalfield_collection.*`.
If the module is present, but not yet enabled, then when the cache is rebuilt, the module configuration is loaded. Then when you later go to enable the module, it fails with an error like

`Configuration objects (tripal.tripal_content_terms.trpcultivate_terms,` etc.`) provided by trpcultivate already exist in active configuration`

This will affect any extension module that adds a new tripal content type or new tripal fields, thus high priority!

The bug also exists in the tripal_layout module for config types `tripal_layout.tripal_layout_default_form.*` and `tripal_layout.tripal_layout_default_view.*`

## Testing?

You will need a non-core extension module. I suggest tripal cultivate core module. Regardless of the module used, the key steps are:
1. Install the module but do not enable it yet
2. Rebuild cache
3. Enable the module

See https://github.com/TripalCultivate/TripalCultivate/issues/99 for detailed steps to test using the tripal cutivate module.

### DIFF
--- a/tripal/src/api/tripal.api.php
+++ b/tripal/src/api/tripal.api.php
@@ -1,9 +1,13 @@
 <?php
 
+/**
+ * @file
+ * Provides an application programming interface (API) for tripal.
+ */
+
 use Drupal\Core\Extension\ExtensionDiscovery;
 use Drupal\Core\Config\InstallStorage;
 use Drupal\Core\Config\FileStorage;
-
 
 /**
  * Loads a Tripal managed Configuration Entity.
@@ -19,7 +23,6 @@ use Drupal\Core\Config\FileStorage;
  *   The ID of a Configuration Entity type.
  */
 function tripal_load_configuration($module, $config_type) {
-
   // Get the config entity storage class and definition class.
   $config_storage = \Drupal::entityTypeManager()->getStorage($config_type);
   $definition = \Drupal::entityTypeManager()->getDefinition($config_type);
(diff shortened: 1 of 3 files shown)


laceysanderson (commented):
I still get this error on this branch :-(

I let devcontainer create a fresh Tripal container on this branch then followed the instructions in the linked Tripal Cultivate issue.
Screenshot 2026-02-03 at 11 25 51 AM

dsenalik:
@laceysanderson the tripal_layout module has the same bug, not sure how I missed that. Fixed.

laceysanderson (approved):
Confirmed that following the same steps as is in my previous review, installation now works beautifully!

Screenshot 2026-02-05 at 1 39 16 PM

PR #2273, opened 2025-09-03
Tripal File Module 📁

dsenalik:
# New Feature

## Depends on 

 * ~~#2260 Allow core fields to work on custom tables~~ **Merged**
 * ~~#2271 Editing single hop property field fixed~~ **Merged**
 * ~~#2280 Update OBO importer to work with EDAM~~ **Merged**
 * ~~#2283 Fix invalid menu machine name~~ **Merged**
 * ~~#2291 Chado custom tables with nullable columns unique constraint~~ **Merged**
 * ~~#2327 Add missing "1:" prefix in select in ChadoRecords et al.~~ **Merged**
 * ~~#2336 Chado custom tables do not have foreign keys~~ **Merged**
 * ~~#2381 Handle foreign keys from database properly~~ **Merged**
 * ~~#2383 OBO importer fixTermMismatch() problem~~ **Merged**
 * ~~#2415 Do not load config if module not active~~ **Merged**
 * ~~#2527 Add an assertion to ChadoFieldTestTrait to check chado table values~~ **Merged**
 * ~~#2564 Allow discover() to pass display options~~ **Merged**
 * #2525 Chado records support for saving a null and check "nulls not distinct"

## Description

This PR ports the Tripal File module to Tripal 4!

This is currently a draft so I have somewhere to create a to-do list.
 - [x] css for the file location formatter table
 - [x] drupal views for displaying files and licenses at `/data-search`
 - [x] Install the EDAM ontology upon module install, and submit the job automatically (tripal 3 you had to run the job).
 - [x] Put all of the "General" linker fields on the file content type.
 - [x] Migrate and update tripal 3 documentation https://tripal-file.readthedocs.io/en/latest/ (Issue https://github.com/tripal/tripal_doc/issues/121) (PR https://github.com/tripal/tripal_doc/pull/126)
 - [x] tests! Qlty coverage diff is at 65.55% - note no coverage on formatters yet 🙈 so other coverage is pretty good
 - [x] File upload option. (this did not exist in Tripal 3 but was mentioned here https://github.com/tripal/tripal_file/issues/5)
 - [x] Function to update tables to add type_id and rank for a migrated tripal 3 site
 
✅ **Documentation** - see tripal_doc issue https://github.com/tripal/tripal_doc/issues/121
 ❌ admin setting to hide filename field like in tripal 3 `includes/tripal_file.admin.inc` - or better yet, formatter setting for which columns to display and in what order - after discussion, we may try to have an embedded view - **will not be implemented in this PR**

[...]

PR #2416, opened 2026-02-02
Make tripal_layout module uninstallable

### LINKED ISSUE #2411: Think about module uninstall
### Task Description

We never think about uninstalling the tripal module, for good reason, but contrib modules we should. For example we could make tripal_layout uninstallable as a test. 

Drupal will automatically handle config uninstall if we include a `dependencies` section in yaml files.
The configuration(s) are given a dependency on the module.
For example `tripal_layout/config/install/tripal_layout.tripal_layout_default_form.general_form.yml`
should include this `dependencies` section right after `description`
```
dependencies:
    module:
        - tripal_layout
    enforced:
        module:
            - tripal_layout
```
~~I have not tested this though~~ Tested below...

### Branch Name

tv4g0-issue2411-config-uninstall

### Related Uninstall Issue

#2417 

### Notes

With tripal_layout module enabled:
```
sitedb=> SELECT name FROM config WHERE name LIKE '%tripal_layout%';
                         name                          
-------------------------------------------------------
 tripal_layout.tripal_layout_default_form.general_form
 tripal_layout.tripal_layout_default_view.general_view
 tripal_layout.tripal_layout_default_view.genomic_view
(3 rows)
```

Uninstall:
`drush pm:uninstall tripal_layout`
`[success] Successfully uninstalled: tripal_layout`

```
sitedb=> SELECT name FROM config WHERE name LIKE '%tripal_layout%';
 name 
------
(0 rows)
```

dsenalik:
# Tripal 4 Core Dev Task

## Description

This adds dependencies in the yaml files in the tripal_layout module, that ensures that if the module is uninstalled, then its configurations are also removed.

This is mainly a test case for best practice of any add-on tripal module on how to make its configurations go away if the module is uninstalled.

## Testing?

See how I tested this in issue #2411

## But?

This is a draft because on 4.x the configurations are deleted after module uninstall even without this change.

### DIFF
--- a/tripal_layout/config/install/tripal_layout.tripal_layout_default_form.general_form.yml
+++ b/tripal_layout/config/install/tripal_layout.tripal_layout_default_form.general_form.yml
@@ -1,6 +1,12 @@
 id: 'general_form'
 label: 'General Content Types'
 description: 'Layouts for general content types (analysis, organism, etc.) based on Chado and added via the Tripal Chado module.'
+dependencies:
+  module:
+    - 'tripal_layout'
+  enforced:
+    module:
+      - 'tripal_layout'
 layouts:
   - tripal_entity_type: "organism"
     field_groups:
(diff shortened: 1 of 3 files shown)


dsenalik:
Closing this PR as this change is unnecessary for my test case.
PR https://github.com/tripal/tripal/pull/2273 is going to be a nice test for module uninstall as both configs and chado tables need to be removed.

PR #2421, opened 2026-02-08
Allow UI to create content types based on chado custom tables

### LINKED ISSUE #2420: Can't use a chado custom table as base table for a new content type
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

If you want to Create a new content type using a chado custom table as its base table, you can't add the first field. This is because currently the form only allows a base table if it is in the core chado schema.

This distinction happens at this line:
https://github.com/tripal/tripal/blob/c4530b518e24d74346751cd2e85e5565b199565a/tripal_chado/src/TripalField/ChadoFieldItemBase.php#L155

We should allow _any_ table currently in chado to be selected as the base table.
To make this change, simply change `'status' => 'base'` to `'status' => 'table'`.

A second aspect, however, is that when we get a table schema definition, the column type can vary depending on the source, 'file', 'tripal', or 'database'. See part 2 below. We need to be aware and handle each variant correctly.

### Steps to reproduce part 1

1. Go to Tripal -> Data Storage -> Chado -> Chado Custom Tables -> Add Custom Table
2. Create a "license" table with this schema:
```
[
    'table' => 'license',
    'fields' => [
      'license_id' => [
        'type' => 'serial',
        'not null' => TRUE
      ],
      'name' => [
        'type' => 'varchar',
        'length' => 1024,
        'not null' => TRUE
      ],
      'summary' => [
        'type' => 'text',
      ],
      'uri' => [
        'type' => 'text',
      ],
    ],
    'primary key' => ['license_id'],
    'unique keys' => [
      'license_c1' => ['name'],
      'license_c2' => ['uri'],
    ],
    'indexes' => [
      'license_idx1' => ['name'],
    ],
  ]
```
3. Create a new content type based on this table
    a. Tripal -> Page Structure -> + Add Custom Type
    b. Name it "License" and give it whatever CV term you want, but there is a License (NCIT:C43611) term.
    c. Try to add the first field for name (Chado String Field Type) - the "license" table is not in the list of possible tables.

### Steps to reproduce part 2

In the devel/php window
```
$table_names = ['featuremap', 'license'];
$sources = ['file', 'tripal', 'database'];

foreach ($table_names as $table_name) {
  foreach ($sources as $source) {
    $schema_def = $chado->schema()->getTableDef($table_name, ['source' => $source, 'format' => 'drupal']);
    print "Table=$table_name Schema source=$source schema="
      . print_r($schema_def['fields']['name'] ?? 'no definition', TRUE) . "\n";
  }
}
```

The output is
```
Table=featuremap Schema source=file schema=Array
(
    [type] => character varying
    [not null] => 
    [size] => 255
)

Table=featuremap Schema source=tripal schema=no definition
Table=featuremap Schema source=database schema=Array
(
    [type] => varchar
    [not null] => 
    [pgsql_type] => character varying
    [length] => 255
)

Table=license Schema source=file schema=no definition
Table=license Schema source=tripal schema=Array
(
    [type] => varchar
    [length] => 1024
    [not null] => 1
)

Table=license Schema source=database schema=Array
(
    [type] => varchar
    [not null] => 1
    [pgsql_type] => character varying
    [length] => 1024
)
```

So the point is that for `type`, sometimes we get `character varying` and sometimes `varchar`.

### Branch

`tv4g2-issue2420-custom-table-content-type`

dsenalik:
# Bug Fix

## Description

The UI currently won't allow creation of a new content type based on a chado custom table. This is because currently the form only allows a base table if it is in the core chado schema.

This PR removes this restriction, and now any chado table can be used, commit https://github.com/tripal/tripal/pull/2421/commits/7cd8e3ecfb71fb211d51c4011985d6d95eae10b2

This PR also allows table columns where the schema returns `varchar` to be handled the same as `character varying`, commit https://github.com/tripal/tripal/pull/2421/commits/69db09baefbd1396dc9030f3b8fa635f127b827b This happens for custom tables, see the issue #2420 for details.

## Testing?

1. Go to Tripal -> Data Storage -> Chado -> Chado Custom Tables -> Add Custom Table
2. Create a "license" table with this schema:
```
  [
    'table' => 'license',
    'fields' => [
      'license_id' => [
        'type' => 'serial',
        'not null' => TRUE
      ],
      'name' => [
        'type' => 'varchar',
        'length' => 1024,
        'not null' => TRUE
      ],
      'summary' => [
        'type' => 'text',
      ],
      'uri' => [
        'type' => 'text',
      ],
      'inttest' => [
        'type' => 'int',
      ],
      'booltesttext' => [
        'type' => 'text',
        'pgsql_type' => 'boolean',
      ],
      'booltestint' => [
        'type' => 'tinyint',
        'pgsql_type' => 'boolean',
      ],
    ],
    'primary key' => ['license_id'],
    'unique keys' => [
      'license_c1' => ['name'],
      'license_c2' => ['uri'],
    ],
    'indexes' => [
      'license_idx1' => ['name'],
    ],
  ]
```
3. Create a new content type based on this table
   Tripal -> Page Structure -> + Add Custom Type
4. Name it "License" and give it whatever CV term you want, but there is a License (NCIT:C43611) term.
5. Try to add the first field for name (Chado String Field Type) - the "license" table is not in the list of possible tables on 4.x, but it is available on this branch.

### DIFF
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoIntegerTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoIntegerTypeDefault.php
@@ -50,6 +50,7 @@ class ChadoIntegerTypeDefault extends ChadoFieldItemBase {
     'smallint',
     'integer',
     'bigint',
+    'int',
     'serial',
   ];
 
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
@@ -49,7 +49,7 @@ class ChadoStringTypeDefault extends ChadoFieldItemBase {
    *   Simple list of valid base column types for this field to manage.
    * @see ChadoFieldItemBase
    */
-  protected static $valid_base_column_types = ['character', 'character varying'];
+  protected static $valid_base_column_types = ['character', 'character varying', 'varchar'];
 
   /**
    * {@inheritdoc}
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
@@ -49,7 +49,7 @@ class ChadoStringTypeDefault extends ChadoFieldItemBase {
    *   Simple list of valid base column types for this field to manage.
    * @see ChadoFieldItemBase
    */
-  protected static $valid_base_column_types = ['character', 'character varying', 'varchar'];
+  protected static $valid_base_column_types = ['character', 'character varying'];
 
   /**
    * {@inheritdoc}
--- a/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
+++ b/tripal_chado/src/Plugin/Field/FieldType/ChadoStringTypeDefault.php
@@ -49,7 +49,7 @@ class ChadoStringTypeDefault extends ChadoFieldItemBase {
    *   Simple list of valid base column types for this field to manage.
    * @see ChadoFieldItemBase
    */
-  protected static $valid_base_column_types = ['character', 'character varying'];
+  protected static $valid_base_column_types = ['character', 'character varying', 'char', 'varchar'];
 
   /**
    * {@inheritdoc}
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -152,7 +152,7 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
       $base_tables[''] = '- Select -';
       $chado = \Drupal::service('tripal_chado.database');
       $schema = $chado->schema();
-      $tables = $schema->getTables(['type' => 'table', 'status' => 'base']);
+      $tables = $schema->getTables(['type' => 'table', 'status' => 'table']);
       foreach (array_keys($tables) as $table) {
         $base_tables[$table] = $table;
       }
--- a/tripal_chado/src/TripalField/ChadoFieldItemBase.php
+++ b/tripal_chado/src/TripalField/ChadoFieldItemBase.php
@@ -555,7 +555,10 @@ abstract class ChadoFieldItemBase extends TripalFieldItemBase {
 
     $table_schema_def = self::getChadoTableDef($table_name);
     foreach ($table_schema_def['fields'] as $field => $properties) {
-      if (!$column_types or in_array($properties['type'], $column_types)) {
+      // We use pgsql_type in preference, because for example a boolean
+      // will have pgsql_type=boolean but type=text.
+      $type = $properties['pgsql_type'] ?? $properties['type'];
+      if (!$column_types or in_array($type, $column_types)) {
         $table_columns[] = $field;
       }
     }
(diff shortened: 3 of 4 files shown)


laceysanderson (changes requested):
So this does allow ✅ any table to be selected for a field and ✅ I am able to add a field where the column type is varchar; ❌ however, I cannot add a field for an int 🙈 

Steps to reproduce
1. Create a content type (I called it counts)
2. Try to add a Chado Integer field where the table is `organism_feature_count` and note that there are no columns that work despite the fact that the custom table schema has a `num_features` column of type int.
3. Then try adding a string field with the same base table and notice this works fine.

Image

Image

dsenalik:
So apparently `int` is returned when we get the schema from the database method rather than `integer`. Simply adding that like I did for `varchar` is an easy fix.
These are generated by `getTableDdl()` in TripalDbx which gets it from `SELECT pg_temp.tripal_get_table_ddl...` which in turn is a fairly complicated postgres function in `tripal/src/TripalDBX/pg-clone-schema/clone_schema.sql`

So let's just see all types that actually exist. I evaluated this code
```
$chado_connection = \Drupal::service('tripal_chado.database');
$sources = ['file', 'database'];
foreach ($sources as $source) {
  $types = [];
  $schema_def = $chado_connection->schema()->getSchemaDef(['source' => $source, 'format' => 'drupal']);
  foreach ($schema_def as $table => $def) {
    foreach ($def['fields'] as $column => $props) {
      $types[$props['type']] = ($types[$props['type']] ?? 0) + 1;
    }
  }
  print "=> For source=\"$source\"\n";
  foreach ($types as $type => $count) {
    print "$type - $count\n";
  }
  print "\n";
}
```

with results
```
=> For source="file"
serial - 208
bigint - 471
timestamp without time zone - 10
text - 122
integer - 95
character varying - 53
boolean - 24
double precision - 7
real - 4
character - 2
smallint - 1
date - 1

=> For source="database"
serial - 210
int - 595
text - 166
varchar - 74
float - 11
char - 2
```

The types not supported before this PR are:
 - varchar
 - char
 - int
 - float (but we don't support real or double precision yet either)

But notice there is no bool or boolean here! Look at the database version for `feature` table column `is_analysis`
```
 ["is_analysis"]=>
  array(4) {
    ["type"]=>
    string(4) "text"
    ["not null"]=>
    bool(true)
    ["pgsql_type"]=>
    string(7) "boolean"
    ["default"]=>
    string(5) "false"
  }
```

So with the current logic we won't be able to support boolean on custom tables.

By using the `pgsql_type` in preference to the `type`, when it is present, we can handle boolean also.
Note that to store a boolean in a chado custom table, this is invalid
```
      'booltestbool' => [
        'type' => 'boolean',
      ],
```
You need to include the pgsql_type. The `type` is not really passed to postgres, it gets stored as a postgres boolean in the database.
```
      'booltesttext' => [
        'type' => 'text',
        'pgsql_type' => 'boolean',
      ],
```
You can also use this, which makes the exact same db table
```
      'booltestint' => [
        'type' => 'tinyint',
        'pgsql_type' => 'boolean',
      ],
```

reynoldtan:
Performed the manual test and verified the that license table is available when creating Chado String field and I am offered Table Column options in creating Chado Integer type.

**String type**
Screenshot 2026-02-20 at 8 03 36 AM

**Integer type**
image

laceysanderson (approved):
✅ Fresh code review looks really solid and shows my concerns as fixed. I agree with this approach.
✅ Validation done by @reynoldtan shows that this is functioning as expected as well.

As such, I approve and feel this is ready for merge! Thanks for the diligence on the extra types to be supported ❤️

PR #2435, opened 2026-03-02
Added field group deprecations  related to theme and plugins.

### LINKED ISSUE #2434: Field Group module deprecations to exclude
### Task Description

We are coming across an additional deprecation being caused by our use of field group and need to add it to our exclude list.

Specifically,

> Not supporting attribute discovery in Drupal\field_group\FieldGroupFormatterPluginManager is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Provide an Attribute class and an Annotation class for BC. See https://www.drupal.org/node/3395582

> Providing a file for theme hook field_group_html_element is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3549500

> Providing template_preprocess_field_group_html_element() is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3504125

These are being thrown when the various TripalCultivate modules are installed but are not specific to those modules but rather appearing there due to the test coverage :-)

### Branch Name

tv4g9-2434-fieldGroupDeprecations

laceysanderson:
# Tripal 4 Core Dev Task

## Description

The field group module is throwing additional deprecation noticed in extension modules. We are not seeing them in core because 1) we do not have as good of coverage and 2) core layouts do not yet use horizontal or vertical tabs.

This PR adds these deprecations to our ignore list. They are being ignored in Tripal Core because Field Group is a core Tripal Layout dependency and these deprecations are being thrown in extension modules when Tripal Layout is used in supported ways.

## Testing?

There are no tests in Tripal core that currently throw these deprecations.

Our tests should show if there is a syntax error in the added ignore patterns though.

If you want to test this in situ, you can do so using the InstallTest in TripalCultivate as follows:

1. Create a TripalCultivate docker container on branch `g5.102-resolveDeprecations`.
2. Replace the `/var/www/drupal/web/modules/contrib/tripal/.tripal-deprecation-ignore.txt` file inside the container with the modified one in this PR.
3. Within the TripalCultivate module directory (`/var/www/drupal/web/modules/contrib/TripalCultivate`) run the PHPUnit test using `phpunit tests/src/Functional/InstallTest.php`
4. Notice that without the change to the Tripal file, there are many deprecations and with there are only 3 deprecations, none of which relate to field group or horizontal tabs.

### DIFF
--- a/.tripal-deprecation-ignore.txt
+++ b/.tripal-deprecation-ignore.txt
@@ -12,6 +12,36 @@
 # Example test that triggers this: testTripalEmptyContentTypes
 %field_group_module_implements_alter without a #\[LegacyModuleImplementsAlter\] attribute%
 
+# Module: Field Group Module
+# Location: /var/www/drupal/web/core/lib/Drupal/Core/Test/HttpClientMiddleware/TestHttpClientMiddleware.php:52
+# Deprecation: Providing a file for theme hook horizontal_tabs is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3549500
+# Example test that triggers this is TripalCultivate::installTest. Triggered whenever a Tripal module with layouts including vertical or horizontal tabs is installed or such layouts are applied.
+%Providing a file for theme hook horizontal_tabs is deprecated%
+
+# Module: Field Group Module
+# Location: /var/www/drupal/web/core/lib/Drupal/Core/Test/HttpClientMiddleware/TestHttpClientMiddleware.php:52
+# Deprecation: Providing template_preprocess_horizontal_tabs() is deprecated in drupal:11.3.0 and is removed from drupal:12.0.0. Use initial preprocess for template_preprocess instead. See https://www.drupal.org/node/3504125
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
If I run as shown in the `TripalCultivate` directory
`phpunit tests/src/Functional/InstallTest.php`
I see a total of 13 deprecations. 

But if I run from the `tripal` directory as
`phpunit ../TripalCultivate/tests/src/Functional/InstallTest.php`
then I do indeed see only the three other deprecations. Without changes of this PR I see 8 deprecations.

Code review looks perfect

So I am approving, but curious is there is anything additional needed either in the testing directions or in the TripalCultivate module?

reynoldtan:
The code changes look good, and I've verified the following:

- [x] Presence of deprecation relating to horizontal tabs before applying the changes.

Screenshot 2026-03-03 at 7 37 12 AM

Screenshot 2026-03-03 at 7 31 22 AM


- [x] **3 Deprecations** none of which relate to horizontal tabs after applying the changes.

Screenshot 2026-03-03 at 7 31 29 AM

laceysanderson:
That sounds like your docker didn't have the phpunit.xml changes @dsenalik. I would guess you either missed checking out the branch indicated or did so after the docker was created 🤔 

Thanks for the reviews both of you! I'm going to merge this as confirmed 😍

PR #2438, opened 2026-03-04
Tripal markup field: provide static text on Tripal Content form and page.

### LINKED ISSUE #2437: Field to provide longer help text / guidance in Tripal Content forms
### Task Description

Currently the only way to provide guidance to users on filling out Tripal content forms is to use the description of the field. This approach only supports very short guidance and is rendered quite small.

With Biological content it can help to provide much more detailed guidance in order to help researchers and curators submit high quality metadata. While Drupal Core does not provide functionality to do this, I had used [Drupal Markup](https://www.drupal.org/project/markup) to accomplish this on KnowPulse. Unfortunately we are now getting deprecation messages and this module is not maintained.

After a discussion in Tripal Cofest, we have decided to fold this functionality into Tripal Core since it has wide applications in our sites and is very simple (only a single field).

This will be done in this issue 😃 

### Design

We will reuse the field classes from the Drupal Markup module and update them to recent Drupal versions and our coding standards.

### Branch Name

tv4g1-2437-tripalMarkupField

laceysanderson:
# New Feature

## Description

Currently the only way to provide guidance to users on filling out Tripal content forms is to use the description of the field. This approach only supports very short guidance and is rendered quite small.

With Biological content it can help to provide much more detailed guidance in order to help researchers and curators submit high quality metadata. This PR will create a "Tripal Markup" field which allows admin to set a single formatted message for a given content type that will be rendered on the form and/or page. This formatted message will be managed like a field which allows it to be moved around and hidden just like any field.

## Implementation details

We will define a field (Tripal Markup) with a single property. This is needed because the field has to have a value in order for the formatter to be called. The single property is:

- hasValue is a boolean stored in the database and indicates if the admin has set markup in the field settings. This is stored in the drupal field table.

We will use a static method, `getMarkupValue()`, to retrieve the value of the markup form element saved in the field settings.

In the formatter and widget we will call this static method to get the content to be rendered and ignore the field values all to

## Tasks

- [x] Create the TripalMarkupTypeItem class and determine what TripalStorage properties it should have.
- [x] Add field settings to allow the static markup to be set.
- [x] Create the static method to return the field setting value as the value of the field
- [x] Create a widget that will render the value rather then display a form element.
- [x] Ensure hasValue is set if the field setting has a value or not
- [x] Create a formatter that will render the value as well.
- [x] Ensure we are not creating security issues.
- [x] Add automated testing

## Testing?

**The following can be done for any Tripal Content type. We've chosen "Project" since it has very few required fields.**

1. Go to Admin > Tripal > Page Structure > Project > Manage fields.
2. Add a new field. The field name does not matter. For the type choose "Tripal Field" > "Tripal Markup". On the field settings page, set the "Markup" form element to whatever you want to display on the form/page. 

    Screenshot 2026-03-09 at 2 02 08 PM

3. Create a project by going to Admin > Tripal > Content > Add Content > Project. The text you added in the "Markup" form element should be rendered on the form.  

    Screenshot 2026-03-09 at 2 06 10 PM

4. After saving the form, the text you entered in the "Markup" form element should be rendered on the page.

    Screenshot 2026-03-09 at 2 07 47 PM

5. Test that you can move this field around in the form and view to change where it is displayed.
6. Test that you can hide this element.
7. Make a second element and confirm that they are independent.


reynoldtan:
Excellent work @laceysanderson I Performed the manual test and verified the following items:

- [x] My test field markup was rendered in the create form.
Screenshot 2026-03-10 at 7 43 29 AM

- [x] My test field markup was rendered after content creation.
Screenshot 2026-03-10 at 7 44 04 AM

- [x] I can change the order it appeared in the page
Screenshot 2026-03-10 at 7 45 05 AM
Screenshot 2026-03-10 at 7 45 28 AM

- [x] I can hide my makup field by disabling it
Screenshot 2026-03-10 at 7 46 59 AM
Screenshot 2026-03-10 at 7 47 05 AM

- [x] Unable to create a second element 
image

tripal/src/Plugin/Field/FieldWidget/TripalMarkupWidget.php
```diff
@@ -0,0 +1,49 @@
...
+  ],
+)]
+class TripalMarkupWidget extends TripalWidgetBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+
+    // The value is set int the field settings, so we need to get it from the
```
reynoldtan:
```suggestion
    // The value is set to int in the field settings, so we need to get it from the
```

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
```diff
@@ -0,0 +1,323 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
+
+/**
+ * Tests the Field Types in the associated Test Info YAML.
+ *
+ * Specifically focused on create + update actions performed on the entity
+ * directly.
+ */
```
reynoldtan, line 18:
```suggestion
 * 
 * @group TripalField
 */
```

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * The modules that this test depends on.
+   *
+   * NOTE: since this is a kernel test, these modules are not being installed
+   * but are available to be installed.
+   *
+   * @var array
+   */
+  protected static $modules = ['system', 'user', 'path', 'path_alias', 'field', 'datetime', 'tripal'];
```
reynoldtan:
```suggestion
  protected static $modules = [
    'system',
    'user',
    'path',
    'path_alias',
    'field',
    'datetime',
    'tripal',
  ];
```

Consider this format where each element is in a new line.

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
```diff
@@ -0,0 +1,323 @@
...
+  protected array $system_under_test;
+
+  /**
+   * The TripalEntityType id of the bundle being used in this test.
+   *
+   * @var string
+   */
+  protected string $bundle_name;
+
+  /**
```
reynoldtan, line 85: To improve readability, consider adding @see to reference your yml test scenario file and maybe moving the $yaml_info_file in line 54 closer to this property declaration

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
```diff
@@ -0,0 +1,323 @@
...
+   *
+   * @param int $current_scenario_key
+   *   The key of the scenario in the YAML.
+   * @param string $current_scenario_label
+   *   The label of the scenario in the YAML.
+   *
+   * @dataProvider provideScenarios
+   */
+  #[DataProvider('provideScenarios')]
+  public function testFieldTypeCrud(int $current_scenario_key, string $current_scenario_label) {
```
reynoldtan, line 178: I think this test is missing a Delete case for the D in CRUD.

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
laceysanderson: We no longer need the group annotation as we are moving to attributes and are not actively running tests locally on versions not supporting attributes. Tripal core is phasing out adding the annotation :-)

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
laceysanderson: Good catch! Addressed in 680265d784d0a58d2dcfcea8cc70ead93e28330a.

tripal/tests/src/Kernel/Plugin/TripalField/TripalMarkupTest.php
laceysanderson: I used a note instead of @see so I could expand providing more context :-)

laceysanderson: See 680265d784d0a58d2dcfcea8cc70ead93e28330a for the change.

laceysanderson:
Regarding the issue in your manual test:

> Unable to create a second element

This is expected behaviour. I can provide a default term which allows you to avoid setting the term for the first field. However, two fields on the same content type cannot have the same term, so for the second one you need to change the term to anything else. Sorry that I forgot to include that instructions! 🙈 Can you please try again while changing the term by expanding the bottom "Controlled Vocabulary Term" fieldset and entering another term.

reynoldtan:
> However, two fields on the same content type cannot have the same term, so for the second one you need to change the term to anything else.

- [x] Verified that instances of markup field are independent from each other.
image

Shows even reordering does not affect other instance.

image

reynoldtan (approved):
Code looks good 👍

dsenalik:
I don't have time to test more thoroughly, but I am seeing that the markup is appearing on the rendered entity in addition to the form. It should be in the disabled section but is not. I thought this was intended to guide users entering data, so would only appear in the widget?
notdisabled

laceysanderson:
Ahhh, @dsenalik it can be used in both cases and by default all fields appear on both form and page. You can disable it on the page easily but I don't think I can change the default 🤔 I'll look into it though! If we add the field using the YAML then we can make it hidden on the page by default.

laceysanderson:
Got it! Commit [6664fbc](https://github.com/tripal/tripal/pull/2438/commits/6664fbc6ff4c4eb828b983eb906c2e43c4319791) makes it not appear on the page by default. However, when you go to Manage Display then the formatter is picked up and it will be enabled on the page. This is the best we can do within Drupal.

dsenalik:
Thinking about this, I think you should leave the default as enabled, I can imagine uses for this field for the formatter too. Maybe it's better to default to visible than to "why doesn't my field show up".

laceysanderson:
Both myself and @reynoldtan have confirmed all is well after the most recent change. As such, this is ready to merge! Thank you! <3

PR #2443, opened 2026-03-16
Gin theme toolbar is going away

### LINKED ISSUE #2439: Gin theme toolbar is going away
### Tripal Version

4.x

### Drupal Version

11.x

### PostgreSQL Version

18

### PHP Version

8.5

### General Description

If I build a docker on the described version combo, it is missing the shortcut toolbar at the top.

If I change only the PHP version to 8.4, then I have the normal toolbar.

### Steps to reproduce

.

### Error messages

```
```

### Screenshots

Image of missing toolbar on PHP 8.5
Image

Also I am seeing a missing Tripal icon
Image

Expected toolbar is present on 8.4
Image

Normally you will see this in the gin theme settings
Image

But when the toolbar is absent, the gin settings look like
Image

It looks like the new Navigation module is responsible
https://www.drupal.org/project/gin/issues/3503575

To fix the missing Tripal icon, we should be using the Drupal Icon API https://www.drupal.org/docs/develop/drupal-apis/icon-api

### Appearance
The old sidebar expanded down thusly:
Image

The new navigation expands to the right (this is after fixing the Tripal icon display):
Image

dsenalik:
# Bug Fix

## Description

The gin theme is losing its toolbar because the new Drupal navigation module is overriding it, Drupal is making the navigation module the new default menu system.

The existing Tripal admin menu contains all options that were present in the toolbar menu, so we have not lost the ability to do anything, however the Tripal icon is broken, we need to use the new Icon API to add the icon to the menu now. That is the main change in this PR.

We do, however, lose access to the Devel menu that was in the toolbar. I will propose in a different PR to include the https://www.drupal.org/project/navigation_extra_tools module as in commit https://github.com/tripal/tripal/pull/2443/commits/e2ec09fbae4bf4be6c0d9ca4a4c091dca661f343 (reverted), which adds a "Tools" menu to the navigation. This causes a bit of complication when updating an existing site, so will need to be combined with a Tripal version increment, thus is not included here.

## Testing?

Build a docker on 11.x php 8.5 - The shortcut toolbar should be gone, but the tripal menu item on the left should include an icon, rather than the "Tr" you would see on 4.x.


laceysanderson (changes requested):
This works perfect in light mode on Drupal 11.x-dev but not in dark mode 🙈 

Image

I also confirmed it works for the old style menu as well and it does. This also demonstrates what it should look like in dark mode ;-p 

Image

### How to enable dark mode in gin

1. Go to Admin > Appearance. 
2. Click on Settings for the gin theme and choose Dark.
3. Save Configuration

dsenalik:
Thanks for catching that! 
So using a Drupal icon as an example of how to set color properly
`web/core/modules/navigation/assets/icons/system-admin-config.svg`
We see
```
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path d="M6.00002 ... 17.25Z" fill="currentColor"/>
</svg>
```

My svg in 
`web/modules/contrib/tripal/tripal/icons/tripal-menu-icon.svg`
was basically a copy of existing tripal icon svg, and contained the color hard coded as black
```
<svg id="svg" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="400" height="354.8387096774194" viewBox="0, 0, 400,354.8387096774194"><g id="svgg"><path id="path0" d="M339.516
...
20.925 " stroke="none" fill="#000000" fill-rule="evenodd"></path></g></svg>
```

I have updated the svg to match exactly how Drupal does it in commit dc44abd  - most importantly using `fill="currentColor"`

```
<svg width="400" height="355" viewBox="0, 0, 400, 355" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path d="M339.516 ... 20.925" fill="currentColor"/>
</svg>
```

2439-fixedicon

laceysanderson (approved):
Works perfect now! Thanks @dsenalik <3 

Image

PR #2447, opened 2026-03-24
Add another EBI OLS known error message

dsenalik:
# Bug Fix

## Description

This adds another error message to watch for since the EBI OLS external database has been having problems lately.

## Testing?

Test should be marked skipped again. As of submission of the PR, the new 503 error is gone, but the original error is still present.

### DIFF
--- a/tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
+++ b/tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
@@ -300,12 +300,17 @@ class OboImporterTest extends ChadoTestKernelBase {
     catch(\Exception $e) {
       // If we get a specific known message due to the EBI OLS external
       // database being down, then mark this test as skipped.
+      $skip_reasons = [];
       foreach ($this->mock_messages as $message) {
         foreach ($this->known_external_messages as $pattern) {
           if (preg_match($pattern, $message)) {
-            $this->markTestSkipped();
+            $skip_reasons[$message] = 1;
           }
         }
+      }
+      if ($skip_reasons) {
+        $this->markTestSkipped("We received only known external error messages and chose to ignore them. Specifically:\n"
+          . implode("\n", array_keys($skip_reasons)));
       }
       // For anything else, rethrow the exception so we see the problem.
       throw $e;
--- a/tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
+++ b/tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
@@ -96,6 +96,18 @@ class OboImporterTest extends ChadoTestKernelBase {
   protected array $scenarios;
 
   /**
+   * Messages encountered when EBI OLS is down or broken.
+   *
+   * @var array
+   *   A list of error messages encountered when the external
+   *   web site is down or not functioning normally.
+   */
+  protected array $known_external_messages = [
+    '/Cannot find the ontology via an EBI OLS lookup/',
+    '/Service Temporarily Unavailable/',
+  ];
+
+  /**
    * {@inheritdoc}
    */
   protected function setUp(): void {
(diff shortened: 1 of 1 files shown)


tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
```diff
@@ -286,11 +298,13 @@ public function testOboImporter(int $current_scenario_key, string $current_scena
...
-      // If we get a specific message due to the EBI OLS external database
-      // being down, then mark this test as skipped.
+      // If we get a specific known message due to the EBI OLS external
+      // database being down, then mark this test as skipped.
       foreach ($this->mock_messages as $message) {
-        if (preg_match('/Cannot find the ontology via an EBI OLS lookup/', $message)) {
-          $this->markTestSkipped();
+        foreach ($this->known_external_messages as $pattern) {
+          if (preg_match($pattern, $message)) {
+            $this->markTestSkipped();
```
laceysanderson: Can the code result in more then one item in `mock_messages`? If yes, right now we might miss a different error... For example, the test would be skipped for the following values in mock messages and we would miss that there is an API format issue.

```php
$this->mock_messages = [
  'Cannot find the ontology via an EBI OLS lookup',
  'Incorrect API format: unrecorgnized key'
];
```

Instead, perhaps we could use something like this:

```php
if (!empty($this->mock_messages)) {
  // Check if there are any error messages that are not known external messages.
  $unexpected_error_messages = array_diff($this->mock_messages, $this->known_external_messages);
  $this->assertEmpty($unexpected_error_messages, "We should not get any unexpected error messages.');
  $this->markTestSkipped('We received only known external error messages and chose to ignore them. Specifically, ' . print_r($this->mock_messages, TRUE));
}
```

laceysanderson (changes requested):

tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
dsenalik: That is a good idea, but I am going to do it a bit differently because I use preg_match to a key phrase, because the full message may not be 100% predictable, specifically the 503 error that is no longer happening.

tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
```diff
@@ -286,15 +301,30 @@ public function testOboImporter(int $current_scenario_key, string $current_scena
...
+        $this->assertTrue($matched, 'An exception caused an error: '
+          . $e->getMessage() . ' and the unexpected logger message was: ' . $message);
+      }
+      if ($skip_triggered) {
+        $this->markTestSkipped("We received only known external error messages and chose to ignore them. Specifically:\n"
+          . implode("\n", $skip_triggered));
+      }
+      else {
+        // If the exception did not create a logger message, rethrow.
+        throw $e;
```
laceysanderson: This still wouldn't catch the case where there are skipped errors and new logged errors. Now that you also have the 'normal' messages caught, I think we can use the else as I had suggested before... like this:


```suggestion
        // For each expected external message...
        foreach ($this->message_actions as $pattern => $action) {
          // If it's in our list with an action of skip
          // then we indicate the test should be skipped.
          if (preg_match($pattern, $message)) {
            // Note: an action of normal will simply be ignored
            // since it would fall into an 'else' here.
            if ($action == 'skip') {
              $skip_triggered[] = $message;
            }
          }
          // If it doesn't match our pattern then we want to
          // immediately rethrow the exception!
          else {
            throw $e;
          }
        }
      }
      // If any of the known external error messages were caught
      // then we want to skip here.
      if ($skip_triggered) {
        $this->markTestSkipped("We received only known external error messages and chose to ignore them. Specifically:\n"
          . implode("\n", $skip_triggered));
      }
      // If the exception did not create a logger message, rethrow.
      else {
        throw $e;
```

tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php
laceysanderson: Moved to a new suggestion 🙈

laceysanderson (approved):
Ahhh, there we go! Thanks! 😍 🙏

PR #2453, opened 2026-03-31
New Chado buddy delete functions

### LINKED ISSUE #2448: Chado Buddies don't have a delete function
### Task Description

To implement issue #2196 (Missing Manage Chado CV Terms) we need to be able to delete a cvterm.
The buddies do not currently have delete functions.
The task is to add these.

### Branch Name

tv4g2-issue2448-buddy-delete-functions

dsenalik:
# New Feature

## Description

This adds several delete functions to chado buddies.

These will be used in an upcoming PR to have views to manage db, cv, and cvterm records.

## Testing?

As these functions are not currently used, phpunit tests will hopefully provide good coverage of the new functions.

### DIFF
--- a/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
+++ b/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
@@ -1191,6 +1191,7 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
    *
    * @throws Drupal\tripal_chado\ChadoBuddy\Exceptions\ChadoBuddyException
    *   Thrown in the following cases:
+   *   - Foreign key references exist to the record.
    *   - The conditions match more then one record.
    *   - SQL error encountered when deleting the cvterm.
    */
--- a/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
+++ b/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
@@ -1068,6 +1068,10 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
    *       Default is FALSE, and in this case, if any such referencing
    *       records exist, the delete will be skipped and this function
    *       will return FALSE.
+   *     - fail_when_referenced (bool; default TRUE)
+   *       If TRUE, throw an exception when the cv indicated is referenced
+   *       by other records and cascade is FALSE. If FALSE, return
+   *       ChadoBuddyPluginBase::FAILURE.
    *
    * @return int
    *   Indicates whether the CV was
@@ -1077,10 +1081,11 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
    *
    * @throws Drupal\tripal_chado\ChadoBuddy\Exceptions\ChadoBuddyException
    *   Thrown in the following cases:
-   *   - the conditions match more then one record.
+   *   - Foreign key references exist to the record.
+   *   - The conditions match more then one record.
    *   - SQL error encountered when deleting the cv.
    */
-  public function deleteCv(array $conditions, array $options = []): ?bool {
+  public function deleteCv(array $conditions, array $options = []): int {
     $valid_tables = ['cv'];
     $valid_columns = $this->getTableColumns($valid_tables);
     $conditions = $this->dereferenceBuddyRecord($conditions);
@@ -1104,9 +1109,14 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
         }
       }
 
-      // If there are referencing records and cascade is not set, do nothing.
+      // If there are referencing records and cascade is not set.
       if ($total_references && !($options['cascade'] ?? FALSE)) {
-        return self::FAILURE;
+        if ($options['fail_when_referenced'] ?? TRUE) {
+          throw new ChadoBuddyException('ChadoBuddy deleteCv error, cannot delete the cv, other records reference it');
+        }
+        else {
+          return self::FAILURE;
+        }
       }
       else {
         // Perform the record deletion. This might fail if
@@ -1168,18 +1178,23 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
    *     - drop_dbxref
    *       If TRUE, then delete the dbxref record used by the term.
    *       Default is FALSE.
+   *     - fail_when_referenced (bool; default TRUE)
+   *       If TRUE, throw an exception when the cvterm indicated is referenced
+   *       by other records and cascade is FALSE. If FALSE, return
+   *       ChadoBuddyPluginBase::FAILURE.
    *
-   * @return bool|null
-   *   Returns TRUE if the cvterm was deleted.
-   *   Returns FALSE if the cvterm was not deleted.
-   *   Returns NULL if the cvterm did not exist.
+   * @return int
+   *   Indicates whether the cvterm was
+   *   - deleted (ChadoBuddyPluginBase::SUCCESS = 4)
+   *   - not deleted (ChadoBuddyPluginBase::FAILURE = 5)
+   *   - did not exist (ChadoBuddyPluginBase::NON_EXISTING = 3)
    *
    * @throws Drupal\tripal_chado\ChadoBuddy\Exceptions\ChadoBuddyException
    *   Thrown in the following cases:
-   *   - the conditions match more then one record.
+   *   - The conditions match more then one record.
    *   - SQL error encountered when deleting the cvterm.
    */
-  public function deleteCvterm(array $conditions, array $options = []): ?bool {
+  public function deleteCvterm(array $conditions, array $options = []): int {
     $valid_tables = ['cvterm', 'cv', 'dbxref'];
     $valid_columns = $this->getTableColumns($valid_tables);
     $conditions = $this->dereferenceBuddyRecord($conditions);
@@ -1205,9 +1220,14 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
         }
       }
 
-      // If there are referencing records and cascade is not set, do nothing.
+      // If there are referencing records and cascade is not set.
       if ($total_references && !($options['cascade'] ?? FALSE)) {
-        return self::FAILURE;
+        if ($options['fail_when_referenced'] ?? TRUE) {
+          throw new ChadoBuddyException('ChadoBuddy deleteCvterm error, cannot delete the cvterm, other records reference it');
+        }
+        else {
+          return self::FAILURE;
+        }
       }
       else {
         $transaction = $this->chado_connection->startTransaction();
--- a/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
+++ b/tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
@@ -560,9 +560,7 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
     if (count($existing_records) < 1) {
       return FALSE;
     }
-    if (count($existing_records) > 1) {
-      throw new ChadoBuddyException("ChadoBuddy updateCv error, more than one record matched the conditions specified\n" . print_r($conditions, TRUE));
-    }
+    $this->throwIfMultipleRecords($existing_records, 'cv.cv_id', 'updateCv', $conditions);
     // Update query will only be based on the cv.cv_id,
     // which we get from the retrieved record.
     $cv_id = $existing_records[0]->getValue('cv.cv_id');
@@ -639,9 +637,7 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
     if (count($existing_records) < 1) {
       return FALSE;
     }
-    if (count($existing_records) > 1) {
-      throw new ChadoBuddyException("ChadoBuddy updateCvterm error, more than one record matched the conditions specified\n" . print_r($conditions, TRUE));
-    }
+    $this->throwIfMultipleRecords($existing_records, 'cvterm.cvterm_id', 'updateCvterm', $conditions);
     // Only update the dbxref if it is being changed.
     $existing_values = $existing_records[0]->getValues();
     $update_dbxref = FALSE;
@@ -748,9 +744,7 @@ class ChadoCvtermBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInterfa
     if (count($existing_records) < 1) {
       return FALSE;
     }
-    if (count($existing_records) > 1) {
-      throw new ChadoBuddyException("ChadoBuddy updateCvterm error, more than one record matched the conditions specified\n" . print_r($conditions, TRUE));
-    }
+    $this->throwIfMultipleRecords($existing_records, 'cvtermsynonym.cvtermsynonym_id', 'updateCvtermSynonym', $conditions);
     // This function will only update the cvtermsynonym table.
     // The update query will only be based on the cvtermsynonym_id,
     // which we get from the retrieved record.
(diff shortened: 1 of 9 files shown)


tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,200 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   */
+  public function deleteCv(array $conditions, array $options = []): ?bool {
+    $valid_tables = ['cv'];
+    $valid_columns = $this->getTableColumns($valid_tables);
+    $conditions = $this->dereferenceBuddyRecord($conditions);
+    $this->validateInput($conditions, $valid_columns);
+    $existing_records = $this->getCv($conditions, $options);
+    if (count($existing_records) > 0) {
+      if (count($existing_records) > 1) {
+        throw new ChadoBuddyException("ChadoBuddy deleteCv error, more than one record matched the specified values\n" . print_r($values, TRUE));
```
reynoldtan: I think the $values is not defined here.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,200 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   */
+  public function deleteCvterm(array $conditions, array $options = []): ?bool {
+    $valid_tables = ['cvterm', 'cv', 'dbxref'];
+    $valid_columns = $this->getTableColumns($valid_tables);
+    $conditions = $this->dereferenceBuddyRecord($conditions);
+    $this->validateInput($conditions, $valid_columns);
+    $existing_records = $this->getCvterm($conditions, $options);
+    if (count($existing_records) > 0) {
+      if (count($existing_records) > 1) {
+        throw new ChadoBuddyException("ChadoBuddy deleteCvterm error, more than one record matched the specified values\n" . print_r($values, TRUE));
```
reynoldtan: Same here $values is undefined

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,200 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+        return FALSE;
+      }
+      else {
+        try {
+          // Perform the record deletion. This might fail if
+          // a foreign key is not defined as ON DELETE CASCADE.
+          $query = $this->chado_connection->delete('1:cvterm');
+          $query->condition('cvterm_id', $cvterm_id, '=');
+          $query->execute();
+        }
```
reynoldtan: Delete query not wrapped in a transaction

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
```diff
@@ -665,4 +665,83 @@ public function associateDbxref(string $base_table, int $record_id, ChadoBuddyRe
...
+   */
+  public function deleteDb(array $conditions, array $options = []): ?bool {
+    $valid_tables = ['db'];
+    $valid_columns = $this->getTableColumns($valid_tables);
+    $conditions = $this->dereferenceBuddyRecord($conditions);
+    $this->validateInput($conditions, $valid_columns);
+    $existing_records = $this->getDb($conditions, $options);
+    if (count($existing_records) > 0) {
+      if (count($existing_records) > 1) {
+        throw new ChadoBuddyException("ChadoBuddy deleteDb error, more than one record matched the specified values\n" . print_r($values, TRUE));
```
reynoldtan: $values is undefined in this line

reynoldtan:
I've reviewed the revisions and it looks good overall. I've left a few minor suggestions/comments for your consideration

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: Good point! fixed in https://github.com/tripal/tripal/pull/2453/commits/4ed74e3d0c7a66998ede8d71c9cd12b95c64804e

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
dsenalik: OOPS! fixed in https://github.com/tripal/tripal/pull/2453/commits/4ed74e3d0c7a66998ede8d71c9cd12b95c64804e

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: OOPS! fixed in https://github.com/tripal/tripal/pull/2453/commits/4ed74e3d0c7a66998ede8d71c9cd12b95c64804e

dsenalik:
Converting to draft as I need to add a dbxref delete function too

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   * records in those tables!
+   *
+   * @param array $conditions
+   *   An associative array of the conditions to find the record to delete:
+   *     - cv.cv_id
+   *     - cv.name
+   *     - buddy_record (object): a ChadoBuddyRecord can be used
+   *       in place of or in addition to other keys.
+   * @param array $options
+   *     - cascade
```
laceysanderson, line 1063: Coding standards ;-p

```suggestion
   *   An associative array of options with the following keys supported:
   *     - cascade (bool; default FALSE):
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   *       If TRUE, then delete even if there are foreign keys in use.
+   *       If ON DELETE CASCADE is defined for the foreign key, then
+   *       those records will also be deleted. If not, an exception will
+   *       be thrown.
+   *       Default is FALSE, and in this case, if any such referencing
+   *       records exist, the delete will be skipped and this function
+   *       will return FALSE.
+   *
+   * @return bool|NULL
```
laceysanderson: Coding standards

```suggestion
   * @return bool|null
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   *   Returns TRUE if the CV was deleted.
+   *   Returns FALSE if the CV was not deleted.
+   *   Returns NULL if the CV did not exist.
```
laceysanderson: Why not use the same approach to return values that we did with the update? Specifically, using the global variable and returning an integer with meaning?

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   */
+  public function deleteCv(array $conditions, array $options = []): ?bool {
+    $valid_tables = ['cv'];
+    $valid_columns = $this->getTableColumns($valid_tables);
+    $conditions = $this->dereferenceBuddyRecord($conditions);
+    $this->validateInput($conditions, $valid_columns);
+    $existing_records = $this->getCv($conditions, $options);
+    if (count($existing_records) > 0) {
+      if (count($existing_records) > 1) {
+        throw new ChadoBuddyException("ChadoBuddy deleteCv error, more than one record matched the specified conditions\n" . print_r($conditions, TRUE));
```
laceysanderson: It would be helpful to also mention the existing records that were matched 😉

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+   *       will return FALSE.
+   *
+   * @return bool|NULL
+   *   Returns TRUE if the CV was deleted.
+   *   Returns FALSE if the CV was not deleted.
+   *   Returns NULL if the CV did not exist.
+   *
+   * @throws Drupal\tripal_chado\ChadoBuddy\Exceptions\ChadoBuddyException
+   *   If an error is encountered. This is most likely that another table
+   *   is referencing this CV.
```
laceysanderson: Be more specific? There are actually two cases.

```suggestion
   *   Thrown in the following cases:
   *   - the conditions match more then one record.
   *   - SQL error encountered when deleting the cv.
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+      $cv_id = $existing_records[0]->getValue('cv.cv_id');
+
+      // Determine if there are referencing records.
+      $table_def = $this->chado_connection->schema()->getTableDef('cv', ['source' => 'database', 'format' => 'default']);
+      // Format is [referencing_table =>
+      // [cv column (cv_id) => referencing_table column], ].
+      $foreign_keys = $table_def['referenced_by'] ?? [];
+      $total_references = 0;
+      foreach ($foreign_keys as $referencing_table => $keydef) {
+        foreach ($keydef as $pkey => $refkey) {
```
laceysanderson: You don't actually use the primary key.

```suggestion
        foreach ($keydef as $refkey) {
```

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1057,202 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+          $query = $this->chado_connection->select('1:' . $referencing_table);
+          $query->condition($refkey, $cv_id, '=');
+          $n = $query->countQuery()->execute()->fetchField();
+          $total_references += $n;
+        }
+      }
+
+      // If there are referencing records and cascade is not set, do nothing.
+      if ($total_references && !($options['cascade'] ?? FALSE)) {
+        return FALSE;
```
laceysanderson: Should we just return FALSE here or should it be an exception? It seems like an error that they tried to delete a CV with referencing records when they indicated they did not want to 🤔 but maybe there just needs to be an option? Like:

` - fail_when_referenced (bool; default TRUE): throw an exception when the cv indicated is referenced by other records and cascade is FALSE`

laceysanderson (changes requested):
I just started with the first method. The comments likely apply to all the methods ;-p

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: I am making a class function to do the check for multiple, so it can be applied everywhere 😁

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
dsenalik: I see this one as okay for the intended use case. A site admin will try to delete a CV on the CV managing form, but does not know that it is in use. We really don't want an exception because we would just have to catch it anyway in our submit function. The FALSE (which will now be changed to `ChadoBuddyPluginBase::FAILURE`) informs the submit code that the CV could not be deleted, and the only common reason is that there are referencing records, so submit can generate the appropriate error message.

laceysanderson: I totally agree with this from a form perspective but am still concerned a bit from the perspective of programmatic use... for example in an importer.

I'm worried that developers will use this function and forget to check the return value 🙈 I know that is on them 😜 but at the same time, this is why the move towards exceptions was made. It's easy enough to catch the exception to provide the same nice form feedback AND it ensures that anyone using this method is suitably warned if they try to delete a CV and it doesn't work 😜 

So I still lean towards making this an exception but allowing an option to provide the silent fail. This way for the form, you just set that option to FALSE but for programmatic use outside the form, it's set to TRUE by default. Anyone else using the method who doesn't want to catch the exception can also set it to FALSE but they have to make the informed choice to do so 😉

dsenalik: Okay, I have made those changes in commit [7cae835](https://github.com/tripal/tripal/pull/2453/commits/7cae835ce55edb7c332d1291741d774be39643ea)

dsenalik: I am now more convinced that your original suggestion to make it an exception is better, and I am going to remove the fail_when_referenced option.

tripal_chado/src/Plugin/ChadoBuddy/ChadoCvtermBuddy.php
```diff
@@ -1057,6 +1045,203 @@ public function associateCvterm(string $base_table, int $record_id, ChadoBuddyRe
...
+          return self::SUCCESS;
+        }
+        catch (\Exception $e) {
+          throw new ChadoBuddyException('ChadoBuddy deleteCv database error ' . $e->getMessage());
+        }
+      }
+    }
+    else {
+      return self::NON_EXISTING;
+    }
```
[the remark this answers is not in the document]
dsenalik: Acknowledge that there is duplicated logic in the various buddies

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
```diff
@@ -665,4 +657,170 @@ public function associateDbxref(string $base_table, int $record_id, ChadoBuddyRe
...
+          return self::SUCCESS;
+        }
+        catch (\Exception $e) {
+          throw new ChadoBuddyException('ChadoBuddy deleteDb database error ' . $e->getMessage());
+        }
+      }
+    }
+    else {
+      return self::NON_EXISTING;
+    }
```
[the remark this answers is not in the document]
dsenalik: Acknowledge that there is duplicated logic in the various buddies

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
```diff
@@ -665,4 +657,170 @@ public function associateDbxref(string $base_table, int $record_id, ChadoBuddyRe
...
+        foreach ($keydef as $refkey) {
+          $query = $this->chado_connection->select('1:' . $referencing_table);
+          $query->condition($refkey, $db_id, '=');
+          $n = $query->countQuery()->execute()->fetchField();
+          $total_references += $n;
+        }
+      }
+      // If there are referencing records and cascade is not set.
+      if ($total_references && !($options['cascade'] ?? FALSE)) {
+        throw new ChadoBuddyException('ChadoBuddy deleteDb error, cannot delete the db, other records reference it');
```
laceysanderson: I would still like to know where the referencing records are. Above I see you go through each referencing table and get a count so maybe saving it to an array of table => count so that you can add that to the exception? Maybe another helper method throwIfReferencingRecords is in order? ;-p

laceysanderson (changes requested):
One more exception suggestion ;-p Previous review comments have all been resolved.

tripal_chado/src/Plugin/ChadoBuddy/ChadoDbxrefBuddy.php
dsenalik: Done in commit https://github.com/tripal/tripal/pull/2453/commits/ece98a2869a8f8d912b7d7080976181f75df35e9
And this helps with code duplication as I had essentially the same code in four places.
Noted a bug, too, in one of the table names, so one of the checks was not working, now fixed.

Example exception:
`Drupal\tripal_chado\ChadoBuddy\Exceptions\ChadoBuddyException: ChadoBuddy deleteCvterm error, records from the following tables reference the cvterm record 3189: 1 record(s) in table "phenotype_cvterm", 1 record(s) in table "stock_cvterm".`

laceysanderson (approved):
✅ Code looks good, exceptions are meaningful 😍 
✅ Did some manual testing in `drush php:cli` and it all worked as expected including exception menssages

I think this is ready to merge! Thanks @dsenalik 😍

PR #2468, opened 2026-04-19
Create default views for tripal content

dsenalik:
# Tripal 4 Core Dev Task

## Description

This PR has the goal of creating a drupal view for each of the tripal chado content types. The views will provide a default way of listing and filtering content.

Views will be loaded for existing content types when cache is rebuilt.
Views will be loaded for new content type collections when they are imported.

## Testing?

There will be several cases to test:
1. Build a new docker.
2. Build a docker on 4.x then switch to this branch.
3. For either of the above, then import a new content type collection.

The new views should be available at _e.g._
`/data-search/analysis`, `/data-search/publication`

And of course you will need some content to properly test the views.

Here is some HTML you can put on your home page or in a block to make getting to these easier
```
<div style="display:flex;gap:15px;">
    <div style="background-color:#f0f0c0;flex:1;padding:20px;">
        <pre><code class="language-plaintext">General</code></pre>
        <ul>
            <li data-list-item-id="e6d0a1142be12ff6c736c06a6df07fc9d">
                <a href="/data-search/analysis">Analyses</a>
            </li>
            <li data-list-item-id="e1775b9da1d3075f2e4050cd1b11d6e45">
                <a href="/data-search/contact">Contacts</a>
            </li>
            <li data-list-item-id="e3a82c24ff6e5a745f226cd412e708d1f">
                <a href="/data-search/organism">Organisms</a>
            </li>
            <li data-list-item-id="e06f5383698dd8a88a0e22b0345dc5978">
                <a href="/data-search/project">Projects</a>
            </li>
            <li data-list-item-id="e800c993583376b3627abdbc074c39048">
                <a href="/data-search/protocol">Protocols</a>
            </li>
            <li data-list-item-id="e3aee53016d1bb163d810fd0e98d6bbc2">
                <a href="/data-search/publication">Publications</a>
            </li>
            <li data-list-item-id="e2cbd676920b608c618bc5e0c7afc9c8e">
                <a href="/data-search/species-tree">Species Trees</a>
            </li>
            <li data-list-item-id="eb3160ad4054be5f2384471ac7d2105b2">
                <a href="/data-search/study">Studies</a>
            </li>
        </ul>
    </div>
    <div style="background-color:#f0f0c0;flex:1;padding:20px;">
        <h5>
            Germplasm
        </h5>
        <ul>
            <li data-list-item-id="e9b0fe8c91612ed74a894cf81aa1284f1">
                <a href="/data-search/germplasm-collection">Germplasm Collections</a>
            </li>
[...]

PR #2470, opened 2026-04-26
Tripal entity permissions bug

### LINKED ISSUE #2469: Bug in tripal content permissions
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

I observe access denied on a drupal view for any tripal content type, _e.g._ analysis, if that view uses tripal content permissions, _i.e._
Image

I have the permission configured as shown here
Image

The bug is on this line
https://github.com/tripal/tripal/blob/44b2c0da1267d2aafa3468ab87a06246f2ce5120/tripal/src/Access/TripalEntityAccessControlHandler.php#L313

This generates a permission `view analysis content`, however the actual permission name is `view all analysis content` as defined here:
https://github.com/tripal/tripal/blob/44b2c0da1267d2aafa3468ab87a06246f2ce5120/tripal/src/Access/TripalEntityPermissions.php#L84

There are also a few cases of uninitialized variables that show up that can be fixed.

And finally, the initial summary text generated is incorrect because the variable here
https://github.com/tripal/tripal/blob/44b2c0da1267d2aafa3468ab87a06246f2ce5120/tripal/src/Plugin/views/access/TripalContentViewAccessHandler.php#L133
is not defined yet, instead we should use `$access_options`

### Steps to reproduce

You can use PR #2468 (which will be dependent on this PR) where the views are defined.
You will need a role that can access one content type but not another.
And then a user with that role.
Then Update the views to use permissions as described here.

### Error messages

```shell
.
```

### Screenshots

.

dsenalik:
# Bug Fix

## Description
Tripal entity permissions are always denied for a view, the linked issue describes in more detail.

## Testing?

On 4.x on Drupal 11.x you can only get this far:
1. Structure -> Views -> +Add view
2. View name: analysis
3. View settings, Show: Tripal Content
4. of type: Analysis
5. Page settings: Create a page checked
6. Save and edit
7. Access: Click on Unrestricted
8. Check Use Tripal Content permissions
9. Apply - nothing happens
10. If you close, and refresh, you get a WSOD:

TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in count() 
(line 214 of modules/contrib/tripal/tripal/src/Plugin/views/access/TripalContentViewAccessHandler.php). 

To test with changes in this PR, you can also use PR https://github.com/tripal/tripal/pull/2468 (which will be dependent on this PR) where the views are defined.
You will need a role that can access one content type but not another.
And then a user with that role.
Then Update the views to use permissions as described here.

### DIFF
--- a/tripal/src/Access/TripalEntityAccessControlHandler.php
+++ b/tripal/src/Access/TripalEntityAccessControlHandler.php
@@ -310,7 +310,7 @@ class TripalEntityAccessControlHandler extends EntityAccessControlHandler {
       // If we have a specific list of content types, then generate the
       // permissions string for each one.
       foreach ($content_types as $content_type) {
-        $permission = $operation . ' ' . $mode . ' ' . $content_type . ' content';
+        $permission = $operation . ' all ' . $content_type . ' content';
         if ($account->hasPermission($permission)) {
           $access_granted++;
         }
--- a/tripal/src/Access/TripalEntityAccessControlHandler.php
+++ b/tripal/src/Access/TripalEntityAccessControlHandler.php
@@ -310,7 +310,7 @@ class TripalEntityAccessControlHandler extends EntityAccessControlHandler {
       // If we have a specific list of content types, then generate the
       // permissions string for each one.
       foreach ($content_types as $content_type) {
-        $permission = $operation . ' ' . $content_type . ' content';
+        $permission = $operation . ' ' . $mode . ' ' . $content_type . ' content';
         if ($account->hasPermission($permission)) {
           $access_granted++;
         }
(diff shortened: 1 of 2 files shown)


vidu-wil (approved):
Performed the manual tests above and verified the following:

- Reproduced the error on 4.x.

Image

- Switched to the `tv4g7-2469-tripal-entity-permissions` branch and created a view as described.
- Created a 'Curator' role with access to analysis content type but no access to organism content type and created a user with curator role.
- Set the access options as shown below.

Image

✅ Was able to successfully apply the changes with no WSOD.

- Logged into the new user account with curator role and tried to access the view. 
✅I was not able to access it as expected. 

Image

- Logged back into the admin account and changed the access options as shown below.

Image

- Logged into the curator account again and tried to access the view. 
✅ I was able to access the view this time as expected.

Image

✅ Everything works as expected and code changes look good! 👍

PR #2474, opened 2026-04-29
TripalEntity methods for better interaction with TripalFields

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PS is working towards better caching of the TripalEntity object for performance reasons. Specifically, this PR will add some of the methods created in the linked PR in order to break the work down into a more testable piece of work 😜 

### Goals

1. ✅ Cache information about Tripal Fields and their backend storage.
2. ✅ Ensure this cache is populated when the entity is first created and that it is updated after the entity is updated.
3. ✅ Make it easy to check the backend storage mappings for chado fields
4. ✅ Improve the testing around the TripalEntity and TripalEntityType component.

#### Next PR

1. Update the TripalField to know a lot more about itself including handling more of it's own storage.
   - Keep track of it's own property types and values
   - Keep the property values sync'd with the drupal ones
   - Clearer interaction between tripal and drupal properties
2. Improve handling of empty fields / properties and default values.
3. Improve the testing around the TripalEntity and TripalEntityType component.

#### Auxiliary Fixes

- [x] Fixes a limitation in `TripalEntityFieldTestTrait` that did not support indicating a field should be required.
- [x] Fixes a bug in TripalFields so they know their terms (pass through idspace plugin to value property type).
- [x] Sets default for the cache status in storage property types.
- [x] Fixes a brittle chado field test which was not using the right test trait.
- [x] Update devcontainer to use PHP 8.5
- [x] #2475 

### Methods

*Not all of these methods were added in this PR but they are the focus of this PR.*

- `registerAllTripalFields()`: caches information about all tripal fields for easy lookup later.
- `registerTripalField()`: called by `registerAllTripalFields()` this focuses on a single field.
- `getTripalStorageFields()`: provides a quick way to get a list of fields managed by a specific storage backend. For example, finding all the chado fields that are required.
- `getTripalFieldPropertyKeys()`: provides a quick list of all the properties for a specific field.
- `getTripalFieldInfo()`: retrieves the previously cached information making it easier to interrogate fields for constraints (e.g ).
- `getTripalFieldStorageInfo()`: retrieves field information specific to tripalstorage bckends.
- `getTripalFieldPropertyInfo()`: retrieves information about the TripalStoragePropertyTypes returned by `tripalTypes()` without having to interrogate objects ;-p
- `getTripalFieldPropertyPath()`: retrieves the path for a given field property type and removes any table alias' defined.

## Testing?

Automated testing coverage of all the methods of interest for this PR:

[...]

PR #2475, opened 2026-04-30
Temporary fix for schema issues caused by Drupal fields on Tripal Content

laceysanderson:
# Bug Fix

## Description

This PR adds a temporary solution to the schema issues described in the linked issue. In short the issue is that our entity schema definitions are not written such that they adopt non-tripal field settings. This causes schema issues whenever you try to test a drupal field on a tripalentity 🙈 

This PR loops through the field storage settings defined by non-tripal fields and adds them to the tripal field storage definition. This is clearly not ideal as there can be conflicts between Drupal fields and now all fields can have all settings (i.e. setting schema are no longer specific to their field).

How we have attempted to mitigate the concerns with this temporary solution:

1. We only add the Drupal field setting to the schema if there is not a tripal field setting with the same name.
2. If there is a tripal field setting with the same name, we check for them to agree.
3. If they don't agree, we throw an exception.

Even so, there are disagreements between fields in drupal 🤦‍♀️ This is why I added a `$skipped_settings` array which allows us to manually inspect any disagreements and add acceptable ones to a whitelist.

Current disagreements:
- `allowed_values`: two Drupal fields define this setting slightly differently but they are compatible. Specifically, one defines the key as an integer and the other a float. We have chosen the float version as it should work for both fields.

### Automated testing

I added `TripalEntityConfigAlterTest` which clears all the config definitions and then gets the tripal entity field definitions specifically. It checks:

1. A setting from a Drupal field which was not available before is now available.
2. A setting from Tripal Fields is still available.
3. A setting shared by both is also available and correct.

### Manual Testing

This can only be tested manually by adding a Drupal field to an existing test and confirming you no longer get schema errors 😜  [I have done that in this dependant PR](https://github.com/tripal/tripal/pull/2474/changes#diff-7ff609b0aeaed77ded07e482b6532e03e3b2c6de639b7b47453e2b2755e217cfR37-R47) and the tests do not show schema errors anymore.

If you want to test the exception, you can comment out the else-block for `allowed_values` and see it fail spectacularly when you clear the cache 😜.

### DIFF
--- a/tripal/tests/src/Kernel/Entity/TripalEntityConfigAlterTest.php
+++ b/tripal/tests/src/Kernel/Entity/TripalEntityConfigAlterTest.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Drupal\Tests\tripal\Kernel\Entity;
+
+use Drupal\KernelTests\KernelTestBase;
+use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
+
+/**
+ * Tests schema alterations for TripalEntity storage configuration.
+ */
+#[RunTestsInSeparateProcesses]
+class TripalEntityConfigAlterTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
+    'field',
+    'tripal',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // IMPORTANT: reset cached schema definitions.
+    $this->container
+      ->get('config.typed')
+      ->clearCachedDefinitions();
+
+  }
+
+  /**
+   * Tests that the config schema is properly altered.
+   *
+   * @see TripalEntityHooks::configSchemaInfoAlter()
+   */
+  public function testSchemaIsAltered() {
+
+    // Fetch the altered schema definition.
+    $definition = $this->container->get('config.typed')->getDefinition('field.storage.tripal_entity.test');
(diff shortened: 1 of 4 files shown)


tripal/tests/src/Kernel/Entity/TripalEntityConfigAlterTest.php
```diff
@@ -0,0 +1,96 @@
...
+      'allowed_values',
+      $settings_mapping,
+      'Setting with conflicting definitions is still available.'
+    );
+    $allowed_values_definition = $settings_mapping['allowed_values'];
+    $this->assertArrayHasKey('sequence', $allowed_values_definition);
+    $this->assertArrayHasKey('mapping', $allowed_values_definition['sequence']);
+    $this->assertArrayHasKey('value', $allowed_values_definition['sequence']['mapping']);
+    $this->assertArrayHasKey('type', $allowed_values_definition['sequence']['mapping']['value']);
+    $this->assertEquals('float', $allowed_values_definition['sequence']['mapping']['value']['type'], 'allowed_values is correctly defined with more permissivefloat keys.');
```
dsenalik:
```suggestion
    $this->assertEquals('float', $allowed_values_definition['sequence']['mapping']['value']['type'], 'allowed_values is correctly defined with more permissive float keys.');
```

dsenalik (approved):
You have everything very well documented and this looks like a nice "temporary fix" for the issue. As such, I hereby approve.

laceysanderson:
Thank you for the quick Review @dsenalik ! 🎉

PR #2480, opened 2026-05-22
Update email addresses for Drupal installation

Ferrisx4:
Per the recommendation, use drupaladmin@local.dev. `local.dev` is technically owned by Google but it seems they promise/guarantee that there is no active mailbox/listener there.

# Bug Fix

## Description

Modifies the default administrator email address so that it passes a newly introduced Drupal requirement for email address to have a full <name>@<domain>.<tld> format. 

## Testing?

Build a brand new Docker image for Drupal. It should not fail when the admin user is automatically created.

### DIFF
--- a/tripaldocker/drupal.Dockerfile
+++ b/tripaldocker/drupal.Dockerfile
@@ -203,10 +203,10 @@ RUN cd /var/www/drupal \
   && sleep 30 \
   && /var/www/drupal/vendor/drush/drush/drush site-install standard \
   --db-url=pgsql://drupaladmin:drupaldevelopmentonlylocal@localhost/sitedb \
-  --account-mail="drupaladmin@localhost" \
+  --account-mail="drupaladmin@local.dev" \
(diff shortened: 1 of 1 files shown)


dsenalik (approved):
The fix is as we discussed in cofest, super simple fix.
Build drupal docker on 4.x:
```
44.74 In SiteInstallCommands.php line 181:
44.74                                                                             
44.74   The email address /drupaladmin@localhost/ is not valid. Use the format    
44.74   user@example.com.                                                         
44.74   The email address /drupaladmin@localhost/ is not valid. Use the format    
44.74   user@example.com.                                                         
44.74                                                                             
```
Build on this branch is successful
```
 => exporting to image                                                                                            12.2s
 => => exporting layers                                                                                           12.1s
 => => writing image sha256:e9b7ae6d04b90ea3abb428582d7081885e0076f526532348ea68870df8fc12bd                       0.0s
```

Issue #2184, opened 2025-03-31
Chado Time field needed


laceysanderson:
### Task Description

This is an offshoot of #1550.

We used to have the following fields in Tripal 3:

- local__time_executed
- local__time_accessioned
- local__time_last_modified

In Tripal 4 these can all be handled by a single field type ChadoTimeDefault which is discoverable on the following columns: timeexecuted, timeaccessioned, timelastmodified.

### Branch Name

tv4g1-issue2184-chadoTimeField

Ferrisx4:
As noted, discovery is important for this field.
In Chado, the following tables have at least one timestamp column:

| Table | Timestamp column(s) |
|---|---|
| `acquisition` | `acquisitiondate` |
| `analysis` | `timeexecuted` |
| `assay` | `assaydate` |
| `cell_line` | `timeaccessioned`, `timelastmodified` |
| `feature` | `timeaccessioned`, `timelastmodified` |
| `library` | `timeaccessioned`, `timelastmodified` |
| `quantification` | `quantificationdate` |

For these tables, when they act as the base table for a Tripal Content type, the column should be detected and the Datetime field used for it automatically. This will need an update hook as well so that existing sites have their Tripal Content types updated.

dsenalik:
PR #2483 created the time field, so we can now implement the following todo's:
https://github.com/tripal/tripal/blob/b319ad5650533af4e5297ca26cbde4a1d23921ee/tripal_chado/src/Plugin/Field/FieldType/ChadoFeatureTypeDefault.php#L186
https://github.com/tripal/tripal/blob/b319ad5650533af4e5297ca26cbde4a1d23921ee/tripal_chado/src/Plugin/Field/FieldType/ChadoAnalysisTypeDefault.php#L182
and also Assay, which doesn't have a todo

Issue #2360, opened 2025-11-16
Move DB query for setting date format from hook_preprocess_html to rebuild


dsenalik:
### Task Description

This query is made **every time** hook_preprocess_html is called

https://github.com/tripal/tripal/blob/21750dc854b3841f9e5f28588dbe356d47a05531/tripal/tripal.module#L292

I suspect we could move it to hook_rebuild.

Update, this is now located here
https://github.com/tripal/tripal/blob/2eba786257cb5a4415c575a9fb419d964ea98ba1/tripal/src/Hook/TripalThemeHooks.php#L183

### Branch Name

tv4g10-issue2360-move-date-format-query

PR #2483, opened 2026-05-26
Tv4g1 issue2184 chado time field

Ferrisx4:
# Tripal 4 Core Dev Task

## Description

This PR adds Tripal and Chado fields that support Datetime values. In the case of Chado, this will add support for `analysis.timeexecuted` and `feature.timeaccessioned` and `feature.timelastmodified`, among others.

## Requirements to move beyond Draft
 - [x] Tripal and Chado Datetime fields implemented
 - [x] Tests for each.
 - [x] ~~Datetime picker implemented~~ - Not worth complexity at the moment.
 - [x] Automatically add to [content type configurations](https://github.com/tripal/tripal/issues/2184#issuecomment-4556000688), with Update Hook 
 - [x] Display format configuration implemented (both Tripal and Chado)
 - [x] Define schema of field settings ([field.settings is deprecated >=13](https://github.com/tripal/tripal/issues/2484))

### Tripal Datetime Field 
On this branch, choose a content type to add a Datetime field to. For example, we can add a "Birthday" field to the Contact type.
1. Navigate to `admin/structure/bio_data/manage/contact/fields` and click "+ Create new field"
2. Choose "Tripal Fields"
3. Provide a name such as "Birthday" and select "Tripal Datetime Field Type"
4. On the following page, since we likely don't have a CV Term for Birthday, we can just use the generic Date term for now: `Date (NCIT:C25164)`
5. Add (`bio_data/add/contact`) or edit an existing Contact, filling out the Birthday field.
6. When saved, a birthday value should appear on the page in the default format (yyyy-mm-dd).
7. Adjust how the date is displayed by visiting `admin/structure/bio_data/manage/contact/display` and using the Gear icon on the Birthday row to change the format. There are useful examples in the help text. Change the format to something like `F j, Y` (January 1, 2000), click Update in the Birthday field, then Save at the bottom of the page. Return to the contact page from earlier and refresh, the Birthday field should have changed its display format.

### Chado Datetime Field
Testing for the Chado Datetime field is largely the same as above except you use Chado fields instead of Tripal fields.
You may already have date/time values in your analysis, feature, or other tables. Let's use 'analysis` as an example.
1. Navigate to `admin/structure/bio_data/manage/analysis/fields` and click "+ Create new field"
2. Choose "Chado Fields"
3. Provide a name such as "Date Executed" and select "Chado Datetime Field Type"
4. On the following page, the current defaults should be OK. The field should identify that, in the analysis table, there is only 1 column that has a timestamp data_type in the database, so it should auto-populate the Table Column dropdown with `timeexecuted`. Just add a CV term for this field, such as `time_executed (local:timeexecuted)`.
5. If you already had existing Chado analysis records with the `timeexecuted` column populated (it's a required field in `chado.analysis`), you should be able to simply (re)publish Analysis content and the values in Chado should appear on the page.
6. If you have no analyses, add a new one, ensuring that the Date Executed field has a value. This is enforced by the field as the underlying Chado column cannot be NULL.
7. When saved, a Date Executed value should appear on the page in the default format (yyyy-mm-dd).
8. Adjust how the date is displayed by visiting `admin/structure/bio_data/manage/analysis/display` and using the Gear icon on the Date Executed row to change the format. There are useful examples in the help text. Change the format to something like `F j, Y` (January 1, 2000), click Update in the Date Executed field, then Save at the bottom of the page. Return to the analysis page from earlier and refresh, the Date Executed field should have changed its display format.

### Unit Testing
Both fields have functional tests implemented, as well as Kernel tests which are modeled after the Phylotree tests.


dsenalik:
When we test this, we should see if the `SET DATESTYLE` command executed for every page as described in https://github.com/tripal/tripal/issues/2360 is even needed. If your formatter and widget work without it we could drop it, it is inherited directly from Tripal 3 and I suspect might not be needed.

dsenalik:
The field is working very nicely except for this case:
I create an analysis but leave the time field blank.
As one might expect, the time field is not displayed.
However, when the field is blank, the current time is saved, you can see it in the db table.
Only if you republish analysis will the time show up.

1. I think we should include a message to the effect "If the field is left blank, then the current time is used."

2. But we are relying on postgres to fill in the default, the chado table definition is set to `Default` - `CURRENT_TIMESTAMP`
so there is no simple way to get this value back without republishing.
Probably the simplest thing to do is to generate the timestamp ourselves in the widget when the field is empty, then we will know what it is and it can get published.
Tripal 3 set the initial default as the current time, that would be a possibility, but would be annoying to clear out to enter a different value.

dsenalik:
Feature request: If I enter `0` in the field, it will store a special value, which the formatter will have in its "hide if equals"
This special value could be either epoch time zero `1970-01-01 00:00:00`, the minimum date supported by postgresql `4713-01-01 BC` (your code doesn't currently support BC), the special value `-infinity` (might need special handling!)
e.g. `update analysis set timeexecuted='-infinity' where analysis_id=2;`

Ferrisx4:
@dsenalik If I'm gathering your thoughts from both comments right, we'd have a field that:
1. When empty (such as on new content), the field's placeholder/suggested value is the current timestamp. This is not the actual value of the field, just the slightly gray text that disappears the moment an actual value is entered.
2. If the user enters `0` into the field, or perhaps (un)checks a checkbox, this is a special designation **not** to expect a legitimate value and to insert a 'special case' value into the database that will be interpreted by Tripal as a 'do not display' flag.

I like the option to use the epoch. This thought occurred to me during initial discussion and the only argument against it would be if we wanted to keep business logic values out of Chado. If someone were to take their Chado database elsewhere, that system would not know to interpret the epoch value, or -infinity, etc, as NULL. It's a shame that this value cannot be NULL. 


---

I vaguely recall participating in an issue in the Tripal 3 era about its behavior of using the current value as the default if no value was provided.

dsenalik:
I now prefer the `-infinity` value because
1. it works
2. the database maintains the "special case" value. Epoch zero is not special to postgres.
And I mean to put the actual value of the current time as the default, not the placeholder. This fixes the problem of it not showing up. My complaint about clearing it if you want a different value is maybe silly, this way you can see and edit to a correct value. This is how Tripal 3 did it, the value started out as the current time.

Ferrisx4:
@dsenalik's changes in https://github.com/tripal/tripal/pull/2519 should fix the failing tests

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
```diff
@@ -433,7 +433,34 @@ fields:
 ## Related: analysis_cvterm, analysis_dbxref, analysis_pub, analysis_relationship, analysisprop,
 ##          phylotree, quantification,
 ##          analysisfeature, nd_experiment_analysis, project_analysis
-## MISSING FIELD: Timestamp
+    -   name: analysis_timeexec
```
dsenalik:
```suggestion
    -   name: analysis_timeexecuted
```
Discovered this because update hook didn't match name here

dsenalik (changes requested):
Only one important fix discovered testing the update hook, then I will approve!

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
Ferrisx4: I had originally made that change in the opposite direction because the machine name ended up being too long, as per Drupal's max length: https://github.com/tripal/tripal/commit/3203d6210dd3fd6d74eebf0a6c863e82584a1ce3
The constraint is FieldStorageConfig::NAME_MAX_LENGTH = 32 in core/modules/field/src/Entity/FieldStorageConfig.php.

tripal/src/Plugin/Field/FieldWidget/TripalDatetimeTypeWidget.php
```diff
@@ -0,0 +1,144 @@
...
+      '#placeholder' => 'YYYY-MM-DD HH:MM:SS',
+      '#attributes' => ['class' => ['js-text-full', 'text-full']],
+      '#element_validate' => [[$this, 'validateDatetimeValue']],
+    ];
+    $element['value']['#description'] = $this->t(
+      'Enter a date and time using the format <code>YYYY-MM-DD HH:MM:SS</code> '
+      . '(e.g., <code>2025-01-15 10:30:00</code>). '
+      . 'The time portion is optional and defaults to <code>00:00:00</code> when omitted. '
+      . 'Sub-second precision is also accepted (e.g., <code>2025-01-15 10:30:00.123456</code>). '
+      . 'If left blank, the current date and time is used. '
```
dsenalik:
```suggestion
```
This is no longer the behavior because of the publishing problem, so this phrase should be removed.

tripal_chado/config/install/tripal.tripalfield_collection.general_chado.yml
Ferrisx4: There are other ones that are longer which gives me confidence that this length is not an issue.

dsenalik (approved):
Approved!

Ferrisx4:
Thanks @dsenalik, I meant to do that earlier 😁 Thanks for your help, input, and reviews.

Issue #1999, opened 2024-10-10
Revisit schema settings for TripalEntity, TripalField and ChadoField


laceysanderson:
### Task Description

Currently we have defined the schema for tripal_entity and its associated fields as we had to in response to automated test schema failures. However, this approach combined with (1) our limited knowledge and (2) lack of Drupal documentation regarding schema definitions, has resulted in a schema definition that does not follow what Drupal expects. 

This has resulted in schema errors for any non-Tripal/Chado fields added to TripalEntities as seen in #1915 and is the reason that PR was closed without merge.

## Observations

Drupal has a number of field specific YAML schema stanzas: `field.storage_settings.[field_type]`, `field.field_settings.[field_type]`, `field.value.[type]`, `field.formatter.settings.[formatter]`, `field.widget.settings.[widget]`.

I think we will want to define stanza's like this for all our fields and include the settings there rather then what we are currently doing. I believe this will allow them to be used auto-magically by Drupal. I'm led to believe this by the stanzas like `field.storage.*.*` which defines the config type for settings as `field.storage_settings.[%parent.type]`. Based on reading that I've done the `[%parent.type]` notation means "fill this token in with the value of the `type` key one level above". In this case, that implies that the definition for the settings in all fields will be `field.storage_settings.[field_type]` which tracks with what I am seeing.

### Notes

You can check the schema definitions for all of Drupal by implementing `hook_config_schema_info_alter` in the .module file and then exploring the $definitions array. Each key in this array matches with the key of a config stanza (e.g. `field.storage.*.*`). You can then add `print_r` statements to look at specific stanzas and the printed values will be shown on the command line when you clear the cache.

You will know if you broke the schema by running the automated tests. If the schema is not defined for anything you get annoying errors on EVERY test using an entity or field. For example, the errors caused by the organism field were:
```
Error adding field, "organism_image", to "organism": Schema errors for field.storage.tripal_entity.organism_image with the following errors: field.storage.tripal_entity.organism_image:settings.default_image missing schema, field.storage.tripal_entity.organism_image:settings.target_type missing schema, field.storage.tripal_entity.organism_image:settings.display_field missing schema, field.storage.tripal_entity.organism_image:settings.display_default missing schema, field.storage.tripal_entity.organism_image:settings.uri_scheme missing schema
```

laceysanderson:
Note: The above linked PR adds a temporary solution to the schema issues described.

[...]

PR #2486, opened 2026-05-27
Drupal 11.4 Deprecations: Field schema fixes

### LINKED ISSUE #2484: Drupal 11.x Deprecations
### Task Description

We have a new batch of deprecation's causing testing on Drupal 11.x to fail 🙈 

See this run for the full list of them: https://github.com/tripal/tripal/actions/runs/26407708641/job/77734919026#step:3:889. For brevity,

1) The 'field.settings' config is deprecated in drupal:11.4.0 and is removed from drupal:13.0.0. Use the `field_purge_batch_size` setting instead. See https://www.drupal.org/node/3494023.

2) The `Drupal\tripal_chado\Controller\ChadoOrganismAutocompleteController` is deprecated in tripal:4.0.0-alpha3 and is removed from tripal:4.1.0. Instead, use `\Drupal\tripal_chado\Controller\ChadoOrganismFormElement`. All the methods in `ChadoOrganismAutocompleteController` now exist in `ChadoOrganismFormElement` class with the same functionality. See https://tripaldoc.readthedocs.io/en/latest/dev_guide/deprecations/organism_autocomplete.html

3) `Drupal\pgsql\Driver\Database\pgsql\Connection::addSavepoint()` is deprecated in drupal:11.4.0 and is removed from drupal:13.0.0. Use `TransactionManager` to start a transaction then call `::commitOrRelease()` or `::rollback()` on it. See https://www.drupal.org/node/3524461

4) `Drupal\pgsql\Driver\Database\pgsql\Connection::releaseSavepoint()` is deprecated in drupal:11.4.0 and is removed from drupal:13.0.0. Use `TransactionManager` to start a transaction then call `::commitOrRelease()` or `::rollback()` on it. See https://www.drupal.org/node/3524461

5) The 'field.settings' config is deprecated in drupal:11.4.0 and is removed from drupal:13.0.0. Use the `field_purge_batch_size` setting instead. See https://www.drupal.org/node/3494023.

### Branch Name

tv4g0-issue2484-drupal114-deprecations

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR addresses only the single deprecation described next. This ended up being a huge rabbit hole that has resulted in us finally fixing the schema errors described in issue #1999 and removing the need for the temporary fix added in #2475.

> The 'field.settings' config is deprecated in drupal:11.4.0 and is removed from drupal:13.0.0. Use the field_purge_batch_size setting instead. See https://www.drupal.org/node/3494023.

### So, what is the right way to do field schema?

1. Every field type, widget and formatter needs it's own schema definition 🙈 See tripal/tests/modules/tripal_test/config/schema/tripal_test.schema.yml for an example for a single field with no custom settings.

   ```yml
   field.field_settings.tripal_test_text_type:
     type: mapping
     mapping: {}

   field.storage_settings.tripal_test_text_type:
     type: mapping
     mapping: {}

   field.widget.settings.tripal_test_text_type:
     type: mapping
     mapping: {}

   field.formatter.settings.tripal_test_text_type:
     type: mapping
     mapping: {}
   ```

2. Tripal defines some settings which should be available to any field --a concept that Drupal is not a fan of 😅 We get around this by defining re-usable chunks of schema definition using distinct patters (see [tripal/config/schema/tripal.schema.yml](https://github.com/tripal/tripal/pull/2486/changes#diff-72a560c72fac3457574dd66b7ddb4168f7e7e12873849ce8073f4882fcf5085cR5-R37) ) and then dynamically including them in existing field schema using [hook_config_schema_info_alter](https://github.com/tripal/tripal/pull/2486/changes#diff-61387b995d0417b75c9bf9336454fb68d2223a2351b5dbd90514d2916431d751R145).

### What to expect in this PR

- [x] field type, storage, widget and formatter schema definition for every single field defined in Tripal Core.
- [x] Implementation of hook_config_schema_info_alter which dynamically adds common chunks of schema definition to ANY field defined within the schema. This cannot be made specific to only fields on tripal_entity content because that is a runtime value which is not available when the schema are setup.
- [x] Fixes to tripalfield collection schema to not validate any of the field type, storage, widget and formatter settings. This allows us to support all fields and is needed because Drupal does not provide a generic field schema type.
- [x] Fix to tripal_test test module to include schema definition for this field.
- [x] Removed tripal/tests/src/Kernel/Entity/TripalEntityConfigAlterTest.php since this specifically tested the temporary fix. It is not needed to test the solution in this PR because without it every single field test failed 😒 and all fields have coverage in our tests.
- [x] Fixed a typo in `tripal_chado/tests/src/Kernel/Plugin/ChadoField/FieldType/ChadoStockCollectionType-TestInfo.yml` where the `base_column` should have been specified but the `type_column` was instead which makes no sense for that field.
- [x] Remove all field related schema altering from tripal_chado/src/Hook/TripalChadoHooks.php since it is no longer needed. These settings were moved into tripal_chado/config/schema/tripal_chado.schema.yml and broken down by field.
- [x] Move field specific settings into their own config
- [x] Move widget specific settings into their own config
- [x] Move formatter specific settings into their own config

## Testing?

As long as all the automated testing works this should be perfect but it is likely a good idea to test fields in general 🙈 

### Automated testing

Note: The 11.x tests will still have a large number of deprecations since I only fixed 2 of 5 deprecations in this PR.


tripal/config/schema/tripal.schema.yml
```diff
@@ -2,6 +2,289 @@
 ## Drupal schema information for various functionality provided by Tripal.
 ## Specifically, focused on Tripal Entities, Tripal Entity Types, and Fields.
 
+## The following reusable schema defininitions follow a specific naming pattern
+## that you can define in your own module to add global settings!
+## Pattern: tripal.YOURMODULE.SPECIFICSETTING
+## In your own YOURMODULE.schema.yml file define a block similar to the one you
+## want to add more settings to but replace `core` with the machine name of
+## your module. All your custom settings will then be merged with cores.
```
dsenalik:
```suggestion
## your module. All your custom settings will then be merged with core's.
```

tripal/config/schema/tripal.schema.yml
```diff
@@ -2,6 +2,289 @@
...
+    termAccession:
+      type: string
+      label: 'Term Accession'
+    debug:
+      type: boolean
+      label: 'Flag to Enable field debugging'
+      nullable: true
+    fixed_value:
+      type: string
+      label: "A fixed value to use for the term Id Space and Accession for this field (e.g. data:1278)."
```
dsenalik:
```suggestion
      label: 'A fixed value to use for the term Id Space and Accession for this field (e.g. data:1278).'
```

tripal/config/schema/tripal.schema.yml
```diff
@@ -2,6 +2,289 @@
...
+    hide_value:
+      type: string
+      label: 'Hide value'
+      nullable: true
+
+##
+## Tripal Configuration Entities!
+##
+##
 ## Described the config entity used for YML-file loading of
```
dsenalik:
```suggestion
## Describes the config entity used for YML-file loading of
```

tripal/src/Hook/TripalEntityHooks.php
```diff
@@ -104,60 +104,101 @@ public function preprocessField(&$variables) {
...
+   *
+   * To work around this limitation, Tripal defines reusable schema fragments
+   * (`tripal.field_settings`, `tripal.field_storage_settings`, etc.) and then
+   * programmatically merges those definitions into the appropriate Drupal
+   * field-type-specific schema definitions here.
+   *
+   * This approach:
+   * - Preserves Drupal core and contrib schema definitions
+   * - Avoids overriding polymorphic schema resolution
+   * - Allows Tripal to define a shared set of settings form many field types
```
dsenalik:
```suggestion
   * - Allows Tripal to define a shared set of settings for many field types
```

tripal/src/Hook/TripalEntityHooks.php
```diff
@@ -104,60 +104,101 @@ public function preprocessField(&$variables) {
...
+      elseif (str_starts_with($key, 'field.field_settings.') && isset($definition['mapping'])) {
+        $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_settings);
+      }
+      elseif (str_starts_with($key, 'field.widget.settings.') && isset($definition['mapping'])) {
+        $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_widget_settings);
+      }
+      elseif (str_starts_with($key, 'field.formatter.settings.') && isset($definition['mapping'])) {
+        $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_formatter_settings);
+      }
+    }
```
dsenalik:
```suggestion
    foreach ($definitions as $key => &$definition) {
      if (isset($definition['mapping'])) {
        if (str_starts_with($key, 'field.storage_settings.')) {
          $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_storage_settings);
        }
        elseif (str_starts_with($key, 'field.field_settings.')) {
          $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_settings);
        }
        elseif (str_starts_with($key, 'field.widget.settings.')) {
          $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_widget_settings);
        }
        elseif (str_starts_with($key, 'field.formatter.settings.')) {
          $definition['mapping'] = array_replace_recursive($definition['mapping'], $field_formatter_settings);
        }
      }
    }
```
This could be ever-so-slightly more efficient, only do the isset once.
Although to be fair, it would only ever be executed once.
Feel free to reject if you like.

dsenalik (approved):
I did a few manual tests creating and editing a few content types, no problems noted

PR #2488, opened 2026-05-28
Organism autocomplete Deprecation fix

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR solves a single deprecation described in #2484; specifically,

> The `Drupal\tripal_chado\Controller\ChadoOrganismAutocompleteController` is deprecated in tripal:4.0.0-alpha3 and is removed from tripal:4.1.0. Instead, use `\Drupal\tripal_chado\Controller\ChadoOrganismFormElement`. All the methods in `ChadoOrganismAutocompleteController` now exist in `ChadoOrganismFormElement` class with the same functionality. See https://tripaldoc.readthedocs.io/en/latest/dev_guide/deprecations/organism_autocomplete.html

This error was being triggered during class discovery 🙈 so the solution was to make a method within the class triggering the deprecation and then to call that in the constructor and each static method.

Additionally, I noticed the old version was used in the phylotree widget so I updated that as well.

## Testing?

Mainly, check that this particular deprecation is no longer in the automated tests.

Also, check the phylotree field to make sure it is still working beautifully :-)

### DIFF
--- a/tripal_chado/src/Controller/ChadoOrganismAutocompleteController.php
+++ b/tripal_chado/src/Controller/ChadoOrganismAutocompleteController.php
@@ -2,8 +2,7 @@
 
 namespace Drupal\tripal_chado\Controller;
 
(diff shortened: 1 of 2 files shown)


vidu-wil (approved):
✅ Code changes look good! 👍
✅ Automated tests run successfully, with no errors related to `ChadoOrganismAutocompleteController` deprecations.

laceysanderson:
Thank you @vidu-wil 😍 🙏

PR #2476, opened 2026-05-09
Add more errors to skip when www.ebi.ac.uk is down

dsenalik:
# Bug Fix

### Issue #2143
( see comment https://github.com/tripal/tripal/issues/2143#issuecomment-4412962914 )

## Description
We have started to see a new error as observed in https://github.com/tripal/tripal/actions/runs/25595777291/job/75141339757 and multiple other workflows:

There was 1 failure:

1\) Drupal\Tests\tripal_chado\Kernel\Plugin\TripalImporter\OboImporterTest::testOboImporter with data set #0 (0, 'TO subset vocabulary')
Unanticipated error: Invalid hostname in URL http://www.ebi.ac.uk/ols/api/ontologies/bfo/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_0000002: cURL error 28: Operation timed out after 30001 milliseconds with 0 bytes received (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://www.ebi.ac.uk/ols4/api/ontologies/bfo/terms/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FBFO_0000002

/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Kernel/Plugin/TripalImporter/OboImporterTest.php:325
/var/www/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:729

And there are multiple other errors observed

## Testing?

Tests on this PR should show the offending test is now skipped.

Issue #2143, opened 2025-03-02
Fix brittle tests that depend on external sites


dsenalik:
### Task Description

This is an extension of issue #1989 which fixed brittle pubmed tests, so that if pubmed happens to be down when tests are run, they will be gracefully skipped.
This is also related to #1501 where the first thing to try is to retry the download a couple of times. In particular, see this comment https://github.com/tripal/tripal/pull/1679#pullrequestreview-1718265923

The task is to both increase the robustness of downloads from external sites, and if the download still fails, process gracefully in automated tests

This is inspired today because the overnight PHPUnit tests are all failing.
Rerunning one of the failed tests completes without error, highly suggestive of this being a symptom of the external site being down.

### Branch Name

tv4g9-issue2143-fix-brittle-tests-depending-on-external-sites

### Example error messages from https://github.com/tripal/tripal/actions/runs/13612833349/job/38052135242
```
There were 3 failures:

1) Drupal\Tests\tripal_chado\Functional\PubSearchQueryImporterTest::testPubSearchQueryImporterSimpleTest
One publication should have been retrieved but was not
Failed asserting that 1 matches expected 0.

/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Plugin/PubSearchQueryImporterTest.php:60

2) Drupal\Tests\tripal_chado\Functional\TaxonomyImporterTest::testTaxonomyImporterSimpleTest
No organism Arabidopsis thaliana found which should have been created
Failed asserting that '0' matches expected 1.

/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Plugin/TaxonomyImporterTest.php:61

3) Drupal\Tests\tripal_chado\Functional\TripalPubLibraryTest::testTripalPubLibraryTestSimpleTest
There should be more than 0 records found for this query
Failed asserting that null is greater than 0.

/var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Functional/Plugin/TripalPubLibraryTest.php:65

--

2 tests triggered 6 PHP warnings:

1) /var/www/drupal/web/modules/contrib/tripal/tripal/src/Plugin/TripalPubLibrary/TripalPubLibraryPubMed.php:287
fopen(): php_network_getaddresses: getaddrinfo for www.ncbi.nlm.nih.gov failed: Temporary failure in name resolution

Triggered by:

[...]

PR #2492, opened 2026-06-03
Add more errors to skip when www.ebi.ac.uk is down part 2

dsenalik:
# Bug Fix

## Description

Another error appears on occasion during the obo importer test, see this comment https://github.com/tripal/tripal/issues/2143#issuecomment-4612271178

This PR adds this to the list of messages (see PR #2476) to cause the test to be skipped.

## Testing?

Just code review because this is so infrequent!


laceysanderson (approved):
Thanks for keeping on these!

PR #2493, opened 2026-06-03
Add a title to the null publication

### LINKED ISSUE #2465: Add a title to the null publication
### Task Description

As a quality of life issue, the null publication should have a title, to avoid the following warning when publishing:

`[warning] WARNING: 1 entities have blank titles. You should either update the source records, or modify the title format tokens, and then republish.`

The GFF3 importer and synonym field find the null publication by its `uniquename` being the text string `null` (not an actual NULL), so any other field can have any value.

### Branch Name

tv4g1-issue2465-null-publication-title

dsenalik:
# New Feature

## Description

This adds a title to the null publication. This publication is used as a workaround for tables that have a pub_id with a NOT NULL constraint, but there is not actually a relevant publication.

During chado install, this record is created with the `miniref` and `uniquenames` set to the text string "null", and the `type_id` set to the "null" cvterm. All other columns without defaults are set to actual NULL values.

But when we publish publications, we get this annoying warning:
`[warning] WARNING: 1 entities have blank titles. You should either update the source records, or modify the title format tokens, and then republish.`

The GFF3 importer and synonym field find the null publication by its uniquename being the text string "null", so any other field can have any value.

This PR adds a title of "No publication".
It is also added to the test chado SQL.

## Testing?

Build a new docker, and publish the publication content type. There should be no warning, and the null pub should have a title now.

### DIFF
--- a/tripal_chado/chado_schema/initialize-1.3.sql
+++ b/tripal_chado/chado_schema/initialize-1.3.sql
@@ -22,7 +22,8 @@ INSERT INTO cvterm (name, cv_id, dbxref_id) VALUES (
   (SELECT dbxref_id FROM dbxref WHERE accession = 'local:null')
 ) ON CONFLICT DO NOTHING;
 
-INSERT INTO pub (miniref, uniquename, type_id) VALUES (
+INSERT INTO pub (title, miniref, uniquename, type_id) VALUES (
+  'No publication',
   'null',
   'null',
   (SELECT cvterm_id FROM cvterm WHERE name = 'null')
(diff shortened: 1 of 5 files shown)


laceysanderson (commented):
This is looking good but until #2495 is fixed, I can't fully test it.

✅ Code review looks good :-) 
✅ I published the `pub` content type and it worked :-)

Image

Image

## What I'm not able to test yet

I think we should check the linker field doesn't make any assumptions on this by creating a project and assigning the null publication to it.

I would like to test the synonym field specifically as well.

Also we should import a GFF3 file to test this... Will any GFF3 trigger a link to the null pub or how do I get that checked?

dsenalik:
The GFF importer uses the null publication for the `feature_cvterm` and `feature_synonym` tables.
Any gff will trigger the null pub lookup, but may not add anything to those tables, not sure how to trigger that.

laceysanderson (approved):
✅ I was able to create a project and assigning the null publication to it.

Image

✅ The synonym field works perfectly

✅ I imported the `tripal_chado/tests/fixtures/gff3_loader/small_gene.gff` and confirmed that it created synonym and cvterm records using the null pub 👍 

Image

PR #2495, opened 2026-06-03
path_item can be null on latest drupal version

### LINKED ISSUE #2494: Call to a member function set() on null
### Tripal Version

4.x

### Drupal Version

11.x-dev

### PostgreSQL Version

18

### PHP Version

8.5

### General Description

New error from `phpunit --filter=ChadoPropertyWidgetFormTest`
```
 ✘ Chado property widget update with data set "0"
   │
   │ Error: Call to a member function set() on null
   │
   │ /var/www/drupal/web/modules/contrib/tripal/tripal/src/Entity/TripalEntity.php:420
   │ /var/www/drupal/web/modules/contrib/tripal/tripal/src/Entity/TripalEntity.php:1664
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:617
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/ContentEntityStorageBase.php:918
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:543
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php:828
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/EntityBase.php:371
   │ /var/www/drupal/web/core/lib/Drupal/Core/Entity/EntityForm.php:294
   │ /var/www/drupal/web/modules/contrib/tripal/tripal/src/Form/TripalEntityForm.php:275
   │ /var/www/drupal/web/modules/contrib/tripal/tripal_chado/tests/src/Kernel/Plugin/ChadoField/Widget/ChadoPropertyWidgetFormTest.php:205
```

This is also triggered entering a publication through the UI

### Branch
`tv4g0-issue2494-property-widget`

dsenalik:
# Bug Fix

## Description

A recent change in drupal results in now getting a NULL for code of the format
`$path_item = $this->path->first();`

In the past, the object would be defined, but be effectively empty, i.e. just a language code.
Now it will be NULL.

The change here simply checks if the object is truthy before proceeding.

Also update `$this->path` to `$this->get('path')` to avoid using the magic path method.

## Testing?

1. Tests should again pass on 11.x
2. You can test creating a publication through the UI with a number of properties included.

### DIFF
--- a/tripal/src/Entity/TripalEntity.php
+++ b/tripal/src/Entity/TripalEntity.php
@@ -398,9 +398,11 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
       if ($entities) {
 
         // Reset the internal path field.
-        $path_item = $this->path->first();
-        $path_item->set('alias', '');
-        $path_item->set('pid', NULL);
+        $path_item = $this->get('path')->first();
+        if ($path_item) {
+          $path_item->set('alias', '');
+          $path_item->set('pid', NULL);
+        }
 
         // Keep track of the duplicates here but DO NOT throw the exception
         // until the end so we can still remove previous alias' if that applies.
@@ -416,8 +418,10 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
       // The field will create the alias for us so we just need to ensure
       // its set to the new one here.
       if ($during_save) {
-        $path_item = $this->path->first();
-        $path_item->set('alias', $new_alias);
+        $path_item = $this->get('path')->first();
+        if ($path_item) {
+          $path_item->set('alias', $new_alias);
+        }
       }
       // We have to create the path alias ourselves.
       else {
@@ -431,9 +435,11 @@ class TripalEntity extends ContentEntityBase implements TripalEntityInterface {
         }
         $new_alias_object->save();
         // And update the internal path field.
-        $path_item = $this->path->first();
-        $path_item->set('alias', $new_alias);
-        $path_item->set('pid', $new_alias_object->id());
+        $path_item = $this->get('path')->first();
+        if ($path_item) {
+          $path_item->set('alias', $new_alias);
+          $path_item->set('pid', $new_alias_object->id());
+        }
       }
     }
     // If an alias already exists, and is different...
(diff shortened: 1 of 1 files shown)


vidu-wil (approved):
I did manual testing for this and everything worked as expected. ✅

- I tried to create a publication content on 4.x and the following error was thrown.

Image

- Then, I switched to the current branch and tried the same and the publication content was created successfully. ✅

Image

- Also created, edited and deleted content across various content types, and all of them work as expected. ✅

Image

Image

- All the automated tests are passing as expected ✅ (specially ran the `testChadoPropertyWidgetUpdate` test locally since it was failing on 4.x before this PR and it is passing now as expected)

Image

laceysanderson (changes requested):
I don't think this is creating the alias when content is created anymore 😬 

1. Set the URL Alias Pattern for the organism content type to `organism/[organism_genus]-[organism_species]`.
2. ❌ Created an organism `Tripalus databasica subspecies postgresqlus` but it had a URL of `bio_data/2` instead of my expected `organism/Tripalus-databasica`.
3. ❌ When I edit this organism, it shows the alias is not set 😬 when I expect to see `organism/Tripalus-databasica`.

Image

dsenalik:
Commit https://github.com/tripal/tripal/pull/2495/commits/b2efbdb48abc9b8403f71b0ca717140e83b7b49a should fix the problem @laceysanderson has noted. It is because we were skipping the set because $path_item was null. Now we use a flag to trigger the generation of the alias if it was not set.

laceysanderson (approved):
Manual testing worked flawlessly this time! Thanks @dsenalik 

Specifically,

✅ Alias was created according to rules when content was created
✅ During content creation, if the alias from our pattern already existed, an error is thrown and the alias is not created again.
✅ During content creation, if our custom already existed, an error is thrown and the alias is not created again.
✅ When editing content, if our custom already existed, an error is thrown and the alias is not assigned.
✅ We can set a custom alias when creating content and it is used, not the alias pattern.
✅ If an error is encountered when creating content and the alias is not created, we can edit the content and the alias pattern will be tried again with any new values. Allowing it to be fixed 😘

PR #2513, opened 2026-06-16
TripalField should control it's own properties and storage

laceysanderson:
# Tripal 4 Core Dev Task

## Description

This PR is working towards better caching of the TripalEntity object for performance reasons. More specifically, it is focused on allowing fields to control their own properties rather than TripalEntity controlling them.

- [x] Add new method `getTripalFieldItems()` to make it easy to grab the values for all Tripal fields on an entity.
- [x] Small fixes to marking fields as cached (see change in TripalStorageBase and DrupalSqlStorage) and set the default cache_status to TRUE for all property types.
- [x] Fix coding standards in `tripal/src/TripalField/TripalFieldItemBase.php`
- [x] Add methods and variables to `TripalFieldItemBase` to support keeping track of tripal properties.
    - `getTripalStoragePropertyTypes()` and `getTripalStoragePropertyType()` retrieves the tripal property type(s)
    - `getTripalStoragePropertyValues()` and `getTripalStoragePropertyValue()` retrieves the tripal property value(s)
- [x] Remember the difference between `exportTripalStoragePropertyValues()` and `getTripalStoragePropertyValues()`.
- [x] Add methods to ensure drupal and tripal properties remain sync'd
    - `updateTripalStoragePropertyValues()` allows you to update the tripal property values when new values are available.
    - `syncTripalStoragePropertyValues()` updates the tripal properties to match the drupal ones.
    - `syncFieldValuesWithTripalStorage()` updates the drupal field items to match the tripal properties. This is often used with `updateTripalStoragePropertyValues()`
    - `clearFieldValuesForTripalStorage()` which removes from drupal field items that are not meant to be cached in the drupal field tables. This should be called after Tripal Storage has done it's job ;-p 
- [x] Update `getValuesArray()` and `saveValuesArray()` methods to interact with the fields when asking for values instead of Tripal Storage directly.
- [x] Add Interface for TripalStorage property types to a) make it easier to say we expect a property type ;-p and b) ensure that all property types implement the methods we expect.
- [x] Update all the property types to use the interface 😒 
- [x] Allow property types to get their default value since they inherently know what it should be ;-p 
- [ ] Now that fields handle their own properties, we can allow them to determine if they are empty like Drupal expects them to be able to do!
- [ ] Update Tripal entity to simply ask the field if it is empty rather than making assumptions
- [x] Add a lot of testing 🤩


dsenalik:
Issue #2550 will benefit from this as it may solve the problem

PR #2516, opened 2026-06-22
Fix Drupal docker build failing by bringing variables into build scope.

### LINKED ISSUE #2515: Drupal Docker builds are failing 🙈
### Task Description

I just ran the workflow to build all the drupal images (base image for our TripalDocker) and see that a large number are failing 😬 

See https://github.com/tripal/tripal/actions/runs/27980660542/job/82809594105

These are the failing images:

Image

They are all failing on [step 19](https://github.com/tripal/tripal/actions/runs/27980660542/job/82809594105).

### Branch Name

tv4g0-issue2515-dockerBuildsFailing

laceysanderson:
# Tripal 4 Core Dev Task
## Description

Building a number of the drupal images (base image for our TripalDocker) was failing due to the phpversion arg not being in scope. This happened due to a change in a recent PR which mistakenly removed a perceived duplicated definition of the arg.

Both declarations were needed to ensure the arguments were in the correct scope. See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact for more information but essentially:

> An ARG declared before a FROM is outside of a build stage, so it can't be used in any instruction after a FROM. To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage.

[These were the failing images](https://github.com/tripal/tripal/actions/runs/27980660542/job/82809594105):

Image

## Testing?

Build some of the failing drupal images locally. For example,

```
docker build -f tripaldocker/drupal.Dockerfile --build-arg drupalversion=11.x-dev --build-arg postgresqlversion=18 --build-arg phpversion=8.5 .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ ARG postgresqlversion='18'
 ARG buildplatform='linux/amd64'
 FROM --platform=${buildplatform} tripalproject/tripaldocker-drupal:drupal${drupalversion}-php${phpversion}-pgsql${postgresqlversion}
 
-## Redefine the core args so that they are within the build scope.
+## Redefine the global args so that they are within the build scope.
+# See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG phpversion='8.4'
 ARG drupalversion='11.3.x'
 ARG postgresqlversion='18'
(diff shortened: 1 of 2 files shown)


dsenalik (approved):
Awesome @laceysanderson - thanks for finding and fixing this.
I'll wait for the docker builds to finish, but this looks ready to merge.

Confirmed also locally, fails on 4.x, succeeds on this branch

PR #2291, opened 2025-09-21
Chado custom tables with nullable columns unique constraint

dsenalik:
# Tripal 4 Core Dev Task

## Description

This PR solves the problem of NULL values "sneaking" through unique constraints and allowing duplicate records to be inserted in a table. With Postgresql 15+ we can use `UNIQUE NULLS NOT DISTINCT` which prevents this.
For some context see chado issue https://github.com/GMOD/Chado/issues/139
The inspiration for this is any new linker table created with a nullable `type_id` column could potentially get duplicate linker records inserted.

Documentation will be added by https://github.com/tripal/tripal_doc/pull/120

## Caveat

If you want to apply a unique key to a single nullable column, this will end badly. For example, if you have a `uri` column, and want it unique, but also want to allow it to be left blank, you can only have NULL in this column once.

## Opt In

So I am making this an "opt-in" behavior, the key `nulls not distinct` must be present and set to TRUE in the schema for this change to be made.

## Testing?
The changes do not affect core chado tables, only custom chado tables.
So to test, create one. Here is a schema you could use:
```
array (
  'table' => 'biomaterial_project',
  'fields' => array (
    'biomaterial_project_id' => array (
      'type' => 'serial',
      'not null' => true,
    ),
    'biomaterial_id' => array (
      'type' => 'int',
      'not null' => true,
    ),
    'project_id' => array (
      'type' => 'int',
      'not null' => true,
    ),
    'type_id' => array (
      'type' => 'int',
    ),
  ),
  'primary key' => array (
    0 => 'biomaterial_project_id',
  ),
  'foreign keys' => array (
    'biomaterial' => array (
      'table' => 'biomaterial',
      'columns' => array (
        'biomaterial_id' => 'biomaterial_id',
      ),
    ),
    'project' => array (
      'table' => 'project',
      'columns' => array (
        'project_id' => 'project_id',
      ),
    ),
  ),
  'indexes' => array (
    'biomaterial_project_idx1' => array (
      0 => 'biomaterial_id',
    ),
    'biomaterial_project_idx2' => array (
      0 => 'project_id',
    ),
  ),
  'unique keys' => array (
    'biomaterial_project_unique_uq1' => array (
      0 => 'biomaterial_id',
      1 => 'project_id',
      2 => 'type_id',
    ),
  ),
  'nulls not distinct' => array(
    'biomaterial_project_unique_uq1' => TRUE,
  ),
)
```
Once the table is created, examine it at a psql prompt and verify the state of the unique constraint
```
[...]

PR #2525, opened 2026-07-06
Chado records support for saving a null and check "nulls not distinct"

### LINKED ISSUE #2419: Chado records doesn't exactly follow db rules for nulls in unique key columns
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

It is legal to create a chado column that allows NULL values, but is also configured as a unique key.

Example: we have a `uri` column, and we **do not** want duplicated uri values. But having a uri is **optional**, so we **do** want to allow duplicated NULL values.

The Postgres default in this case is to consider each NULL as a distinct value, and so more than one record can have a NULL value. This is probably not desirable if the column is part of a multi-column unique key, see closed PR #2291, however if the unique key is for a single key, such as an **optional** URI, then you would likely want to allow multiple records to have a NULL value.

Chado Records currently does not allow this because it does not check the Postgres 'nulls not distinct' setting here, but assumes that multiple NULL values are not allowed.
https://github.com/tripal/tripal/blob/c4530b518e24d74346751cd2e85e5565b199565a/tripal_chado/src/TripalStorage/ChadoRecords.php#L1467-L1493

Also as they work currently, the text|string widgets return an empty string, they do not ever return a NULL, so chado records won't be able to store a NULL.

### Steps to reproduce

1. Go to Tripal -> Data Storage -> Chado -> Chado Custom Tables -> Add Custom Table
2. Create a table with this schema:
```
[
    'table' => 'license',
    'fields' => [
      'license_id' => [
        'type' => 'serial',
        'not null' => TRUE
      ],
      'name' => [
        'type' => 'varchar',
        'length' => 1024,
        'not null' => TRUE
      ],
      'summary' => [
        'type' => 'text',
      ],
      'uri' => [
        'type' => 'text',
      ],
      'uri2' => [
        'type' => 'text',
      ],
    ],
    'primary key' => ['license_id'],
    'unique keys' => [
      'license_c1' => ['name'],
      'license_c2' => ['uri'],
      'license_c3' => ['uri2'],
    ],
    'nulls not distinct' => [
      'license_c3' => TRUE,
    ],
    'indexes' => [
      'license_idx1' => ['name'],
    ],
  ]
```
3. Create a new content type based on this table
  a. Tripal -> Page Structure -> + Add Custom Type
  b. Name it "License" and give it whatever CV term you want, but there is a License (NCIT:C43611) term.
  c. Add fields: name (Chado String Field Type) required, summary (Chado Text Field Type), uri (Chado Text Field Type), and uri2 (Chado Text Field Type)

Image

4. Update the page title format for your new content type Tripal -> Page Structure -> License -> Edit
5. Add a new License entity, e.g. "Public Domain" and leave the URI empty.
6. Add a second License entity, e.g. "No License" and again leave the URI empty. Observe the following error:
Image

dsenalik:
# Bug Fix

## Description

It is legal to create a chado column that allows NULL values, but is also configured as a unique key.

Example: we have a `uri` column, and we **do not** want duplicated uri values. But having a uri is **optional**, so we **do** want to allow duplicated NULL values.

The Postgres default in this case is to consider each NULL as a distinct value, and so more than one record can have a NULL value. This is probably not desirable if the column is part of a multi-column unique key, see closed PR #2291, however if the unique key is for a single key, such as an **optional** URI, then you would likely want to allow multiple records to have a NULL value.

Chado Records currently does not allow this because it does not check the Postgres 'nulls not distinct' setting here, but assumes that multiple NULL values are not allowed. The logic has been updated in this PR to only enforce if a 'nulls not distinct' constraint is present.

Also as they work currently, the text|string widgets return an empty string, they do not ever return a NULL, so chado records are not able to store a NULL. The two string widgets have been modified to have a field setting to specify if an empty string should be converted into a NULL.

To support testing this, allow passing of widget settings to TripalEntityFieldTestTrait and ChadoFieldTestTrait, see commit https://github.com/tripal/tripal/pull/2525/commits/1b51e51892ad8445ea8283adb0ca25f7448bf9dd

Proper phpunit testing of this will be added to PR #2273 in association with #2527 because that will be the first case where it is needed.

## Testing?

1. Go to Tripal -> Data Storage -> Chado -> Chado Custom Tables -> Add Custom Table
2. Create a table with this schema:
```
[
    'table' => 'license',
    'fields' => [
      'license_id' => [
        'type' => 'serial',
        'not null' => TRUE
      ],
      'name' => [
        'type' => 'varchar',
        'length' => 1024,
        'not null' => TRUE
      ],
      'summary' => [
        'type' => 'text',
      ],
      'uri' => [
        'type' => 'text',
      ],
      'uri2' => [
        'type' => 'text',
      ],
    ],
    'primary key' => ['license_id'],
    'unique keys' => [
      'license_c1' => ['name'],
      'license_c2' => ['uri'],
      'license_c3' => ['uri2'],
    ],
    'nulls not distinct' => [
      'license_c3' => TRUE,
    ],
    'indexes' => [
      'license_idx1' => ['name'],
    ],
  ]
```
3. Create a new content type based on this table
  a. Tripal -> Page Structure -> + Add Custom Type
  b. Name it "License" and give it whatever CV term you want, but there is a License (NCIT:C43611) term.
  c. Add fields: name (Chado String Field Type) required, summary (Chado Text Field Type), uri (Chado Text Field Type), and uri2 (Chado Text Field Type)

Image

4. Update the page title format for your new content type Tripal -> Page Structure -> License -> Edit
5. Add a new License entity, e.g. "Public Domain" and leave the URI and URI2 empty.
6. Add a second License entity, e.g. "No License" and again leave the URI empty. Observe the following error:
Image
7. Now in the field settings for the URI field, set the text filter to "plain text" so that we don't get paragraph tags, and turn on the store NULL setting.

8. For the other URI2 field set it to "plain text" but leave the store NULL at its default unchecked state.
9. Try adding a second License, URI empty, but put something unique for URI2
10. Now add another License, URI empty, but put something unique for URI2
11. Check the database table to confirm a null was stored in each case. The first record has an empty string since it was created before we changed the widget setting.

```
sitedb=> \pset null '*NULL*'
Null display is "*NULL*".
sitedb=> select * from license;
 license_id |     name      |      summary       |  uri   |  uri2  
------------+---------------+--------------------+--------+--------
          1 | Public Domain | <p>Free to use</p> |        | 
          2 | test1         | <p>something</p>   | *NULL* | abcd
          3 | test2         | <p>something</p>   | *NULL* | defg
```


reynoldtan:
Completed the manual test and verified the following items:

- [x] Adding a second License (No License) entity with URI empty triggers an error.
image

- [x] uri fields store null with set field text filter.

image


image

- [x] The automated tests are passing.
image


Overall, the changes look good. I just have a few minor suggestions for improvement.

tripal/src/Plugin/Field/FieldWidget/TripalStringTypeWidget.php
```diff
@@ -35,4 +35,50 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
...
+    return $elements;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsSummary() {
+    $summary = parent::settingsSummary();
+    $summary[] = $this->t('Store NULL if empty: @setting',
+      ['@setting' => $this->getSetting('null_if_empty') ? $this->t('yes') : $this->t('no')]);
```
reynoldtan:
```suggestion
      ['@setting' => $this->getSetting('null_if_empty') ? 'yes' : 'no']);
```
I think the outer t() call handles all translation, consider removing the nested t()s.

tripal/src/Plugin/Field/FieldWidget/TripalStringTypeWidget.php
```diff
@@ -35,4 +35,50 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
     ];
     return $element;
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
+    foreach ($values as $key => $item) {
+      if ($item['value'] === '' && $this->getSetting('null_if_empty')) {
```
reynoldtan:
```suggestion
      if (array_key_exists('value', $item) && $item['value'] === '' && $this->getSetting('null_if_empty')) {
```
Consider adding an extra check that makes sure there is 'value' element in $item.

tripal/src/Plugin/Field/FieldWidget/TripalTextTypeWidget.php
```diff
@@ -132,7 +145,8 @@ public function settingsSummary() {
 
     $summary[] = $this->t("Text Format: @format", ['@format' => $format_label]);
     $summary[] = $this->t("Number of Rows: @rows", ['@rows' => $num_rows]);
-
+    $summary[] = $this->t('Store NULL if empty: @setting',
+      ['@setting' => $this->getSetting('null_if_empty') ? $this->t('yes') : $this->t('no')]);
```
reynoldtan:
```suggestion
      ['@setting' => $this->getSetting('null_if_empty') ? 'yes' : 'no']);
```
Same suggestion as in the StringType widget

tripal/src/Plugin/Field/FieldWidget/TripalStringTypeWidget.php
dsenalik: That is an interesting question! I have now tested this, I set up a second language (Italian) and with the code as it is, the Yes and No (but for No you can't tell the difference) get translated
with_t

If I remove the t() then it does not get translated
without_t

So I think it is correct as I have it. Although in other places where yes|no is used, the first letter is capitalized, so I changed to that form Yes|No for consistency. https://github.com/tripal/tripal/pull/2525/commits/4bb7375fc33ba15d97d687df55b5ad868f98ca4d

tripal/src/Plugin/Field/FieldWidget/TripalTextTypeWidget.php
dsenalik: Same response as in the StringType widget 😁

tripal/src/Plugin/Field/FieldWidget/TripalStringTypeWidget.php
reynoldtan: Interesting indeed! Thanks for testing that - that's really helpful to know. I'll keep that in mind and we can leave it as is.

PR #2541, opened 2026-07-15
Tv4g1 issue2540 tripal entity type list builder sort method

### LINKED ISSUE #2540: TripalEntityTypeListBuilder should use class sort method
### Task Description

As @laceysanderson noted in Slack, [TripalEntityTypeListBuilder](https://github.com/tripal/tripal/blob/ca59f180380f5ffee702155c0c4a6784649fb200/tripal/src/ListBuilders/TripalEntityTypeListBuilder.php#L71-L77) is using the wrong sort method and should instead use [the one provided in the TripalEntityType](https://github.com/tripal/tripal/blob/ca59f180380f5ffee702155c0c4a6784649fb200/tripal/src/Entity/TripalEntityType.php#L855) class.

### Branch Name

tv4g1-issue2540-TripalEntityTypeListBuilder_sort_method

Ferrisx4:
# Bug Fix

## Description

Sorting Tripal Content Types on the Category listing page uses the incorrect sorting method. 

The changes include removing a redundant function (`load()`) that held sorting logic and instead rename the actual sorting function in the TripalEntityType class from `sortByCategory()` to just `sort()`. Also, this change reverses the sort order so that it is in alphabetical order (previously it was reversed, z->a).

## Testing?

Visit `admin/structure/bio_data` on this branch and view the order of the content types. They should be in order, alphabetically, first by category and then by label. "General" is a special case which always gets sorted to the top. If you have 1 or fewer categories installed, use the "+ Import type collection" button at the top to install more categories.

### DIFF
--- a/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
+++ b/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
@@ -7,7 +7,9 @@ use Drupal\tripal\Entity\TripalEntityType;
 use PHPUnit\Framework\Attributes\Group;
 
 /**
- * Tests TripalEntityType::sort(), used to order Tripal Content Types in
+ * Tests TripalEntityType::sort().
+ * 
+ * This function is used to order Tripal Content Types in
  * both TripalEntityTypeListBuilder and TripalEntityUIController.
  *
  * @group TripalEntityType
--- a/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
+++ b/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
@@ -5,6 +5,7 @@ namespace Drupal\Tests\tripal\Kernel\Entity;
 use Drupal\Tests\tripal\Kernel\TripalTestKernelBase;
 use Drupal\tripal\Entity\TripalEntityType;
 use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
 
 /**
  * Tests TripalEntityType::sort().
--- a/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
+++ b/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
@@ -66,7 +66,7 @@ class TripalEntityTypeSortTest extends TripalTestKernelBase {
   /**
    * Within the same category, types should be sorted alphabetically by label.
    */
-  public function testTypesWithinACategoryAreSortedByLabel() {
+  public function testTypesWithinCategoryAreSortedByLabel() {
 
     $types = [
       $this->createType('t_orange', 'Orange', 'Genetic'),
--- a/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
+++ b/tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
@@ -16,6 +16,7 @@ use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
  * @group TripalEntityType
  */
 #[Group('tripal-content')]
+#[RunTestsInSeparateProcesses]
 class TripalEntityTypeSortTest extends TripalTestKernelBase {
 
   /**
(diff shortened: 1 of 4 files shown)


tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+namespace Drupal\Tests\tripal\Kernel\Entity;
+
+use Drupal\Tests\tripal\Kernel\TripalTestKernelBase;
+use Drupal\tripal\Entity\TripalEntityType;
+use PHPUnit\Framework\Attributes\Group;
+
+/**
+ * Tests TripalEntityType::sort(), used to order Tripal Content Types in
+ * both TripalEntityTypeListBuilder and TripalEntityUIController.
```
dsenalik:
```suggestion
 * Tests TripalEntityType::sort().
 * 
 * This function is used to order Tripal Content Types in
 * both TripalEntityTypeListBuilder and TripalEntityUIController.
```

dsenalik: Just a coding standards change.

tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
```diff
@@ -0,0 +1,110 @@
...
+      ['General', 'Expression', 'Genetic', 'Genomic', 'Germplasm'],
+      $ordered_categories,
+      'Categories should be sorted with "General" first and all other categories in ascending alphabetical order.'
+    );
+  }
+
+  /**
+   * Within the same category, types should be sorted alphabetically by label.
+   */
+  public function testTypesWithinACategoryAreSortedByLabel() {
```
dsenalik:
```suggestion
  public function testTypesWithinCategoryAreSortedByLabel() {
```
Another coding standards - they don't like two capitals together

tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace Drupal\Tests\tripal\Kernel\Entity;
+
+use Drupal\Tests\tripal\Kernel\TripalTestKernelBase;
+use Drupal\tripal\Entity\TripalEntityType;
+use PHPUnit\Framework\Attributes\Group;
```
dsenalik, line 7:
```suggestion
use Drupal\Tests\tripal\Kernel\TripalTestKernelBase;
use Drupal\tripal\Entity\TripalEntityType;
use PHPUnit\Framework\Attributes\Group;
use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
```
1) /var/www/drupal/web/core/tests/Drupal/KernelTests/KernelTestBase.php:221
Kernel test classes must specify the #[RunTestsInSeparateProcesses] attribute, not doing so is deprecated in drupal:11.3.0 and will throw an exception in drupal:12.0.0. See https://www.drupal.org/node/3548485

tripal/tests/src/Kernel/Entity/TripalEntityTypeSortTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Tests TripalEntityType::sort(), used to order Tripal Content Types in
+ * both TripalEntityTypeListBuilder and TripalEntityUIController.
+ *
+ * @group TripalEntityType
+ */
+#[Group('tripal-content')]
+class TripalEntityTypeSortTest extends TripalTestKernelBase {
```
dsenalik, line 20:
```suggestion
#[Group('tripal-content')]
#[RunTestsInSeparateProcesses]
class TripalEntityTypeSortTest extends TripalTestKernelBase {
```

dsenalik (approved):
I tested also by installing the tripal_file module and its content type group also sorted correctly to the end. I modified the sort order by changing the comparison in `tripal/src/Entity/TripalEntityType.php` that you had fixed, and sorting was descending but otherwise in correct order.

Sorting also verified at `/bio_data/add` to be correct

Some minor coding standards and deprecation in the test, but I verified the tests work by messing with the sort function to reverse it again or break the check for General and the test caught it!

Approving because my change requests are minor

PR #2556, opened 2026-08-05
Bundles must have at least one required field to permit publish

### LINKED ISSUE #2555: Warn user if they made a new content type with no required fields
### Task Description

Scenario: A user creates a new content type, and neglects to set any of the fields as required.

Publish depends on having at least one required field to detect existing published entities.
At the following function, we silently return an empty array if there are no required fields
https://github.com/tripal/tripal/blob/4cc159344e3fe7be161154d3e28c171825301946/tripal/src/Services/TripalEntityLookup.php#L276-L284

The result is that every time publish is run, everything is published again, creating duplicates

We need to trigger an error here and stop publishing in this case.

Probably throw an exception here, and publish can wrap in a try-catch to provide a nice error message.

### Branch Name

tv4g1-issue2555-error-if-no-required-fields

dsenalik:
# Bug Fix

## Description

Scenario: A user creates a new content type, and neglects to set any of the fields as required.

Publish depends on having at least one required field to detect existing published entities.

Existing behavior is to republish every entity every time publish is run, creating duplicate published entities for a single chado record.

Because we need at least one required field to detect entity_id, the new behavior will be to display an error message and halt publishing.

A serious problem is that the `biomaterial` table does not have any required columns other than `biomaterial_id` so we can't republish this either.
Also `featuremap` table lacks any required columns.
Resolved this by making the name field required for both of these, and added an update hook to do likewise. https://github.com/tripal/tripal/pull/2556/commits/6a5fb208c80a1c4229a248629b2b9abeda124b59

## Testing?

Phpunit testing has been extended to cover this.

To test manually, create a new content type and don't make any field required.
e.g use contact table, create content type "Lab", and add only the `name` field (Chado String Field Type)
Make sure it is not required!

not_required

Fix the page title format if you want, but this is not required to test
You can enter a new "Lab" through the UI
Republish should fail thusly:
```
drush trp-chado-pub lab --republish
 [notice] Initializing publish
 [notice] Finding all candidate records in the "contact" chado table
 [error]  ERROR: Bundle "lab" does not have any required fields. At least one is needed to lookup entity ID from record ID. 
[site http://default] [TRIPAL] ERROR: Bundle "lab" does not have any required fields. At least one is needed to lookup entity ID from record ID.
 [notice] There are no records to publish for this content type
```

To test the update hook, you would need to build a docker on 4.x, and import the Expression, Genetic, and Genomic content type collections. Then you can switch to this branch, and run the update hook. Verify that the name field is now required on biosample, genetic map, and physical map.

### DIFF
--- a/tripal_chado/tests/src/Kernel/Plugin/TripalBackendPublish/ChadoPublishTest.php
+++ b/tripal_chado/tests/src/Kernel/Plugin/TripalBackendPublish/ChadoPublishTest.php
@@ -35,7 +35,7 @@ class ChadoPublishTest extends ChadoTestKernelBase {
   protected $chado_publish;
 
   /**
-   * The most recent error message from the mocked tripal logger.
+   * The most recent warning and error messages from the mocked tripal logger.
    *
    * @var string
    */
@@ -55,9 +55,14 @@ class ChadoPublishTest extends ChadoTestKernelBase {
 
     // Create a mocked logger so we can access error messages from the Tripal logger.
     $mock_logger = $this->getMockBuilder(TripalLogger::class)
-      ->onlyMethods(['warning'])
+      ->onlyMethods(['warning', 'error'])
       ->getMock();
     $mock_logger->method('warning')
+      ->willReturnCallback(function ($message, $context, $options) {
+          $this->mock_warning .= str_replace(array_keys($context), $context, $message);
+          return NULL;
+      });
+    $mock_logger->method('error')
       ->willReturnCallback(function ($message, $context, $options) {
           $this->mock_warning .= str_replace(array_keys($context), $context, $message);
           return NULL;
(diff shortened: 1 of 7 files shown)


tripal_chado/tests/src/Kernel/Plugin/TripalBackendPublish/ChadoPublishTest.php
```diff
@@ -386,6 +391,20 @@ public function testTripalPublishServiceSingleJob() {
...
+
+    // Test that an exception is thrown if the bundle has no required fields.
+    // We will just remove the cached values for simplicity.
+    $cache_id = 'tripal_required_fields';
+    $cache = \Drupal::cache()->get($cache_id);
+    $data = $cache->data;
+    $data['organism'] = [];
+    \Drupal::cache()->set($cache_id, $data, time() + 3600);
+
+    // Now publish, and expecte a logger error message to be generated.
```
vidu-wil:
```suggestion
    // Now publish, and expect a logger error message to be generated.
```

vidu-wil (commented):
The manual testing works as expected. ✅

- I created a "Lab" content type with no required fields as instructed and executed `drush trp-chado-pub lab --republish` command. The expected error was shown.

Image

Code changes look good. ✅ Just a minor typo. :)

PR #2564, opened 2026-08-13
Allow discover() to pass display options

### LINKED ISSUE #2563: Allow a field's discover() to specify display settings
### Task Description

Use case: A field formatter that presents information as a table.
In this case we may not want the field label to be displayed, because the table has column headings.
In this case, the field `discover()` function should hide the field label by default.

Image

### Branch Name

tv4g1-issue2563-display-for-field-discovery

dsenalik:
# New Feature

## Description

Use case: A field formatter that presents information as a table.
In this case we may not want the field label to be displayed, because the table has column headings.
In this case, the field `discover()` function should hide the field label by default.

Image

## Testing?

This will be used in the tripal file module, PR #2273, not on any current fields.
You can test this feature on that branch by discovering fields on various content types, e.g. organism or project, and adding the file field.
After the file field is added, the display settings should look like:
file_on_organism_is_hidden


vidu-wil (approved):
I did manual testing for this PR and it works as expected. 🎉

Before adding this PR, the label for the organism_file field is shown by default as mentioned.

Image

Image

After adding the changes on this PR, the label is hidden by default as shown below.

Image

Image


Code changes look good as well ✅

PR #2573, opened 2026-08-18
Organism buddy find by scientific name infraspecies fix

### LINKED ISSUE #2572: getOrganismFromScientificName() with and without infraspecies
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

The function `getOrganismFromScientificName()` mishandles this particular case:
You have an existing organism with an infraspecific name e.g. _Tripalus bogusii_ subsp. _fakus_.
You call `getOrganismFromScientificName()` with just the genus and species. You do not specify infraspecific type or name.
You get back a single organism that is this organism with the subspecies.
That is not the one that should be returned, you should get back no organism.
In the case where you have both _Tripalus bogusii_ and _Tripalus bogusii_ subsp. _fakus_, you get back an error because more than one record matches.

This function should add null placeholders for type_id and infraspecific_name when the organism name does not have them.

### Steps to reproduce

1. Import one organism by NCBI taxidthat has an infraspecific type_id and name:
Use this value: 79196
2. Run the job
3. Generate a tree: Tripal -> Data Loaders -> Taxonomy Tree Generator
4. Run the job
5. Note that this association is made, but should not be because we did not import _Daucus carota_, we only imported the subspecies.
...
Wrong: `Import phylotree: Associated <em class="placeholder">Daucus carota</em> to organism_id: <em class="placeholder">1</em>`
Correct: `Import phylotree: Associated <em class="placeholder">Daucus carota subsp. carota</em> to organism_id: <em class="placeholder">1</em>`
`Import phylotree summary: <em class="placeholder">2</em> nodes were successfully associated to content, <em class="placeholder">0</em> nodes could not be associated`
...

Image

dsenalik:
# Bug Fix

## Description

This fixes the case where if you ony have organism "Genus1 species2 subsp. infra3", and try to lookup by scientific name "Genus1 species2", you should not get any matches returned.
Right now on 4.x, you will get the organism with infraspecies returned.

## Testing?

I have phpunit now catching this. Scenario 3 in provideOrganismFromScientificNameScenarios should return no organisms, rather than two.

I also now test getOrganism with and without specifying infraspecies.

Manual testing can be done easily by generating a tree
1. Import one organism by NCBI taxidthat has an infraspecific type_id and name:
Use this value: 79196
2. Run the job
3. Generate a tree: Tripal -> Data Loaders -> Taxonomy Tree Generator
4. Run the job
5. This line should NOT be in the output any more:
`Import phylotree: Associated <em class="placeholder">Daucus carota</em> to organism_id: <em class="placeholder">1</em>`

### DIFF
--- a/tripal_chado/src/Plugin/ChadoBuddy/ChadoOrganismBuddy.php
+++ b/tripal_chado/src/Plugin/ChadoBuddy/ChadoOrganismBuddy.php
@@ -497,20 +497,27 @@ class ChadoOrganismBuddy extends ChadoBuddyPluginBase implements ChadoBuddyInter
     // else specified by $options will be checked.
     // Split our string up into a maximum of 4 substrings or parts.
     $parts = preg_split('/\s+/', $scientific_name, 4);
-    // $scientific_name could be a single word, so make sure this is defined.
-    $parts[1] = $parts[1] ?? '';
-
+    // $scientific_name could be a single word, so make sure a
+    // placeholder for species is defined.
+    $parts[1] ??= '';
+    // Supply null defaults for organisms without infraspecies.
+    $parts[2] ??= NULL;
+    $parts[3] ??= NULL;
+    if ($parts[2]) {
+      // Expand the abbreviation from rank if it exists.
+      $parts[2] = $this->unabbreviateInfraspecificRank($parts[2]);
+    }
     // Setup our conditions for lookup.
     $conditions = [
       'organism.genus' => $parts[0],
       'organism.species' => $parts[1],
+      'organism.infraspecific_name' => $parts[3],
     ];
-    // Remove the abbreviation from rank if it exists.
-    if (array_key_exists(2, $parts)) {
-      $conditions['cvterm.name'] = $this->unabbreviateInfraspecificRank($parts[2]);
+    if ($parts[2]) {
+      $conditions['cvterm.name'] = $parts[2];
     }
-    if (array_key_exists(3, $parts)) {
-      $conditions['organism.infraspecific_name'] = $parts[3];
+    else {
+      $conditions['organism.type_id'] = NULL;
     }
 
     // Check 'case_sensitive' option and pass it through to our getter.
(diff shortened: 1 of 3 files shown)


tripal_chado/src/Plugin/ChadoBuddy/ChadoOrganismBuddy.php
```diff
@@ -497,20 +497,27 @@ public function getOrganismFromScientificName(string $scientific_name, array $op
...
       'organism.genus' => $parts[0],
       'organism.species' => $parts[1],
+      'organism.infraspecific_name' => $parts[3],
     ];
-    // Remove the abbreviation from rank if it exists.
-    if (array_key_exists(2, $parts)) {
-      $conditions['cvterm.name'] = $this->unabbreviateInfraspecificRank($parts[2]);
+    if ($parts[2]) {
+      $conditions['cvterm.name'] = $parts[2];
     }
```
carolyncaron: Consider rearranging so that there is only a single `if ($parts[2])` lookup:
```suggestion
    // Setup our conditions for lookup.
    $conditions = [
      'organism.genus' => $parts[0],
      'organism.species' => $parts[1],
      'organism.infraspecific_name' => $parts[3],
    ];
    if ($parts[2]) {
      // Expand the abbreviation from rank if it exists.
      $parts[2] = $this->unabbreviateInfraspecificRank($parts[2]);
      $conditions['cvterm.name'] = $parts[2];
    }
```

carolyncaron (approved):
Looks and works great, thank you for fixing this!

I made one suggestion to simplify the code, but otherwise doesn't impact functionality 😄

tripal_chado/src/Plugin/ChadoBuddy/ChadoOrganismBuddy.php
dsenalik: Nice suggestion, and I will even skip setting parts[2] as it is unnecessary = https://github.com/tripal/tripal/pull/2573/commits/f82c6305bb92c708318f0131bf87998ead29518f

PR #2581, opened 2026-09-03
getOrganismFromScientificName() without infraspecific_name fix

### LINKED ISSUE #2578: getOrganismFromScientificName() without infraspecific_name returns incorrect results
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

getOrganismFromScientificName() does not return the correct organism when infraspecific_name is not provided. Instead, the lookup behaves as though the organism does not exist in the database, even when a matching record is present. This bug was introduced in [Organism buddy find by scientific name infraspecies fix PR](https://github.com/tripal/tripal/pull/2573) in commit [Commit b945d78](https://github.com/tripal/tripal/commit/b945d780602fc642721cac84e2fb75934d84f722). This occurs because the query looks for NULL values in infraspecific_name. 

### Steps to reproduce

Use the getOrganismFromScientificName() method when infraspecific_name is not given for an organism.

vidu-wil:
## Description

getOrganismFromScientificName() does not return the correct organism when infraspecific_name is not provided. Instead, the lookup behaves as though the organism does not exist in the database, even when a matching record is present. This bug was introduced in [Organism buddy find by scientific name infraspecies fix PR](https://github.com/tripal/tripal/pull/2573) in commit [Commit b945d78](https://github.com/tripal/tripal/commit/b945d780602fc642721cac84e2fb75934d84f722). This occurs because the query looks for NULL values in infraspecific_name. 

## Current Status

The fix we need to ensure the correct organism is returned when the infraspecific name is not provided breaks the original issue the linked PR was created to solve 😫 As such we could use help @dsenalik 🙈 We've included the testing instructions for both situations below.

### Testing the new issue introduced

1. Two organisms to test both cases: 
  - No infraspecific name) Genus: `Tripalus`, Species: `databasica`
  - Has infraspecific name) Genus: `Daucus`, Species: `carota`, Type: `subspecies (TAXRANK:0000023)`, infraspecific name: `sativus`
3. Open up an interactive php client using `drush php:cli` and use the following code to attempt to pull out this organism.
```php
$organism_buddy = \Drupal::service('tripal_chado.chado_buddy')->createInstance('chado_organism_buddy', []);
$noinfra = $organism_buddy->getOrganismFromScientificName('Tripalus databasica');
$infra = $organism_buddy->getOrganismFromScientificName('Daucus carota subspecies sativus');
```

<details>
  <summary>Linked PR instructions</summary>

1. Import one organism by NCBI taxid that doesn't have an infraspecific name: Use this value: 4039
2. Run the job
4. Generate a tree: Tripal -> Data Loaders -> Taxonomy Tree Generator
5. Run the job. The nodes should associate to the content successfully. 

Screenshot 2026-09-08 at 12 03 43 PM

</details>

<details>
  <summary>Original detection instructions on TripalCultivate Germplasm</summary>

1. Clone the [TripalCultivate-Germplasm](https://github.com/TripalCultivate/TripalCultivate-Germplasm) module.
4. Create the organism `Tripalus databasica` by navigating to Tripal->Content->Add Tripal Content -> Organism
6. Create new germplasms  "121S", "122S" and "124S" of type "Breeding Cross Progeny" by navigating to Tripal > Content > Add Tripal Content > Breeding Cross Progeny with the same organism you created in step 2.
7. Navigate to the Germplasm Cross Importer in your site: Tripal > Data Loaders > Tripal Cultivate: Germplasm Cross Importer
8. Import [crosses_simple.tsv](https://github.com/user-attachments/files/31971042/crosses_simple.tsv).
9. Before this fix, the result should be the following as it doesn't recognize the existing organism.
Screenshot 2026-09-08 at 12 11 52 PM
10. Checkout this branch on the tripal module within the `TripalCultivate-Germplasm` module.
11. Repeat steps 4 and 5 above. It should create a job as shown below.
Screenshot 2026-09-08 at 12 23 53 PM
12. Run the job and verify that it was imported successfully. 
Screenshot 2026-09-08 at 12 59 25 PM

</details>

### DIFF
--- a/phpunit.11.xml
+++ b/phpunit.11.xml
@@ -74,7 +74,7 @@
       <file>tripal_chado/tripal_chado.module</file>
       <directory suffix=".php">tripal_layout/src</directory>
       <file>tripal_layout/tripal_layout.module</file>
-      <directory suffix=".php">tripal_search_viewsh_views/src</directory>
+      <directory suffix=".php">tripal_search_views/src</directory>
       <file>tripal_search_views/tripal_search_views.module</file>
     </include>
     <exclude>
--- a/phpunit.12.xml
+++ b/phpunit.12.xml
@@ -86,7 +86,7 @@
       <file>tripal_chado/tripal_chado.module</file>
       <directory suffix=".php">tripal_layout/src</directory>
       <file>tripal_layout/tripal_layout.module</file>
-      <directory suffix=".php">tripal_search_viewsh_views/src</directory>
+      <directory suffix=".php">tripal_search_views/src</directory>
       <file>tripal_search_views/tripal_search_views.module</file>
     </include>
     <exclude>
--- a/tripal/src/TripalDBX/TripalDbxSchema.php
+++ b/tripal/src/TripalDBX/TripalDbxSchema.php
@@ -316,7 +316,7 @@ WHERE pg_attribute.attnum > 0
 AND NOT pg_attribute.attisdropped
 AND pg_attribute.attrelid = :key::regclass
 AND (format_type(pg_attribute.atttypid, pg_attribute.atttypmod) = 'bytea'
-OR pg_get_expr(pg_attrdef.adbin, pg_attribute.attrelid) LIKE 'nextval%')
+OR pg_get_expr(pg_attrdef.adbin, pg_attribute.attrelid)::text LIKE 'nextval%')
 EOD;
         $result = $this->connection
           ->query($sql, [
@@ -479,8 +479,8 @@ EOD;
           AND n.nspname = :schema
         )
       WHERE
-        p.proname ILIKE :function_name
-        AND pg_get_function_identity_arguments(p.oid) ILIKE :func_args
+        p.proname::text ILIKE :function_name
+        AND pg_get_function_identity_arguments(p.oid)::text ILIKE :func_args
     ";
     $args = [
       ':function_name' => $function_name,
@@ -599,10 +599,10 @@ EOD;
     $constraint_exists = $this->connection->query("
       SELECT TRUE
       FROM information_schema.table_constraints
-      WHERE table_name ILIKE :table
-        AND table_schema ILIKE :schema
-        AND constraint_type ILIKE :type
-        AND constraint_name ILIKE :name
+      WHERE table_name::text ILIKE :table
+        AND table_schema::text ILIKE :schema
+        AND constraint_type::text ILIKE :type
+        AND constraint_name::text ILIKE :name
       ;",
       [
         ':table' => $table,
@@ -674,8 +674,8 @@ EOD;
     $constraint_exists = $this->connection->query(
       "SELECT TRUE
       FROM information_schema.table_constraints
-      WHERE table_name ILIKE :table
-        AND table_schema ILIKE :schema
+      WHERE table_name::text ILIKE :table
+        AND table_schema::text ILIKE :schema
         AND constraint_type = 'PRIMARY KEY';",
       [
         ':table' => $table,
--- a/tripal/src/TripalDBX/pg-clone-schema/clone_schema.sql
+++ b/tripal/src/TripalDBX/pg-clone-schema/clone_schema.sql
@@ -485,8 +485,8 @@ BEGIN
   LOOP
     cnt := cnt + 1;
     IF data_type = 'USER-DEFINED' THEN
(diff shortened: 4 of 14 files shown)


vidu-wil:
I did manual testing for this PR after @dsenalik added a fix. I am getting all the expected results now 😍 The tests I did and the results are as follow:

1. I created two organisms to test both cases as mentioned:

- No infraspecific name) Genus: Tripalus, Species: databasica
- Has infraspecific name) Genus: Daucus, Species: carota, Type: subspecies (TAXRANK:0000023), infraspecific name: sativus

2. executed the following on the PHP interactive terminal as shown.

Screenshot 2026-09-17 at 9 43 22 AM

Both cases now return Organism Buddy objects as expected ✅

I also tested this on our TripalCultivate Germplasm module and it is also working as expected now with the new changes as shown below. ✅

Screenshot 2026-09-17 at 10 22 25 AM
Screenshot 2026-09-17 at 10 22 47 AM

I also tested using the Original issue's testing instruction using `Daucus carota` and `Daucus carota subsp. sativus`. The job completed successfully. ✅

Screenshot 2026-09-17 at 10 58 02 AM

dsenalik (approved):
@vidu-wil has tested and since it has been thoroughly tested I will approve, although really I did most of the code changes, so some conflict of interest there, so I won't be the one to actually merge

tripal_chado/tests/src/Kernel/Plugin/ChadoBuddy/ChadoOrganismBuddyTest.php
```diff
@@ -646,20 +646,35 @@ public static function provideOrganismFromScientificNameScenarios() {
       0,
     ];
 
-    // #1: An organism without infraspecific rank.
+    // #1: An organism with empty infraspecific name.
```
carolyncaron, line 649: I would suggest keeping the original test scenario and adding a new one to specifically test with an empty infraspecific name as both scenarios should be valid.

carolyncaron (changes requested):
I followed the testing instructions with both clone sites and this solution does work 🎉 

However (and I hate to be the bringer of bad news 😢) I tried further testing with a combo of genus + species + infraspecies (so no Infraspecific type) and `getOrganismFromScientificName()` does not seem to be able to pick up this situation while manual testing, even though we now have an automated test for it. 

```
> $noinfra = $organism_buddy->getOrganismFromScientificName('Tripalus databasica NULL');

= []
```

Whereas when I run the Taxonomy Tree Generator, it does successfully detect both 'Tripalus databasica' and 'Tripalus databasica NULL':

Image

tripal_chado/tests/src/Kernel/Plugin/ChadoBuddy/ChadoOrganismBuddyTest.php
dsenalik: So when parsing a scientific name, the baked-in assumption is that the third word is the infraspecific type. So a scientific name string would probably need a "special" placeholder if there was no type. Right now "NULL" in the string is just text. There is actually a term "no_rank" that could potentially be used for this. But that term caused me so much trouble in tripal3 that I finally converted them to NULLs and I don't want to go back to that. But I will think about this some more.

dsenalik: FWIW I do have one actual case with no type_id but an infraspecific name https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=278110
```
select * from organism where type_id is null and infraspecific_name is not null;
 organism_id |          abbreviation           | genus |  species   | common_name | infraspecific_name | type_id | comment 
-------------+---------------------------------+-------+------------+-------------+--------------------+---------+---------
         649 | Apium graveolens Rapaceum Group | Apium | graveolens |             | Rapaceum Group     |         | 
(1 row)
```

laceysanderson: @dsenalik can you create a new issue for this as it sounds non-trivial to handle in this PR and I'd like to get this fix merged 😘

laceysanderson: I've moved this issue to https://github.com/tripal/tripal/issues/2598 to be dealt with at a later date.

PR #2591, opened 2026-09-16
Fix undefined variable in contact widget

### LINKED ISSUE #2589: Undefined variable in contact widget
### Tripal Version

4.x

### Drupal Version

any

### PostgreSQL Version

any

### PHP Version

any

### General Description

In this section
https://github.com/tripal/tripal/blob/45e1c134e833b960dc79ff89e8f89d96fc402093/tripal_chado/src/Plugin/Field/FieldWidget/ChadoContactWidgetDefault.php#L95-L108
the variable `$item` should be `$item_vals`. There is no `$item` variable.

### Steps to reproduce

You would need to do the chado migrations for this to even show up. But the bizarre thing is that the code actually works, because the undefined variable is inside an `if empty()`.

Nevertheless it should be fixed.

1. Data storage -> Chado -> Manage Chado Schemas
2. Apply Migrations -> Apply Migrations
3. Run the job
4. Tripal -> Content -> + Add Tripal Content
5. Add a Project
6. No visible error is thrown but magically everything works.

EthemKD:
# Bug Fix

### Issue #

Closes #2589

## Description

The contact widget's `linker_type_id` handling references an undefined `$item` variable in three places. This change uses the existing `$item_vals` array consistently for checking, assigning, and reading the linker type value.

The change is scoped to `tripal_chado/src/Plugin/Field/FieldWidget/ChadoContactWidgetDefault.php`.

## Testing?

- Verified there are no remaining `$item['linker_type_id']` references in the file.
- Verified the three affected references use `$item_vals['linker_type_id']`.
- `git diff --cached --check` passed.

PHP is not installed in the local environment, so `php -l` and the Drupal/PostgreSQL test matrix were not run locally. The repository CI can exercise the supported PHP/Drupal/PostgreSQL matrix.


dsenalik (approved):
Thank you for the fix.Code changes are perfect.
I updated the issue, as even the incorrect code managed to work, because `$item` is first checked inside `if empty()` but this fix is as it should be. Can be merged once tests pass.

